PR #3, opened 2020-03-27
Pushstate & Back Button

evantahler:
Closes https://www.pivotaltracker.com/story/show/171760073

1. When entering a list page for the first time, use `Router.replace` to add the "?" to the url.  This is not a pushstate change so the back button will take you the previous page
1. When navigating between list pages, and the browser back button is clicked, we will compare the new url hash with the pagination saved in the component's state. If it's not a match, we update the offset, which will then trigger a reload


andyjih:
@evantahler looks like more than a one-line change 😅

PR #4, opened 2020-03-30
A profile property rule cannot be saved if the source is missing required options

evantahler:
closes https://www.pivotaltracker.com/story/show/172007155

Screen Shot 2020-03-30 at 9 22 57 AM


evantahler:
Waiting on CI... 

Screen Shot 2020-03-30 at 9 28 05 AM

PR #11, opened 2020-03-31
Mixins

evantahler:
Goals:

* implement a strategy for mixins which provides both the type definitions and the prototype overrides for javascript.  
* share code between models for common relations `Options` and `Mapping`

Notes:
* The `recommended` way of doing Mixins in typescript is kind of crazy - the interfaces combine nicely but you still need to manually mess with the prototypes of your classes: https://www.typescriptlang.org/docs/handbook/mixins.html
* This Sequelize advice is out of date and does not work: https://github.com/RobinBuschmann/sequelize-typescript/issues/487
* Sequelize messes with the insntances classes as part of the `constructor`.  Therefore, we should not run mixin constructors as part of `applyMixins ` 
* There are cyclc load order problems with Sequelize models requiring eachother before the mixins are applied.  This looks like a null class rather than raising a proper error :/


evantahler:
closing in favor of https://github.com/grouparoo/grouparoo/pull/12

PR #13, opened 2020-04-01
Draft mode

evantahler:
- sources, profile property rules, schedules, & destinations gain draft mode
- you don't need to choose a name to start working on a draft
- remove all modals from the UI

Closes https://www.pivotaltracker.com/story/show/171906943


core/api/src/migrations/000065-addState.js
```diff
@@ -0,0 +1,31 @@
...
+  "schedules",
+  "sources",
+];
+
+module.exports = {
+  up: async function (migration, DataTypes) {
+    for (const i in tables) {
+      await migration.addColumn(tables[i], "state", {
+        type: DataTypes.STRING(191),
+        defaultValue: "ready",
```
bleonard, line 14: i wonder if the default should be "draft"

bleonard: if it's not moving "from draft to ready" will it skip the validations in the transition?

core/api/src/models/ProfilePropertyRule.ts
```diff
@@ -152,21 +159,43 @@ export class ProfilePropertyRule extends LoggedModel<ProfilePropertyRule> {
...
+    await StateMachine.transition(instance, STATE_TRANSITIONS);
+  }
+
+  @BeforeCreate
+  static async ensureSourceIsReady(instance: ProfilePropertyRule) {
+    const source = await Source.findOne({
+      where: { guid: instance.sourceGuid },
+    });
+
+    if (source.state !== "ready") {
```
bleonard, line 197: Some of the models have `VALID_STATES` or similar. I wonder if it might be useful to have `isReady()` or something on all the models to abstract it away to other models as they evolve.

core/api/src/modules/stateMachine.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  export async function transition(
+    instance: Model,
+    transitions: Array<StateTransition>
+  ) {
+    const klass = modelName(instance);
+    const newState: string = instance["state"];
+    //@ts-ignore
+    const oldState: string = instance._previousDataValues["state"];
+    if (!newState || !oldState || newState === oldState) {
```
bleonard: i'm worried `!oldState` here misses initial validation (running `checks` on creation. Is that true?

core/api/src/migrations/000065-addState.js
evantahler: actually, the table has a default value of 'ready', but the model's default is 'draft'.  I did that to make the migration easier... and will likely revert that when I squash the models a bit later.

core/api/src/models/ProfilePropertyRule.ts
evantahler: oh! we can remove `validStates` now!

core/api/src/modules/stateMachine.ts
evantahler: great catch! https://github.com/grouparoo/grouparoo/pull/13/commits/67bdef65ea8dc6dea891ce3931876912c02f1d7d should take care of that

PR #19, opened 2020-04-03
If there are no schedules, after creating a source mapping, create a schedule automatically

evantahler:
closes https://www.pivotaltracker.com/story/show/172010602

### DIFF
--- a/core/web/components/forms/source/mapping.tsx
+++ b/core/web/components/forms/source/mapping.tsx
@@ -1,6 +1,7 @@
 import { useState, useEffect } from "react";
 import { useApi } from "../../../hooks/useApi";
 import { Row, Col, Table, Form, Button } from "react-bootstrap";
+import { createSchedule } from "../schedule/add";
 import Router from "next/router";
 
 export default function ({ apiVersion, errorHandler, successHandler, query }) {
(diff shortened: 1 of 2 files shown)


core/web/components/forms/source/mapping.tsx
```diff
@@ -105,6 +113,19 @@ export default function ({ apiVersion, errorHandler, successHandler, query }) {
     if (response?.source) {
       successHandler.set({ message: "Source updated" });
       setSource(response.source);
+
+      // this source can have a schedule, and we have no schedules yet
+      if (scheduleCount === 0 && response.source.scheduleAvailable) {
```
bleonard, line 118: If I read it right, the this actually makes a schedule. The [story](https://www.pivotaltracker.com/story/show/172010602) is more about redirecting to that page, not making one explicitly. Or did I read this wrong?

evantahler: yes... but there was another theme to remove as many of the the modals as possible.  Since schedules no longer need tiles, there's no user input needed to create a schedule any more... so there's no schedule/new page.   To be validated later with @andyjih

bleonard: I guess the issue is that there needs to be a schedule to take them to that page for the same reason we need all this draft business?

evantahler: That's what the `createSchedule` method does!

PR #23, opened 2020-04-06
Allow next to load & transpile plugins from node_modules

evantahler:
Closes https://www.pivotaltracker.com/story/show/171189466

Allows the app to be distributed via NPM and overwrites next.js' default webpack loader to allow to transpilaition of files within node_modules.  This includes `@grouparoo/core` and plugins.

This PR also statically writes a special loader for the frontend to find plugins.  Since webpack pre-scans the directories for files to transpile, we need to only have static references to paths that exist in this deployment.  `if`/`switch` statements are not allowed.  The web app uses the `env` building step to write a new typescript helper at runtime, before the webpack process begins. 

Related to https://github.com/zeit/next.js/issues/11411 and https://github.com/evantahler/next-subproject-example

### DIFF
--- a/core/web/client/client.ts
+++ b/core/web/client/client.ts
@@ -1,7 +1,7 @@
 import Axios, { AxiosRequestConfig, Method } from "axios";
 import Router from "next/router";
 import { isBrowser } from "../utils/isBrowser";
-import PackageJSON from "../../../package.json";
+import PackageJSON from "../../package.json";
 
 interface ClientCacheObject {
   locked: boolean;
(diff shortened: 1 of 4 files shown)


bleonard (approved):
💥

core/web/client/client.ts
```diff
@@ -1,7 +1,7 @@
 import Axios, { AxiosRequestConfig, Method } from "axios";
 import Router from "next/router";
 import { isBrowser } from "../utils/isBrowser";
-import PackageJSON from "../../../package.json";
+import PackageJSON from "../../package.json";
```
bleonard, line 4: Oh I see, it was pulling from the wrong package.json (in the monorepo)

PR #31, opened 2020-04-08
Deployment & Docs

evantahler:
- [x] Heroku
- [x] Docker
- [x] Docker Compose
- [x] AWS EKS


bleonard:
note that the `tools` for lerna used to look for `PLUGINS.md` in its old spot. Might need to update that.

evantahler:
> note that the `tools` for lerna used to look for `PLUGINS.md` in its old spot. Might need to update that.

Thanks!

PR #32, opened 2020-04-08
groups can transition from updating to initializing

evantahler:
solves https://www.pivotaltracker.com/story/show/172216579


bleonard:
can this cancel the current run? (or does that already happen)

PR #41, opened 2020-04-13
Sailthru Plugin

bleonard:


plugins/@grouparoo/sailthru/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "node": "12.x.x"
+  },
+  "scripts": {
+    "prepare": "rm -rf dist && tsc --declaration",
+    "test": "jest",
+    "pretest": "npm run lint && npm run prepare",
+    "lint": "prettier --check src "
+  },
+  "dependencies": {
+    "mailchimp-api-v3": "^1.13.1",
```
evantahler: we probably don't need this one

plugins/@grouparoo/sailthru/src/lib/client.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export default class SailthruClient {
+  sailthru: any;
+
+  constructor(options: SimpleAppOptions) {
+    const apiKey = options.apiKey;
+    const apiSecret = options.apiSecret;
+
+    this.sailthru = Sailthru.createSailthruClient(apiKey, apiSecret);
+    //this.sailthru.enableLogging();
```
evantahler, line 12: Maybe this choice should be an app or connection option?  Folks might have API trouble in production?

plugins/@grouparoo/sailthru/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,217 @@
+// ~ jump over to core for this test ~
```
evantahler, line 1: I like this style of unit? testing a lot!

plugins/@grouparoo/sailthru/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,217 @@
...
+process.chdir(`${__dirname}/../../../../../core/api`);
```
evantahler, line 2: Since you aren't booting the app, I don't think you need to `chdir`

PR #47, opened 2020-04-14
only show ready profile property rules in the group rule builder

evantahler:
solves https://www.pivotaltracker.com/story/show/172315807


evantahler:
one more spot https://github.com/grouparoo/grouparoo/blob/master/core/api/__tests__/actions/groups.ts#L238

PR #49, opened 2020-04-14
Profile property rule filters + Plugin API Updates

evantahler:
Allows users to filter their data when creating Profile Property Rules for mySQL and Postgres tables.

Screen Shot 2020-04-14 at 4 55 31 PM

Also:
* Simplifies the method signatures of most plugin methods to provide a single argument hash with data, vs a lot of arguments (see `plugins.ts` for new signatures)
* Grouparoo exports plugin method signatures which will automatically include argument types & sub-method signatures

TODO:
- [x] filter tests for PG 
- [x] filter tests for MySQL 
- [x] api tests for filterOptions 
- [x] model tests for filterOptions

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# 🦘 Grouparoo
+# Grouparoo
 
-Achieve Marketing Autonomy
+**Achieve Marketing Autonomy**
 
 Want to move faster? Take control of your data and your destiny with the open source customer data platform that puts marketers in charge.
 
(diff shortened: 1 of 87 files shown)


evantahler:
cc @bleonard I like the clarity added in https://github.com/grouparoo/grouparoo/pull/49/commits/b6e60cfdec99e4dc4edb206b4d4c878543dffc83. Good suggestion!

PR #50, opened 2020-04-15
default scopes filter state=ready

evantahler:
Pros: 
* Objects will be "ready" by default; less `{where: {state: 'ready'}}`

Cons: 
* All the actions that might want to deal with not-yet-ready objects need to remove the default scope.  
* The list views should de-scope by default (so drafts can be shown), but lots of other pages use the `profilePropertyRules:list` action, so consumers still need the scope filter in the params.
* All the default `$get` methods use the default scope... so if you have a parent who isn't ready (ie: tests), things are going to get complicated

Misc:
* Added `findByGuid` to all models.  This helper should be used when finding one instance... it will ignore scopes and throw if it can't be found.  this helps mitigate Con 1.

### DIFF
--- a/core/api/src/classes/loggedModel.ts
+++ b/core/api/src/classes/loggedModel.ts
@@ -146,4 +146,24 @@ export abstract class LoggedModel<T> extends Model<T> {
   }
 
   abstract apiData(): Promise<{ [key: string]: any }>;
+
+  /**
+   * Find an instance of this class, regardless of scope
+   */
+  static async findByGuid(guid: string): Promise<any> {
+    // static class definitions or type defining are not yet available in TS.  See:
+    // * https://github.com/microsoft/TypeScript/issues/14600
+    // * https://github.com/microsoft/TypeScript/issues/34516
+    // * https://github.com/microsoft/TypeScript/issues/33892
+
+    // So, each model will implement this method
+
+    throw new Error("not implemented");
+
+    // const instance: T = await this.scope(null).findOne({ where: { guid } });
+    // if (!instance) {
+    //   throw new Error(`cannot find ${this.name} ${guid}`);
+    // }
+    // return instance;
+  }
 }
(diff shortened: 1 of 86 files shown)


core/api/src/models/Group.ts
```diff
@@ -62,6 +63,9 @@ const STATE_TRANSITIONS = [
   { from: "ready", to: "deleted", checks: [] },
 ];
 
+@DefaultScope(() => ({
+  where: { state: "ready" },
```
bleonard: I think this will drop "initializing" and "updating" --- that's probably not the intent.

evantahler: Yep! once the tests are green, I'm going to investigate compound or ~= scopes for this one

core/api/src/classes/loggedModel.ts
```diff
@@ -146,4 +146,24 @@ export abstract class LoggedModel<T> extends Model<T> {
...
+    // So, each model will implement this method
+
+    throw new Error("not implemented");
+
+    // const instance: T = await this.scope(null).findOne({ where: { guid } });
+    // if (!instance) {
+    //   throw new Error(`cannot find ${this.name} ${guid}`);
+    // }
+    // return instance;
+  }
```
evantahler, line 168: If we can define this here we won't need to on every model... I don't think it's possible though at the moment in TS.

bleonard: what! boo.
I assume the issue is knowing that it's `Group.scope...` vs. `Destination.scope...`
What if you made this method do
```
const klass = this.klass // or this.klass();
const instance: T = await klass.scope(null).findOne({ where: { guid } });
if (!instance) {
  throw new Error(`cannot find ${klass.name} ${guid}`);
}
return instance;
```
And then implemented `this.klass = ` or `function klass()` in the children. I bet that would pay off on the next thing somewhat like this.

bleonard: Related: when I make the Group class it does this: `export class Group extends LoggedModel<Group>`
Doesn't the code have access to that `<Group>` thing somewhere?

evantahler: You would think so...
Screen Shot 2020-04-15 at 3 46 51 PM

PR #53, opened 2020-04-16
Set NULL to the value when there are no rows.

bleonard:
I came across a table that 1 to 1 with my users table but sparsely populated. That is, it didn't have all the user ids in it. So I needed an `exact` query, but it crashed when there was no row for that user.
I'm not exactly sure the right thing to do in this spot, but I think it's something like this to make that value `NULL`.

For example users table has 1000 rows.
User_demographics table has 300 with a unique indexed `user_id` in it for the users that have filled out their info. The property was using `user_demographics.income` or whatever. So now 300 will have their income and 700 will have `NULL` in the property.

### DIFF
--- a/plugins/@grouparoo/mysql/src/lib/table-import/profileProperty.ts
+++ b/plugins/@grouparoo/mysql/src/lib/table-import/profileProperty.ts
@@ -66,13 +66,11 @@ export async function profileProperty(
   const client = await connect(appOptions);
   try {
     const rows = await client.asyncQuery(parameterizedQuery);
-    if (rows) {
-      row = rows[0];
-      for (const remoteKey in sourceMapping) {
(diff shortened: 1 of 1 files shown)


evantahler:
This would be a possible solution to https://www.pivotaltracker.com/story/show/172320909.

In general, I think that I would rather move the handling of the `null` case to the Source vs each plugin.  I'd rather the plugin return `null` then `{}`.  To me that `null` indicates a row can't be found whereas `{}`
 indicates a row was found, but it has no content. 

Also, we should make a similar change to the mysql-query import, and the PG source(s)

evantahler:
will be covered by https://github.com/grouparoo/grouparoo/pull/49/commits/7106b8cb638aef2d90ef054dd9abdfd35537d7ba within https://github.com/grouparoo/grouparoo/pull/49

PR #54, opened 2020-04-16
Idea for resiliency

bleonard:
This might help being resilient to the key name changing while things are in flight.


evantahler:
Let's talk about this one!   Mappings are [stored by guid rather than key](https://github.com/grouparoo/grouparoo/blob/180d43e54cd89e1a58c72a2438068b49b41573b7/core/api/src/models/Mapping.ts#L34-L37) and this method exists to return a convenient hash.  At some point you'll need to change guids to names before communicating with the destination...

bleonard:
Can do later.

PR #55, opened 2020-04-16
Styling changes to cards

andyjih:
@evantahler I have a couple issues I'd like some help on, when you get a chance. 😅 

1. Changing the card color
2. Centering the image on the card
Screen Shot 2020-04-16 at 11 44 35 AM

### DIFF
--- a/core/web/components/forms/app/add.tsx
+++ b/core/web/components/forms/app/add.tsx
@@ -37,13 +37,14 @@ export default function ({ apiVersion, errorHandler }) {
     <>
       <Form id="form" onSubmit={create}>
         <CardDeck>
-          {types.map((_app) => {
+          {types.map((_app, idx) => {
+            const className =
+              _app.name === app.type ? "button-color-selected" : "button-color";
             return (
               <div
                 style={{
                   width: "150px",
                   borderRadius: "4px",
-                  backgroundColor: "#242436",
                   borderColor: "gray",
                   color: "white",
                   margin: "10px",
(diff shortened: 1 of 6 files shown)


andyjih:
Designs for reference: 
Screen Shot 2020-04-16 at 11 47 43 AM

core/web/components/forms/app/add.tsx
```diff
@@ -1,6 +1,6 @@
 import { useState, useEffect } from "react";
 import { useApi } from "../../../hooks/useApi";
-import { Form, Button, Row, Col, Card } from "react-bootstrap";
+import { Form, Button, Row, Col, Card, CardColumns } from "react-bootstrap";
```
evantahler: I had no idea about `CardColumns`! Nice find!

evantahler: I think this means you shouldn't use `Row` and `Col` within the `CardColumns` section... it handles that all for you.

evantahler:
As far as colors go, without coming up with something custom, we've got the normal bootstrap color "variants": 
```
[
  'Primary',
  'Secondary',
  'Success',
  'Danger',
  'Warning',
  'Info',
  'Light',
  'Dark',
]
```
You can see them applied to cards here: https://react-bootstrap.github.io/components/cards/#background-color

core/web/components/forms/app/add.tsx
evantahler: For centering, it might be easiest to use text-align rather then messing with the Float properties.  Perhaps on the body: `<Card.Body style={{textAlign: 'center'}}>`

andyjih:
@evantahler Made a new react component called "Selector" I'll update the Source and Destination creation pages next, but wanted to make sure this was directionally how you think it should work.

andyjih:
Screen Shot 2020-04-21 at 10 42 56 AM

andyjih:
Screen Shot 2020-04-21 at 11 33 05 AM

PR #66, opened 2020-04-21
fixing routing after making a new source and also filtering out draft profile property rules

andyjih:


evantahler:
Nice @andyjih! I made one change to request only `ready` profile property rules from the API, rather than filtering them on the front-end. (https://github.com/grouparoo/grouparoo/pull/66/commits/89e00c427224312cb6888f640726e5e40fd78384)

As a note, *most* APIs now will exclude "draft"-state objects, except for the `list` ones, as we'll want to show a list of draft ones on a page at some point (or include them in an overall list).  But, you can pass a `state` filter in your API request.

evantahler (approved):

PR #70, opened 2020-04-23
simpler exports and conditional includes on destination#apiData

evantahler:


core/api/src/models/Destination.ts
```diff
@@ -219,25 +219,34 @@ export class Destination extends LoggedModel<Destination> {
     });
   }
 
-  async apiData() {
-    const app = await this.$get("app");
+  async apiData(includeApp = true, includeGroups = true) {
```
bleonard, line 222: This pattern might get complicated over time with lots of booleans in hard-to-know order.
Maybe pass `options` instead: `async apiData({ includeApp? : boolean, includeGroups? : boolean} )`

PR #76, opened 2020-04-28
Update Package workflow

evantahler:
* relax peer deps
* use "latest" in examples
* top-level project has no version
* remove tools/lerna

---

Some notes about the relaxing of peer deps: https://github.com/lerna/lerna/issues/1575


.github/workflows/bump-internal-repo.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@master
         with:
           repository: grouparoo/internal
-          token: ${{ secrets.GITHUB_ACCESS_TOKEN }}
```
evantahler, line 15: it now seems that secrets in github actions cannot contain `github` (the string)

documents/deployment/getting-started.md
```diff
@@ -9,18 +9,18 @@
...
-    "@grouparoo/core": "0.1.1-alpha.14",
-    "@grouparoo/mysql": "0.1.1-alpha.14",
-    "@grouparoo/postgres": "0.1.1-alpha.14",
-    "@grouparoo/mailchimp": "0.1.1-alpha.14",
-    "@grouparoo/csv": "0.1.1-alpha.14"
+    "@grouparoo/core": "latest",
+    "@grouparoo/mysql": "latest",
+    "@grouparoo/postgres": "latest",
+    "@grouparoo/mailchimp": "latest",
+    "@grouparoo/csv": "latest"
```
evantahler, line 23: the version `latest` should be OK, as the version installed by users will be locked in place by `package-json.lock` (or yarn.lock).

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "author": "Grouparoo Inc <hello@grouparoo.com>",
   "name": "@grouparoo/grouparoo",
-  "version": "0.1.1-alpha.14",
+  "version": "0.0.0",
```
evantahler, line 4: This package is never published, (private=true) so the version doesn't matter

PR #92, opened 2020-05-01
Event Basics

evantahler:
* Add events documentation
* Event initializer and prototype classes 
* Event base APIs
* SequelizeEventBackend


documents/development/events.md
```diff
@@ -0,0 +1,108 @@
...
+- _userId_ - Set by the client, optional. If the client knows the primary identifying profile property of the current user, it should be sent. This is /your/ notion of userId. This property takes precedence over `anonymousId` when present.
+- _anonymousId_ - Set by the client, required (unless `userId` is present). If the `userId` is not known (the user is not yet logged in), a unique ID should be generated by the client. These can be device GUIDs in the case of mobile phones, or just another UUID.
+
+### `eventData` table
+
+This is a “has-many” data table logging the data properties of each event. Each event can have 0 or many rows in this table.
+
+- _guid_ - Set by the Grouparoo server. The primary Key for this portion of an event’s data. Generated by the Grouparoo server. Should /not/ be auto-incremented but others unique (we use UUIDv4) + a `edt_` prefix.
+- _key_- Set by the client, required. The named key of this portion of an event’s data.
+- _value_- Set by the client, required. The value of this portion of an event’s data. All data is stored as strings.
```
bleonard, line 41: want to `rawValue` here like in profile properties?

documents/development/events.md
```diff
@@ -0,0 +1,108 @@
...
+- _type_ - Set by the client, required. The named type of the event. Examples include `pageview`, `item-added-to-cart`, `purchase`, or `click`
+- _userId_ - Set by the client, optional. If the client knows the primary identifying profile property of the current user, it should be sent. This is /your/ notion of userId. This property takes precedence over `anonymousId` when present.
+- _anonymousId_ - Set by the client, required (unless `userId` is present). If the `userId` is not known (the user is not yet logged in), a unique ID should be generated by the client. These can be device GUIDs in the case of mobile phones, or just another UUID.
+
+### `eventData` table
+
+This is a “has-many” data table logging the data properties of each event. Each event can have 0 or many rows in this table.
+
+- _guid_ - Set by the Grouparoo server. The primary Key for this portion of an event’s data. Generated by the Grouparoo server. Should /not/ be auto-incremented but others unique (we use UUIDv4) + a `edt_` prefix.
```
bleonard, line 38: missing `eventGuid` to point to parent event.

documents/development/events.md
```diff
@@ -0,0 +1,108 @@
...
+- _updatedAt_ - The time the Grouparoo Server last modified the event. Set by the grouparoo server.
+- _occurredAt_ - Set by the client, optional. This likely differs from `createdAt` as it should store the local time of the event. Event’s may be batched when sent from the client, so we cannot rely on the time server received the event to be the time the event occurred. If there is no `occurredAt`, the server will apply `createdAt`
+- _type_ - Set by the client, required. The named type of the event. Examples include `pageview`, `item-added-to-cart`, `purchase`, or `click`
+- _userId_ - Set by the client, optional. If the client knows the primary identifying profile property of the current user, it should be sent. This is /your/ notion of userId. This property takes precedence over `anonymousId` when present.
+- _anonymousId_ - Set by the client, required (unless `userId` is present). If the `userId` is not known (the user is not yet logged in), a unique ID should be generated by the client. These can be device GUIDs in the case of mobile phones, or just another UUID.
+
+### `eventData` table
```
bleonard, line 34: should it have `createdAt/updatedAt` ?

documents/development/events.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+Grouparoo will create a new `Profile` to track all the activity we can for “person_1”, and attempt to decorate their profile with any additional information we can load from your `Profile Property Rules`.
+
+Eventually, after visiting a few more pages, the user will sign in, and we’ll know their `userId`
+
+```js
+Gropuaroo.track({
+  userId: 123,
+  occurredAt: new Date(),
```
bleonard, line 81: If the library doesn't handle silently sending the anonyous id with every request, they'd need to send it here to do the mapping. I think it's ok to always be sending it and we have the user `userId` take precedence.

In the [segment api](https://segment.com/docs/connections/spec/identify/) this is actually an `identify` call, but is to the same endpoint.

documents/development/events.md
```diff
@@ -0,0 +1,108 @@
...
+    // ...
+  },
+});
+```
+
+From this point forward, we don’t need to send `anonymousId`. On the backend, Grouparoo will now be able to check if user 123 already has a `Profile`, and merge it with the profile we created for “person_1”. In this way, we will have a combined history for all the sessions of this user. If the same user signs in via another client / browser / phone, we will continue to create temporary profiles with an `anonymousId`, and eventually merge the profiles (and events!) once we have enough data.
+
+## Creating events outside of Grouparoo
+
+Grouparoo will not only process events created though our APIs, but will also periodically scan the events table for new events that have been added by an external system. This is a helpful if you want to write events directly from your backend applications. Grouparoo will look for events lacking a `profileGuid` and associate the events, making new `Profile`s as needed.
```
bleonard: It would be good to note that the schema should match and that we will write back the `profileGuid`.

documents/development/events.md
```diff
@@ -0,0 +1,108 @@
...
+
+By default, Grouparoo will check for new, un processed events each minute.
+
+## Apps & Sources
+
+Grouparoo comes with a built-in `app` to use with events. We ask that you provide the `Profile Property Rule` that matches `userId`s sent from events so we can identify your users.
+
+Once you have events flowing into Grouparoo, you can create `Sources` again an event `type`. You can look for counts, sums, diffs, latest values, etc between events to create `Profile Properties` to use within the rest of Grouparoo.
```
bleonard: "against" :typo:

bleonard (commented):
Looks great.

documents/development/events.md
evantahler: I don't think so.  The reason it works with Profile Properties is we know how to cast back because of the Profile Property Rule.  We'd need something like that for each possible eventData#key to cast property.  I think I'd rather move the casting step to the source, and cast based on the typeof the profile property rule you are trying to make

documents/development/events.md
evantahler: 👍

documents/development/events.md
evantahler: I think that's up to the backend.  Grouparoo core probably won't do anything with it

documents/development/events.md
evantahler: But... if `userId` takes precedence... why require it?

bleonard: For a while, the user is `anonymousId: person-1` and then they login and now are `userId: 10`
If no event ever has both a `userId` and `anonymousId`, then Grouparoo doesn't know that `person-1` and `10` are the same profile. The `10` just looks like a completely separate logged in user to the server, so the `person-1` and `10` data wouldn't be merged.

documents/development/events.md
bleonard: I'm not sure of the answer here. I agree, but it would be nice to know the type in the source somehow to make profile property creation easier. Probably not worth it so they just have to pick the type when making the rule as you said.

PR #93, opened 2020-05-02
Simplify plugin & test Model imports

evantahler:
This PR allows plugins to import models directly from `@grouparoo/core` without worrying about `dist` or `src`!  Similarly, this PR ensures compatibility between all models regardless of running within Typescript or JS.  This means that the types exported by the SpecHelper and direct import will no longer conflict: 

```ts

import { Profile } from `@grouparoo/core` // <-- just use the package directly
import { helper } from "@grouparoo/core/tests" // <-- I'm a new import helper
let profile: Profile

// later in the test
profile = await helper.factories.profile() // OK NOW! 
```

1) Any plugin which might access the models from `@grouparoo/core` (either in test or directly) needs to add a start method to it's initializer, `plugin.mountModels()`. This method checks each model in *this* export context (likely JS), and ensures that it has been attached to `api.sequelize` so it knows how to talk to the database.  `api.sequelize` is equally likely built in the TS context (dev mode or test) or JS (npm start).  Either way, this method will make sure things are OK.

2) We can no longer use `private` methods on our Models.  From `documents/databases.md`:

> ## Grouparoo Model Guidelines
> 
> - No private methods. This breaks compatibility between Typescript-exported model definitions (to JS) and the TS files themselves. We require this compatibility as plugins both source and export to core... so a single runtime's require path may be `core/dist -> plugin/src` or even `core/dist -> plugin/dist -> core/src`
>  - Instead, Pretend it's 1999 and prefix methods you want to be private with `_`, i.e. `async _privateMethod()`
> - New models need to be enumerated a few places:
>   - `Index.ts`
>   - `SpecHelper.ts`
>  - `modules/plugin.ts`

3) `@grouparoo/core` now exports the specHelper too so factories and setup for tests should be easy to use: `import { helper } from "@grouparoo/core/tests"`
---

goal: simplify how Models are imported by plugins. 
problems: `Model not initialized: Member "findOne" cannot be called. "ProfilePropertyRule" needs to be added to a Sequelize instance`

This problem happens when the plugin uses the `dist` but we are running the `src`.  There are hacks (like `plugin.ensureModelsMounted()`) but needing to call that within every plugin is annoying.

Ideas: 
* Can the spec helper do something like `plugin.ensureModelsMounted()`?


evantahler:
https://github.com/grouparoo/grouparoo/pull/93/commits/f9f4615fc939c8bb2a05d75a771737ffefa11b0f demonstrates something interesting: 

Even though this branch has `@gropuaroo/core` exporting the models, which works great for the 'normal' plugin files, the integration tests still need to explicitly load the models from `src` - this is because our specHelper, factories, etc from core all source `src`, as the expect to be used within core's test suite.

PR #100, opened 2020-05-04
API Keys and new Permission System

evantahler:
This PR upgrades Grouparoo's permission system.  Closes https://www.pivotaltracker.com/story/show/172570932

Actions now extend `AuthenticatedAction` which allows each action to specify their topic and if they require `read` or `write` access on that topic. 

```ts
export class AppCreate extends AuthenticatedAction {
  constructor() {
    super();
    this.name = "app:create";
    this.description = "create an app";
    this.outputExample = {};
    this.permission = {topic: "app", mode: 'write'}; // <--- Here
    this.inputs = {
      name: { required: false },
      type: { required: true },
      state: { required: false },
      options: { required: false },
    };
  }

  async run({ params, response }) {
   // ...
  }
}
```

Under the hood, `AuthenticatedAction` uses a new middleware, `authenticated-action` which replaces the previous `role-*` and `team-member` middleware. Depending on if the request to the action included an `apiKey` or not, the `authenticated-action` middleware will go down one of two paths to attempt to authenticate the user - session cookie + csrf token or API key. 

A new generic `Permission` model is added which can be used by both `Team`s and `ApiKey`s.  Both can enumerate their topics and opt in and out of read and write for each.

The list of all topics is held by `Permission.topics()` and we validate each action against that list.

* Teams now use per-topic permissions (read/write)
  * permissions can be `locked`. this means that they cannot be changed.  This is helpful for the `deletable=false` default admin team so you can't lock yourself out of the tool.
* Api Keys exist!
  * API Keys use the same per-topic permission model as teams

Notes
* A few Actions don't use `AuthenticatedAction`, like signIn and status, as they are not authenticated actions. 
* Even a few more actions do require a logged-in-user, but should never 
disable a user from accessing them, regardless of permission set (ie: `account:edit`).  These actions set their `permissionTopic = '*'`, which will still require an authenticated team member, but not check permissions for that topic. 

TODO: 
- [x] UI for team permissions
- [x] API Key actions
- [x] API Key UI
- [x] Tests for new permission system
- [x] Tests for APIKey models
- [x] Tests for APIKey actions


evantahler:
Team Permissions Table replaces the old global `read`, `write` and `administer` settings

Screen Shot 2020-05-05 at 10 57 34 AM

evantahler:
Permission system is re-usable for `APIkeys`

Screen Shot 2020-05-05 at 12 18 42 PM
Screen Shot 2020-05-05 at 12 18 51 PM

PR #106, opened 2020-05-07
Swagger UI

evantahler:
Now that we have API keys and global way to authorize API Requests... lets easily expose our API! 

All of the needed data can be sourced from `package.json` or the action definitions. 

Screen Shot 2020-05-06 at 5 53 23 PM


evantahler:
There is one outstanding issue with the swagger form reading user input.  see https://github.com/swagger-api/swagger-ui/issues/5990

PR #111, opened 2020-05-07
Big query Source

bleonard:
- Query Source
- Table Source


plugins/@grouparoo/bigquery/.npmignore
```diff
@@ -0,0 +1,10 @@
+# General
```
evantahler, line 1: Normally, npm will ignore whatever .gitignore ignores.  The existence of an npm ignore file tells npm to use this instead of the .gitignore.  I think it will be better to only have one ignore file for the project (.gitignore).  What's different here than in the .gitigonore

plugins/@grouparoo/bigquery/__tests__/.env.example
```diff
@@ -0,0 +1,5 @@
+# Note: These are only needed when recording new nock fixtures
```
evantahler, line 1: Just checking that none if this data is secret... looks like a real key?

plugins/@grouparoo/bigquery/src/lib/validateQuery.ts
```diff
@@ -0,0 +1,24 @@
+export function validateQuery(sql: string) {
+  const lowerCaseSQL = sql.toLowerCase();
+
+  if (lowerCaseSQL.indexOf(";") >= 0) {
+    throw new Error("only provide a single query");
+  }
+
+  if (lowerCaseSQL.indexOf("--") >= 0) {
```
evantahler, line 8: Might also want to check for "#" https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#comments

evantahler (commented):
Woo! 
... and that is some very through testing!

Just some minor comments.

plugins/@grouparoo/bigquery/.npmignore
bleonard: I didn't even know this was here. it's just from copying the postgres plugin. looks like most plugins have an `.npmignore` - should we make a story to standardize having them or not or something?

plugins/@grouparoo/bigquery/__tests__/.env.example
bleonard: it's fake!

plugins/@grouparoo/bigquery/.npmignore
evantahler: OH! forget everything I said here - we *do* want NPM to think differently than git!  We want git to ignore `/dist` but we do want that published to NPM.  It's all good!

Sorry!

PR #112, opened 2020-05-08
disable schedule options when schedule state is ready

evantahler:
Screen Shot 2020-05-07 at 7 32 28 PM


evantahler:
fixes https://www.pivotaltracker.com/story/show/171580415

PR #132, opened 2020-05-12
Event Sources & Profile Property Rules

evantahler:
## WIP Source:
Screen Shot 2020-05-12 at 1 21 51 PM
Screen Shot 2020-05-12 at 1 21 54 PM

## WIP Profile Property Rule
Screen Shot 2020-05-12 at 1 20 51 PM

## Notes (compared to [Figma Design](https://www.figma.com/file/9cLCZV4sAMLZADjzSUNd0g/Wireframes?node-id=1068%3A8))

**Source**

* Adding a schedule on an event source isn't Necessary. We'll be re-importing the profile & groups at the association of the  Profile.
* There's a new property of a plugin source connection - `skipSourceMapping`.  This is when you are able to provide a source preview, but somehow the source knows already how find out which Profile each event is about.  
  * Side effect - you cannot "Bootstrap" your server with an Event source.  This makes sense, as you'll already need a unique profile property configured from another source to se up the Event App.

**Profile Property Rule**

* Since the Source already chose the event type, this is not needed at the profile propety
* At the profile property, exact doesn't really make sense without choosing "oldest" or "newest". To reflect this more strongly, I've moved to `exact-most-recent` and `exact-least-recent`, which removes the need for the tertiary selection
* The "real" columns of the event are rather sparse, and most of them don't make sense to expose in this view (userId, profileAssociatedAt, etc).  With that in mind, we can pre-expand the keys of "data", to have a unified "column" selection.


evantahler:
TODO: use the font-awesome icon and replace the icon for the Event app

PR #139, opened 2020-05-13
Do not hoist dependencies; have lockfiles for each package

evantahler:
Closes https://www.pivotaltracker.com/story/show/172822053

Pros:
* Lockfiles! We'll actually have consistent versions of our packages
* Rely on simpler plumbing and install process that comes with Lerna (removes custom build script)

Cons:
* Doubles the time on CI to `npm install` and set up the project
* ~4x (80MB to 400MB) the size of the package cache, as there are now many redundant modules


evantahler:
Without a hoisted `actionhero`, this requires a truly "global" actionhero/api.  Waiting on https://github.com/actionhero/actionhero/pull/1459

evantahler:
Now that the install step is so slow on CI, we can do some fancier caching, per https://github.com/circleci/circleci-docs/issues/3933

PR #158, opened 2020-05-19
Combine core tests and plugin tests to have simpler & faster CI

evantahler:
Should be simpler and faster!

Screen Shot 2020-05-19 at 11 09 33 AM

For plugins, we can rely on lerna to run `npm test` in each package for us:
```
lerna run test --stream --concurrency 1 --ignore "@grouparoo/core"
```

---

This PR also simplifies top-level `npm test` which will test all the packages in this monorepo

Screen Shot 2020-05-19 at 11 48 37 AM


evantahler:
closes https://github.com/grouparoo/grouparoo/pull/157

PR #161, opened 2020-05-20
Events become a regular Grouparoo Model

evantahler:
This PR greatly simplifies the event system.  
* Removes `eventBackend`s
* Creates models for `Event` and `EventData`
  * Moves complex logic to model hooks
* Creates migrations for `Event` and `EventData` tables
* Adds model tests for `Events`
* Show event totals on Dashboard


evantahler:
When this PR is merged in, if you will need to either:
- drop your `events` and `eventData` tables and let the app re-create them from the migrations 
- add `000029-createEvents.js` to your `SequelizeMeta` table, to pretend the migration already happened

PR #186, opened 2020-05-29
Type fixes

evantahler:
Fixes various bugs that appear in client installed apps with a newer/stricter version of typescript


evantahler:
Then this branch is merged to master, a new prerelease release should be issued.

PR #191, opened 2020-05-31
Throw clear error if SERVER_TOKEN is not set & remove default value

evantahler:
SERVER_TOKEN needs to be set both at run time and build time (for next build).  This PR adds clear warnings if `process.env.SERVER_TOKEN` is missing.


evantahler:
This PR required that `SERVER_TOKEN` be set on CI as well.

PR #208, opened 2020-06-04
Web Client Library

### LINKED ISSUE #227: Make an example Client lib that acts like we think segment does
And the `identify` call too (track the event, but don't update the profile properties?)

evantahler:
Closes #227 

--- 
# @grouparoo/client-web

Grouparoo provides a javascript library for tracking events from browsers and sending them to your Grouparoo server.

## Install

You can include this library in a few ways:

**Javascript / Typescript Module**

For use in environments that will be compiled, ie: React or Webpack.

```ts
import { GrouparooWebClient } from "@grouparoo/client-web";

const client = new GrouparooWebClient({
  host: "https://grouparoo.my-company.com",
  apiKey: "abc123zzz",
});
```

**Script Tag**

```html
<html>
  <head>
    <script
      type="application/javascript"
      src="https://grouparoo.my-company.com/public/javascript/grouparooWebClient.js"
    ></script>
  </head>

  <script>
    const client = new GrouparooWebClient({
      host: "https://grouparoo.my-company.com",
      apiKey: "abc123zzz",
    });
  </script>
</html>
```

To initialize a client (per the above examples), you'll need to provide:

- **host** (required): The publicly-available URL of your Grouparoo server/cluster.
- **apiKey** (required): An API Key that you have generated in your Grouparoo server, with `write` access to the `events` topic.
- **userId** (optional): A unique and permanent Id for this logged-in use (if this user is signed in). Otherwise, you can use the `identify` method later.
- **path** (optional): The route to post events too. Default: `/api/v1/track`
- **errorHandlers** (optional): An array of functions to pass errors too. This is helpful if you want to display tracking errors to the user, or log them with a tool like New Relic.
- **logging** (optional): Should tracking events be logged to browser console? Default: `false`
- **userIdCookie** (optional): What is the name of the cookie to store the `userId`? Default: `grouparoo-user-id`
- **anonymousId** (optional): A unique ID you would like to use for the `anonymousId` for this client. By default, we will generate a unique UUID for this browser and store it in a cookie for future use. However, if your application already has the notion of a unique ID for every visitor, you can apply it here.
- **anonymousIdCookie** (optional): What is the name of the cookie to store the `anonymousId`? Default: `grouparoo-anonymous-id`
- **anonymousIdCookieTTL** (optional): How long should the tracking cookie exist for? Set in ms. Default: `(1000 * 60 * 60 * 24 * 90)`, i.e: 90 days.
- **sessionId** (optional): A unique ID you would like to use to identify the session for this client. By default, we will generate a unique UUID for this session and store it in a cookie for future use. However, if your application already has the notion of a unique session ID for every visitor, you can apply it here.
- **sessionIdCookie** (optional): What is the name of the cookie to store the `sessionId`? Default: `grouparoo-session-id`
- **sessionIdCookieTTL** (optional): How long should the session cookie exist for? Set in ms. Default: `(1000 * 60 * 60)`, i.e: 1 hour.
- **retrySleep** (optional): How long to sleep, in ms, between http errors before trying to send again. Default: `(1000 * 3)`, i.e.: 3 seconds.
- **maxErrorCount** (optional): We will stop trying to send events after seeing this many HTTP errors. Default: 10.

## Creating Events

**Track**

To create an event, call the `track` method on the client with a `type` and any additional data you want to store about the event:

```ts
// track an arbitrary event
// API: client.track(event-name, data?)
client.track("button-click", { page: "/home", button: "learn-more" });
```

This will send the event, time, and associated user information (`anonymousId` and `userId` if set) to your Grouparoo server. We will also always include the values for `anonymousId` and `sessionId` with every `track` call.

**Identify**

A special event called `identify` should be called whenever you know a way to de-anonymize this user. This usually corresponds with a `sign-in` event. The `identify` event requires your application's unique id for this user set as the `userId` argument.

```ts
// The user successfully signed in as user #123
// API: client.identify(userId, data?)
client.identify(123, { firstName: "Mario", lastName: "Mario" });
client.track("sign-in", { page: "/sign-in" });
```

From this point onward, all events tracked will be linked to user `#123`. Grouparoo will also associate all previous events from this session (with the same `anonymousId`) to user `#123`.

The second argument to `identify`, `data`, will _not_ override profile data that Grouparoo has obtained from your Sources. However, it may be useful to record related identifying information about the profile at the time of `identify`. This information will be stored as data on the Event, and able to be later utilized within Grouparoo.

**Reset**

The `reset` method will delete all data stored, and reset the `userId` and `anonymousId`. Use this method when a user logs out.

```ts
// The user has logged out
// API: client.reset(newAnonymousId?)
client.reset();
```

## Higher-Order Events

This library also comes with a few higher-order events geared towards making it easy to track the performance of your websites.

**Page**

The `page` method will create a track event pre-filled with some useful data about the page. The `page` method will also append all query string variables. This will ensure that data about your email, social media, and ad campaigns is properly associated to the right visitors.

```ts
// track a page
// API: client.page(category?, title?, data?)
client.page("help-pages");

// which is equivalent to
client.track("page", {
  category: "help-pages", // from "page"
  title: window.document.title, // from "title"
  url: window.location.href,
  path: window.location.pathname,
  referrer: window.document.referrer,
  // ...and data from the URL Query String if present
  utm_source: "adwords",
  utm_medium: "cpc",
  utm_campaign: "fall-campaign-mushroom-keywords",
});
```

### Notes on building this package

We build both a `dist/grouparooWebClient.js`, which is to be used in the browser, and `dist/index.js` which is to be consumed by other node or webpack projects which will then be transpiled. The code within both libraries is exactly the same, except for the method of bundling and exporting the `Client` class.


evantahler:
Intersting Learnings:
1. To really be able to have this be usable in the browser directly (from a script tag):
   * no external modules can be used
   * we need to use webpack & typescript to bundle everything up in to a single file
2. The implementation of UUID isn't /technically/ up to spec, but should be OK for our use cases.  The ES6 version is better than the ES4 version.
3. At the moment, this library relies on `fetch`.  We may need to also include an XHR version for older browsers?
4. At the moment, this tracking library is implemented entirely in JS, and there is no fallback to image-based tracking for those users who have disabled javascript.

TODO: 
* Tests!
* Demo.

clients/@grouparoo/web/README.md
```diff
@@ -0,0 +1,121 @@
...
+  <script>
+    const client = new GrouparooWebClient({
+      host: "https://grouparoo.my-company.com",
+      apiKey: "abc123zzz",
+    });
+  </script>
+</html>
+```
+
+To initialize a client, you'll need to provide:
```
bleonard: probably need an example of all of these jsut to be clear what's expected.

clients/@grouparoo/web/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+- **host** (required): The publicly-available URL of your Grouparoo server/cluster.
+- **apiKey** (required): An API Key that you have generated in your Grouparoo server, with `write` access to the `events` topic.
+- **path** (optional): The route to post events too. Default: `/api/v1/track`
+- **anonymousId** (optional): An ID you would like to use for the `anonymousId` for this client. By default, we will generate a unique UUID for this browser and store it in a cookie for future use.
```
bleonard: is this a boolean? I can't figure out how it would make sense to give an actual value, but the words suggest you could pass an ID.

bleonard: maybe move this one down in the doc to buy the cookie and cookieTTL as they are all the same. I wonder if it should be one setting actually
```
anonymousId: false
```
or 
```
anonymousId: true
```
or
```
anonymousId: { cookie: "grouparoo-anonymous-id", ttl: 90*24*60*60 }
```

clients/@grouparoo/web/README.md
```diff
@@ -0,0 +1,121 @@
...
+- **errorHandlers** (optional): An array of functions to pass errors too. This is helpful if you want to display tracking errors to the user, or log them with a tool like New Relic.
+- **logging** (optional): Should tracking events be logged to browser console? Default: `false`
+- **cookie** (optional): What is the name of the cookie to store the `anonymousId`? Default: `grouparoo-anonymous-id`
+- **cookieTTL** (optional): How long should the tracking cookie exist for? Default: 90 days.
```
bleonard: unclear if this is seconds or milliseconds, or other.

clients/@grouparoo/web/README.md
```diff
@@ -0,0 +1,121 @@
...
+client.page("help-pages");
+
+// which is equivalent to
+client.track("page", {
+  category: "help-pages", // from "page"
+  title: window.document.title, // from "title"
+  url: window.location.href,
+  path: window.location.pathname,
+  referrer: window.document.referrer,
+  // ...and data from the URL Query String
```
bleonard: "if present"

clients/@grouparoo/web/README.md
```diff
@@ -0,0 +1,121 @@
...
+  UTM_SOURCE: "adwords",
```
bleonard: funny that these are uppercase.

clients/@grouparoo/web/README.md
```diff
@@ -0,0 +1,121 @@
...
+client.reset();
+```
+
+## Higher-Order Events
+
+This library also comes with a few higher-order events geared towards making it easy to track the performance of your websites.
+
+**Page**
+
+The `page` method will create a track event pre-filled with some useful data about the page. The `page` event will also create a `new_session` event for the first time that an `anonymousId` has viewed any page.
```
bleonard: I wonder if `new_session` should be a property on the event and not a whole other event.

bleonard: I also wonder if segment or others has a "sessionId" on every event allowing aggregating by session in an interesting way.

bleonard (commented):
comments on readme.

clients/@grouparoo/web/README.md
evantahler: I'll clarify that's what's happening in the above examples

clients/@grouparoo/web/README.md
evantahler: Actually, they aren't the same:

Say your app has your own notion of `anonymousId`, (like TR) that might be hydrated into the page, from a cookie, in the URL (J_sessionId), or something else.  This allows you to pull that in directly and not use our random UUID thing.    Once you set it, the Grouparoo client will store that in our cookie, but it doesn't need to originate there.

... I'll add this all to the readme!

clients/@grouparoo/web/README.md
evantahler: AFAIK, this isn't something segment does.  This pattern I'm taking from Google Analytics... but it might be overkill.  We count unique anonymousId's just as easily as we can count `new_session` events.  We can also easily find the first events for any given anonymousId with some property.  I think I'll remove this.

evantahler:
closes https://github.com/grouparoo/grouparoo/pull/208

evantahler:
closes 227

evantahler:
Closes #227

evantahler:
This PR includes a static test page for events in the browser.  This page is linked to from APIkeys as a way to test them out.

Screen Shot 2020-06-10 at 12 06 15 PM
Screen Shot 2020-06-10 at 12 15 23 PM

PR #335, opened 2020-06-11
Plugin connection pool

### LINKED ISSUE #231: Plugin connections can be shared and reused (speed up all connections & sources)

### LINKED ISSUE #357: When making a new destination, go to the edit page, not the tab.
https://github.com/grouparoo/grouparoo/blob/master/core/web/pages/destination/new.tsx#L21-L24

evantahler:
Closes #231
Closes https://github.com/grouparoo/grouparoo/issues/357

For plugins with persistent connections (mySQL, Postgres) or those that need to obtain a long-lived token (BigQuery), this PR adds `connect` and `disconnect` methods to Plugin Apps.  

Grouparoo will try to connect if the plugin#app#connect method exists, and store the resulting object for later use.  The object is either the real MySQL client or an object that contains the keys you want to store.  This will then be passed back to all future methods called on the plugin.

This PR: 
- simplifies plugin logic to not need to connect and disconnect for every query
- speeds up queries as connections do not need to be established
- reduces the number of connections that Grouparoo makes to each source/destination as we are using a shared connection, and not creating a new one for each task/worker/"thread". 

The connection objects are stored per-server.  it's possible that not every server in your grouparoo cluster will connect to all sources/destinations.

### DIFF
--- a/core/api/__tests__/actions/schedules.ts
+++ b/core/api/__tests__/actions/schedules.ts
@@ -94,6 +94,7 @@ describe("actions/schedules", () => {
         expect(error).toBeUndefined();
         expect(schedule.guid).toBeTruthy();
         expect(schedule.name).toBe("new schedule name");
+        expect(schedule.state).toBe("draft");
       });
 
       test("an administrator can view a schedule", async () => {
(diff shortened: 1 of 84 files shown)


evantahler:
This PR also includes 2 other speed boosts (via https://github.com/grouparoo/grouparoo/pull/335/commits/34a01e4bc74382a5081cd489d71be41154a8d2cf):
* `promise.all` our way though sources and profile properties, rather than looping 1-by-1 (allows for more parallelism, and moves the blocker to the the sequelize connection pool)
* cache the results of `profilePropertyRule#validateOptions` in redis.  For every profile we import, for every profilePropertyRule we want to make sure that it is valid before trying to import... but we can cache that :D. It's a slower operation because it involves mustache and getting the full set of options from the source, which often involves pulling example rows to load columns etc. 

After this PR, on my localhost, importing a profile with 6 properties from 2 sources went from ~400ms to ~200ms (with the cache warm).  200% more better!

PR #381, opened 2020-06-19
Store previous page as global prop; never user Router.back

### LINKED ISSUE #349: After creating a new profile property, when I am taken back to the source overview, the list of profile properties is not updated.
Router.back() doesn't reload the page

evantahler:
closes https://github.com/grouparoo/grouparoo/issues/349

We now provide the `intitialPageProp` of `previousRoute` to all pages.  This contains the previous page's URL.  This should be used with `Router.push(previousPage)` vs `Router.back()`.  This will really re-hydrate pages and load new data, and won't mess with the back button's behavior.

### DIFF
--- a/core/web/pages/_app.tsx
+++ b/core/web/pages/_app.tsx
@@ -1,6 +1,6 @@
 import App from "next/app";
 import { useState, useEffect } from "react";
-import { useRouter } from "next/router";
+import Router, { useRouter } from "next/router";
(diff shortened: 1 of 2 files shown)


evantahler:
@bleonard just a note - I experimented with requiring a shared page under 2 routes, but that caused a few issues (tab behavior, [guid] query params changed, etc).

PR #399, opened 2020-06-23
Documentation updates from Edmundo's review

evantahler:
* move topology to own page
* clarify "user" vs "developer" docs
* add Heroku deploy button


evantahler:
Thanks again @edmundito!

PR #403, opened 2020-06-24
Custom matchers for Group Rules and Clearer choices for date ranges

### LINKED ISSUE #396: Use different words when adding a Group Rule for a date or timestamp
When adding a rule that we know is a timestamp or date, let's use different operators/comparison:

- Is in the past [Number] [Days/weeks/months/years]
- Is in the future [Number] [Days/weeks/months/years]
- Before (date & timestamp)
- After (date & timestamp)
- Current [Day/week/month/year]
- On (date & timestamp)
- Between (two dates and timestamps)
- Does not exist (is null)
- Exists (is not null)

evantahler:
closes https://github.com/grouparoo/grouparoo/issues/396

* Removes the `ProfilePropertyRuleOps._dictionary` in favor of a more verbose, but flexible collection of `op` + `description` pairs.  This allows the same SQL operation (ie: `gt`) to have different descriptions per Profile Property Rule type (ie: "greater than" for numbers and "after" for dates)
* For relative date matches, UI is simplified to only ask "is in the past [n] [units]" or "is in the future [n] [units]".  
  * These queries also now transparently add a second filter with the inverse `op` based on today.  For example, asking "... where updatedAt is in the past 1 month" also adds "... where updatedAt is less than now()".  These range statements are less flexible than before, but clearer to understand.

Screen Shot 2020-06-23 at 5 30 18 PM

TODO: 

- [x] tests


evantahler:
@andyjih I'm not going to do the "between" case because it's so different than all the rest or the rules, both in UI and logic.  We can come back to this later as it's own story. Also, the same logic can be created by adding to absolute comparisons as 2 rules:

Screen Shot 2020-06-23 at 5 37 34 PM

andyjih:
@evantahler makes sense on the `between` use case. Looks great!

evantahler:
closes T-21

PR #408, opened 2020-06-26
Store Typeahead value when focus is lost

### LINKED ISSUE #252: The Typeahead doesn't accept values that are just entered, even if they match a pre-existing option.

evantahler:
When the mouse is moved away from the `<AsyncTypeahead />` (using the `onBlur` DOM event), we treat the current value in the form as a valid submission. 
* In the case that the user chose a valid autocomplete value, it will just re-save
* In the case that the user typed a custom value, but hasn't clicked "New Selection..." from the drop down, we'll save the value as if they had. 

movie

This is not a problem with `<Typeahead />` as we do not accept user-provided values.

Closes https://github.com/grouparoo/grouparoo/issues/252

### DIFF
--- a/core/api/src/actions/events.ts
+++ b/core/api/src/actions/events.ts
@@ -30,9 +30,7 @@ export class EventsList extends AuthenticatedAction {
     const where = {};
     const includeWhere = {};
 
-    if (params.type) {
-      where["type"] = params.type;
-    }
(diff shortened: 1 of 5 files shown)


evantahler:
Here is how I make these super-sweet gifs https://gist.github.com/evantahler/10d2c62ff47dc53b0fe0934233c995ad

PR #436, opened 2020-07-02
App Parallelism and Dynamic export retrying

evantahler:
1. **Allow plugin Apps to define their own parallel limit**.  For example Mailchimp has a hard limit of [10-requests at a time](https://mailchimp.com/developer/guides/get-started-with-mailchimp-api-3/) (not per second or minute, at a time).  For these types of plugins, we increment and decrement a redis key per App of how many tasks are running that talk to it.  If this task would go over the limit, we re-enqueue the job for 5 seconds from now, and the worker can move on to something else.

2. **Allow `plugin#expportProfile` to respond with if the export should be retried**.   The export method can now respond with `{success: boolean, retryDelay: number, error: any}`.  If `success=false` the plugin can determine how long to delay the export task to be tried again.  Plugins like Hubspot return headers with rate limit information that can be used to dynamically set this retry time.  Or, we can be lazy and just sleep the task for a few seconds, if the Destination won't ban us for constantly retrying.

3. **Multiple export steps per Destination**. The `export` step is now split, in that there is a single `profile#export` job that creates the export records for all the destinations this profile should be sent to, and then individual `export#send` jobs for each Profile + Export pair.  This was done so that a failing or rate-limited export doesn't block all the Destinations that a profile should be sent to.

4. **Run Tracking**. Runs now are smarter, and will track individual Export + Destination pairs, not profiles Exported.  I.E.: if a Group of 10 profiles was sent to 2 Destinations, there will be 20 exports.

5. **Run Looping**, The smarter Runs will not enqueue the next patch (default size: 100) until all the Export + Destination pairs the previous run created are sent.  This is important so that one slow Destination doesn't fill up all of your Redis.  We want to slow down the whole export to give things time to send & complete.

Also:
* Separated the 'save' button from the 'sync' button when viewing /destination/data.  The destination still syncs the whole group automatically when a group is changed, but not again... until you hit the sync button!

Future Work:
* Use Batch APIs, if the plugin supports them


evantahler:
Closes T-142
Closes T-53
Closes T-147

evantahler:
Closes T-144

PR #452, opened 2020-07-10
Try to report errors with telemetry to the telemetry service

evantahler:
If there's an error generating the Telemetry Payload, we should try to send that error to the Telemetry service.

### DIFF
--- a/core/api/src/tasks/telemetry.ts
+++ b/core/api/src/tasks/telemetry.ts
@@ -1,5 +1,7 @@
 import { Task, api, config } from "actionhero";
 import fetch from "isomorphic-fetch";
+import { plugin } from "../modules/plugin";
+import os from "os";
 
 export class Telemetry extends Task {
   constructor() {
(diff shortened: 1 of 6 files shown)


evantahler:
Closes T-180

core/api/src/tasks/telemetry.ts
```diff
@@ -13,14 +16,70 @@ export class Telemetry extends Task {
...
+    return error.stack
+      .split(os.EOL)
+      .map((line: string) => {
+        // It's likely the part of the path with 'grouparoo' in it would be safely down in the path structure to not include app-information
+        // There may be many instances of the word "grouparoo" in the file system, so we'll only take the last part of the path
+        const chunks = line.split("grouparoo" + path.sep);
+        return chunks[chunks.length - 1];
+      })
+      .join(os.EOL);
   }
```
evantahler, line 84: @bleonard Can you think of a situation where there won't be `grouparoo/` in the path structure indicating that we are safely within `node_modules`?  We should be within `node_modules/@grouparoo/...` for client installs and `grouparoo/grouparoo/core...` for our own development

bleonard: I suppose the core (monorepo) case doesn't _require_ it to be in a directory called `grouparoo` but it's certainly a convention. If they are running the whole monorepo app it could be `/Users/someone/Desktop/testing/core/...`

evantahler: True.  I guess I’m less worried about monorepo users when compared to normal “npm” users.  Have you heard about folks using custom NPM module paths or something like that?

PR #467, opened 2020-07-15
Move mutating methods to ModelOps [Refactor]

evantahler:
In order to keep models _smaller_, we make the distinction between the Model itself and ModelOps
  - Table/Column definitions, Class methods, Setters, Getters, and Validations belong on the Model
  - Methods that mutate the Model (or more than one Model), or interact with a Plugin, belong in a ModelOp

For Example, rather than being a class method on the Model, `Profile.merge(profile:Profile, otherProfile:Profile)` is moved to `ProfileOps.merge(profile:Profile, otherProfile:Profile)`.   When appropriate, some model methods still exist as a pass-through to a ModelOp, ie: 

```ts
// from models/Profile.ts

  async import(toSave = true, toLock = true) {
    return ProfileOps._import(this, toSave, toLock);
  }

  async export(force = false, groupsOverride?: Group[]) {
    return ProfileOps._export(this, force, groupsOverride);
  }

```

This PR also cleans up the model files, and collects all the `static` Class methods together at the ends of the file.

### DIFF
--- a/core/api/__tests__/models/profile.ts
+++ b/core/api/__tests__/models/profile.ts
@@ -7,6 +7,7 @@ import { GroupMember } from "./../../src/models/GroupMember";
 import { App } from "./../../src/models/App";
 import { Source } from "./../../src/models/Source";
 import { plugin } from "./../../src/modules/plugin";
+import { ProfileOps } from "../../src/modules/ops/profile";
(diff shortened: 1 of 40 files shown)


evantahler:
This cleanup will help the ProfileProperties to Arrays work be easier.

PR #468, opened 2020-07-15
Single-line if statements [Refactor]

evantahler:


bleonard:
Over time, I've found those extra parentheses prevent bugs (say when adding another thing in the `if` statement) more than they annoy me. What you think?

bleonard:
I was ok with always have ing `{` because of that but was also fine with "ok if on the same line, but if there are multiple lines, add the `{`"

evantahler:
Yeah, I agree - there's something interesting here about `Prettier`'s line length... in theory these should all be single line...

evantahler:
Now... just with single-line IFs! 😄

bleonard:
Is this in prettier or in lint? My favorite is when we don't have to think about it.

evantahler:
Yeah, it's totally valid (now?) with default prettier!  
It wasn't before, and that's how we ended up with with all those multi-lines.   No changes needed to anyone's tooling!

PR #470, opened 2020-07-16
Profile Property Arrays

evantahler:
- [x] Profile Property Rules can be toggled `hasArray` which will allow multiple values to be stored for the Profile Property
  - [x] A Profile Property Rule cannot be both `unique` and `hasArray`
- [x] Profile Properties are stored in _in order_ (with a position column).
- [x] All Methods and APIs which use `profile.properties` now return a `values` array, vs a single `value`:

```ts
interface ProfilePropertyType {
  [key: string]: {
    guid: string;
    values: Array<string | number | boolean | Date>;
    type: string;
    unique: boolean;
    createdAt: Date;
    updatedAt: Date;
  };
}
```


evantahler:
Closes T-113

evantahler:
closes T-108

evantahler:
closes T-109

evantahler:
closes T-209

evantahler:
closes T-106

PR #506, opened 2020-07-28
Destination Plugins can opt into receiving profile property arrays

evantahler:
Plugin Connections gain a new method, `exportArrayProperties` which allows them to define if they want arrays of profile properties, or just the one value to their `exportProfile` method.  In the case where some fields are returned (`['purchases']`) or all (`[*]`) `newProfileProperties` and `oldProfileProperties` will be arrays of values... otherwise the previous behavior of getting single strings/numbers/booleans/dates will remain.

This is an aync method, as the plugin may need to check an API or database to know if arrays are acceptable or not.

### DIFF
--- a/core/api/__tests__/actions/destinations.ts
+++ b/core/api/__tests__/actions/destinations.ts
@@ -225,6 +225,34 @@ describe("actions/destinations", () => {
       );
     });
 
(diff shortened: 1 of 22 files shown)


evantahler:
Closes T-214

PR #511, opened 2020-07-28
Support Postgres schemas other than 'public'

### LINKED ISSUE #510: The @grouparoo/postgres plugin should be able to select a schema other than `public`
The `@grouparoo/postgres` plugin currently only can connect to the `public` schema in a database.   But, there are other possible schemas https://www.postgresql.org/docs/9.1/ddl-schemas.html

A Postgres `App` should be able to provide a `schema` as well as a `database`

N/A

**Additional context**
N/A

evantahler:
Allows the `@grouparoo/postgres` plugin to work with schemas other than `public` (both sources and destinations)

Screen Shot 2020-07-28 at 11 20 16 AM

Closes https://github.com/grouparoo/grouparoo/issues/510
Closes T-226

### DIFF
--- a/plugins/@grouparoo/postgres/src/lib/connect.ts
+++ b/plugins/@grouparoo/postgres/src/lib/connect.ts
@@ -1,4 +1,5 @@
 import { Client } from "pg";
+import format from "pg-format";
 import { ConnectPluginAppMethod } from "@grouparoo/core";
 
 export const connect: ConnectPluginAppMethod = async ({ appOptions }) => {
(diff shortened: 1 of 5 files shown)


plugins/@grouparoo/postgres/src/lib/connect.ts
```diff
@@ -1,8 +1,14 @@
 import { Client } from "pg";
+import format from "pg-format";
 import { ConnectPluginAppMethod } from "@grouparoo/core";
 
 export const connect: ConnectPluginAppMethod = async ({ appOptions }) => {
   const client = new Client(appOptions);
   await client.connect();
+
+  if (appOptions.schema) {
+    await client.query(format(`SET search_path TO %L;`, appOptions.schema));
```
bleonard, line 10: does this change it for the "shared" connection. That might be a problem.

evantahler: We have a shared connection per App, and this option is per App.  Said another way, if you want to use 2 Schemas, you need 2 Apps.   This is the same security model as having a user/connection per-database.

bleonard: Score. Thanks.

PR #513, opened 2020-07-29
Update Docs to use new `npx grouparoo` CLI tool

evantahler:


evantahler:
closes T-225

evantahler:
Went from 5 steps to 4 for `running locally with node.js`!

PR #515, opened 2020-07-30
Redshift Plugin (Source and Destination)

evantahler:
Redshift is Postgres-Compatible (with some caveats), so the Redshift plugin can consume the Postgres Plugin:

```ts
// from plugins/@groupaoo/redshift/src/initializers/plugin.ts

import { test } from "@grouparoo/postgres/dist/lib/test";
import { connect } from "@grouparoo/postgres/dist/lib/connect";
import { disconnect } from "@grouparoo/postgres/dist/lib/disconnect";
import { exportProfile } from "@grouparoo/postgres/dist/lib/export/exportProfile";
import { exportArrayProperties } from "@grouparoo/postgres/dist/lib/export/exportArrayProperties";

//...
```

The only change from the existing plugin is that Redshift cannot handle the `on conflict` directive (as it's a distributed system).  To get around this, when writing the multiple rows for a exported Profile's group, we can do the following (but it's slow):
* Old: `INSERT INTO %I (%I) VALUES (%L) ON CONFLICT DO NOTHING`
* New: `INSERT INTO %I (%I) SELECT %L WHERE NOT EXISTS (SELECT %L FROM %I WHERE %I = %L AND %I = %L)`

The new query looks like: 
```sql
INSERT INTO out_groups (user_id,group_name) SELECT '429','High Value Customers' WHERE NOT EXISTS (SELECT 'user_id' FROM out_groups WHERE user_id = '429' AND group_name = 'High Value Customers')
```

---

Notes on using Redshift:

* To allow access from the 'outside world'
  1. You need a Security Group attached with the port open
  2. You need an IAM role attached
  3. You need to click "Publicly accessible" in the properties
* There are no auto-increment primary keys.  If you aren't writing to this column, the write will fail... so just don't have a primary key
* There are no `default NOW()` timestamp columns.


evantahler:
Closes T-215

PR #533, opened 2020-08-03
Adding guides

andyjih:


evantahler:
These are awesome!   

Maybe we source these (and some of the other docs) into WWW now?

andyjih:
@evantahler Sure! How do you want to do that?

PR #595, opened 2020-08-11
Destination draining

evantahler:
This PR Links Destinations to Runs though a new `ExportRun` model.  `Group` -> `Run` -> `ExportRun` -> `Export` -> `Destination`.

In this way, we can now get a more direct link to the Runs that have exported a profile to a given destination.   This allows a few new features:

1. We can display a Destination's runs in the UI.  The "Creator" of the Runs remain the Group, but we can show which runs have exported a profile to this Destination.
 Screen Shot 2020-08-11 at 11 24 56 AM

2. We can prevent a Destination from being deleted if any run is still using it for a run
Screen Shot 2020-08-11 at 10 31 26 AM

3. We can prevent a Destination from being deleted if a Group which used to be tracked by this destination is updating.  This is the case when you've  _just_ told a Destination to track no Groups, but we are in the early phases of a Run for that group which still needs to tell the Destination to remove the Profiles it used to Track.  Once there are exports created, we are in condition `#2` above
Screen Shot 2020-08-11 at 10 40 16 AM

Closes T-54
Closes T-263

### DIFF
--- a/core/api/src/migrations/000036-createExportRuns.ts
+++ b/core/api/src/migrations/000036-createExportRuns.ts
@@ -1,0 +1,46 @@
+module.exports = {
+  up: async function (migration, DataTypes) {
+    await migration.createTable("exportRuns", {
+      guid: {
+        type: DataTypes.STRING(40),
+        primaryKey: true,
+      },
+
+      exportGuid: {
+        type: DataTypes.STRING(40),
+        allowNull: false,
+      },
+
+      runGuid: {
+        type: DataTypes.STRING(40),
+        allowNull: false,
+      },
+
+      createdAt: {
+        type: DataTypes.DATE,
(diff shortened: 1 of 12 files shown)


core/api/src/migrations/000036-createExportRuns.ts
```diff
@@ -0,0 +1,46 @@
...
+        allowNull: false,
+      },
+    });
+
+    await migration.addIndex("exportRuns", ["exportGuid", "runGuid"], {
+      unique: true,
+      fields: ["exportGuid", "runGuid"],
+    });
+
+    await migration.removeColumn("exports", "runGuids");
```
evantahler, line 35: ! This is a destructive migration, but the export data isn't too important to keep around...

PR #598, opened 2020-08-12
Destination Batch Exports

evantahler:
## Overview

Plugins which export can not provide either `plugin.methods.exportProfile` to be sent a single profile at a time, or `plugin.methods.exportProfiles` to be sent a collection of profiles which should be exported.  

If both methods are provided by a plugin, `exportProfile` will be used in the case when just one profile is being updated, and `exportProfiles` will be used within the context of a run, however only one method is needed.

A Run's batch will not proceed until all outstanding exports are exported.  For example, if 1001 profiles were imported in batch, we might do 11 calls to `plugin#exportProfle` (10 x 100 profiles, and then 1 call with the 1 remaining profile).  There's an upper bound on the batch size (via a new setting), but no lower bound.

Handling errors within a batch is tricky:
* If an error is thrown, all exports in the batch will have the error applied
* To apply an error to only some exports, return the `errors` object, which is an instance of `Error`, with an additional `errors` array per profileGuid:

```ts
const errors: Error[] = [];

const specificError = new Error("the API didn't like this profile");
specificError.profileGuid = 'pro_abc123';
errors.push(specificError);

return {success: false, errors}
``` 

The Logger plugin has been updated to use the new batch methods as an example, to minimize file writes. 

## Details
* Use combined ExportedProfile interface in destinations
* Adds new setting `core/export-profile-batch-size` (default 100) to control how large exportProfiles batches can be
* There's a new common type, `ExportedProfile`, which both `plugin#exportProfle` and `plugin#exportProfiles` use.  `plugin#exportProfiles` is passed `exports: ExportedProfile[]` while `plugin#exportProfle` is passed `export: ExportedProfile`
  * All existing plugins have been updated to use these new method signatures.
```ts
export interface ExportedProfile {
  profile: Profile;
  oldProfileProperties: { [key: string]: any };
  newProfileProperties: { [key: string]: any };
  oldGroups: Array<string>;
  newGroups: Array<string>;
  toDelete: boolean;
}
```
* `Run#afterBatch` now exists as the "hook" the tasks will call after each batch of a run.  This now runs both `Run#processBatchExports()` and `Run#determinePercentComplete()`, and is where we can add more methods in the future.

Closes T-274

Todo:
- [x] Update existing plugins for new method signatures
- [x] Update 1 plugin to use `exportProfiles` method
- [x] Update tests for old features
- [x] Test for new features


bleonard:
Fixes https://github.com/grouparoo/grouparoo/issues/655

PR #600, opened 2020-08-13
Destination UI Fixes

evantahler:
* When the MYSQL and Postgres are exporting a user who has already been exported, don't use null for columns that are unmapped 
* Resque error when appending message to a destination error 
* When removing a destination mapping for MySQL (and others?), a different mapping was reset 
* Error appears when switching a destination from 'track all groups' to a single group


evantahler:
Closes T-287

evantahler:
Closes T-289

evantahler:
Closes T-291
Closes T-293

PR #603, opened 2020-08-14
Phone Number Profile Property Type

evantahler:
This PR adds a new Profile Property type to Grouparoo - `phoneNumber`. 

* We will attempt to format phone numbers when they are saved according to [libphonenumber-js](https://www.npmjs.com/package/libphonenumber-js).  If the library doesn't think the phone number is valid, rather than throw an error, we will store the phone number as provided by the source.
  * To aid in this formatting, a new setting, `core/default-country-code` is required (default: "US").  When phone numbers are found without a `+#` explicitly setting the country code, we need to guess which country the phone number is from.

### DIFF
--- a/core/api/__tests__/actions/groups.ts
+++ b/core/api/__tests__/actions/groups.ts
@@ -243,6 +243,7 @@ describe("actions/groups", () => {
           "email",
           "float",
           "integer",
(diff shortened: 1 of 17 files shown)


evantahler:
Closes T-294

PR #609, opened 2020-08-17
Require that phoneNumber and email  Profile Properties be valid

evantahler:
Throw and error and do not save `phoneNumber` or `email`-type Profile Properties if they are not valid.

Closes T-303
Closes T-304


core/api/src/models/ProfileProperty.ts
```diff
@@ -14,6 +14,7 @@ import { Profile } from "./Profile";
 import { ProfilePropertyRule } from "./ProfilePropertyRule";
 import { parsePhoneNumberFromString, CountryCode } from "libphonenumber-js/max";
 import { plugin } from "../modules/plugin";
+import isEmail from "validator/lib/isEmail";
```
evantahler, line 17: `validator` comes with `sequelize`

PR #625, opened 2020-08-25
Destination Export Types and Export Typecasting

evantahler:
This pull request allows Destinations to request specific data types that don't necessarily map 1-1 to the Grouparoo data types.  For example, Zendesk requires a `remote-id` profile property that is of type `string`.  However, it would be valid to send the Profile Property `userId` that is type `integer`, but convert it to a string before the hitting the Destination's `exportProfile()` method.    

This PR also now ensures that all old and new Profile Properties sent to a Destination's `exportProfile(s)` method are of the type expected (string, integer, date, etc).  The only exception is if the type `any` is provided, than no typecasting will take place.

The following [table](https://docs.google.com/spreadsheets/d/1Fbkdsq_IR8deOYF4QpVq_XIdqpvAah3tJbVP2K5nCdc/edit?pli=1#gid=0) shows the valid type casting at this time.

Screen Shot 2020-08-26 at 9 27 20 AM

This PR also introduced a new 'relaxed' type of `number` which a destination can request, which is valid for both `integer` and `float`.

Closes T-297


evantahler:
@bleonard I'm curious if you agree to the logic in the table.  Is that what you would expect?

bleonard:
I updated the sheet a bit and put some comments. I'm not sure about the (converting to) Boolean column. Technically any of them could be `!!` or should it be 1/0 from integers and "true"/"false" from String? Or should it be red?

evantahler:
Thanks!  

I like the additions of  `boolean` -> `int` and `float` (1/0).  I also agree that we probably shouldn't coerce a string into a boolean with `!!` (that's from the airbnb JS style guide).  What I think we should do is:
 1. 'true' (/lowercase/) is true 
 2. '1' is true
 3. 'false' is false
 4. '0' is false
 5. any other string with a length > 0 is true
 6. ... otherwise false.

Some choices I made that I wanted to call out explicitly:
* **[`float` -> `int`]**: In JS if you `parseInt(float)` it actually more closely resembles `Math.floor()` in that 3.999 becomes 3.  I'm guessing that our users would expect the behavior of `Math.round()` in which 3.999 becomes 4.
* **[`date` -> `number` (and `int` + `float`)]**: We deal in JS, so all times are in milliseconds (since epoch).  However, most other systems use seconds (since epoch).  I chose to keep things in milliseconds, but we could just as easily convert these to seconds-since-epoch. 
* **[`date` -> `string`]**: I went with `date.toISOString()` ie: `1970-01-01T00:00:00.123Z`... but there are lots of other formats we could have used.  Zendesk and Postgres seem to like this out-of-the-box, so I thought it was a good choice.

bleonard:
I guess I'm thinking we do a first pass only with "lossless" cases. Maybe float rounding to integer is ok since it's predictable (agree with rounding), but coercing a string to a boolean in the generic sense seems like something to punt to later.

evantahler:
TODO: 
* remove float -> int (it's lossy)
* remove string -> bool (it's confusing)

PR #659, opened 2020-08-31
Destinations can only track one Group

evantahler:
This PR updates the API and UI to only allow a single group to be tracked by a Destination at a time.

Now that Groups can be built in ways that would allow a group to contain all profiles (https://github.com/grouparoo/grouparoo/pull/641), Destinations can be simplified to only track a single Group at a time, and a tacking group (ie: 'people to send to Mailchimp') can be built.  

Closes T-415
Closes T-408


evantahler:
Now that there's only 1 group for each destination, the model DestinationGroups has been removed. There's a new column groupGuid on Destination.

PR #660, opened 2020-08-31
Revert "Bump typedoc from 0.18.0 to 0.19.0"

evantahler:
Reverts grouparoo/grouparoo#645


evantahler:
Looks like this won't quite work until we hit TS 4.0

PR #688, opened 2020-09-03
Destination batch updates

bleonard:
* do better at ensuring parallelism is right
* name things more clearly


evantahler:
Thank you!

PR #690, opened 2020-09-04
Do not block Group deletion if DestinationGroupMemberships exist

evantahler:
Allows Groups to be deleted when Destinations may be sending membership in this group to Destinations.  Groups remain unable to be deleted if a destination is Tracking it.

Closes T-408


evantahler:
@andyjih I like it when we can see in the tests that we've changed out minds about how a feature should work

PR #691, opened 2020-09-04
When Destination Group Memberships are removed by deleting a Group, a run is started

evantahler:
When Destination Group Memberships are removed by deleting a group, a run should be started to re-export the profiles to that destination with the group tag removed.

Closes T-450


evantahler:
Closes T-461

PR #693, opened 2020-09-04
Check that the run still exists before incrementing after sending exports

evantahler:
It's possible the run that created this export has been removed (it was from a deleting group).  Check that the run still exists!

Closes T-450


core/api/src/modules/ops/destination.ts
```diff
@@ -722,6 +699,28 @@ export namespace DestinationOps {
     }
   }
 
+  /**
+   * It's possible the run that created this export has been removed (it was from a deleting group).
+   */
+  async function safelyIncrementExportRunRuns(
```
bleonard, line 705: still might have a timing issue. you might want to catch "not found" or something if you really want it to be fully safe (and not have to check existence again first)

PR #694, opened 2020-09-04
Move mock events to new demo plugin

bleonard:


plugins/@grouparoo/demo/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "author": "Grouparoo Inc <hello@grouparoo.com>",
+  "name": "@grouparoo/demo",
+  "description": "The Grouparoo Demo Setup Plugin",
+  "version": "0.1.12-alpha.1",
+  "license": "MPL-2.0",
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
```
evantahler, line 10: Maybe we don't actually want to publish this one?  
Lerna knows not to publish things that are `private` (defined here)

evantahler (commented):
Legit!

plugins/@grouparoo/demo/package.json
bleonard: maybe. I didn't do it on purpose per se, but I can imagine a world where someone wants to load it up with data. doesn't seem to hurt.

bleonard: whichever you think.

evantahler: I'm indifferent at the moment - just calling out the options we have.   It'll get published tonight!

PR #716, opened 2020-09-09
Do not sort by Random()

evantahler:
Sorting by `RAND()` or `RANDOM()` depending on your database, can be very slow with large datasets.  This is due to the need to first apply a random number to each row in memory before sorting.  
* https://stackoverflow.com/questions/8674718/best-way-to-select-random-rows-postgresql
* https://stackoverflow.com/questions/1244555/how-can-i-optimize-mysqls-order-by-rand-function

There are a few cases when a Random sort was being used:
* Internally (against the Grouparoo database) 
  1. To find a random Profile to either test an import against or display.  We now switch to sorting by `guid`, which is semi-random but deterministic.  For a static collection of Profiles, the same profile will now always be returned.  
  2. Collecting random Profile Properties to display as examples to what values exist (like a list of first names).  We now switch to sorting by `guid`, which is semi-random but deterministic.  This will generate the same list each time for a static collection of Profiles
* For Source Previews
  * We re-use the `sourcePreview()` method in a number of places (previewing a source's data, creating a source's mapping, building or editing a Profile Property) to show samples of the data in the Source.  In these cases, showing a random collection of data remains a good idea.  In these cases we now:
    1. Only show the first 10 records in the source table, ordered naturally
  ~1. `count()` how large the table is~
  ~2. Generate random numbers (in JS) between 0 the number of rows in the table~
  ~3. Select rows with an `offset` 1-by-1 (so 10 unique select statements) to use the natural sort of the table, avoiding a `sort by` call.~
  * This closes T-86 as a side effect.

Closes T-482

### DIFF
--- a/core/api/src/actions/profilePropertyRules.ts
+++ b/core/api/src/actions/profilePropertyRules.ts
@@ -54,7 +54,7 @@ export class ProfilePropertyRulesList extends AuthenticatedAction {
           profilePropertyRuleGuid: rule.guid,
           rawValue: { [Op.not]: null },
         },
-        order: api.sequelize.random(),
+        order: [["guid", "asc"]],
         limit: 5,
       });
       const exampleValues = examples.map((e) => e.rawValue);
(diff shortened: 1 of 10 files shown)


evantahler:
Updated the description to reflect that we are now only showing the first 10 records from sources.

PR #718, opened 2020-09-10
Demo data

bleonard:
Set up a demo world.

Main command: `npm run actionhero -- grouparoo demo dataPurchases`
- 🔴 Clears the current grouparoo database 
- Creates admin team and a demo user (user: demo@grouparoo.com / pw: password)
- Adds `demo` schema with 1000 users and their purchases
- Creates Postgres source for users and purchases
- Sets up events and creates a few in an eCommerce funnel
- Creates event source that reads those pageview events
- Creates groups that use a few properties

Future work:
- Postgres Query source
- Tests

Need a pattern to opt into other things with credentials (env?):
- BigQuery and other sources
- Mailchimp and other supported destinations


plugins/@grouparoo/demo/src/events.ts
```diff
@@ -0,0 +1,309 @@
...
+    where: {
+      type: "events",
+    },
+  });
+  if (!eventsApp) {
+    eventsApp = App.build({
+      type: "events",
+      name: "events",
+      state: "draft",
+    });
```
evantahler, line 134: In what cases is the event app not being built?  I guess if you have a totally blank database and start with the `cli` commands the start steps won't run...

If you want to share that code, you could add an `export` in front of addEventsApp() in the core initializer and you can require it thought he node module, something like 
`import {addEventsApp} from "@gruoparoo/core/dist/initializers/events"`

evantahler (commented):
This is super fun!

plugins/@grouparoo/demo/src/events.ts
bleonard: Yes, that was the case.
CLI doesn't call those initalizers in that way.
I think it's ok for now.

PR #727, opened 2020-09-13
Relax the validation requirements of email addresses

evantahler:
Relax the validation requirements of email addresses - Email addresses can now:
* Be longer than 254 chars
* The user (part of the email address before the `@`) and the domain can be longer than 64 and 254 bytes respectively
* Use an internal-looking domain (like `@company.internal`) 
* Include the display name (like `<evan>`) before the real email address
* Be against an IP address (like `person@1.2.3.4`)

Requires https://github.com/validatorjs/validator.js/pull/1435
Closes T-510


evantahler:
For now, let's just simply validate ourselves:
- there's an @ in the string
- there's a . after the @

andyjih:
:+1:

PR #736, opened 2020-09-15
Spec Helper Plugin

evantahler:
This Pull Request moves the Grouparoo Spec Helper and Factories into their own plugin.  This will make it possible for plugins outside of the monorepo to include `@grouparoo/spec-helper` and run tests which require models, a running server, etc. 

A typical test file's shape:
```ts

describe("test suite", () => {
  let actionhero;

  beforeAll(async () => {
    const env = await helper.prepareForAPITest();
    actionhero = env.actionhero;
  }, 1000 * 30);

  afterAll(async () => {
    await helper.shutdown(actionhero);
  });

  test('it works', () => {
    expect(1 + 1).toBe(2);
  })
});
```

There's a fair amount of trickery going on, to basically `chidir` into `@gropuaroo/core` and pretend that's where are running the test from there.  This also involves tricking Actionhero to know where the config files are. Since `@grouparoo/spec-helper` is only ever to be used in test, we don't need to compile it - we can use the typescript packages directly.

In your test files, you'll need to import the file inline (`import "@grouparoo/spec-helper";`) or import a module which is **actually used** elsewhere in the codebase (`import {helper} "@grouparoo/spec-helper"`... the module resolver will skip it if you don't use `helper`) .

For tests, if you need the Grouparoo API to load in the plugin your are testing for an integration test, you can now do with with the Environment Variable `GROUPAROO_INJECTED_PLUGINS`:

```ts
import path from "path";
process.env.GROUPAROO_INJECTED_PLUGINS = JSON.stringify({
  "@grouparoo/bigquery": { path: path.join(__dirname, "..", "..") },
});
```

- [x] Make a new `@grouparoo/spec-helper` plugin
- [x] Update Core to use `@grouparoo/spec-helper`
- [x] Update Monorepo plugins to use `@grouparoo/spec-helper`
- [x] Update external plugins to use `@grouparoo/spec-helper`
- [x] ~Solve circular dependency warnings~


evantahler:
T-360

evantahler:
`@grouparoo/spec-helper` tested on https://github.com/grouparoo/internal/pull/62.  Until the plugin the `npm link` workflow is required

PR #739, opened 2020-09-15
Do not freak out if unknown marketo type

bleonard:
Closes T-514


andyjih:
For future reference, the destination page is throwing an error from the unknown Marketo type `lead_function`, which is a type that Salesforce uses. Thanks for the fix, Brian!

Issue #649, opened 2020-08-31
Feature Request: Salesforce Destination


bleonard:
**Is your feature request related to a problem? Please describe.**
Add Salesforce as a destination for profiles and groups.

bleonard:
It turns out they have 100 tools. The two main options for Grouparoo seems likely to be either 



*   stock Salesforce.com which has a bunch of config variability that I want to discover
*    Salesforce Marketing Cloud, which is a whole different thing

For this document, I’m going to assume Regular Salesforce.

This is basically just a generic database with a REST API. A table is called an [SObject](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_sobject_basic_info.htm).

It comes with many built in SObject types, but this [API](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_describeGlobal.htm) gets you the whole list.

**Destination Options**

Maybe this list is used in Destination Objects to set the “Contact” (default).

Then they should pick the “external id” identifier (default to “ExternalId” if it exists).

It should be possible to make multiple destinations for an app.

**Profiles**

There is the [Contact](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_contact.htm) schema. It’s hard to know the difference. We need to see what people are using.

I’m generally not assuming we use the [User](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_user.htm) schema even thoughost of the things reference that. It looks to be for internal users.

There is a ton of stuff that hangs off of this. Other _hasMany_ objects. It’s unclear how we deal with those. Special case or does the DESCRIBE call let use know the references? 

For example: [UserPreference](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_userpreference.htm), [UserDevice](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_userdevice.htm), etc

[UserProfile](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_userprofile.htm) is more of a _hasOne_ kind of thing.

But it can still be generic. We can get the fields by [describing](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_sobject_describe.htm) the SObject.

[...]

PR #754, opened 2020-09-19
Salesforce Destination

bleonard:
Adds a generic salesforce destination to sync over objects from Grouparoo.
For Issue https://github.com/grouparoo/grouparoo/issues/649

Salesforce is a pretty generic thing. So the first destination I made is the most generic one. You pick a variety of objects and keys and it knows what to do. 

So many settings!
- `profileObject`: Which object in Salesforce represents a Grouparoo profile?
- `profileMatchField`: Which field in the profile Object is used to match Grouparoo profiles?
- `groupObject`: Which object in Salesforce represents a Grouparoo group?
- `groupNameField`: Which field in the group Object is used for the name of a Grouparoo group?
- `membershipObject`: Which object in Salesforce maps the profile object to the group object?
- `membershipProfileField`: Which field in the membership Object is the reference to the profile?
- `membershipGroupField`: Which field in the membership Object is the reference to the group?
- `profileReferenceField`: Is there a reference field on the profile Object to fill out?
- `profileReferenceObject`: If there is a reference field, which Object should be created to apply to the profile?
- `profileReferenceMatchField`: If there is a reference field, how should it be matched to Grouparoo profiles?


bleonard:
In the future, we could make more targeted ones if we see common patterns regarding usage of building in SObjects. Some thoughts for the future:
- Lead/Contact assume email field and that it makes campaigns (Only have to select lead or campaign)
- Contacts that add reference to Account. (set foreign lookup keys for Contact and Account)
- Custom sync using Grouparoo specific objects like GrouparooProfile, GrouparooGroup, GrouparooGroupMember.

bleonard:
closes T-537

evantahler:
@bleonard rebased with master, version updated, and force-pushed.

PR #783, opened 2020-09-28
Ensure that we do not touch updatedAt when incrementing a model and do not await SQL when a Transaction is in play

evantahler:
This PR prevents a class of bugs that resulted in errors like `SequelizeConnectionAcquireTimeoutError` or `SequelizeTimeoutError`.

1. **Do not update `updatedAt` when using `increment()`** 

We now explicitly ask the `increment()` method not to bump the `updatedAt`, and then update `updatedAt` in a subsequent SQL statement.  By Default, Sequelize will try to update `updatedAt` every time a row is altered (which makes sense). 
 However, in the cases we think a row might be altered rapidly, we use the `model.increment()` method to avoid collisions...
but the increment command also tries to set `updatedAt` - which creates contention!

Screen Shot 2020-09-28 at 2 54 02 PM

A negative side effect of this change: we may be a few seconds off for `run.updatedAt`... but that's OK.

2. **Ensure that no other SQL statements are await-ed when a transaction is in play**
`Group.getRules()` was potentially used within a transaction, but not included in the transaction.  This was bad.  Also, rather than making many references to single rows in a transaction which may be changing, it's better to make one big SQL statement to update the bulk of items at once.

3. **Provide an environment variable, `SEQUELIZE_POOL_SIZE` to up the Sequelize Connection Pool Size (default: 5)**
ie: `SEQUELIZE_POOL_SIZE=10`

4. **Allow the Event#associate method to be retried once**
With many workers all processing events from the same anonymous profile, multiple workers may attempt to create the same profile at once.  One will fail as there is a unique database constraint.  We now retry this specific type of failure so the newly created Profile can found and linked.  

--- 

Things to watch out for in the future:

* As indicated in https://github.com/sequelize/sequelize/issues/11024 and https://github.com/sequelize/sequelize/pull/11633 it appears that if you attempt to run more Sequelize transactions than there are available slots in the connection pool, you'll get a deadlock maybe?

Closes T-541
Closes T-553

### DIFF
--- a/core/api/src/config/sequelize.ts
+++ b/core/api/src/config/sequelize.ts
@@ -62,6 +62,12 @@ export const DEFAULT = {
       password: password,
       models: [join(__dirname, "..", "models")],
       migrations: [join(__dirname, "..", "migrations")],
+      pool: {
+        max: process.env.WORKERS ? parseInt(process.env.WORKERS) + 1 : 5,
+        min: 0,
+        acquire: 30000,
+        idle: 10000,
+      },
     };
   },
 };
--- a/core/api/src/config/sequelize.ts
+++ b/core/api/src/config/sequelize.ts
@@ -63,7 +63,9 @@ export const DEFAULT = {
       models: [join(__dirname, "..", "models")],
       migrations: [join(__dirname, "..", "migrations")],
       pool: {
-        max: process.env.WORKERS ? parseInt(process.env.WORKERS) + 1 : 5,
+        max: process.env.SEQUELIZE_POOL_SIZE
+          ? parseInt(process.env.SEQUELIZE_POOL_SIZE)
+          : 5,
         min: 0,
         acquire: 30000,
         idle: 10000,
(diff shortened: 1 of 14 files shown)


evantahler:
This PR also makes the `grouparoo demo scale` CLI command configurable.

ie: `npm run actionhero -- grouparoo demo scale --scale 10` produces 10,000 Users.

CC @bleonard

core/api/src/config/sequelize.ts
```diff
@@ -62,6 +62,14 @@ export const DEFAULT = {
...
       models: [join(__dirname, "..", "models")],
       migrations: [join(__dirname, "..", "migrations")],
+      pool: {
+        max: process.env.SEQUELIZE_POOL_SIZE
+          ? parseInt(process.env.SEQUELIZE_POOL_SIZE)
+          : 5,
+        min: 0,
+        acquire: 30000,
+        idle: 10000,
+      },
```
evantahler, line 72: I don't think the average Grouparoo cluster will need to change the timeouts... but the pool size seems likely.

PR #813, opened 2020-10-05
Button, Form, and Table Loading states

### LINKED ISSUE #792: UI Performance - Placeholders, Spinners, and Loading/Button states

* More, smaller components
* More front-end tests

N/A

**Additional context**
N/A

evantahler:
* Revamp the `<LoadingTable/>` component to always show the headers of the table and denote that the `tbody` is loading
* Use new component `<LoadingButton />` any time the hitting a button hits the API.  The button now shows a loading state and is disabled.
  * Fixes bugs in which some buttons were using `active` to indicate that the button was disabled rather than `disabled`
* Form Fields become disabled when the form is submitting
* For Sources and Destinations, we no longer attempt to load the options as part of page hydration, as it may be slow
  * We now load the options after loading the page, and show a banner as the options are loading

|   |   |   | 
|---|---|---|
| table | sign-in  | 
loading

(network calls were slowed down artificially to make these videos)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,6 @@ jobs:
       - image: circleci/node:12
     steps:
       - checkout
-      - run: sudo apt-get install -y rsync
       - run:
           name: npm-install
           command: npm install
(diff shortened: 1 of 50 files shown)


evantahler:

PR #814, opened 2020-10-05
Cache helper that uses guids

bleonard:
* add guid to plugin methods
* create `objectCache` that uses guids and fancy key stuff (anything -> md5) for smarter caching
* uses `objectCache` methods in plugins that were caching
* switches `SimpleOptions` to be string explicitly (was already the case, just not in typescript)


core/api/__tests__/modules/cache.ts
```diff
@@ -0,0 +1,120 @@
...
+      };
+      const first = objectCache({ objectGuid, cacheKey }, methodToGetValue);
+      const second = objectCache({ objectGuid, cacheKey }, methodToGetValue);
+
+      const values = await Promise.all([first, second]);
+      expect(values).toEqual([3, 3]);
+      expect(callee).toHaveBeenCalledTimes(1);
+    });
+
+    test("if a lock is in use, it will call them in parallel", async () => {
```
evantahler: I think the test title should be "if a lock is /not/ in use"?

core/api/__tests__/modules/cache.ts
```diff
@@ -0,0 +1,120 @@
...
+  beforeAll(async () => {
+    await helper.factories.profilePropertyRules();
+  });
+
+  beforeEach(async () => {
+    await api.resque.queue.connection.redis.flushdb();
+  });
+
+  describe("caching", () => {
+    function checkCaching(name, cacheKey) {
```
evantahler, line 29: I love your recursive tests!

core/api/src/index.ts
```diff
@@ -1,3 +1,4 @@
+import { ObjectCacheInvalidateMethod } from "./modules/cache";
```
evantahler: Why have the import up here if you export * below?

core/api/src/modules/cache.ts
```diff
@@ -0,0 +1,171 @@
...
+
+function useRedis() {
+  const running = !!api?.redis?.clients;
+  if (!running && process.env.NODE_ENV === "test") {
+    return false;
+  }
+  return true;
+}
+
+export const objectCacheInvalidate: ObjectCacheInvalidateMethod = async ({
```
evantahler, line 111: This method is great - thanks for keeping the `:` separated namespace (that's what Actionhero does internally)

Two notes for the future:
* `keys` is pretty slow and blocking if there are a lot of cache entries.  The `scan` method is the way to go.  Here's how node-resque uses it: https://github.com/actionhero/node-resque/blob/master/src/core/connection.ts#L116-L146.  
* Instead of making N redis calls to delete each key (jobs array), you could try a redis pipeline to do it in one call: https://github.com/luin/ioredis#pipelining (our redis connection objects are form `ioredis`)

bleonard: I did what `cache.clear()` in actionhero does.
https://github.com/actionhero/actionhero/blob/ef6d1f7a9d4a7d10cdfe5f6689445367ee8ad2b2/src/modules/cache.ts#L74-L83

Maybe we should update that and allow an optional prefix (on top of the actionhero one)

evantahler: On it!

evantahler: https://github.com/actionhero/actionhero/pull/1611

evantahler: ... and added toGrouparoo via https://github.com/grouparoo/grouparoo/pull/817

core/api/src/index.ts
bleonard: VS code typeahead is aggressive at adding import statements in weird ways. removing.

core/api/src/modules/cache.ts
bleonard: made some comments there. we can update this on another PR when things are all merged.

bleonard:
@evantahler  - i also made `SimpleOptions` be a string so it was easier to use them as a cache key. A few things (and possible bugs fixed) flowed out of that. Yeah types!

evantahler (commented):
I think there's still a miss-named test here.  I'll also get the new version of Actionhero added today!

evantahler:
Merge in https://github.com/grouparoo/grouparoo/pull/821 and then rebase, and you should be good to go!

bleonard:
Closes T-429

PR #815, opened 2020-10-06
Remove Router and Query from _app

### LINKED ISSUE #792: UI Performance - Placeholders, Spinners, and Loading/Button states

evantahler:
This PR removes all instances of `router` and it's associated properties (`query`, `pathname`, etc) from our base `_app.tsx` page.  We had been double-rendering all pages on changes of router and path which has the nasty side effect of resetting all page-state to "empty" before routing to a new page.  This looks bad (and is unnecessary).  

This PR brings much of the logic about reading in `query`, `pathname`, etc into the pages and components themselves:

- [x] Do not import Router directly, use the `{useRouter}` hook
- [x] Create a new helper for `useOffset`
- [x] Create a new helper for `updateURLParams`
- [x] Fix a bug that doesn't property render charts now on initial page load
- [x] Update usage of previousPath (without storing it in `_app`'s state; likely a cookie?)
- [x] Fix Tests

Closes T-586


evantahler:

PR #816, opened 2020-10-07
Sub-Pages for selecting new Sources and Destinations

### LINKED ISSUE #792: UI Performance - Placeholders, Spinners, and Loading/Button states

evantahler:
When creating new Apps, you just click on the card for your app and a new draft App is created (no second click on conformation button)

When creating new Sources and Destinations you are first shown a list of possible Apps that can be used for the Source/Destination.  When you click on one, you are taken to a new page which lists the possible Connections which use that app.  From there, clicking the "Create Source/Destination" Button will create the Draft for you and take you to the edit page.

Screen Shot 2020-10-07 at 11 15 26 AM

Screen Shot 2020-10-07 at 11 15 28 AM

Closes T-558


evantahler:

PR #817, opened 2020-10-07
Typed API Responses - Actionhero v24.0.0-alpha.2

### LINKED ISSUE #792: UI Performance - Placeholders, Spinners, and Loading/Button states

evantahler:
With Actionhero v24.0.0, Actions now have a return Typescript Type.  This PR:
1. Updates all Actions to return a value (rather than using the old `data.response` object)
2. Creates a helper in `core/web/utils/apiData` to import all the Actions and determine thier response type by resolving the promise on the prototype, ie: 

```ts
import { GroupCreate } from "../../api/src/actions/groups";

export namespace Actions {
  export type GroupCreate = AsyncReturnType<typeof GroupCreate.prototype.run>
}
```

3. Type all the uses of `execApi` in the frontend, IE:

```ts
    const response: Actions.GroupCreate = await execApi(
      "post",
      `/group`,
      Object.assign({}, data, { state })
    );
```

4. Fix the bugs that were found 🐞 

Closes T-587


evantahler:

PR #820, opened 2020-10-08
Updated bowtie

andyjih:


evantahler:
I updated the 'social image' from `/Settings` for all of our Github repos

PR #826, opened 2020-10-09
In-App Notifications

### LINKED ISSUE #781: Feature Request: Upgrade Notifications and System Messages

As Grouparoo is an installed application within our customer's data centers, we need a way to help them with the upgrade process when new versions are out.

1. We should let customers know that the versions of Grouparoo that they are running are outdated.  We currently list all the versions in use on the `about` page.  Perhaps we could list the `latest` version on NPM there as well?  Actually doing the upgrade for our customers is out-of-scope (and likely impossible given Docker-based deployment)
2. We should alert customers to check the `/about` page if there are new versions somehow.  This hints that we need a new section in the Grouparoo UI to place "system messages", of which the need to upgrade will be the first one.  This is likely a new page that Administrators can see. 
  
For the first version of this feature, messages will be:
* Shared by all uses of the Grouparoo Cluster - The inbox is Global to all Administrators
* Marked as read/unread
* Deletable. 

N/A

**Additional context**
N/A

evantahler:
This PR adds the concept of `Notifications` to Grouparoo. The starting place for this is the `Notifications` model & API.  Notifications are global to the cluster (as opposed to per-user).  Unread Notifications present as an alert in the Navigation.  Any Team Member (with sufficient permissions) reading a notification removes the alert.

Notifications have a `from`, `subject` and `body`.  Notification bodies can be written in plain-text or markdown, which will be rendered to HTML when displayed.   Notifications can also optionally have a `cta` and `ctaLink` which can be used to display a button to click on. 

Screen Shot 2020-10-09 at 3 02 50 PM

Screen Shot 2020-10-09 at 2 52 32 PM

- [x] Rename `messages` to `notifications` (I'm sure we'll use 'messages' for something else in the future)
- [x] Permissions for `notifications` api methods
- [x] Add both a `cta` and `ctaLink` column (null-able) - we'll need something like "see errors" and "/url/to/errors"
- [x] UI - inbox page and and message-details page
- [x] notifications badge


evantahler:
closes T-596

PR #827, opened 2020-10-09
Share import code across sources

bleonard:
Most of the code in the various import plugins are the same. For example: big query, postgres, snowflake, mysql, redshift, etc.
This extracts the common bits to a new plugin so it can be shared and we can focus on what each does differently.

It starts with the snowflake plugin. Then we can do each one by one.


core/api/src/modules/plugin.ts
```diff
@@ -295,6 +295,7 @@ export namespace plugin {
   export async function getProfileData(
     profile: Profile
   ): Promise<{ [remoteKey: string]: any }> {
+    // TODO: we could do these types better to be string | number | etc | Array<string | etc>
```
evantahler, line 298: I think you can get this type from `AsyncReturnType<Profile.prototype.properties>`

plugins/@grouparoo/app-templates/src/source/query/options.ts
```diff
@@ -0,0 +1,11 @@
...
+export * from "./profilePropertyRuleOptions";
+export * from "./sourceOptions";
+export * from "./profileProperty";
+
+export {
+  PluginConnectionProfilePropertyRuleOption,
+  PluginConnection,
+  SourceOptionsMethod,
+  ProfilePropertyPluginMethod,
+} from "@grouparoo/core";
```
evantahler, line 11: Why re-export from core?

plugins/@grouparoo/snowflake/package.json
```diff
@@ -27,6 +27,7 @@
     "lint": "prettier --check src __tests__"
   },
   "dependencies": {
+    "@grouparoo/app-templates": "^0.1.17",
```
evantahler: Make this a specific version please (no `^`)

plugins/@grouparoo/app-templates/src/source/query/profileProperty.ts
```diff
@@ -0,0 +1,77 @@
...
+    profilePropertyRule,
+    profilePropertyRuleOptions,
+  }) => {
+    const ruleQuery = profilePropertyRuleOptions[queryKey];
+    const query = await plugin.replaceTemplateProfileVariables(
+      ruleQuery,
+      profile
+    );
+    if (!validateQuery) {
+      validateQuery = validateGenericQuery;
```
evantahler: If a validation query is not provided, I don't think that the generic one should be used without a way to opt-out.  I'd suggest moving this to the arguments so callers can provide `null` if they don't want the default:

```ts
export const getProfileProperty: GetProfilePropertyMethod = ({
  executeQuery,
  validateQuery = validateGenericQuery,
}) => {}

// then to opt-out

getProfileProperty(() => {}, null)
```

plugins/@grouparoo/app-templates/src/source/query/profileProperty.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (rows && rows.length > 0) {
+        if (!profilePropertyRule.isArray) {
+          const row = rows[0];
+          response = [Object.values(row)[0]];
+        } else {
+          response = rows.map((row) => Object.values(row)[0]);
+        }
+      }
+    } catch (error) {
+      throw new Error(`Error with Query: - \`${query}\`, Error - ${error}`);
```
evantahler: Since this is going to be used generically, the error message might want to include the source + app names/guids

plugins/@grouparoo/app-templates/src/source/table/getExamples.ts
```diff
@@ -0,0 +1,119 @@
...
+    connection: any;
+    appOptions: SimpleAppOptions;
+    tableName: string;
+    getSampleRows: GetSampleRowsMethod;
+    getColumns: GetColumnDefinitionsMethod;
+  }): Promise<
+    Array<{
+      key: string;
+      description?: string;
+      examples?: Array<any>;
```
evantahler: This might be type `examples?: DataResponse`?

plugins/@grouparoo/app-templates/src/source/table/getExamples.ts
```diff
@@ -0,0 +1,119 @@
...
+} from "./pluginMethods";
+
+export interface GetExampleRowsMethod {
+  (argument: {
+    connection: any;
+    appOptions: SimpleAppOptions;
+    tableName: string;
+    columns?: ColumnDefinitionMap;
+    getSampleRows: GetSampleRowsMethod;
+    getColumns: GetColumnDefinitionsMethod;
```
evantahler, line 19: I think `getColumns` is optional (?) if columns is provided?

plugins/@grouparoo/app-templates/src/source/table/getExamples.ts
```diff
@@ -0,0 +1,119 @@
...
+  if (!rows || rows.length === 0) {
+    if (!columns) {
+      columns = await getColumns({ connection, appOptions, tableName });
+    }
+    rows = makeRowsFromColumns(columns);
+  }
+  return rows;
+};
+
+function makeRowsFromColumns(columns: ColumnDefinitionMap): DataResponseRow[] {
```
evantahler, line 69: Won't this method null-out all the data passed in?

plugins/@grouparoo/app-templates/src/source/table/pluginMethods.ts
```diff
@@ -24,8 +30,6 @@ export enum FilterOperation {
   NotContain = "does not contain",
 }
 
-export type DataResponse = string | number | boolean | Date;
-
 export interface MatchCondition {
   columnName: string;
   filterOperation: FilterOperation;
```
evantahler, line 35: The `relative*` ProfilePropertyRuleFiltersWithKey keys are for relative dates, ie: "greater than 3 days ago".  This logic is implemented in Group rules, but not yet in filters.

evantahler (commented):
General Thoughts:

* I really love how this simplifies the plugins. Goal Accomplished!  
  * For example, I see that `buildConnection()` really handles a lot for you, like knowing when to call `getSampleRows` for the various cases we need a preview. 
  * I think this will have a positive effect on the future work of importing Profile Properties in bulk - we can define a a prototypical method once and extend
* With Snowflake as the example, I'm curious if you expect methods `makeWhereClause` to propagate generically to all sources, or is this something special for those sources whose connections don't type-cast results to JS and back properly?  For example, `getChangedRowCount` has quite a few steps.
* Since you are adding a lot of new Types and Methods, I think block comments above each type would be helpful, as they propagate though the editor 

I'd suggest doing PG or MySQL next to get a sense for what they look like.

plugins/@grouparoo/app-templates/src/source/table/getExamples.ts
bleonard: that's true, but this is an internal method so I kept it simple. Not even sure how to type that.

plugins/@grouparoo/app-templates/src/source/table/getExamples.ts
bleonard: The point of these was to return _something_ for previews even on empty tables.
It only sets keys on `sample` and doesn't affect what's passed in. 
It's the same as https://github.com/grouparoo/grouparoo/blob/3dff7b5b567b2d9289015ed5e79e228edffe67af/plugins/%40grouparoo/bigquery/src/lib/util.ts#L159-L164

plugins/@grouparoo/app-templates/src/source/table/pluginMethods.ts
bleonard: Well cool. Then we can change it here in one place if it does!

PR #845, opened 2020-10-12
Move @types to dependencies so they are available for client install process

evantahler:
Normally, when distributing a typescript project, you ship the compiled JS to clients with declaration (`*.d.ts`) files.  Since you consumers aren't using the typescript files directly, you also shouldn't include any `@types` helper packages as dependencies, as they won't be used, or, will already have the relevant types exported to the `*.d.ts` files.

However, `@grouparoo/core` does have our users really re-compiling the typescript project from scratch.  We do this in case there are plugins installed that modify/create pages for the UI.  Our ORM, [`sequelize-typescript`](https://github.com/RobinBuschmann/sequelize-typescript) makes use of a few peer dependencies when determining the types of its responses, `@types/bluebird` being the most important.  Without `@types/bluebird`, the response types from built-in model methods, like `findAll()` and `count()` are `any`.  

Sometimes, the type-behavior when operating on something `any` is dramatically different than the real type:

**with `@types/bluebird`, comparing 2 numbers is OK**
Screen Shot 2020-10-12 at 2 57 27 PM

**without `@types/bluebird`, things get wacky**
Screen Shot 2020-10-12 at 2 57 58 PM
Screen Shot 2020-10-12 at 2 57 53 PM

In this particular example, mapping an array of `any` somehow becomes an array of __exactly__ 10 `unknown`.   Why 10?  Who knows.
Screen Shot 2020-10-12 at 3 06 39 PM
Screen Shot 2020-10-12 at 3 06 45 PM

This PR moves our `@types` packages from `devDependencies` to `dependencies` to prevent this class of bug.


evantahler:
While `sequelzie-typescript` (without `@types/bluebird`) is responsible for returning `any` from it's model methods... it's Typescript in general which makes the array of exactly 10 elements

ts

evantahler:
@notwoods answered the question of why TS returns an array of exactly 10 `unknown` elements for us on his blog!  https://tigeroakes.com/posts/any-wreaks-havoc-on-overloads/

PR #848, opened 2020-10-14
Test fixes

evantahler:
Closes T-619
Closes T-620
Closes T-623
Closes T-614
Closes T-618


evantahler:
https://github.com/grouparoo/grouparoo/pull/848/commits/5e0c54b212e2e4c2528872458fac14380f9ebeb2 is interesting - using `<form autoComplete="off">` seems to be a good way to tell the browser not to remember what the previously submitted content in the form was when leaving the page and coming back to it.  I only did the "edit" pages for now, but we may want to extend this to other forms too.  

The edit pages are the ones most likely to have local state that hasn't been submitted.

PR #852, opened 2020-10-14
Use the same code for the event stream as the demo

bleonard:
sharing the code and getting the same types of events
Closes T-550


evantahler:
Want to work together to craft a test for this?

bleonard:
Not against it, per se, and maybe for more of the demo stuff. But not at the top of the list. 
I'm was trying to quickly respond to @andyjih to have the stream match up with the demo data.

evantahler:
@bleonard please add labels to the PRs so they appear in release notes

evantahler (approved):

PR #853, opened 2020-10-15
App.connect() should be locked to ensure that only one worker builds the connection

evantahler:
Prevents errors like the following at startup (likely visible in Resque):

```
Error: Error with Postgres SQL Statement: Query - SELECT COALESCE(SUM("price"), 0) as __result FROM %I WHERE "user_id" = %L AND "state" = %L LIMIT 1, Error - Error: Client was closed and is not queryable
```

Closes T-625

### DIFF
--- a/core/api/src/modules/ops/app.ts
+++ b/core/api/src/modules/ops/app.ts
@@ -134,7 +134,6 @@ export namespace AppOps {
   }
 
   async function getConnectionLock(app: App) {
-    console.log("waiting...");
     const key = `app:${app.guid}:connection:${id}:`;
     return waitForLock(key);
   }
--- a/core/api/src/modules/ops/app.ts
+++ b/core/api/src/modules/ops/app.ts
@@ -1,38 +1,65 @@
-import { api, log } from "actionhero";
+import { api, log, id } from "actionhero";
 import { App, SimpleAppOptions } from "../../models/App";
+import { waitForLock } from "../locks";
 import { OptionHelper } from "../optionHelper";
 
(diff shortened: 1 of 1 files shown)


core/api/src/modules/ops/app.ts
```diff
@@ -1,38 +1,65 @@
...
-      const connection = await pluginApp.methods.connect({
-        app,
-        appGuid: app.guid,
-        appOptions: options ? options : appOptions,
-      });
-      app.setConnection(connection);
-      return connection;
+    if (!pluginApp.methods.connect) return;
+
+    const appOptions = await app.getOptions();
```
bleonard, line 19: i'd suggest a `finally` for all lock situations.

core/api/src/modules/ops/app.ts
```diff
@@ -103,4 +132,9 @@ export namespace AppOps {
 
     return { success, message, error };
   }
+
+  async function getConnectionLock(app: App) {
+    const key = `app:${app.guid}:connection:${id}:`;
```
bleonard: wonder if we should make a `objectLock` like we did for `objectCache`

core/api/src/modules/ops/app.ts
evantahler: 👍

core/api/src/modules/ops/app.ts
evantahler: Possibly! This one's a little special though, as it's a lock key per node in the cluster.  `id` is something like `hostname+pid`

PR #875, opened 2020-10-19
Add tests for getChangedRowCount and rename `__hwm`

evantahler:
Remove a space between `COUNT (*)` in mySQL, but it wasn't the cause of any problems...

This PR also renames `__hmw ` to `__hwm `


evantahler:
Closes T-635

PR #876, opened 2020-10-19
MySQL and Postgres apps use connection pooling

evantahler:
When connecting to Postgres and MySQL Sources (and Destinations), we should use a connection pool.  This will create more connections between the node.js process and the database, which should speed up imports and exports.  

Using a connection pool in this way /may/ also help solve a class of bugs in which a connection is "polluted" by a previous request, like `Cannot enqueue Query after invoking quit` or `Cannot enqueue Query after fatal error` for MySQL connections.

---

From https://github.com/mysqljs/mysql
> With Pool, disconnected connections will be removed from the pool freeing up space for a new connection to be created on the next getConnection call.

and https://github.com/mysqljs/mysql/issues/1694 hits that if we aren't using the pool, we'll need to manually check if `error.fatal` and rebuild the connection ourselves


andyjih:
Closes T-644

PR #877, opened 2020-10-20
Store Profile Property Rules Cache in Redis

evantahler:
Closes T-642


bleonard:
Seems like a good case to refactor to `objectCache` at some point.

PR #881, opened 2020-10-21
Demo events get `addToCart` and `purchase` events.

bleonard:
Also added to demo stream.

`npm run actionhero -- grouparoo demo eventStream`
`npm run actionhero -- grouparoo demo dataPurchases`

This also fixes a bug that concurrent events brought up related to `SequelizeConnectionAcquireTimeoutError` and T-484.


plugins/@grouparoo/demo/src/events.ts
```diff
@@ -34,7 +33,7 @@ export async function events() {
 
   const apiKey = await getApiKey();
   const categories = await getPurchaseCategories();
-  const purchases = await getPurchases();
+  const purchases = await getPurchases(250);
```
evantahler, line 36: Should this be a function of `--scale?`

plugins/@grouparoo/demo/src/bin/grouparoo/demo/eventStream.ts
```diff
@@ -61,7 +61,7 @@ export class Console extends CLI {
           sessions[i] = runner.makeSession(i);
         }
       }
-      setTimeout(tick, sleep);
+      await setTimeout(tick, sleep);
```
evantahler: I don't think you can await `setTimeout`.  You'll have to promise-ify it

core/api/src/modules/ops/event.ts
```diff
@@ -139,4 +136,22 @@ export namespace EventOps {
       }
     }
   }
+
+  async function createProfileFromAnonymousId(
```
evantahler, line 140: This looks like what we had on the Tuple call, but it wasn't enough to prevent the deadlocks... was there something else?

plugins/@grouparoo/demo/src/events.ts
bleonard: well this is base for `scale` and then that number can scale it up.

core/api/src/modules/ops/event.ts
bleonard: it worked with this and the `catch` left in. this made sure it was in both spots.

evantahler:
`internal` label added

PR #889, opened 2020-10-23
Allow the Grouparoo Database to use a self-signed SSL cert with DATABASE_SSL_SELF_SIGNED

evantahler:
You can set the environment variable `DATABASE_SSL_SELF_SIGNED=true` to use self-signed SSL certificates when connecting to a Grouparoo Database.


andyjih:
👍

PR #944, opened 2020-11-06
Properly consider alpha versions in semver ranges

evantahler:
We now use the [`compare-versions`](https://www.npmjs.com/package/compare-versions) package to determine if a package is out of date for the `/about` page and new-version notifier.  `compare-versions` knows how to support alpha releases with "extended" patch semver names like `1.2.3-alpha.4`


evantahler:
Closes T-689

PR #965, opened 2020-11-10
Grouparoo can use SQLite

### LINKED ISSUE #935: Feature Request: Grouparoo can be run locally without Postgres or Redis
To aid the developer onboarding & trial experience, we should be able to run Grouparoo without the need for Redis and Postgres.  We can use `fakeredis` and `sqlite` locally in this mode.

evantahler:
This PR makes it possible to run Grouparoo against a SQLite database.  This enables both in-memory operation and persisting the SQLite database to disk.  Per #935, this will make it easier to get started with Grouparoo.

## Features

- [x] With the first-time developer experience in mind, this PR changes the output of `grouparoo generate` to use SQLite by default
- [x] This PR removes any custom queries which would only work on Postgres
- [x] This PR removes any custom aggregations that only work on Postgres (or knows how to format things for both PG and SQLite)
- [x] This PR removes all instances of `findOrCreate` globally in the app... they keep causing problems! 
- [x] Re run (some) tests in CI against sqlite

## Questions
* ~~SQLite does not like the creation of new records in @`afterSave`/@`afterCreate` hooks.  This makes the LoggedModel hard to use.  We can choose to not `await` these, so they will be outside of the save transaction... but then we can't guarantee that the logs are written...  See https://github.com/grouparoo/grouparoo/pull/965/commits/1cbe2ac5c2b642b8959330d41027bb80b7807e39 for more info~~
* Hooks which add or create models (like using an `@afterSave` to create a Log record) need to use the transaction of the parent model, if there is a transaction in play.  SQLite only has one transaction available, so can't `await Log.create()` in a hook without the transaction... it will never complete.
* `iLike` (case-insensitive like queries) do not work for sqlite. - there are now a few conditionals scattered about the codebase based on `config.sequelize.dialect`

Part of #935
Closes T-673

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -240,6 +240,29 @@ jobs:
           command: cd core/web && ./../node_modules/.bin/jest --ci --maxWorkers 2
           environment:
             SELENIUM_REMOTE_URL: http://localhost:4444/wd/hub
+  test-core-local:
+    docker:
+      - image: circleci/node:12
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+      - image: redis:latest
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+    steps:
+      - checkout
+      - restore_cache:
+          <<: *node-module-cache-options
+      - restore_cache:
+          <<: *dist-cache-options
+      - restore_cache:
+          <<: *core-cache-options
+      - run:
+          name: test-api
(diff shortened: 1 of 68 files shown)


evantahler:
This PR is blocked by https://github.com/sequelize/sequelize/issues/12823

evantahler:
We can work around sequelize/sequelize#12823 by changing our migrations to include a default value for non-null columns in one migration. See https://github.com/grouparoo/grouparoo/pull/965/commits/5c1bb1bb5e02b53f82af959d1a9322048d6e1dcf for more info

PR #969, opened 2020-11-11
Mailchimp source and new destination

### LINKED ISSUE #790: Feature Request: Mailchimp Member Updates without email addresses
Plan:
- source that talks mailchimp to import generated mailchimpId and another other merge var
- that allows it to make to an internal id
- new destination for mailchimp that uses mailchimpId to update existing users.

bleonard:
Executing this plan: https://github.com/grouparoo/grouparoo/issues/790

- source that talks mailchimp to import generated mailchimpId and another other merge var
- that allows it to make to an internal id
- new destination for mailchimp that uses mailchimpId to update existing users.


plugins/@grouparoo/marketo/__tests__/integration/destination-mapping-options.ts
```diff
@@ -61,6 +61,6 @@ describe("marketo/destinationMappingOptions", () => {
       expect(urgency.type).toBe("float");
       expect(urgency.important).toBe(false);
     },
-    30 * 1000
+    1000 * 30
```
evantahler: should be `helper.setupTime` to match other tests

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -106,6 +106,8 @@ const models = [
 ];
 
 export namespace helper {
+  export const setupTime = 60 * 1000;
+
```
evantahler: That's a real shame.  If this is a problem just on CircleCI we can detect that with `process.env. CIRCLECI`.  I *think* there is value in keeping the timeout shorter normally to more quickly detect boot problems maybe?

plugins/@grouparoo/mailchimp/__tests__/data/profiles.csv
```diff
@@ -0,0 +1,11 @@
+EMAIL,FNAME,LNAME,UserID,LTV
+xdiston9@grouparoo.com,Dorthy,Iston,10,290.53
```
evantahler, line 2: lol, why `X`s?

plugins/@grouparoo/mailchimp/__tests__/import/profiles.ts
```diff
@@ -0,0 +1,243 @@
...
+  SimpleDestinationOptions,
+} from "@grouparoo/core";
+
+const nockFile = path.join(__dirname, "../", "fixtures", "import-profiles.js");
+
+import { profiles } from "../../src/lib/import/profiles";
+
+// these comments to use nock
+const newNock = false;
+require("./../fixtures/import-profiles");
```
evantahler, line 27: FYI I've been changing these to be `require (nockFile)`

plugins/@grouparoo/mailchimp/__tests__/import/profiles.ts
```diff
@@ -0,0 +1,243 @@
...
+    importsCount,
+    sourceOffset: nextSourceOffset,
+  };
+}
+
+describe("mailchimp/import/profiles", () => {
+  beforeAll(async () => {
+    const env = await helper.prepareForAPITest();
+    actionhero = env.actionhero;
+  }, 1000 * 60);
```
evantahler: `helper.timeout` probably?

plugins/@grouparoo/mailchimp/__tests__/integration/mailchimp-export-id.ts
```diff
@@ -0,0 +1,350 @@
...
+    try {
+      const mailchimpId = generateMailchimpId(email);
+      await client.delete(`/lists/${listId}/members/${mailchimpId}`);
+    } catch (err) {}
+  }
+}
+
+async function cleanUp(suppressErrors) {
+  await deleteUsers(suppressErrors);
+}
```
evantahler, line 81: `suppressErrors` isn't used?

plugins/@grouparoo/mailchimp/__tests__/integration/mailchimp-import.ts
```diff
@@ -0,0 +1,446 @@
...
+        session = await specHelper.buildConnection();
+        session.params = { email: "mario@example.com", password: "P@ssw0rd!" };
+        const sessionResponse = await specHelper.runAction(
+          "session:create",
+          session
+        );
+        expect(sessionResponse.error).toBeUndefined();
+        csrfToken = sessionResponse.csrfToken;
+
+        // create a sheet app with an uploaded file
```
evantahler: // mailchimp app

evantahler: There are a couple places in this file that say `sheet`

plugins/@grouparoo/mailchimp/src/lib/import/profileProperty.ts
```diff
@@ -0,0 +1,7 @@
+import { ProfilePropertyPluginMethod } from "@grouparoo/core";
+
+export const profileProperty: ProfilePropertyPluginMethod = async () => {
+  // we always do nothing when called by profile update... we only exist to be used by the schedule & runs
+  // TODO: should this be made optional in the plugin definition?
```
evantahler, line 5: https://grouparoo.height.app/general/T-707

core/api/src/tasks/schedule/run.ts
```diff
@@ -29,6 +30,10 @@ export class ScheduleRun extends RetryableTask {
     await run.afterBatch();
 
     if (importsCount > 0) {
+      if (process.env.NODE_ENV === "test") {
+        // prevents "job already enqueued" in tests
+        params.count = (params.count || 0) + 1;
+      }
```
evantahler: Which test was running into this error?  There's `specHelper.deleteEnqueuedTasks()` from `actionhero` which might be helpful

evantahler: https://docs.actionherojs.com/modules/spechelper.html#deleteenqueuedtasks

evantahler (commented):
Lots to like here! 
* Thanks for fixing up the test timeouts
* I think this pattern of adding new Connections for different source/destination behavior is the right thing to do (we'll make it better in the UI one day)

Things to talk about 
* ~~What's the migration path look like for existing `mailchimp-export` destination connection users look like?~~ All good! no name change

plugins/@grouparoo/mailchimp/__tests__/data/profiles.csv
bleonard: mailchimp remembers old ones i deleted, so I burned all the original ones in an early iterations 🤷

plugins/@grouparoo/mailchimp/__tests__/import/profiles.ts
bleonard: does that work? I thought I saw issues there originally.
Like it was run in some pre-compile step or something.

evantahler: seems OK here? https://github.com/grouparoo/grouparoo/blob/master/plugins/%40grouparoo/bigquery/__tests__/table-import/import-properties.ts#L19-L28

core/api/src/tasks/schedule/run.ts
bleonard: the ones that looks this seems to be a risk
```
        // check that the run is enqueued
        const found = await specHelper.findEnqueuedTasks("schedule:run");
        expect(found.length).toEqual(1);
        expect(found[0].args[0].scheduleGuid).toBe(schedule.guid);

        // run the schedule
        const run = await Run.create({
          creatorGuid: schedule.guid,
          creatorType: "schedule",
          state: "running",
        });
        await specHelper.runTask("schedule:run", {
          runGuid: run.guid,
          scheduleGuid: schedule.guid,
        });
```

If it re-enqueues itself and falls on the same millisecond.
Maybe that's why there are these sleeps in the mysql one
```
      await helper.sleep();
      let foundAgain = await specHelper.findEnqueuedTasks("schedule:run");
      expect(foundAgain.length).toEqual(2);
      await specHelper.runTask("schedule:run", foundAgain[1].args[0]);

      await helper.sleep();
      foundAgain = await specHelper.findEnqueuedTasks("schedule:run");
      expect(foundAgain.length).toEqual(3);
      await specHelper.runTask("schedule:run", foundAgain[1].args[0]);
```

bleonard: I'll remove the change to the task and add sleeps.

PR #977, opened 2020-11-14
Update Batch Imports methods for BigQuery + Harden Batch Imports tests

evantahler:
Fixes a bug in which BigQuery batch imports were failing to import more than one person

Closes T-713


plugins/@grouparoo/bigquery/src/lib/table-import/getPropertyValues.ts
```diff
@@ -90,7 +90,11 @@ export const getPropertyValues: GetPropertyValuesMethod = async ({
     types
   );
   if (addAnd) query += ` AND`;
-  query += ` ${inClause}`;
+  // for BigQuery we need to use UNNEST: `id in UNNEST(1,2,3)`
+  // See https://github.com/googleapis/nodejs-bigquery/blob/master/samples/queryParamsPositionalTypes.js#L37
+  query += ` ${inClause}`.replace(" IN (", " IN UNNEST (");
```
bleonard, line 95: it feels like  this change in be `filterOperation` where the type is `FilterOperation.In` - would that work?

evantahler: Oh! Great idea! Thanks!
https://github.com/grouparoo/grouparoo/pull/979

PR #993, opened 2020-11-18
Grouparoo can be run locally without Redis

### LINKED ISSUE #935: Feature Request: Grouparoo can be run locally without Postgres or Redis

evantahler:
Grouparoo can be run without the `redis` dependency, by opting into using [`ioredis-mock`](https://github.com/stipsan/ioredis-mock) instead.  

⚠️  DANGERZONE - Using `ioredis-mock` will not persist any data.  You cannot use Grouparoo in a cluster, or more than once without data loss between runs. ⚠️

Depends on https://github.com/actionhero/actionhero/pull/1653 and https://github.com/actionhero/node-resque/pull/475

Closes T-674
Closes #935


evantahler:
@bleonard check out https://github.com/grouparoo/grouparoo/pull/993/commits/6c309e01139f4fbad966ea25ad503e329e8e4163 I made some updates to `demo/dataPurchases`

1. We no longer need to `api.resque.startQueue()` - Actionhero does this when running a CLI command now
2. Since the schedule you make is recurring, I removed `runAction("schedule:run", { guid: made.guid })` - it'll happen soon anyway
3. I ended the script with `stopRuns()` - Since we are starting from a place of 0 profiles in the database, the `internalRuns` created by the new Profile Property Rules aren't doing anything - there are no profiles that need to be updated.

2 and 3 were problematic in a world without Redis - There would be a run with no corresponding task when the server boots up...

evantahler:
Unlike Sqlite (https://github.com/grouparoo/grouparoo/pull/965) I don't think we need anything additional in the test suite for `ioredis-mock`.  We have no custom logic depending on if you are using "real" Redis or "mock" Redis.   The library itself is tested for compatibility within Actionhero - https://github.com/actionhero/actionhero/blob/master/__tests__/integration/ioredis-mock.ts and node-resque - https://github.com/actionhero/node-resque/blob/master/__tests__/integration/ioredis-mock.ts

apps/staging-public/.env.example
```diff
@@ -22,6 +22,10 @@ SERVER_TOKEN=change_me
 ## REDIS ##
 ###########
 
+# To use an in-memory redis mock (no persistence)
+# REDIS_URL=""
```
bleonard: I wonder if you should have to in more explicitly instead of leaving it off. It would be less surprising if it was like `REDIS_URL=memory://` or something.

bleonard (approved):

apps/staging-public/.env.example
evantahler: good idea! https://github.com/grouparoo/grouparoo/pull/993/commits/ac8e05884392553d0c30ef0740857528a8efcd58

PR #996, opened 2020-11-19
Demo can handle SQLite data

bleonard:
If you are using SQLite, you'll now need to set something like this for where to put the demo users and purchases.
`DEMO_DATABASE_URL="postgresql://localhost:5432/grouparoo_development"`


evantahler:
Woo!

evantahler:
Closes T-700

PR #1019, opened 2020-11-24
App, Source, Destination, and Profile Property Rule options gain displayName property

evantahler:
With the `displayName` property, that means we can have both a `key` and `disaplayName` for the options. `aggregationMethod` would be the key and `Aggregation Method` would be the display name.

Screen Shot 2020-11-23 at 4 21 34 PM

This PR includes fixes for the common options:
* displayName / Display Name
* sortColumn / Sort Column
* column / Column Name

Closes T-738


evantahler:
This is a help to https://github.com/grouparoo/grouparoo/pull/1001, which means the config files can use `{aggregationMethod: "sum"}` as keys for options, not `{"aggregation method":"sum" }`

PR #1021, opened 2020-11-25
Demo config

bleonard:
Use the code config code path to create demo data.
`npm run actionhero -- grouparoo demo config` write config to `config` directory.


plugins/@grouparoo/demo/package.json
```diff
@@ -28,8 +28,10 @@
   },
   "dependencies": {
     "csv-parse": "4.14.1",
+    "fs-extra": "9.0.1",
     "isomorphic-fetch": "3.0.0",
     "pg": "8.5.1",
+    "prettier": "2.2.0",
```
evantahler, line 34: Why move prettier to `dependencies`?

plugins/@grouparoo/demo/src/util/shared.ts
```diff
@@ -16,6 +17,20 @@ export async function execSync(level: number, command) {
...
+  );
+  const pConfig = path.resolve(
+    path.join(__dirname, "..", "..", "..", "..", "..", ".prettierrc")
+  );
+  await execSync(
+    2,
+    `'${pCmd}' --config '${pConfig}' --write '${fileOrDirPath}'`
+  );
+}
+
```
evantahler, line 33: Ah! I see why you use prettier now :D
It might be better to use the built-in API rather than shelling out - https://prettier.io/docs/en/api.html

evantahler (commented):
It seems a shame that there wasn't more code you could delete - to make the apps or sources?

evantahler:
Closes T-758

plugins/@grouparoo/demo/src/util/shared.ts
bleonard: I don't see a way to do the whole directory in the API

evantahler: Try out the `glob` package to get the list of files in a directory. 

For code customers might run, you can't rely on the path structure you are using here.  
`node_modules/.bin` doesn't exist on windows for example, or if you opt to install your packages globally.   In general, we should avoid 'shelling out' whenever possible (tests and other internal tools being the exception).

PR #1022, opened 2020-11-25
Sweep un-associated  profileProperties

evantahler:
There are rare occurrences when it is possible to leave an un-associated Profile Property around after the deletion of it's parent, either a Profile or Profile Property Rule.  These may occur when an import task is running as a Profile Property Rule is deleted, or a 2 Profiles have been merged.

This PR introduces a new periodic task `profileProperties:sweep` to clean these up.

An alternative approach was attempted using foreign key constraints, but that was not possible (see https://github.com/RobinBuschmann/sequelize-typescript/issues/755 for similar issue)

Closes T-772

### DIFF
--- a/core/api/__tests__/actions/groups.ts
+++ b/core/api/__tests__/actions/groups.ts
@@ -227,7 +227,7 @@ describe("actions/groups", () => {
       });
 
       beforeAll(async () => {
-        await Profile.destroy({ truncate: true });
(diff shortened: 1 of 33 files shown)


bleonard:
Looks good. I wonder if there's also something to do in the workers that are writing to that table. "Oh, that profile or rule doesn't exist anymore. Exit gracefully."

PR #1026, opened 2020-11-26
code configure demo event property rules

bleonard:


plugins/@grouparoo/demo/src/configFiles.ts
```diff
@@ -64,3 +69,28 @@ function updatePurchases(configDir) {
...
+  // unlock these for the demo so they can be shared
+  const models = api.sequelize.models;
+  for (const name in models) {
+    const Model = models[name];
+    // if (["Team"].includes(name)) {
+    //   continue;
+    // }
+    const attributes = Model.rawAttributes;
+    if (attributes.locked) {
+      log(3, `Unlocking ${name}`);
```
evantahler, line 83: probably want to remove this log?

plugins/@grouparoo/demo/src/configFiles.ts
```diff
@@ -64,3 +69,28 @@ function updatePurchases(configDir) {
   const out = JSON.stringify(contents);
   fs.writeFileSync(appPath, out);
 }
+
+async function unlockAll() {
```
evantahler, line 73: I don't think I understand what this method is used for.  I'm guessing in the demo you want to use code-config to create all the objects but not have them "locked" so that they can be modified in the UI?

Is there a more general use-case for `locked = false` in code config?

plugins/@grouparoo/demo/src/configFiles.ts
bleonard: This is the demo system log that only logs when they want it more verbose.

plugins/@grouparoo/demo/src/configFiles.ts
bleonard: Yes, I wanted the demo to allow the user to change it.

It also makes it possible to do some of them at a time. It was necessary to run the `userId` rule and related source. Then run some event code. Then run the events config. If the lock was still there it was deleting the earlier ones.

evantahler (approved):

PR #1033, opened 2020-11-30
Use JSON5 to parse Code Config files

evantahler:
JSON5 is a superset of JSON which allows comments and multi-line strings.  Learn more about JSON5 at https://json5.org

The following config file is now valid for Grouparoo:
```json5
[
  {
    // Comments are allowed
    "id": "setting_cluster_name",
    "class": "Setting",
    "pluginName": "core",
    "key": "cluster-name",
    "value": "Evan's Cluster"
    /* Even multiline comments
       Are ok.
    */
  }
]

```

### DIFF
--- a/core/api/src/modules/configLoaders/all.ts
+++ b/core/api/src/modules/configLoaders/all.ts
@@ -1,5 +1,6 @@
 import { log, config, env } from "actionhero";
 import path from "path";
+import fs from "fs";
 import glob from "glob";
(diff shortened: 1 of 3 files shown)


evantahler:
Added to code-config docs via https://github.com/grouparoo/www.grouparoo.com/pull/93

evantahler:
Closes T-801

PR #1039, opened 2020-11-30
`model.locked` is a string indicating why the model is locked

### LINKED ISSUE #932: Feature Request: Grouparoo's code config can be validated
A CLI command, `grouparoo validate` can be run to test your code-based configuration.  We will test your `Apps`, and perhaps even a profile import and export. 

Requires #934

evantahler:
This PR changes the database schema to now store *why* a model is locked.  This separates locked models created by `config:code` from other types of locked models, mainly Teams which were created using the `team:initialize` action/UI.   This way, those models not created by code config will not be modified or deleted by code config. 

This PR also addresses https://github.com/sequelize/sequelize/issues/12823, which is a bug in Sequelize which prevents migrating any table in SQLite which has a compound index.  We previously had used compound indexes to guarantee there was only one of something across a join table.  For example, we previously used a compound index to ensure that there was only one Profile Property per Profile and Profile Property Rule at any position; ensuring you couldn't have two `email` properties (unless `email` was an array property).  This was a strong validation.  

These indexes had to be removed in order to be compatible with SQLite + Sequelize.    The validations are now done in a `@BeforeSave` hook on the model.  If we observe race condition errors with more than one property appearing for the same profile + property + position, we can add a singly-unique "validation column" whose string value is a concatenation of all the columns we want to be unique. 

Closes T-804
Closes T-806


core/api/src/migrations/000053-locekdString.ts
```diff
@@ -0,0 +1,49 @@
...
+      if (table !== "teams") {
+        await migration.sequelize.query(
+          `UPDATE "${table}" SET locked='config:code' where locked IS NOT NULL`
+        );
+      } else {
+        await migration.sequelize.query(
+          `UPDATE "${table}" SET locked='team:initialize' where locked IS NOT NULL and name = 'Administrators' and "permissionAllWrite" = true and "permissionAllRead" = true`
+        );
+        await migration.sequelize.query(
+          `UPDATE "${table}" SET locked='config:code' where locked IS NOT NULL and name != 'Administrators'`
```
evantahler: These migrations are a bit gnarly, but I don't know a better way at the moment to distinguish between an Initialized team and a code-config Team.

core/api/src/migrations/000053-removeCompoundindex.ts
```diff
@@ -0,0 +1,68 @@
+module.exports = {
```
evantahler, line 3: @bleonard I'm pretty sad about this whole situation.  Have any alternative ideas?

bleonard: because it's not possible in SQLite?

bleonard: I wonder if it could only do the first one in SQL lite somehow but we could still optimize postgres. The postgres instances are much better, so I'd want to optimize them a bit. Some `dynamicIndex` method that did something else if in SQLlite.

evantahler: That's an interesting idea - We could:
* Always include model-level (in code) validations which would work for PG and SQLite
* If the `config.seqeulize.dialect === 'postgres'`, we could also add compound index, to be extra safe

bleonard: Sounds good. It will be super clean if we wrap that in a method that is more optional and no-op on sqlite.

evantahler:
Documentation relating to Compound Indexes added to WWW via https://github.com/grouparoo/www.grouparoo.com/pull/100

core/api/src/migrations/000054-lockedString.ts
```diff
@@ -0,0 +1,48 @@
...
+    for (const i in tables) {
+      const table = tables[i];
+      await migration.changeColumn(table, "locked", {
+        type: DataTypes.STRING(191),
+        allowNull: true,
+        defaultValue: null,
+      });
+      if (table !== "teams") {
+        await migration.sequelize.query(
+          `UPDATE "${table}" SET locked='config:code' where locked IS NOT NULL`
```
bleonard, line 26: @evantahler I think this was a bug. They were `false` so this set everything as 'config:code'
I'd consider switching this to `where locked = "true"` (or something close to that)

evantahler: well shit.

Thanks for finding that!

evantahler: Will be resolved by https://github.com/grouparoo/grouparoo/pull/1048

Issue #1043, opened 2020-12-01
Error creating Leads in Salesforce


bleonard:
**Describe the bug**
When syncing Leads to Salesforce, user is seeing an error "cannot reference converted lead"

**To Reproduce**
1. convert a lead to a contact in salesforce (this makes the lead read-only a produces error when editing via UI or API)
2. sync to leads object using the same email address.

**Expected behavior**
The ideal case is that the Contact (which has many of the same fields as Lead) is updated instead of the Lead.
The user doesn't want to sync every profile to Contacts, just update the ones that were leads.

bleonard:
A few options come to mind:

1) Add Salesforce as a source to get an understanding who is already a Contact. Make it so no one that is already a Contact is part of the group being sent to the Leads destination.
2) Maybe this just shouldn't sync to Leads because they aren't leads any more. Ignore that error, skipping that record.
3) Have the generic Salesforce Object sync recognize this error and update the Contact record where the fields overlap.
4) Do `2` above (skip the leads error) and have the user also sync to contacts. There would be options added to Salesforce destinationOptions to only update existing users. They also wouldn't want users to be deleted. This would have the effect of updating all the ones that would have been leads otherwise.
5) Add this behavior (with or without option) to the generic salesforce object destination. There's some decisions to make about how fields line up. Probably start with the ones that have the same names.
6) Make a new Salesforce destination for Leads, that has this explicit behavior (with or without an option). Assume the fields names line up like in `5`.

bleonard:
Did solution 2 here: https://github.com/grouparoo/grouparoo/pull/1051
Will add on solution 4 in future PR.

bleonard:
Did solution 4 here: https://github.com/grouparoo/grouparoo/pull/1052

PR #1051, opened 2020-12-03
Salesforce: converted leads are not errors

bleonard:
Solution 2 from this issue.
https://github.com/grouparoo/grouparoo/issues/1043

If we encounter a lead that have been converted:
* don't error out
* make sure it's not in any groups

Closes T-829


bleonard:
An interesting learning from this is that converted leads can be removed from campaigns, but not added.
It's also interesting that in the UI they show up as their (now converted) Contact, but the `CampaignMember` record still has the `LeadId`.

PR #1053, opened 2020-12-04
@grouparoo/ui package

evantahler:
The Web UI is now in its own `@grouparoo/ui` package.

This is a breaking change as `@grouparoo/ui` will need to be added to client `package.json` files to continue to use the Grouparoo web UI.

1. `npm install --save @grouparoo/ui`
2. enable `@grouparoo/ui` in your `package.json`

Closes T-774


evantahler:
The failing test in CI is for our CLI tool `npx grouparoo generate` - which is trying to install the new `@grouparoo/ui` package from NPM before it is published

PR #1069, opened 2020-12-07
Destination refactor and errorLevel

bleonard:
`exportProfile` and `exportProfiles` can now say an error that says it is just "info" via `error.errorLevel = "info"` - this is true if it it is thrown or returned.
if this happens, it will not retry it. but it will recorded on the Export.
It will also not be factored into future exports. For example, it will not count towards `hasChanges`.

Also a pretty major refactoring to combine the code paths for single and batch exports.


core/src/models/Export.ts
```diff
@@ -137,7 +150,18 @@ export class Export extends Model<Export> {
   @BelongsTo(() => Profile)
   profile: Profile;
 
-  async markMostRecent() {
+  async setError(error: Error) {
+    this.errorMessage = error.toString();
+    if (error["errorLevel"]) {
+      this.errorLevel = error["errorLevel"];
+    }
+    // QUESTION: are we sure this shouldn't be complete?
```
bleonard: @evantahler maybe not for this PR, but some interesting semantics to talk about here.

core/src/models/Destination.ts
```diff
@@ -484,7 +484,10 @@ export class Destination extends LoggedModel<Destination> {
   @BeforeDestroy
   static async waitForPendingExports(instance: Destination, { transaction }) {
     const pendingExportCount = await instance.$count("exports", {
```
bleonard, line 486: It seemed important to be able to delete a destination if it ever had an error, but I still tried to keep the same semantics. Still some possible race conditions that would create errors in Resque.

core/src/classes/plugin.ts
```diff
@@ -225,6 +225,7 @@ export interface ExportProfilesPluginMethod {
 
 export interface ErrorWithProfileGuid extends Error {
   profileGuid: string;
+  errorLevel: "error" | "info";
```
evantahler: Fun Typescript Tricks - Since you made `errorLevels` a const (Export model) you could do `typeof errorLevels[number]` which will let you source that Enum from one place

core/src/models/Export.ts
```diff
@@ -72,6 +76,15 @@ export class Export extends Model<Export> {
   @Column
   errorMessage: string;
 
+  @Is("ofValidErrorLevel", (value) => {
+    if (value && !errorLevels.includes(value)) {
+      console.log("here2");
```
evantahler: Errant log statement

core/src/models/Export.ts
```diff
@@ -72,6 +76,15 @@ export class Export extends Model<Export> {
...
+      throw new Error(`errorLevel must be one of: ${errorLevels.join(",")}`);
+    }
+  })
+  @Column
```
evantahler: If you wanted even stronger typing of the column based on the allowed `errorLevels` - `@Column(DataType.ENUM(...errorLevels))`

evantahler:
Closes T-836

core/src/migrations/000055-errorLevelForExports.ts
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  up: async function (migration, DataTypes) {
+    await migration.addColumn("exports", "errorLevel", {
+      type: DataTypes.STRING(191),
+      allowNull: true,
+      defaultValue: null,
+    });
+
```
evantahler, line 8: From chat: Maybe we add another column to keep track of the Export really being "done" - the retries are completed

core/src/models/Destination.ts
evantahler: There's forthcoming work to change how Destinations get deleted in general - to be more like Groups and they can "drain" and delete themselves fully

core/src/models/Export.ts
evantahler: From chat: probably a new column to know that the export is "done" being retried

core/src/tasks/export/send.ts
```diff
@@ -18,7 +19,28 @@ export class ExportSend extends RetryableTask {
...
 
-    await destination.sendExport(_export);
+    const { success, retryDelay, error } = await destination.sendExport(
+      _export
+    );
+
+    if (!success) {
+      if (retryDelay) {
+        const app = await destination.$get("app");
+        return task.enqueueIn(
```
evantahler, line 34: From chat:  It's good to move all retry logic to the Task.

PR #1070, opened 2020-12-07
Grouparoo CLI Commands

evantahler:
```
→ (npx) grouparoo
Usage: grouparoo [options] [command]

Options:
  -V, --version                  output the version number
  -h, --help                     display help for command

Commands:
  console                        Start an interactive REPL session with the api object in-scope
  start                          Run the Grouparoo server
  status                         Display the status of your Grouparoo cluster
  demo-config                    Writes config to current app
  demo-data-purchases [options]  Load eCommerce users and purchases into a source database.
  demo-event-stream              Makes a continuous stream of user browsing events
  demo-setup                     Start from empty bootstrap
  generate [options] [path]      Generate a new Grouparoo project
  upgrade [path]                 Upgrade an existing Grouparoo project
  help [command]                 display help for command
```

Revamps the `grouparoo` CLI command in a number of ways:
* Can source additional CLI commands from the locally installed version of `@grouparoo/core` and plugins
* Cleans up the UI and output for `grouparoo generate`
* Stop including the default actionhero CLI commands (like `action generate`) to make our CLI only about Grouparoo

In addition, this PR:
* Removes the need for the environment variable `GROUPAROO_MONOREPO_APP` , moving this information into the `package.json` file for each app. 
* Upgrades to [Actionhero v25.0.0-alpha.5](https://github.com/actionhero/actionhero/releases/tag/v25.0.0-alpha.5), which includes support for these new CLI commands based on `commander`
* General improvements to logging to only use the actionhero logger, especially at startup with notes about ENV.  This will be useful later when adding verbosity arguments to the CLI

## Notes:
* All commands now occupy a single namespace, and cannot have spaces.  The `@grouparoo/demo` commands have ben renamed like `demo-config`, `demo-data-purchases`, etc.  We should use this pattern going forward.

# New Commands

## `grouparoo console`

Start an interactive CLI session in which the `api` object is already in-scope:
```
→ (npx) grouparoo console
[🦘::development ] >> api.sequelize.models.App.count().then(console.log)
Promise [Object] {
  _bitField: 0,
  _fulfillmentHandler0: undefined,
  _rejectionHandler0: undefined,
  _promise0: undefined,
  _receiver0: undefined
}
[🦘::development ] >> 2
```

## `grouparoo status`

Display Status about your Grouparoo cluster, similar to the UI Dashboard

```
→ (npx) grouparoo status

Status for evan's dev cluster (development):

Profiles: 1000 (0 pending)
Groups: 3
  People with Email Addresses: 998 members (most recent addition at 2020-12-09T18:25:10.934Z)
  A Small Group: 50 members (most recent addition at 2020-12-09T18:25:10.924Z)
  High Value Customers: 185 members (most recent addition at 2020-12-09T18:04:57.760Z)
Active Runs:
  People with Email Addresses - 15%
Pending Imports: 0
Pending Exports: 0
```

## `grouparoo start`

Start the Grouparoo server, the same as `npm start`


core/__tests__/bin/statuts.ts
```diff
@@ -0,0 +1,56 @@
...
+    beforeEach(() => {
+      messages = [];
+      spy = jest
+        .spyOn(console, "log")
+        .mockImplementation((message) => messages.push(message));
+    });
+
+    afterEach(() => {
+      spy.mockRestore();
+    });
```
evantahler, line 29: This is pretty awesome - you can spy on `console.log` in tests!

cli/src/utils/loadLocalCommands.ts
```diff
@@ -0,0 +1,142 @@
...
+  } catch (error) {
+    process.chdir(process.env.INIT_CWD);
+    throw error;
+  }
+}
+
+// --- UTILS ---
+
+function getActionhero() {
+  process.chdir(
```
bleonard, line 48: what's the story here? should we put this back?
could we require without changing dirs?

core/src/actions/groups.ts
```diff
@@ -4,7 +4,7 @@ import { Group, GROUP_RULE_LIMIT, TopLevelGroupRules } from "../models/Group";
 import { ProfilePropertyRuleOpsDictionary } from "../modules/RuleOpsDictionary";
```
bleonard, line 4: is this part of this PR?

bleonard (approved):

andyjih:
@evantahler Awesome! For `grouparoo status` is that interactive or static? Meaning will those numbers continually update or is it a snapshot of a single point in time?

andyjih (approved):
Looks good! More to iterate on for sure though. :)

cli/src/utils/loadLocalCommands.ts
evantahler: We change back to the original CWD here https://github.com/grouparoo/grouparoo/pull/1070/files#diff-e42b162691ea4dfc92fcb12a614b7cf16041729a65b645373f1520ea8d85323aR40-R42

And no... There's a lot of shenanigans we are doing about running the main app form /within/ `node_modules`, that we really need to always run everything from that working directory (it's why we `cd` in the package.json commands too).  It's further complicated because actionhero is a child of grouparoo is a child of the root project...

core/src/actions/groups.ts
evantahler: Yeah, because I moved the logic out of the action into the op so the new status command could use it as well.

evantahler:
@andyjih for sure!  None of the commands here are "done" - this is just a rough scaffold of how things will work to test that they can work :D

cli/src/utils/loadLocalCommands.ts
evantahler: Oh, and we want to *stay* in the changed working directory if we are running any command that needs to start up the application

PR #1073, opened 2020-12-08
Test that Group Rule comparisons are valid SQL before saving

evantahler:
If you try to add a Group Rule comparison that makes no sense, like ` group.setRules([{ key: "ltv", match: "fish", operation: { op: "gt" } }])` (comparing a number column to a string), the Group Rule will not be saved and the transaction reverted. 

Closes T-763.

### DIFF
--- a/core/__tests__/models/group/calculatedGroup.ts
+++ b/core/__tests__/models/group/calculatedGroup.ts
@@ -1,5 +1,5 @@
 import { helper } from "@grouparoo/spec-helper";
-import { api, specHelper } from "actionhero";
+import { api, specHelper, config } from "actionhero";
 import { Log } from "../../../src/models/Log";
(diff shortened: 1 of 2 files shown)


bleonard:
this handle invalid `"gt"` situations?
this is a great solve.
over time, we'd want to say `"gtx" is not valid, choose from "[gt, gte, etc]"`

evantahler:
We've already got that case covered above when we check that the rule's op is in the dictionary: https://github.com/grouparoo/grouparoo/blob/bb1fcca3e55f82637a2cc8b6f45c570eb2263c72/core/src/models/Group.ts#L236-L250

evantahler:
Hilariously, this comparison is valid for SQLite...
https://www.sqlite.org/faq.html#q3

PR #1079, opened 2020-12-10
SQLite default path within project root

evantahler:
Now, when you `grouparoo generate`, the default `.env` will place the SQLite database within your project's root directory.   

Closes T-875

### DIFF
--- a/cli/src/lib/generate.ts
+++ b/cli/src/lib/generate.ts
@@ -5,7 +5,7 @@ import { Templates } from "../utils/templates";
 import { buildLogger } from "../utils/logger";
 
 export default async function Generate(
-  workDir: string = process.cwd(),
+  workDir: string = process.env.INIT_CWD,
   program
 ) {
   const opts: {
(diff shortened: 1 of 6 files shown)


bleonard:
I think there's value in having that convention and the env works for all people out of the box.
I'd also want the env to work if they renamed the project directory. Breaking would be surprising.

Because of that, I actually thought we should change what `sqlite://grouparoo_development.sqlite` means.
What if that meant in the CWD of the `npm start` directory?

And `sqlite://db/grouparoo_development.sqlite` meant in the db subdirectory?

I suppose there could be some way to be anywhere on the system. Or maybe that's not that important. I'd prefer to keep it all in the same.
If it is necessary, relative could be `sqlite://./grouparoo_development.sqlite` and `sqlite://./db/grouparoo_development.sqlite` and one of those would be in the `.env.example`

evantahler:
> If it is necessary, relative could be `sqlite://./grouparoo_development.sqlite` and `sqlite://./db/grouparoo_development.sqlite` and one of those would be in the `.env.example`

Love this - this would be default flexible (rename the dir would work), but also follow shell conventions indicating it was a relative path

PR #1085, opened 2020-12-12
Rename `Profile Property Rules` to `Properties`

evantahler:
This was a long time coming, and pushed over the edge by the [CLI/Code Config](https://github.com/grouparoo/grouparoo/discussions/1054) work - typing `profilePropertyRules` every time is tedious!

This is a **holistic** change - including the database, models, plugin APIs, actions, routes, UI, etc. 

This would be a breaking change if you were already using Code Config

Closes T-878

**Notes:**

* The prefix for `Property` Guids remains `rul_` - We already had `pro_` (Profile) and `prp_` (Profile Property)... I couldn't think of a better one.  Also, not changing the prefix means we don't need a find/replace migration.

Screen Shot 2020-12-11 at 5 15 12 PM


evantahler:
@andyjih does this mean we can make the left-nav thinner?

evantahler:
Code Config Updates:

```js
    {
      id: "users_table", // guid -> `src_data_warehouse`
      name: "Users Table",
      class: "Source",
      type: "postgres-table-import",
      appId: "data_warehouse", // appGuid -> `app_data_warehouse`
      options: {
        table: "users",
      },
      mapping: {
        id: "user_id",
      },
      bootstrappedProperty: {
        name: "User Id",
        type: "integer",
        id: "user_id", // guid -> `rul_user_id`
        options: {
          column: "id",
        },
      },
    },

    {
      id: "email", // guid -> `rul_email`
      name: "email",
      class: "Property",
      type: "email",
      unique: true,
      identifying: true,
      isArray: false,
      sourceId: "users_table", // sourceGuid -> `src_users_table`
      options: {
        column: "email",
        aggregationMethod: "exact",
        sort: null,
      },
      filters: [],
    },
```

evantahler:
Also made the navigation slightly thinner after discussing with @andyjih 
Screen Shot 2020-12-14 at 11 08 52 AM

PR #1106, opened 2020-12-17
Grouparoo validate & Transactionification

evantahler:
The `grouparoo validate` command inspects your code config and reports any errors.

```
➜  grouparoo validate --help
Usage: grouparoo validate [options]

Validate your code config

Options:
  --externally-validate [externally-validate]  Should we validate the config by connecting to sources and destinations?
  -h, --help                                   display help for command
```

Screen Shot 2020-12-16 at 5 13 14 PM

Screen Shot 2020-12-16 at 5 22 31 PM

- [x] command works and reports errors
- [x]  test the command
- [x]  collect all errors into a single report (might not be possible with dependencies)
- [x] optional flag `--externally-validate` to skip option validation against external sources (ie: if you don't want the command to fail if your Hubspot API Key is missing) 

On the technical side, this is accomplished by attempting to apply the entire config set in one big Transaction.  This allows us to "really try" applying the changes and mixing them with other data in the database.  We then revert the transaction when we are done.  

* All hooks (@beforeCreate, @beforeSave, etc) and their dependent methods now use the transaction of the parent sequelize method.  Lots of optional `transaction` arguments have been added to many methods
* The `StateMachine.transition` method now expects `checks` to be real methods rather than method names on the instance.  This is required so that we can format the incoming transaction from the hook to the format the method wants, ie: 

```ts
// from models/Profile.ts

const STATE_TRANSITIONS = [
  { from: "draft", to: "pending", checks: [] },
  { from: "draft", to: "ready", checks: [] },
  {
    from: "pending",
    to: "ready",
    checks: [
      (instance: Profile, transaction?: Transaction) =>
        instance.validateProfilePropertiesAreReady(transaction),
    ],
  },
  { from: "ready", to: "pending", checks: [] },
];
```

Closes T-881
Closes T-884


cli/src/utils/loadLocalCommands.ts
```diff
@@ -150,7 +150,7 @@ async function convertCLIToCommanderAction(cli, program) {
     if (input.required && !input.default) {
       command.requiredOption(`--${key} <${key}>`, input.description);
     } else {
-      command.option(`--${key} <${key}>`, input.description, input.default);
+      command.option(`--${key} [${key}]`, input.description, input.default);
```
evantahler, line 153: Update CLI help syntax to use `[]` for optional commands, and `<>` for required options. options

Issue #1107, opened 2020-12-18
Feature Request: Facebook Lookalike Audiences Destination


bleonard:
Building on this feature request: https://github.com/grouparoo/grouparoo/issues/915

Enabled two more facebook concepts
* [Value-based custom audiences](https://developers.facebook.com/docs/marketing-api/audiences/guides/value-based-lookalike-audiences): `is_value_based=1` and `subtype=CUSTOM` along with `LOOKALIKE_VALUE` field
* [Lookalike audiences](https://developers.facebook.com/docs/marketing-api/audiences/guides/lookalike-audiences): These need a `lookalike_spec`

spatialy:
+1

Issue #915, opened 2020-10-28
Facebook Audiences Destination


bleonard:
**Is your feature request related to a problem? Please describe.**
Add Facebook Audiences to Grouparoo as a destination

**Describe the solution you'd like**
Sync profiles and groups with Facebook.

bleonard:
Here is the Facebook information on Audiences:

[https://developers.facebook.com/docs/marketing-api/audiences/guides/custom-audiences/](https://developers.facebook.com/docs/marketing-api/audiences/guides/custom-audiences/)

It looks like we share [hashed](https://developers.facebook.com/docs/marketing-api/audiences/guides/custom-audiences/#hash) data of key things like email. 

“Before you hash your data, normalize it so we can handle it. Only First name FN and Last Name LN support special characters and non-Roman alphabet. For best match results, provide the Roman alphabet translation with no special characters.” \


You can add an unlimited number of records to an audience, but only up to 10000 at a time. Changes to your Custom Audiences don't happen immediately and usually take up to 24 hours. The number of records you request to remove and/or the number of Custom Audiences your account contains will increase the amount of time it takes to process this request.

**Profiles & Groups**

This is a strongly group-based destination, so it would probably make sense to do this in batches.

You add and remove based on these hashed properties.

It’s tempting to use the EXTERN_ID to remove, but I think it’s less than ideal that the data retention period for EXTERN_ID is 90 days. So now we can’t remove them after that? Weird.

Here are the fields we can send…


<table>
  <tr>
   <td><code>EMAIL</code>
<p>
criteria: email addresses
   </td>
   <td>Hashing required.
<p>
Trim leading, trail whitespace, and convert all characters to lowercase.
   </td>
  </tr>
  <tr>
   <td><code>PHONE</code>
<p>
criteria: phone numbers
   </td>
   <td>Hashing required.
<p>
Remove symbols, letters, and any leading zeroes. You should prefix the country code if <code>COUNTRY</code> field is not specified.
   </td>
  </tr>
  <tr>
   <td><code>GEN</code>
<p>
criteria: gender
   </td>
   <td>Hashing required.
<p>
Use these values: <code>m</code> for male and <code>f</code> for female.
   </td>
  </tr>
  <tr>
   <td><code>DOBY</code>
<p>
criteria: birth year
   </td>
   <td>Hashing required.
<p>
Use the <code>YYYY</code> format from 1900 to current year.
   </td>
  </tr>
  <tr>
[...]

PR #1108, opened 2020-12-18
Facebook Custom Audiences Destination

bleonard:
Feature Request https://github.com/grouparoo/grouparoo/issues/915

Adds a facebook app and destination for custom audience. 
Lookalike audiences will be in a [future feature request](https://github.com/grouparoo/grouparoo/issues/1107).


bleonard:
One thing to note is that facebook won't allow you to delete users from an audience if it is fairly small. It's hard to find the right number, but it's 100 or more of actual facebook members.
This makes things somewhat complicated for true syncing.

`Audience Size too low: You cannot remove users from this audience because it will result in a low audience size.` is the error message.

To account for this I've done a few things:
* not let this failure stop syncing to other groups for a user
* Added the option at the destination level to be "Additive" only. That is, don't even try to delete.

bleonard:
Closes T-897

plugins/@grouparoo/app-templates/src/destination/shared/batch.ts
```diff
@@ -0,0 +1,93 @@
...
+  },
+  config: any
+) {
+  // build up groups situation of group names to addition and removal
+  let oldGroups = exportedProfile.oldGroups || [];
+  let newGroups = exportedProfile.newGroups || [];
+  if (exportedProfile.toDelete) {
+    // if anyone asks, removing from all groups as part of deletion
+    oldGroups = Array.from(new Set(oldGroups.concat(newGroups)));
+    newGroups = [];
```
evantahler, line 29: Think this should be something that core does when `toDelete` is true?

plugins/@grouparoo/facebook/src/lib/export/fields.ts
```diff
@@ -0,0 +1,177 @@
+import { GroupSyncModeData } from "@grouparoo/app-templates/dist/destination/group";
```
evantahler, line 1: This whole file is fascinating.  Can you walk me through it next week?

plugins/@grouparoo/facebook/src/lib/export/utils.ts
```diff
@@ -0,0 +1,332 @@
+/**
```
evantahler, line 1: What's different here than the facebook lib?  Could we patch it vs re-implemeting?

bleonard: for some reason, it was splitting first and last name it to 5 characters. I think there will be a few other changes too.

plugins/@grouparoo/app-templates/src/destination/shared/batch.ts
bleonard: Not sure. For now, I lean on the side of leaving it as it is. knowing which groups they were actually sent (oldGroups) before.

PR #1111, opened 2020-12-18
Fix bug with Profile Details page not showing properties

andyjih:
All of the values in this table were pointing at the `property`, not the `profileProperties`


andyjih:
Closes T-899

evantahler (approved):
Made some additional changes to simplify the lookup of each property

PR #1159, opened 2021-01-07
Task updates after CLSTask introduction

evantahler:
After https://github.com/grouparoo/grouparoo/pull/1156:
*  `export:enqueue` task should only one once and then commit.  Then the Task should be enqueued again.
* `import:associateProfile`  should not be un in a transaction


bleonard:
Maybe it should enqueue another one right away if there are more left. 10 seconds between could be a lot if there are a lot to do.

evantahler:
@bleonard nice! https://github.com/grouparoo/grouparoo/pull/1159/commits/992b89abd6d0dec8601ee92b88b994ccda264a9f

PR #1160, opened 2021-01-08
`grouparoo generate` command

evantahler:
`grouparoo generate` command

## Example Usage

```bash
➜ grouparoo generate --help
Usage: grouparoo generate [options] [template]

Generate new code config files from templates

Options:
  --path [path]            The location of the config directory (default: "/Users/evan/workspace/grouparoo/grouparoo/apps/staging-public/config")
  --list [list]            Display the available config templates to use with config-generate (default: "false")
  --overwrite [overwrite]  Overwrite existing files? (default: "false")
  -h, --help               display help for command

Example:
  grouparoo generate templateName -- --id="the-id" --name="the-name" --app="app-id"

# --------------------------------

➜ grouparoo generate --list

notice: Modified your runtime environment with /Users/evan/workspace/grouparoo/grouparoo/apps/staging-public/.env
info: using SQLite database: /Users/evan/workspace/grouparoo/grouparoo/apps/staging-public/grouparoo_development.sqlite
info: active plugins: @grouparoo/core/manual, @grouparoo/core/events, @grouparoo/bigquery, @grouparoo/csv, @grouparoo/facebook, @grouparoo/files-s3, @grouparoo/google-sheets, @grouparoo/hubspot, @grouparoo/logger, @grouparoo/mailchimp, @grouparoo/marketo, @grouparoo/mysql, @grouparoo/postgres, @grouparoo/redshift, @grouparoo/sailthru, @grouparoo/salesforce, @grouparoo/snowflake, @grouparoo/zendesk

🦘 Grouparoo: generate

Available Templates:

group:calculated
  Config for a calculated Grouparoo Group
  * id (required) - The ID of this group, used to determine it's guid
  * name - The name of the Group (default: "New Group")
  * rules - The rules for the group, JSON-encoded (default: "[]")

group:manual
  Config for a manual Grouparoo Group
  * id (required) - The ID of this group, used to determine it's guid
  * name - The name of the Group (default: "New Group")

# --------------------------------

➜ grouparoo generate group:calculated 

notice: Modified your runtime environment with /Users/evan/workspace/grouparoo/grouparoo/apps/staging-public/.env
info: using SQLite database: /Users/evan/workspace/grouparoo/grouparoo/apps/staging-public/grouparoo_development.sqlite
info: active plugins: @grouparoo/core/manual, @grouparoo/core/events, @grouparoo/bigquery, @grouparoo/csv, @grouparoo/facebook, @grouparoo/files-s3, @grouparoo/google-sheets, @grouparoo/hubspot, @grouparoo/logger, @grouparoo/mailchimp, @grouparoo/marketo, @grouparoo/mysql, @grouparoo/postgres, @grouparoo/redshift, @grouparoo/sailthru, @grouparoo/salesforce, @grouparoo/snowflake, @grouparoo/zendesk

🦘 Grouparoo: generate

Error: Missing required input "id"

# --------------------------------

➜ grouparoo generate group:calculated -- --id tmp-group --name "New Group" --rules='{"propertyId": "email","operation": { "op": "like" },"match": "%@%"}' --overwrite

notice: Modified your runtime environment with /Users/evan/workspace/grouparoo/grouparoo/apps/staging-public/.env
info: using SQLite database: /Users/evan/workspace/grouparoo/grouparoo/apps/staging-public/grouparoo_development.sqlite
info: active plugins: @grouparoo/core/manual, @grouparoo/core/events, @grouparoo/bigquery, @grouparoo/csv, @grouparoo/facebook, @grouparoo/files-s3, @grouparoo/google-sheets, @grouparoo/hubspot, @grouparoo/logger, @grouparoo/mailchimp, @grouparoo/marketo, @grouparoo/mysql, @grouparoo/postgres, @grouparoo/redshift, @grouparoo/sailthru, @grouparoo/salesforce, @grouparoo/snowflake, @grouparoo/zendesk

🦘 Grouparoo: generate

✅ wrote /Users/evan/workspace/grouparoo/grouparoo/apps/staging-public/config/group/calculated/new_group.js

```

## Making Templates:

1. In core, or in your plugin, place files in `/src/templates` and extend `import {ConfigTemplate} from "@grouparoo/core"`
2. Your run method should return a hash of `{[filename: string]: bodyContent}`.
3. For the nominal case, where a collection of Mustache templates is validated and returned, use the built-in `this.mustacheAllFiles(params)`
4. You can do custom stuff in the `run()` method, like formatting inputs!

```ts
export class CalculatedGroupTemplate extends ConfigTemplate {
  constructor() {
    super();
    this.class = "group";
    this.name = "group:calculated";
    this.description = "Config for a calculated Grouparoo Group";
    this.rootPath = path.join(
      __dirname,
      "..",
      "..",
      "public",
      "templates",
      "groups"
    );
    this.files = [path.join("calculated", "*.template")]; // <--- could be 1 or many files
    this.inputs = {
      id: {
        required: true,
        default: null,
        description: "The ID of this group, used to determine it's guid",
      },
      name: {
        required: true,
        default: "New Group",
        description: "The name of the Group",
      },
      rules: {
        required: true,
        default: "[]",
        description: "The rules for the group, JSON-encoded",
      },
    };
  }

  async run({ params }) {
    if (params.rules) {
      params.rules = JSON.stringify(JSON.parse(params.rules), null, 2); // <--- custom param formatting
    }

    return this.mustacheAllFiles(params);
  }
}
```

And the template file itself. Note, you an use mustache templates **in** the file name, like `{{name}}.js`
Include lots of comments to help out users!

```js
  return [
    {
      class: "Group",
      id: "{{id}}",
      name: "{{name}}",
      type: "calculated",

      /**
       The Rules for this group.  This is an Array.  Learn more about the options @ www.grouparoo.com/TODO:
        Examples:

        rules: [
          {
            propertyId: "email",
            operation: { op: "exists" },
          },
          {
            propertyId: "ltv",
            operation: { op: "gt" },
            match: 100,
          },
        ]
      */
      rules: {{rules}},
    },
  ];
};
```

## Notes

- We can't store the actual JS templates within `/src` - they aren't valid JS and `tsc` gets mad. That's whey they are kept in public
- To prevent prettier and friends from freaking out on semi-valid template files, all templates should end with `.template` which will be removed when the file is written, ie: `{{name}}.js.template`
- We are formatting the generated output with Prettier, so it needs to be valid JS/JSON
- I went with `:` delimiters for the template names.  There are some shenanigans with parsing spaces in ARGS vs OPTS with how Actionhero is using Commander. See https://github.com/tj/commander.js/blob/master/docs/options-taking-varying-arguments.md.  This should be addressed in another story.

Closes T-843
Closes T-883
Closes T-919

## Also in this PR: 

* rename `grouparoo config-validate` and `config-apply` to just `grouparoo  config` (Closes T-925)
* `grouparoo run` gains `--reset-high-watermarks` flag to reset the high water marks on each schedule, so that you can run your sync from scratch.  (Closes T-924)
* `grouparoo destroy` also now resets the highwatermarks on any schedules that have them
* Subscribing new team members fires in a `CLS.afterCommit`
* Fixes a bug that prevented `grouparoo run` from working with SQLlite (actionhero wasn't started fully when the config was being applied) - https://github.com/actionhero/actionhero/pull/1695


core/src/bin/run.ts
```diff
@@ -66,8 +75,8 @@ export class RunCLI extends CLI {
   async waitForReady() {
     return new Promise(async (resolve) => {
       async function check() {
-        if (api.process.running) return resolve(null);
-        setTimeout(() => check(), CHECK_TIMEOUT);
+        if (api.process.started) return resolve(null);
```
evantahler, line 78: Newly added in Actionhero v25.0.3

core/public/templates/groups/calculated/{{name}}.js.template
```diff
@@ -0,0 +1,30 @@
+exports.default = async function buildConfig() {
+  return [
+    {
+      class: "Group",
+      id: "{{id}}",
```
bleonard: something funny here about null values and escaping. i wonder if they should be 
```
id: {{id}}
```
and (by defualt or by opting in when variable declared - depending on what's more common), it does `JSON.encode()` to get null or add the parens/escape.

core/src/modules/cli.ts
```diff
@@ -57,6 +58,21 @@ export namespace GrouparooCLI {
...
 
+  export async function resetHighWatermarks() {
+    log("Resetting Schedule high water marks...", "warning");
+    const schedules = await Schedule.findAll();
+    for (const i in schedules) {
+      const runs = await Run.findAll({
+        where: {
+          creatorGuid: schedules[i].guid,
+          state: { [Op.ne]: "running" },
+          highWaterMark: { [Op.ne]: null },
```
bleonard, line 74: probably just destroying all i guess, so don't need this.

core/src/modules/cli.ts
```diff
@@ -57,6 +58,21 @@ export namespace GrouparooCLI {
...
   }
 
+  export async function resetHighWatermarks() {
+    log("Resetting Schedule high water marks...", "warning");
+    const schedules = await Schedule.findAll();
+    for (const i in schedules) {
+      const runs = await Run.findAll({
+        where: {
+          creatorGuid: schedules[i].guid,
+          state: { [Op.ne]: "running" },
```
bleonard, line 73: maybe "pending" here, too? Or otherwise, whatever means "not started" yet.

core/public/templates/groups/calculated/{{name}}.js.template
```diff
@@ -0,0 +1,30 @@
...
+            operation: { op: "exists" },
+          },
+          {
+            propertyId: "ltv",
+            operation: { op: "gt" },
+            match: 100,
+          },
+        ]
+      */
+      rules: {{{rules}}},
```
bleonard: what's this do?

bleonard: something hard there about arrays and such. I'm not sure what the template defualt should be for groups, though it generally begs for some combo of the following:
1. empty array
2. example rules, ideally from actual current properties
3. a way to say which properties to make the rules out of in the args

That generally suggests the `CalculatedGroupTemplate` can eventually preprocess the options a big and try to be smart about things. I think the class model will enable that, so that's good.

core/public/templates/groups/calculated/{{name}}.js.template
bleonard: probably a `type` (integer, string) expectation in there too.

bleonard (commented):
This is looking great. exciting times. Random thoughts
1. Does this scale to generating tests? That might be a thing, too
2. I agree there's something more to do with these colons, I'd love `grouparoo generate group(s) calculated` and grouparoo groups -h` eventually. Maybe for now can it work without the option: `grouparoo generate group:calculated` that might be good enough. I think that's where Rails is.
3. Probably some interesting stuff around file collisions in there that I didn't. `overwrite` is good start.
4. Should we clean up files on error? Seems like it might be a lot of trouble to do.

bleonard:
Eventually, it would be nice to see the options _per template_ 
The doc suggests there might a list of many of them in the `grouparoo generate -h` command. That should probably leave them out and there is a `grouparoo generate group:calculated -h` that gives just its templates.

core/src/modules/cli.ts
evantahler: For determining the next high-water mark, only `compelte` ones are considered.  We can leave the running ones to finish.

core/public/templates/groups/calculated/{{name}}.js.template
evantahler: `{{{ }}}` is a mustache thing - it generally means "don't try to HTML escape".  we want this here because we are writing pre-formatted JSON

evantahler:
> I agree there's something more to do with these colons ...
> Eventually, it would be nice to see the options per template ...

Agree! I'm going to look at making each generator appear "higher up" as it's own CLI command so we can list them and get the inputs the normal way.

evantahler:
Update! 

* Now, the first argument (not option) is the template name.  No need for `--template`.
* We don't show all template options in the help output now.  Downside: Template options are now passed after a `--`, ie: `grouparoo generate templateName -- --id="the-id" --name="the-name" --app="app-id"`
* We can now get per-template help output:

```
➜ grouparoo generate group:calculated --help
Usage: grouparoo generate [options] [template]

Generate new code config files from templates

Options:
  --path [path]            The location of the config directory (default: "/Users/evan/workspace/grouparoo/grouparoo/apps/staging-public/config")
  --list [list]            Display the available config templates to use with config-generate (default: "false")
  --overwrite [overwrite]  Overwrite existing files? (default: "false")
  -h, --help               display help for command

Example:
  grouparoo generate templateName -- --id="the-id" --name="the-name" --app="app-id"

group:calculated
  Config for a calculated Grouparoo Group
  * id (required) - The ID of this group, used to determine it's guid
  * name - The name of the Group (default: "New Group")
  * rules - The rules for the group, JSON-encoded (default: "[]")
```

bleonard:
amazing. I'm actually ok with the colons. I think it gets us pretty far.

Think about this one?
> something funny here about null values and escaping. i wonder if they should be
> 
> ```
> id: {{id}}
> ```
> 
> and (by defualt or by opting in when variable declared - depending on what's more common), it does `JSON.encode()` to get null or add the parens/escape.

bleonard (commented):
Looking great! String work!

Random things still on my mind for this or future PR
1. the null/JSON encoding thing mentioned.
2. i know it's absurd but what's this `--` double dash thing?

evantahler:
I never knew what `--` was called!  So I guess it's the `operand delimiter`?  
* Options are "known" (and possibly required by the command)
* Operands are "whatever".  As far as the generate CLI command is concerned, only `template`, `overwrite` and `path` are `Options`... the template variables are `operands`?

> POSIX.1-2017
> 12.2 Utility Syntax Guidelines
> Guideline 10:
> The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.

(from https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean)

evantahler:
I think we are OK with quotes.  `{{}}` in mustache HTML-escapes, so:

`grouparoo generate group:calculated -- --id tmp-group --name 'New "Group"' --overwrite` produces `name: "New &quot;Group&quot;"`

If we would have used `{{{}}}}` in the IDs/Names, it would be a problem.

evantahler:
npm start -- --cool

PR #1181, opened 2021-01-12
Simplify retry logic of event:associateProfile Task

evantahler:


bleonard:
Do we know why the `transaction` issue is happening in the first place?
Are two threads processing the same event?

evantahler:
Yep, it's parallelism.  It's not really an issue per-se - there used to be all sorts of complex locking and retry logic in `event#associate` if 2 events were trying to create the same profile.  Now, because of transactions, I removed it... and can just rely on "if there's an error, sleep and try again".

bleonard:
This code is probably better but it will probably still show errors in demo/development environments. I think maybe it doesn't do a retry.

evantahler:
I disagree?  Looks like it's doing the right thing...

```
info: [ worker ] task success workerId=2 class=event:associateProfile queue=events args={"eventGuid":"evt_2b3fc0db-14d3-40b2-90f4-0bcbe4a49042"} duration=352
info: [ worker ] task success workerId=1 class=event:associateProfile queue=events args={"eventGuid":"evt_cc5d46a0-2497-491f-83a5-a4fb12fad2db"} duration=181
info: re-enqueuing association of event evt_8d53d338-8275-4623-9041-226327124183 // <-- HERE
info: [ worker ] reEnqueue task workerId=2 plugin={"delay":1000,"remainingAttempts":2,"err":{}} // <-- HERE
info: [ worker ] task success workerId=2 class=event:associateProfile queue=events args={"eventGuid":"evt_8d53d338-8275-4623-9041-226327124183"} duration=26
info: [ worker ] task success workerId=1 class=event:associateProfile queue=events args={"eventGuid":"evt_0f2cf8bf-c62b-4c71-9e6d-89948a200636"} duration=24
```

PR #1196, opened 2021-01-18
Upgrade Actionhero, Sequelize and Sequelize Typescript to latest versions

evantahler:
Sequelize Typescript now supports v6 of Sequelize!

This upgrade requires related updates to Actionhero and ah-sequelize-plugin as well.


core/src/actions/cluster.ts
```diff
@@ -73,12 +72,11 @@ export class ClusterReset extends AuthenticatedAction {
     const counts: { [model: string]: number } = {};
 
     for (const i in models) {
-      const model = models[i];
+      //@ts-ignore
+      const model: typeof App = models[i]; // pick one of the Models so that the types are the same.  TODO: make this better
```
evantahler, line 76: Now with the stricter typing of sequelize@v6, if there are places where we are looping though a collection of models, we may see errors that they are of different types, and need to do a little trick like this.

This type of bug only occurs a few times in the codebase.  There's probably a better way to do this with a `<typeof class extends Model>`... but we can sort that out later

PR #1200, opened 2021-01-20
Customer Profile Testing

evantahler:
Here's the proposal for customers testing a profile:

```ts
import { helper, relaxedSnapshot } from "@grouparoo/spec-helper";

describe("snapshot", () => {
  // utility to start and stop the server within the test suite
  helper.grouparooTestServer();

  test("a profile snapshot can be tested", async () => {
    const { profile, snapshot } = await helper.getProfile({
      email: ["test-person@example.com"],
    });

    // You can do snapshot testing
    expect(snapshot).toMatchSnapshot(relaxedSnapshot(snapshot));

    // Or you can test the properties of the snapshot directly
    expect(snapshot.properties.userId.values).toEqual([100]);
    expect(snapshot.groups.length).toBe(1);
    expect(snapshot.groups[0].name).toBe("People with Email Addresses");
    expect(profile.state).toBe("ready");
  });
});
```

Of note:
* Customers who want to to testing should include `@groupraoo/spec-helper` in their `devDependencies`
* `helper.grouparooTestServer()` is a new method that injects a beforeAll and afterAll into the test to replace our current server management utils.  Way better! I'll update all the tests with this
* I am asking customers to know a little about our Models... but I think that's OK
* `relaxedSnapshot()` is a new utility that by default knows how to tell jest that keys in objects matching our guids and timestamps can be "relaxed" when matching: IE that they should be strings or numbers, but we don't care if we match the previous snapshot values exactly

---

Here's what a snapshot looks like

```
    {
      properties: {
        userId: {
          guid: 'rul_user_id',
          state: 'ready',
          values: [Array],
          type: 'integer',
          unique: true,
          isArray: false,
          identifying: true,
          valueChangedAt: 2021-01-22T21:31:56.733Z,
          confirmedAt: 2021-01-22T21:31:56.733Z,
          stateChangedAt: 2021-01-22T21:31:56.733Z,
          createdAt: 2021-01-22T21:31:56.632Z,
          updatedAt: 2021-01-22T21:31:56.775Z
        },
        'last name': {
          guid: 'rul_last_name',
          state: 'ready',
          values: [Array],
          type: 'string',
          unique: false,
          isArray: false,
          identifying: false,
          valueChangedAt: 2021-01-22T21:31:56.642Z,
          confirmedAt: 2021-01-22T21:31:56.642Z,
          stateChangedAt: 2021-01-22T21:31:56.642Z,
          createdAt: 2021-01-22T21:31:56.642Z,
          updatedAt: 2021-01-22T21:31:56.775Z
        },
        'first name': {
          guid: 'rul_first_name',
          state: 'ready',
          values: [Array],
          type: 'string',
          unique: false,
          isArray: false,
          identifying: false,
          valueChangedAt: 2021-01-22T21:31:56.652Z,
          confirmedAt: 2021-01-22T21:31:56.652Z,
          stateChangedAt: 2021-01-22T21:31:56.652Z,
          createdAt: 2021-01-22T21:31:56.652Z,
          updatedAt: 2021-01-22T21:31:56.775Z
        },
        email: {
          guid: 'rul_email',
          state: 'ready',
          values: [Array],
          type: 'email',
          unique: true,
          isArray: false,
          identifying: false,
          valueChangedAt: 2021-01-22T21:31:56.753Z,
          confirmedAt: 2021-01-22T21:31:56.753Z,
          stateChangedAt: 2021-01-22T21:31:56.753Z,
          createdAt: 2021-01-22T21:31:56.662Z,
          updatedAt: 2021-01-22T21:31:56.775Z
        }
      },
      groups: [
        {
          guid: 'grp_email_group',
          name: 'People with Email Addresses',
          type: 'calculated',
          rules: [Array],
          matchType: 'all',
          state: 'updating',
          locked: 'config:code',
          profilesCount: 1,
          calculatedAt: null,
          nextCalculatedAt: 1611351116801,
          createdAt: 1611351116156,
          updatedAt: 1611351116398
        }
      ],
      exports: [
        {
          guid: null,
          destination: null,
          profileGuid: 'pro_dadd6f69-9103-446b-963c-74a3c83924c6',
          force: true,
          createdAt: null,
          startedAt: 1611351116825,
          completedAt: null,
          oldProfileProperties: {},
          newProfileProperties: [Object],
          oldGroups: [],
          newGroups: [Array],
          toDelete: false,
          hasChanges: true,
          mostRecent: false,
          errorMessage: undefined,
          errorLevel: undefined
        }
      ]
    }
```

The profile snapshot method:
```ts
  async snapshot(toSync = false) {
    if (toSync) await this.sync(); // import the profile and recalculate groups

    const properties = await this.properties();
    const groups = await this.$get("groups");
    const groupApiData = await Promise.all(groups.map((g) => g.apiData()));
    const exports = await this.export(true, [], false); // build the next exports for all groups, but to not save them
    const exportsApiData = await Promise.all(
      exports.map((e) => e.apiData(false))
    );

    return { properties, groups: groupApiData, exports: exportsApiData };
  }
```

Closes T-882


bleonard (commented):
Just brainstorming some questions about this:


- maybe the ability to make a (default relaxed) snapshot given a profile property (ideally without the `[]`) is in the helper and they only know about the helper. So a test would be: get snapshot for email. I thin kthe default would be to `sync` Example below




```ts
import { helper, snapshotProfile, relaxed } from "@grouparoo/spec-helper";

describe("snapshot", () => {
  // turns on/off actionhero server
  helper.grouparooTestServer();

  test("my user looks right", async () => {
    // first arg (profile seed data): keys needed for the properties to work
    // second arg (options): shows defaults and could have been left off.
    const profileData = await snapshotProfile({userId: 10}, {sync: true});
    expect(profileData).toMatchSnapshot(relaxed(profileData));
  });
});
```


* does relaxed care about the order in arrays? there could be more than one export, for example.
  * Maybe we need to sort things by a known (and predictable) key in all of those.

evantahler:
I went with a new `helper.getProfile()` method to return both the profile and latest snapshot.  Example updated above

PR #1201, opened 2021-01-20
Fetch up to 1000 mail chimp lists

bleonard:
The previous (default) number was 10. Some users have more than that.
1000 seems reasonable and better than paging logic.


evantahler:
Closes T-966

PR #1202, opened 2021-01-22
Remove stale settings on init

seancdavis:
Before the settings are registered, remove any lingering settings that don't match the expected set. This uses the `key` attribute to determine if a setting is stale or not.

---

Closes T-832

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -83,4 +83,20 @@ cd apps/staging-public
 (npx) grouparoo demo-event-stream
 ```
 
+## Running Tests
+
+Grouparoo uses [Jest](https://jestjs.io/) as our testing framework. To run tests, first make sure you've installed all dependencies in the root of the project.
+
+    cd ../
+    pnpm install
+
+Next, create the test databases:
+
+    cd core
+    ./bin/create_test_database
+
+Then you can run the tests. We recommend you run only the test files you're working on. Let our CI server do the bulk of the work:
+
+    npm run test path/to/test/file
+
 🦘
--- a/core/src/initializers/settings.ts
+++ b/core/src/initializers/settings.ts
@@ -171,6 +171,14 @@ export class Plugins extends Initializer {
       },
     ];
 
+    const settingKeys: string[] = [
+      ...coreSettings,
+      ...interfaceSettings,
+      ...telemetrySettings,
+    ].map(({ key }) => key);
+
+    await plugin.cleanSettings(settingKeys);
+
     await this.registerSettingsArray(coreSettings, "core");
     await this.registerSettingsArray(interfaceSettings, "interface");
     await this.registerSettingsArray(telemetrySettings, "telemetry");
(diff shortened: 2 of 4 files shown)


evantahler (commented):
Looks like a good starting point!  

I like the approach you picked to use the set of known Setting keys to check what should exist, rather than use a migration - this will work for all future changes Settings changes too!

Suggestions:
* It doesn't look like you are calling `cleanSettings()` anywhere?  
* A test could be added for the `cleanSettings()` method that a bogus setting which is added is then removed.

core/src/modules/plugin.ts
```diff
@@ -75,6 +76,14 @@ export namespace plugin {
     });
   }
 
+  export async function cleanSettings(staleKeys: string[]) {
```
bleonard: I think this name could be better. I think it's `inUseKeys`.

core/src/initializers/settings.ts
```diff
@@ -171,6 +171,14 @@ export class Plugins extends Initializer {
       },
     ];
 
+    const settingKeys: string[] = [
+      ...coreSettings,
+      ...interfaceSettings,
+      ...telemetrySettings,
+    ].map(({ key }) => key);
```
bleonard, line 178: not a major thing but i think there is `.keys()` that will do this.

core/src/modules/plugin.ts
seancdavis: Great catch! Fixed.

core/src/initializers/settings.ts
seancdavis: In this case, I believe [`.keys()` would return `[0, 1, 2, ...]`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/keys).

This is an array of objects, and I'm pulling out the `key` _property_ from each one.

bleonard: Ah, I see. brackets not `{` and I misunderstood what was in `telemetrySettings`. good work!

core/README.md
```diff
@@ -83,4 +83,20 @@ cd apps/staging-public
...
+    pnpm install
+
+Next, create the test databases:
+
+    cd core
+    ./bin/create_test_database
+
+Then you can run the tests. We recommend you run only the test files you're working on. Let our CI server do the bulk of the work:
+
+    npm run test path/to/test/file
```
evantahler, line 100: No change needed, but as a note: `npm test {args}` will also run our `prestest` command, which will run `prettier` and `tsc`.  If you really just want to run one test file, `./node_modules/.bin/jest path/to/test/file` is probably the way to go.

bleonard: I'd be supportive of `npm run jest` to be able to stop all the typing for this case in the dev iterations.

evantahler:
Oh! I added the Label `bug`.  Without the labels, the PR isn't included in the release notes for the next version.

PR #1203, opened 2021-01-22
Update helper text for cli generate command

seancdavis:
Changes here:

- T-972: Give the user a notice when the `generate` command changes the `id`. I used `Colors` directly in the `generate.ts` file to give it some emphasis.
- T-997: Add example to partial match on `grouparoo generate --list`. I created this one on the fly as I was working. Seemed like it would help.
- T-971 / T-974: The description for the `generate` command is ... _lengthy_ now. I don't love the way I did this, but it does shove the info right in the dev's face. I played around with using a library like [columnify](https://www.npmjs.com/package/columnify), but it seemed a little heavy-handed. I figured we could keep it a little uglier like this and come back to it as we think more about workflows in the CLI.
- T-980: Real simple note on the mailchimp `id` destination based on our conversation from yesterday. As a dev coming in blind, I feel like that note would be enough to tell me what to do.
- T-975: Reworked the four source templates with bootstrapped properties. This was the biggest change I made. The way I formatted it would have helped me work through it yesterday. But it's a decent-sized change to how you had those files setup, so I won't be offended if you want to keep playing with it.

### DIFF
--- a/core/src/bin/generate.ts
+++ b/core/src/bin/generate.ts
@@ -1,4 +1,5 @@
 import { GrouparooCLI } from "../modules/cli";
+import Colors from "colors/safe";
 import {
   ConfigTemplate,
   ConfigTemplateRunResponse,
@@ -83,11 +84,17 @@ export class Generate extends CLI {
 
     const template = await this.getTemplate(params.template);
 
+    const preparedParams = template.prepareParams({ ...params });
+
+    if (preparedParams.id !== params.id) {
+      console.log(
+        `${Colors.yellow("notice")}: ID was changed to ${preparedParams.id}`
+      );
+    }
+
     let fileData: ConfigTemplateRunResponse = {};
     try {
-      fileData = await template.run({
-        params: template.prepareParams(params),
-      });
+      fileData = await template.run({ params: preparedParams });
     } catch (error) {
       this.fatalError(error.message);
     }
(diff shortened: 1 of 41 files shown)


evantahler (approved):
I like all of these changes 👍 .

There's a minor philosophy question embedded here: Should we use `...` to represent all parent/child IDs in these templates?  It's a great sign to folks that they need to change it and put something there.  But, without an example value (ie: `appId: data_warehouse`), they might get lost?

I'm generally pro `...` if in a comment nearby we include an example.  If we want to go this route, then I would suggest changing *all* the templates to match.  There are others...

evantahler:
Added `enhancement` label

seancdavis:
@evantahler I added a handful of placeholders to these templates. It's not exhaustive, but it moves a little closer to being consistent with usage across the board. I'm thinking it's enough to wrap these tasks and move toward v0.2, though I'd like to discuss potential steps down the road to make these even more consistent.

Probably worth another scan since I adjusted so many files with the [latest commit](https://github.com/grouparoo/grouparoo/pull/1203/commits/81e0d31d6267fc5ee8331bab115ce707faac2c13).

evantahler (approved):
Good pattern: `appId: "...", // The ID of the App this Source uses - e.g. appId: "facebookApp"`

I think you found enough of them to move this forward!

Issue #651, opened 2020-08-31
Feature Request: Iterable Destination


bleonard:
**Is your feature request related to a problem? Please describe.**
Add Iterable as a destination for Grouparoo profiles and groups.

bleonard:
Iterable is one of the more modern marketing automation tools.

[Here](https://api.iterable.com/api/docs) are the API docs.

Core concepts: Users, Lists

**What do profiles map to?** 

Users are the iterable concept.

It’s interesting that their [updateUser](https://api.iterable.com/api/docs#users_updateUser) API has a much higher rate limit. Excited to work with you!

This is an upsert API, so it will create the user if necessary

I _think_ we will want mergeNestedObjects=true set so our calls are only additive, but I’m not sure.

They support either email or userId. By default it says “Either email or userId must be passed in to identify the user. If both are passed in, email takes precedence.”  \
There is a boolean to preferUserId, but we will have to test how it works when “upgrading” an email to having a userId. Does it update the same profile? I think it might make a new profile.

This makes me think that only “email” should be a required field in Grouparoo.

This is probably the way to go:



*   If grouparoo says the email has changed, call the [updateEmail](https://api.iterable.com/api/docs#users_updateEmail) API
    *   Handle the errors. For example, maybe that email wasn’t even in the system to begin with
*   Call the upsert [updateUser](https://api.iterable.com/api/docs#users_updateUser) API with the email and optional userId (to make things a bit smarter)
*   Get the lists ids by name (cached)
*   Add/Delete from lists with their ids and user email

Let’s be sure to test the email change case to make sure there are no timing issues (the changed email needs to take effect right away)

**Properties**

Properties can be defined at runtime in Iterable, so allowOptionalFromProperties = true like Sailthru.

[Here](https://support.iterable.com/hc/en-us/articles/208183076-Field-Data-Types) are the data types.

We won’t be able to handle geo_location, but everything else looks straightforward.

[getUserFields](https://api.iterable.com/api/docs#users_getUserFields) should be used to populate the list of “known” properties. I’m curious to see what comes back from nested (object) properties.

There are also arrays that we want to be able to handle.

There are also nested Objects. We need to investigate the best way to set these Objects. Do they show up in the [getUserFields](https://api.iterable.com/api/docs#users_getUserFields) call? 

Is there a way to define the user fields ahead of time or is it only via the API? I don’t see one.

There are also arrays of objects, which seems way more than we can handle.

[...]

PR #1204, opened 2021-01-23
feat: add the new Iterable plugin.

pauloouriques:
- add new iterable plugin as a destination;
- add simple api test call to test itarable api integration.


plugins/@grouparoo/iterable/README.md
```diff
@@ -0,0 +1,5 @@
+# @Grouparoo/Iterable
+
+## Attribution
+
+The Iterable icon used in this plugin is from http://getlogo.net/
```
bleonard: ideally the whole url for attribution of where downloaded.

plugins/@grouparoo/iterable/__tests__/.env.example
```diff
@@ -0,0 +1 @@
+HUBSPOT_API_KEY = "a31ef454-1835-4b88-8d3a-eefc0b688ded"
```
bleonard: will want to switch to `ITERABLE_API_KEY` when do tests.

plugins/@grouparoo/iterable/src/initializers/plugin.ts
```diff
@@ -0,0 +1,72 @@
...
+    super();
+    this.name = packageJSON.name;
+  }
+
+  async initialize() {
+    plugin.registerPlugin({
+      name: packageJSON.name,
+      icon: "/public/@grouparoo/iterable/iterable.svg",
+      templates: [
+        new AppTemplate("hubspot", [
```
bleonard: `hubspot` -> `iterable` updates throughout this file.

plugins/@grouparoo/iterable/src/initializers/plugin.ts
```diff
@@ -0,0 +1,72 @@
...
+          ],
+          methods: { test },
+        },
+      ],
+      connections: [
+        {
+          name: "iterable-export",
+          direction: "export",
+          description:
+            "Export Profiles to Iterable and add them to Contact Lists.",
```
bleonard: "Export profiles to Iterable as Users and put them in static Lists"

plugins/@grouparoo/iterable/src/lib/connect.ts
```diff
@@ -0,0 +1,5 @@
+import { SimpleAppOptions } from "@grouparoo/core";
+
+export async function connect(appOptions: SimpleAppOptions) {
+  return require('node-iterable-api').create(appOptions.apiKey);
```
bleonard: Let's `import Iterable from 'node-iterable-api'` (ideally) or `cosnt Iterable = require('node-iterable-api')` (if necessary) above and then do `Iterable.create()` here.

plugins/@grouparoo/iterable/src/lib/export/destinationOptions.ts
```diff
@@ -0,0 +1,5 @@
+import { DestinationOptionsMethod } from "@grouparoo/core";
+
+export const destinationOptions: DestinationOptionsMethod = async () => {
+  return {};
```
bleonard, line 4: will be empty (as-is)

plugins/@grouparoo/iterable/src/lib/test.ts
```diff
@@ -0,0 +1,9 @@
+import { TestPluginMethod } from "@grouparoo/core";
+import { connect } from "./connect";
+
+export const test: TestPluginMethod = async ({ appOptions }) => {
+  const client = await connect(appOptions);
+  const success = await client.lists.get();
+  const message = success ? 'Connected to Iterable api.' : null;
```
bleonard: What is success here? An array of lists?
We try to have a message something like.

```
  const response = await client.lists.get();
  const success = !!response.lists;
  const message = success ? 'Found ${response.lists.length} lists' : `Error getting lists`;
```

bleonard (commented):
Looking good!

Next step: destination mapping options using the fields

bleonard:
For Issue: https://github.com/grouparoo/grouparoo/issues/651

plugins/@grouparoo/hubspot/__tests__/integration/hubspot-export.ts
```diff
@@ -1,4 +1,4 @@
-import path from "path";
+import * as path from "path";
```
bleonard: Is this some autoformatting tool you have?

plugins/@grouparoo/iterable/README.md
```diff
@@ -0,0 +1,5 @@
+# @Grouparoo/Iterable
+
+## Attribution
+
+The Iterable icon used in this plugin is from http://getlogo.net/iterable-inc-logo-vector-svg
```
bleonard: I think maybe these are usually square. Can you find one of those. `png` is fine.

plugins/@grouparoo/iterable/__tests__/.env.example
```diff
@@ -0,0 +1 @@
+ITERABLE_API_KEY = ""
```
bleonard: i usually put an example in here (not the real one of course) but it helps others to know what it should look like.

plugins/@grouparoo/iterable/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,75 @@
...
+  "../",
+  "fixtures",
+  "destination-mapping-options.js"
+);
+
+// these comments to use nock
+// const newNock = false;
+// require("./../fixtures/destination-mapping-options");
+// or these to make it true
+const newNock = true;
```
bleonard: will comment out for CI

plugins/@grouparoo/iterable/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,75 @@
...
+    destinationGuid: null,
+    destinationOptions: null,
+  });
+}
+
+describe("iterable/destinationMappingOptions", () => {
+  test("can fetch user fields", async () => {
+    const client = await connect(appOptions);
+    const fields = await getUserFields(client);
+    const text = fields.find((f) => f.key === "phoneNumber");
```
bleonard: these variable names don't look quite right. I would think this would would be `phoneNumber` too.

plugins/@grouparoo/iterable/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,75 @@
...
+> => {
+  const fields = await client.users.getFields();
+  const out = [];
+
+  for (const [key, value] of Object.entries(fields.fields)) {
+    if (value !== "object" && key !== "email") {
+      const type: DestinationMappingOptionsResponseTypes = mapTypesFromIterableToGrouparoo(
+        value
+      );
+      out.push({ key, type, important: true });
```
bleonard: only push `if (type)`

plugins/@grouparoo/iterable/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,75 @@
...
+const mapTypesFromIterableToGrouparoo = (iterableType) => {
+  const map = {
+    string: "string",
+    long: "integer",
+    date: "date",
+    boolean: "boolean",
+  };
+  const grouparooType = map[iterableType];
+  if (!grouparooType) {
+    throw `Unknown iterable type: ${iterableType}`;
```
bleonard: we found it's better to `log` and allow this function to return `undefined` or `null`

plugins/@grouparoo/iterable/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,75 @@
...
+      allowOptionalFromProperties: true,
+    },
+  };
+};
+
+export const getRequiredFields = (): Array<{
+  key: string;
+  type: DestinationMappingOptionsResponseTypes;
+}> => {
+  return [{ key: "email", type: "string" }];
```
bleonard: there is a type called "email" that this should use.

plugins/@grouparoo/iterable/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,75 @@
...
+};
+
+const mapTypesFromIterableToGrouparoo = (iterableType) => {
+  const map = {
```
bleonard, line 53: if there is a special "known" ones that should be a `phoneNumber`, can can pass the key in here and catch it earlier.

plugins/@grouparoo/zendesk/__tests__/integration/destination-mapping-options.ts
```diff
@@ -1,13 +1,10 @@
-import path from "path";
```
bleonard: can you undo the changes to this file (and all the zendesk ones)? I feel like you have something removing unused things or something.
that gets weird when we do this comment in and out the nock thing.

plugins/@grouparoo/iterable/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,75 @@
...
+  test("can load all destinationMappingOptions", async () => {
+    const options = await runDestinationMappingOptions({});
+    const { properties } = options;
+    const { required, known } = properties;
+
+    expect(required.length).toBe(1);
+
+    const requiredFieldEmail = required.find((f) => f.key === "email");
+    expect(requiredFieldEmail.key).toBe("email");
+    expect(requiredFieldEmail.type).toBe("string");
```
bleonard: should be "email"

plugins/@grouparoo/iterable/__tests__/utils/shared.ts
```diff
@@ -0,0 +1,8 @@
+import { helper } from "@grouparoo/spec-helper";
+
+const delayTime = 3 * 1000;
+
+export async function indexContacts() {
+  // search index takes a while to catch up after important change or add
```
bleonard, line 4: it's ok to put this in the one test file.
Also only sleep `if newNock`

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,77 @@
...
+  export: {
+    toDelete,
+    newProfileProperties,
+    oldProfileProperties,
+    newGroups,
+    oldGroups,
+  },
+}) => {
+  // if we received no mapped data... just exit
+  if (Object.keys(newProfileProperties).length === 0) {
```
bleonard, line 31: Is this in others? That's weird. There still might be something to remove. I guess there's also no email?

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const email = newProfileProperties["email"]; // this is how we will identify profiles
+  if (!email) {
+    throw new Error(`newProfileProperties[email] is a required mapping`);
+  }
+
+  try {
+    let user;
+    try {
+      user = await client.users.get({ email });
```
bleonard: I thin it's possible you won't have to do this because it uses the email to upsert and add/remove from lists.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,77 @@
...
+    } catch (error) {
+      if (!error.toString().match(/Request failed with status code 404/)) {
+        throw error;
+      }
+    }
+    if (toDelete) {
+      if (user) {
```
bleonard: if we don't have the user, can use just `deleteByEmail` and be sure to have a test where the user doesn't actually exist, so we can handle that.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (Object.keys(newProfileProperties).length === 0) {
+    return { success: true };
+  }
+  const client = await connect(appOptions);
+
+  const email = newProfileProperties["email"]; // this is how we will identify profiles
+  if (!email) {
+    throw new Error(`newProfileProperties[email] is a required mapping`);
+  }
+
```
bleonard, line 26: There's something do around here about email changing.

bleonard (commented):
Looking good!
Made some comments and hints for the future.

apps/staging-public/package.json
```diff
@@ -31,6 +31,9 @@
     "@grouparoo/ui": "^0.2.0-alpha.15",
     "@grouparoo/zendesk": "^0.2.0-alpha.15"
   },
+  "devDependencies": {
```
bleonard: this should be in the regular list (not dev). i may have said "dev" incorrectly before.

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,387 @@
...
+  } catch (err) {
+    if (!suppressErrors) {
+      throw err;
+    }
+  }
+}
+
+async function deleteLists(suppressErrors) {
+  try {
+    for (const groupToDelete in ["List One", "List Two", "List Three"]) {
```
bleonard: let's make instance variables out of these above to make sure they are the same as in use.

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,387 @@
...
+  });
+}
+
+describe("iterable/exportProfile", () => {
+  beforeAll(() => {
+    jest.setTimeout(helper.longTime);
+  });
+  beforeAll(async () => {
+    apiClient = await connect(appOptions);
+    await deleteUsers(false);
```
bleonard: other tests have standardized on a `cleanUp` function that does both of these things (users and lists)

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,387 @@
...
+      newProfileProperties: { email, name },
+      oldGroups: [],
+      newGroups: [],
+      toDelete: false,
+    });
+    await indexContacts(newNock);
+    user = await getUser(email);
+
+    expect(user).not.toBe(null);
+    expect(user.email).toBe(email);
```
bleonard, line 171: check name

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,387 @@
...
+        email,
+        userId,
+        name,
+        phoneNumber,
+        signupDate: exampleDate,
+      },
+      newProfileProperties: {
+        email,
+        userId,
+        name: newName,
```
bleonard: i deally, we can test all the types here (iterable types: long, integer, etc)

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,387 @@
...
+      oldProfileProperties: {
+        email,
+        userId,
+      },
+      newProfileProperties: {
+        email,
+        userId,
+      },
+      oldGroups: [],
+      newGroups: ["List One", "List Two"],
```
bleonard: use global instance variable

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,387 @@
...
+      toDelete: false,
+    });
+    await indexContacts(newNock);
+
+    const user = await getUser(email);
+    expect(user.dataFields.emailListIds.length).toBe(2);
+
+    const listOneId = await getListId("List One");
```
bleonard: in other tests, we save this (`listOneId`) globally and use it the rest of the time so don't look up each time.

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,387 @@
...
+
+    const user = await getUser(email);
+
+    const listTwoId = await getListId("List Two");
+    const listOneThree = await getListId("List Three");
+
+    expect(listTwoId).not.toBe(null);
+    expect(listOneThree).not.toBe(null);
+
+    expect(user.dataFields.emailListIds).toContain(listTwoId);
```
bleonard, line 336: Is this one supposed to be checking for listOne (the already subscribed list that wasn't passed in)

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,387 @@
...
+        email: alternativeEmail,
+        userId,
+      },
+      oldGroups: [],
+      newGroups: [],
+      toDelete: false,
+    });
+    await indexContacts(newNock);
+
+    const user = await getUser(alternativeEmail);
```
bleonard, line 381: check that the other email is not there.

plugins/@grouparoo/iterable/src/initializers/plugin.ts
```diff
@@ -0,0 +1,72 @@
...
+      ],
+      apps: [
+        {
+          name: "iterable",
+          options: [
+            {
+              key: "apiKey",
+              displayName: "Iterable API Key",
+              required: true,
+              description: "Iterable apikey (api) key.",
```
bleonard: Maybe we could say where to find it here.

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,387 @@
...
+        phoneNumber,
+        signupDate: exampleDate,
+      },
+      newProfileProperties: {
+        email,
+        userId,
+        name: newName,
+        phoneNumber: newPhoneNumber,
+      },
+      oldGroups: [],
```
bleonard, line 218: This plugin also supports random un-time properties. I'm not sure if we should literally make up a random string (not currently in getFields) and send it or not and make sure it works.

plugins/@grouparoo/iterable/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "prepare": "rm -rf dist && tsc --declaration",
+    "test": "jest",
+    "pretest": "npm run lint && npm run prepare",
+    "lint": "prettier --check src",
+    "watch": "tsc --watch"
+  },
+  "dependencies": {
+    "@grouparoo/app-templates": "^0.2.0-alpha.15",
+    "axios": "0.21.1",
+    "hubspot-api": "2.15.3",
```
bleonard: can remove this hubspot now and maybe axios too.
are use using app-templates?

plugins/@grouparoo/iterable/public/templates/app/{{{id}}}.js.template
```diff
@@ -0,0 +1,13 @@
+exports.default = async function buildConfig() {
+  return [
+    {
+      class: "app",
+      id: {{{id}}},
+      name: {{{id}}},
+      type: "{{{__pluginName}}}",
+      options: {
+        hapikey: "abc123", // Iterable hapikey (api) key
```
bleonard: i think this name is wrong (with the h)

plugins/@grouparoo/iterable/public/templates/destination/{{{id}}}.js.template
```diff
@@ -0,0 +1,28 @@
+exports.default = async function buildConfig() {
+  return [
+    {
+      id: {{{id}}},
+      name: {{{id}}},
+      class: "destination",
+      type: "{{{__pluginName}}}-export",
+      appId: "hubspotApp", // The id of the App this Source uses
```
bleonard: todo

plugins/@grouparoo/iterable/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,77 @@
...
+  appOptions,
+}) => {
+  const client = await connect(appOptions);
+  const required = getRequiredFields();
+  const known = await getUserFields(client);
+
+  return {
+    labels: {
+      property: {
+        singular: "Iterable Contact Property",
```
bleonard: Does iterable UI call them property or fields?

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,98 @@
...
+  const currentEmail = oldProfileProperties["email"];
+  if (!email) {
+    throw new Error(`newProfileProperties[email] is a required mapping`);
+  }
+
+  // Remove email from profile properties to avoid duplicates.
+  delete oldProfileProperties.email;
+  delete newProfileProperties.email;
+
+  try {
```
bleonard: maybe make a submethod (like intercom) for this whole thing and try/catch in the root.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      // remove from lists
+      for (const group of oldGroups) {
+        if (!newGroups.includes(group))
+          await removeFromList(client, appGuid, appOptions, email, group);
+      }
+
+      if (currentEmail && currentEmail !== email) {
+        const emailPayload = { currentEmail, newEmail: email };
+        if (newProfileProperties.userId) {
```
bleonard: it's interesting that this could also be the `old` userId.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,98 @@
...
+        await addToList(client, appGuid, appOptions, email, groupToAdd);
+      }
+
+      // remove from lists
+      for (const group of oldGroups) {
+        if (!newGroups.includes(group))
+          await removeFromList(client, appGuid, appOptions, email, group);
+      }
+
+      if (currentEmail && currentEmail !== email) {
```
bleonard: doesn't this email change have to happen _before_ using the email to add/remove from groups?

bleonard: let's make sure we have tests with to cover that case and then fix. I'm interested to know if this method will have to wait or not like `indexContacts` to make it work.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,98 @@
...
+  const client = await connect(appOptions);
+
+  const email = newProfileProperties["email"]; // this is how we will identify profiles
+  const currentEmail = oldProfileProperties["email"];
+  if (!email) {
+    throw new Error(`newProfileProperties[email] is a required mapping`);
+  }
+
+  // Remove email from profile properties to avoid duplicates.
+  delete oldProfileProperties.email;
```
bleonard: usually we don't change these inputs, but skip "email" in payload building if that is why you did it.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,98 @@
...
+  delete newProfileProperties.email;
+
+  try {
+    if (toDelete) {
+      await client.users.delete(email);
```
bleonard: it's ok to return early here to prevent the indentation for the rest of the method.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,98 @@
...
+      // create the user and set properties
+      const deletePropertiesPayload = {};
+      const newPropertyKeys = Object.keys(newProfileProperties);
+      Object.keys(oldProfileProperties)
+        .filter((k) => !newPropertyKeys.includes(k))
+        .forEach((k) => (deletePropertiesPayload[k] = null));
+
+      const dataFields = Object.assign(
+        {},
+        newProfileProperties,
```
bleonard: it's a minor thing, but I would suggest the other other order to make sure that `newProperfileProperties` always "wins"
```
const dataFields = Object.assign(
        {},
        deletePropertiesPayload
        newProfileProperties
)
```

plugins/@grouparoo/iterable/src/lib/export/listMethods.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export async function removeFromList(
+  client,
+  appGuid,
+  appOptions,
+  email,
+  groupName
+) {
+  const listId = await getListId(client, appGuid, appOptions, groupName);
+  await client.lists.unsubscribe({ listId, subscribers: [{ email }] });
```
bleonard: Do we ahve a test if the user i not in this list? We might have to catch it.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,98 @@
...
+      const payload = Object.assign(
+        { email },
+        { dataFields: formattedDataFields }
+      );
+
+      await client.users.update(payload);
+
+      // add to lists
+      for (const groupToAdd of newGroups) {
+        await addToList(client, appGuid, appOptions, email, groupToAdd);
```
bleonard: does the response from `client.users.update(payload);` above have the listIds the user is in? You could use that here to only add if they aren't in the list (and remove if only if they are in the list).
This would save API calls.

bleonard (commented):
On the right track!
Lots of comments, but nothing major.

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
pauloouriques: yes

plugins/@grouparoo/iterable/src/lib/export/listMethods.ts
```diff
@@ -0,0 +1,85 @@
...
+
+async function getListId(
+  client,
+  appGuid,
+  appOptions,
+  groupName
+): Promise<string> {
+  groupName = (groupName || "").toString().trim();
+  if (groupName.length === 0) {
+    return `Iterable empty groupName`;
```
bleonard: not sure about this.

bleonard: Not sure if the other plugins handle this. it's a good idea. --- let's try something more user facing like `"Grouparoo Default"`

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
pauloouriques: nope, the response shows only the status.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,101 @@
...
+  const client = await connect(appOptions);
+
+  const email = newProfileProperties["email"]; // this is how we will identify profiles
+  const currentEmail = oldProfileProperties["email"];
+  if (!email) {
+    throw new Error(`newProfileProperties[email] is a required mapping`);
+  }
+
+  if (toDelete) {
+    await client.users.delete(email);
```
bleonard: note, it could also be the old user to delete.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
bleonard: It's interesting that the `getUser` (from tests) call seems to have that data.
There's some number (2+) of lists where making that call is the more efficient thing. But then there is also the timing issue of the user showing up. Not great.

Maybe it's actually better to put that back at the beginning and use that data when making the list. It could also be combined with the change email situation to see that they are already there.


```
  let existingUser = await client.users.get(email);
  if (currentEmail && email !== currentEmail) {
    const oldUser = await client.users.get(oldEmail);
    if (oldUser) {
      ....
      await client.users.updateEmail(emailPayload);
      existingUser = oldUser;
      existingUser.email = email;
    }
  }
```
This makes the switchover much smarter at the expense of these 2 calls. and one of them makes the update lists case possible much more efficent.

Then you'll have `existingUser` lower down to use if it's not `null`. use `existingUser.dataFields.emailListIds` to make add/remove smarter.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
pauloouriques: can you explain better?

bleonard: The user that needs to be deleted could be `currentEmail` from `oldProperties`. If you went with the code I referenced in https://github.com/grouparoo/grouparoo/pull/1204#discussion_r566486570, then you'd maybe add deleting to that concept or do something like that to see who is really in there.

plugins/@grouparoo/hubspot/__tests__/utils/nockHelper.ts
```diff
@@ -1,7 +1,7 @@
 import { SimpleAppOptions } from "@grouparoo/core";
```
bleonard: One more file to put back as before.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,115 @@
+import { ExportProfilePluginMethod } from "@grouparoo/core";
+import { connect } from "../connect";
+import { addToList, removeFromList } from "./listMethods";
+
+export const exportProfile: ExportProfilePluginMethod = async (args) => {
+  try {
+    return sendProfile(args);
+  } catch (error) {
+    if (error?.response?.status === 429) {
+      return { error, success: false, retryDelay: 1000 * 11 }; // the most common rate-limit error from hubspot is in 10-second intervals
```
bleonard: is this how iterable really works?
just noting this still says "hubspot"

pnpm-lock.yaml
```diff
@@ -516,6 +519,42 @@ importers:
       prettier: 2.2.1
       ts-jest: 26.4.4
       typescript: 4.1.3
+  plugins/@grouparoo/iterable:
+    dependencies:
+      '@grouparoo/app-templates': link:../app-templates
+      axios: 0.21.1
+      hubspot-api: 2.15.3
```
bleonard: weird this is still here.

bleonard (commented):
Quick hubspot check. will look more soon.

plugins/@grouparoo/iterable/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,75 @@
...
+    destinationOptions: null,
+  });
+}
+
+describe("iterable/destinationMappingOptions", () => {
+  test("can fetch user fields", async () => {
+    const client = await connect(appOptions);
+    const fields = await getUserFields(client);
+    const phoneNumber = fields.find((f) => f.key === "phoneNumber");
+    expect(phoneNumber.type).toBe("string");
```
bleonard: seems likely this should be `phoneNumber` and the noted change in the mapping. It's ok to have default columns with known names like this.

Then maybe one more "first name" or something here to test for string.

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,463 @@
...
+  } catch (err) {
+    if (!suppressErrors) {
+      throw err;
+    }
+  }
+}
+
+async function deleteLists(suppressErrors) {
+  try {
+    for (const groupToDelete in [listOne, listTwo, listThree]) {
```
bleonard: probably `of` here. I think you are deleting "0" and "1" right now.

bleonard: `listFour` in this list, too.

plugins/@grouparoo/iterable/src/initializers/plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+          direction: "export",
+          description:
+            "Export profiles to Iterable as Users and put them in static Lists.",
+          app: "iterable",
+          options: [],
+          methods: {
+            exportProfile,
+            destinationOptions,
+            destinationMappingOptions,
+            exportArrayProperties,
```
bleonard, line 63: Note: next we'll want to look more into array properties and `getUserFields`

plugins/@grouparoo/iterable/src/lib/export/destinationMappingOptions.ts
bleonard: It's usually a case statement at the top then the falls through to this mapping in the default type. You'd pass in the field name. SO if it's `phoneNumber` jsut said it's that type.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,115 @@
...
+        const emailPayload = { currentEmail, newEmail: email };
+        if (newProfileProperties.userId) {
+          emailPayload["currentUserId"] = newProfileProperties.userId;
+        }
+        await client.users.updateEmail(emailPayload);
+      }
+    }
+
+    await client.users.update(payload);
+    const currentUser = await getUser(client, email);
```
bleonard: When this user is added, don't we have the issue of it taking N seconds. I think if you had a test that added a new user _and_ put them groups we might have a problem here in that it would be null. Or if you changed the email and added groups, it wouldn't find them here.

The code I suggested before had this above `if (currentEmail && currentEmail !== email)` and reassigned the `currentUser` to the changed one.
```
  let currentUser = await client.users.get(email);
  if (currentEmail && email !== currentEmail) {
    const oldUser = await client.users.get(oldEmail);
    if (oldUser) {
      ....
      await client.users.updateEmail(emailPayload);
      currentUser = oldUser;
      currentUser.email = email;
    }
  }
```

With that approach a null `currentUser` will be known to not be in any groups. and you won't have that `indexContacts` type issue.

plugins/@grouparoo/iterable/src/lib/export/listMethods.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+  return null;
+}
+
+export async function addToList(client, appGuid, appOptions, user, groupName) {
+  const listId = await getListId(client, appGuid, appOptions, groupName);
+  if (
+    !user.dataFields.emailListIds ||
+    !user.dataFields.emailListIds.includes(listId)
+  )
```
bleonard: add `{` here to be clear it's all in the `if`

maybe even better to add semantics for clarity.
```
    const currentListIds = user?.dataFields?.emailListIds || [];
    if(currentListIds.includes(listId)) {
      return;
    }
    await client.lists.subscribe({
      listId,
      subscribers: [{ email: user.email }],
    });
```

plugins/@grouparoo/iterable/src/lib/export/listMethods.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export async function removeFromList(
+  client,
+  appGuid,
+  appOptions,
+  user,
+  groupName
+) {
+  const listId = await getListId(client, appGuid, appOptions, groupName);
```
bleonard, line 86: same clarity idea as `add`, but in reverse.

bleonard (commented):
maybe still a few things about timing, especially on new user and change email.

plugins/@grouparoo/iterable/__tests__/fixtures/destination-mapping-options.js
```diff
@@ -0,0 +1,42 @@
+const nock = require('nock');
```
bleonard, line 1: note: you'll want to prettier format these `nocks`

plugins/@grouparoo/iterable/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,77 @@
...
+  const fields = await client.users.getFields();
+  const out = [];
+
+  for (const [key, value] of Object.entries(fields.fields)) {
+    if (value !== "object" && key !== "email") {
+      const type: DestinationMappingOptionsResponseTypes = mapTypesFromIterableToGrouparoo(
+        value
+      );
+      if (type) {
+        out.push({ key, type, important: true });
```
bleonard: they probably aren't all `important`. We probably need a whitelist with these in it:
name
phoneNumber
userId
signupDate

I don't think we want the ones with `"."` to be `important`
Maybe that's a way to do it.

I think we also want to leave out these ones:
emailListIds
unsubscribedChannelIds
unsubscribedMessageTypeIds

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -69,32 +69,27 @@ export const sendProfile: ExportProfilePluginMethod = async ({
...
-          emailPayload["currentUserId"] = newProfileProperties.userId;
-        }
-        await client.users.updateEmail(emailPayload);
-        currentUser = oldUser;
-        currentUser.email = email;
+      const emailPayload = { currentEmail, newEmail: email };
+      if (newProfileProperties.userId) {
+        emailPayload["currentUserId"] = newProfileProperties.userId;
       }
+      await client.users.updateEmail(emailPayload);
```
bleonard: let's get a test to understand what happens here if the `old` email address is not in the system. Might have to catch.

PR #1223, opened 2021-01-26
Use mapping values as prereq for determining code config order

seancdavis:
This closes T-984 by using values (instead of keys) as the prerequisite ID value when building the config dependency tree thing.

I looked through all the templates and it appears that we are consistently using the mapping values as the internal ID, so this is a change we want and can safely make.

That said, I don't see any tests for the file I adjusted, except when used through the code config module. But I also don't see tests for any of the files in `core/src/classes`. Do we want specs to cover this file? If so, I can add to this this PR.

### DIFF
--- a/core/__tests__/classes/codeConfig.ts
+++ b/core/__tests__/classes/codeConfig.ts
@@ -1,0 +1,35 @@
+import path from "path";
+import { loadConfigObjects } from "../../src/modules/configLoaders";
+import { getParentIds } from "../../src/classes/codeConfig";
+
+describe("classes/codeConfig", () => {
+  describe("#getParentIds", () => {
+    let configObjects;
+
+    beforeAll(async () => {
+      const dir = path.join(
+        __dirname,
+        "..",
+        "fixtures",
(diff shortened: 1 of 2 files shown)


evantahler:
I think you found the source of the bug for sure!   Yes please add a test!

* It looks like we are implicitly testing the `sortConfigurationObject` method in the integration tests here: https://github.com/grouparoo/grouparoo/blob/master/core/__tests__/modules/codeConfig/codeConfig.ts.  A good starting point would be to modify the codeConfig test fixtures to include a destination with a mapping that would have broken previously.  
* and here's an example of. using the `sortConfigurationObject` method directly: https://github.com/grouparoo/grouparoo/blob/master/core/__tests__/modules/codeConfig/validate.ts#L64-L65 if you want to do some unit testing

seancdavis:
I didn't find an easy way to recreate this. I expected moving the order of elements around would do the trick, but that didn't do it. Therefore, I created a spec specifically for the code config class and added a couple unit tests to protect against the case this solves.

PR #1224, opened 2021-01-26
Better description for table v query sources on CLI

seancdavis:
I don't know if these are accurate or any good? 🤷‍♂️ 

Closes T-979


plugins/@grouparoo/app-templates/src/source/table/templates.ts
```diff
@@ -18,7 +18,7 @@ export class TableSourceTemplate extends ConfigTemplate {
   ) {
     super();
     this.name = `${name}:table:source`;
-    this.description = `Config for a ${name} Table Source`;
+    this.description = `Config for a ${name} Table Source. Construct properties from the data in the table.`;
```
evantahler: Maybe `... without writing SQL`?

evantahler (approved):

PR #1234, opened 2021-01-27
`parseInt` should always be passed a radix

evantahler:
It's generally safer to pass a radix (ie: the base of the number system you are using) to `parseInt`.  See https://stackoverflow.com/questions/6611824/why-do-we-need-to-use-radix-parameter-when-calling-parseint for all sorts of ways the default guess of the radix can go bad.

Closes T-902

### DIFF
--- a/core/__tests__/modules/groupExport.ts
+++ b/core/__tests__/modules/groupExport.ts
@@ -94,44 +94,44 @@ describe("modules/groupExport", () => {
       // mario
       let properties = await mario.properties();
       expect(rows[0].guid).toBe(mario.guid);
-      expect(parseInt(rows[0].createdAt)).toBeGreaterThan(0);
+      expect(parseInt(rows[0].createdAt, 10)).toBeGreaterThan(0);
       expect(rows[0]["email"]).toBe(properties.email.values[0]);
       expect(parseFloat(rows[0]["ltv"])).toBe(properties.ltv.values[0]);
       lastLoginAt = properties.lastLoginAt.values[0] as Date;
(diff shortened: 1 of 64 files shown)


evantahler:
But... ESLint (https://eslint.org/docs/rules/radix) is saying that in ES5+ the radix is fixed at `10`... so maybe this isn't needed?

evantahler:
@seancdavis Do you know a good resource for checking what ES versions each version of node.js uses?  I can find the V8 versions... but not the underlying ES versions :(

Ideally, it would be awesome to not do this pull request. Grouparoo requires node.js v12+

bleonard:
There's some hints on that Stack OVerflow post that 10 is now the default

bleonard:
and if we do have to do it, let's think about declaring another function or something

seancdavis:
@evantahler Not off the top of my head. I found [this](https://bytearcher.com/articles/how-to-check-if-node-implements-es6-language-feature/), which makes it sound like the easiest way to get there is to look at specific features available.

Otherwise you're stuck digging through first mapping node versions to v8 versions and then looking up compatibility. But even there, still, I couldn't find specific ES version refs, only a way to find available features at that point in time.

[This thing is also cool](https://kangax.github.io/compat-table/es6/). Might be a faster way to get there, but it is also feature-based.

evantahler:
Since we have pretty good control over the inputs to `parseInt()`, and as time goes on, we are going to get a default radix of base-10 (or we have already?), I'm personally comfortable not merging this in.

PR #1241, opened 2021-01-28
Script to write group property ops to JSON file

seancdavis:
⚠️  **DEPENDS on grouparoo/www.grouparoo.com#176** ⚠️ 

---

The script is `./core/bin/build_static_refs`. It writes two JSON files to `./core`. There are tests to cover the script.

This still breaks up the rule ops file into a function and an object. That was a simple way for me to be able to switch between pg and sqlite and output two separate files.

Lastly, this changes the filename of `RuleOpsDictionary` to `ruleOpsDictionary`. I had some issues with this and VS Code. VSC keep capitalizing the name, but only within the editor — it doesn't seem to mess with it on the file system.

---

Lastly, note that this is only really a third of the #176694391 story. I still have to add the page to www and then notes to link to it from the calculated groups template.


evantahler (commented):
Nice! 

Just adding out conversation notes here for reference:

* This is the first step to get data out of core in a way that the WWW project can use.  The output of this file will be coppied into  `www/data` for use there, to power content for new `www/docs` pages.
* We tried a few ways of requiring `@grouparoo/core` into `www`, but it just won't work - there are too many assumptions that `@grouparoo/core` is running in a "node" environment, not a "browser" environment
* In the future, we'll use CI to run this script weekly? and copy the data over to the matching file in the WWW project via a GitHub action or similar.

Suggestions:
* Add the output of the script to `gitignore` maybe?

seancdavis:
This is now ready to be merged. Artifacts added to gitignore, and calculated groups template updated to refer to the www docs (which is why this now depends on the other PR).

evantahler (approved):

PR #1243, opened 2021-01-29
Split the core tests into a chunk for each test dir

evantahler:
This should help us identify and perhaps solve some flaky tests on. CI


bleonard:
and probably be faster!

PR #1251, opened 2021-02-02
Set the Sequelize pool size via  WORKERS or SEQUELIZE_POOL_SIZE, whichever is larger

evantahler:
This PR will ensure that Grouparoo created N+1 connections between your application and the Grouparoo postgres database, where N is the number of workers running within this process.  This should help reduce the class of bugs related to Seqeulize connections, timeouts, etc

### DIFF
--- a/core/src/config/sequelize.ts
+++ b/core/src/config/sequelize.ts
@@ -97,12 +97,11 @@ export const DEFAULT = {
       migrations: [join(__dirname, "..", "migrations")],
       storage, // only used for sqlite
       pool: {
-        max:
-          dialect === "sqlite"
-            ? 1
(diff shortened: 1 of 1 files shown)


core/src/config/sequelize.ts
```diff
@@ -97,12 +97,11 @@ export const DEFAULT = {
       migrations: [join(__dirname, "..", "migrations")],
       storage, // only used for sqlite
       pool: {
-        max:
-          dialect === "sqlite"
```
bleonard, line 101: lose this sqlite affordance on purpose?

evantahler: Good catch! We already have a check that ensures you only have 1 worker for SQLite, but the `+1` gets you in trouble...

PR #1254, opened 2021-02-03
Iterable plugin Destination

pauloouriques:


plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,121 @@
+import { ExportProfilePluginMethod } from "@grouparoo/core";
+import { connect } from "../connect";
+import { addToList, removeFromList } from "./listMethods";
+
+export const exportProfile: ExportProfilePluginMethod = async (args) => {
+  try {
+    return sendProfile(args);
+  } catch (error) {
+    if (error?.response?.status === 429) {
+      let retryIn = 60; // per minute
```
bleonard: I noticed their api is actually per second (not per minute). so maybe this should just a random number between say, 2 and 10 seconds (*1000)

bleonard:
Fixes https://github.com/grouparoo/grouparoo/issues/651

PR #1255, opened 2021-02-03
Rename GUIDs to IDs

evantahler:
Throughout Grouparoo, we replace the concepts of "GUID" with "ID"
This will make code-config easier to reason about, because the user-provided IDs match what ends up in the database. '

To aid the migration, a new `TaskUtils.reEnqueueIfGuidParams` module is included.  This runs at the beginning of all tasks and checks for `guid`-style inputs.  If they are found, we convert them to `id`-style inputs and re-enqueue the task.

This PR also:
- Removes the requirement that code-config objects use the model prefixes 
- Adds missing 'delete' pass to the `grouparoo apply` command


evantahler:
Rebased after https://github.com/grouparoo/grouparoo/pull/1254 to include Iterable plugin

PR #1260, opened 2021-02-04
Update stale docs references

seancdavis:
And add one new one. This is a result of rearranging the docs taxonomy (grouparoo/www.grouparoo.com#184) and comes alongside the changes in grouparoo/www.grouparoo.com#190.


evantahler (approved):
👍

PR #1267, opened 2021-02-06
Destination Sendgrid 

pauloouriques:


plugins/@grouparoo/sendgrid/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,91 @@
...
+  type: DestinationMappingOptionsResponseTypes;
+}> => {
+  return [{ key: "email", type: "email" }];
+};
+
+const isImportant = (key): Boolean => {
+  return importantFields.includes(key);
+};
+
+const mapTypesFromIterableToGrouparoo = (fieldKey, iterableType) => {
```
bleonard: iterable reference.

plugins/@grouparoo/sendgrid/README.md
```diff
@@ -0,0 +1,5 @@
+# @Grouparoo/Sendgrid
+
+## Attribution
+
+The Sendgrid icon used in this plugin is from https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcT2Y8hmS6roynof_6ktxCKHxLt45oz1Whl8qQ&usqp=CAU
```
bleonard: probably want the website that linked to this and not the cdn image.

plugins/@grouparoo/sendgrid/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,71 @@
...
+    destinationId: null,
+    destinationOptions: null,
+  });
+}
+
+describe("sendgrid/destinationMappingOptions", () => {
+  test("can fetch user fields", async () => {
+    const client = await connect(appOptions);
+    const fields = await getUserFields(client);
+    const user = await client.getUser("ourixilva@gmail.com");
```
bleonard: probably don't need this.

plugins/@grouparoo/sendgrid/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const requiredFieldEmail = required.find((f) => f.key === "email");
+    expect(requiredFieldEmail.key).toBe("email");
+    expect(requiredFieldEmail.type).toBe("email");
+
+    const knownFieldPhoneNumber = known.find((f) => f.key === "phone_number");
+    expect(knownFieldPhoneNumber.type).toBe("phoneNumber");
+    expect(knownFieldPhoneNumber.important).toBe(true);
+
+    const knownFieldPhoneSignupDate = known.find((f) => f.key === "created_at");
```
bleonard, line 70: maybe check a "Number" and "Text" from sendgrid too.

plugins/@grouparoo/sendgrid/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (!suppressErrors) {
+      throw err;
+    }
+  }
+}
+
+async function cleanUp(suppressErrors) {
+  await deleteUsers(suppressErrors);
+  await deleteLists(suppressErrors);
+  await indexContacts(newNock, 30 * 1000);
```
bleonard, line 111: Was this found to be needed?

plugins/@grouparoo/sendgrid/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "pretest": "npm run lint && npm run prepare",
+    "lint": "prettier --check src",
+    "watch": "tsc --watch"
+  },
+  "dependencies": {
+    "@grouparoo/app-templates": "^0.2.0-alpha.17",
+    "@sendgrid/client": "^7.4.2"
+  },
+  "peerDependencies": {
+    "@grouparoo/core": "^0.2.0",
```
bleonard: There's been an update on master that removed `peerDependencies` 
It any case, need to make this whole file look more like the plugin/package.json on master branch.

plugins/@grouparoo/sendgrid/public/templates/destination/{{{id}}}.js.template
```diff
@@ -0,0 +1,28 @@
...
+      appId: "sendgridApp", // The id of the App this Source uses
+      groupId: "high_value_customers", // The id of the group whose members you want to export
+
+      options: {},
+
+      // Mappings are how you choose which properties to export to this destination.
+      // Keys are the name to display in the destination, values are the IDs of the Properties in Grouparoo.
+      mapping: {
+        email: "email",
+        firstname: "firstName",
```
bleonard: ideally, there would be the default field names that sendgrid actually uses. `first_name` for example.

plugins/@grouparoo/sendgrid/src/initializers/plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+          ],
+          methods: { test },
+        },
+      ],
+      connections: [
+        {
+          name: "sendgrid-export-marketing",
+          direction: "export",
+          description:
+            "Export profiles to Sendgrid marketing as Contacts and put them in static Groups.",
```
bleonard: "Lists" ?

plugins/@grouparoo/sendgrid/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,91 @@
...
+  switch (fieldKey) {
+    case "phone_number":
+      return "phoneNumber";
+  }
+  const map = {
+    Text: "string",
+    Date: "date",
```
bleonard, line 54: `Number` ?

bleonard: You might have to make a field in Sendgrid that is a number to test it correctly. Put the instrucitons in the `.env.example` -- see examples in other plugins. Generally, we'd called it `number_field`

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,86 @@
...
+    oldProfileProperties,
+    newGroups,
+    oldGroups,
+  },
+}) => {
+  if (Object.keys(newProfileProperties).length === 0) {
+    return { success: true };
+  }
+  const client = await connect(appOptions);
+  const email = newProfileProperties["email"]; // this is how we will identify profiles
```
bleonard: I note in the documentation that: `The email field is case sensitive.`
So we'll probably want to trim and lowercase it if it's not null. (the `old` one too)

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,86 @@
...
+  const currentEmail = oldProfileProperties["email"];
+  if (!email) {
+    throw new Error(`newProfileProperties[email] is a required mapping`);
+  }
+
+  if (toDelete) {
+    let userToDelete = null;
+    if (currentEmail && currentEmail !== email) {
+      userToDelete = client.getUser(currentEmail);
+    } else {
```
bleonard, line 48: let's change this from an `else` to to a `if(!userToDelete)` so it will find the user if their email is already changed in the system.

If we end up using this result in the `set` case below, then we would move that "find user" situation above the `if(toDelete)` check. For example, if you need it to handle the lists.

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,86 @@
...
+      .forEach((k) => (deletePropertiesPayload[k] = null));
+
+    const dataFields = Object.assign(
+      {},
+      deletePropertiesPayload,
+      newProfileProperties
+    );
+    const formattedDataFields = {};
+    for (const key of Object.keys(dataFields)) {
+      formattedDataFields[key] = formatVar(dataFields[key]);
```
bleonard, line 70: I think this is not going to be quite right for custom fields. We should make a few and make sure we test them.
The documentation suggests they will be nested within `custom_fields` object.

bleonard: That means there will either:
1) be a whitelist of "top-level" ones
2) the key itself knows which ones are custom fields (as returned from `destinationPropertyMappings`) by it being like `custom_fields.number_field`.

plugins/@grouparoo/sendgrid/src/lib/export/listMethods.ts
```diff
@@ -0,0 +1,91 @@
...
+function filterLists(allLists, groupName) {
+  allLists = allLists || [];
+  const matchingList = allLists.filter((list) => list.name === groupName)[0];
+  if (matchingList) {
+    return matchingList.id;
+  }
+  return null;
+}
+
+export async function addToList(client, appId, appOptions, email, groupName) {
```
bleonard: In general, I think when we read through the documentation, we saw that the adding would be done in the `PUT` call.

bleonard (commented):
Reminder of the eval we did of the documenation for finishing up `exportProfile`.
This generally worked around the async nature of the API.

1. Look up by email to see if they exist and get id
2. If delete, delete them and return
3. Make all the lists that we will need and get their ids
4. Either way (If the user did exist or not), call upsert api with each of the newGroup lists in the listIds field.
5. Need to check if this is only additive. If it is not additive (and you are just setting the full list), then we combine the lists from the search from #1 with newGroups and call that - but the docs say it is additive.
6. If the user did exist, the call the api to remove them from the groups they should be removed from.

This probably means moving the `getUser` call to the top and handling email changes there.

plugins/@grouparoo/sendgrid/__tests__/utils/shared.ts
```diff
@@ -0,0 +1,8 @@
+import { helper } from "@grouparoo/spec-helper";
+
+export async function indexContacts(isNewNock, delayTime = 30 * 1000) {
```
bleonard, line 3: can delete this file.

plugins/@grouparoo/sendgrid/src/lib/apiHelper.ts
```diff
@@ -0,0 +1,110 @@
+import { Client } from "@sendgrid/client";
```
bleonard: let's call this file `client.ts` - a few other plugins have that.

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,117 @@
...
+  if (!email) {
+    throw new Error(`newProfileProperties[email] is a required mapping`);
+  }
+
+  const user = await client.getUser(email);
+
+  if (toDelete) {
+    let userToDelete = null;
+    if (oldEmail && oldEmail !== email) {
+      userToDelete = client.getUser(oldEmail);
```
bleonard: not doing `await` here.

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,117 @@
...
+    for (const key of Object.keys(dataFields)) {
+      formattedDataFields[key] = formatVar(dataFields[key]);
+    }
+
+    let listsToAdd = [];
+    let existingLists = user?.["list_ids"] || [];
+
+    // change email
+    if (!user && oldEmail && oldEmail !== email) {
+      const oldUser = await client.getUser(oldEmail);
```
bleonard, line 75: we're fetching `oldUser` in the regular and delete cases. consider moving to top.

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,117 @@
...
+      if (oldUser) {
+        formattedDataFields = Object.assign({}, oldUser, formattedDataFields);
+        formattedDataFields["id"] = undefined;
+        formattedDataFields["_metadata"] = undefined;
+        listsToAdd = oldUser["list_ids"] || [];
```
bleonard: probably don't want to carry these through. just cause the old user was on it, doesn't mean the new user will be. that's not our job.

bleonard: Same fo existing lists.

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,117 @@
...
+        listsToAdd.push(listId);
+      }
+    }
+    await client.addOrUpdateUser(formattedDataFields, listsToAdd);
+
+    // remove from lists
+    if (user) {
+      for (const group of oldGroups) {
+        if (!newGroups.includes(group)) {
+          await removeFromList(client, appId, appOptions, user["id"], group);
```
bleonard: i waonder if we can check their current lsits here to see if they are really in it.

bleonard (commented):
Looking good!

plugins/@grouparoo/sendgrid/__tests__/.env.example
```diff
@@ -0,0 +1,9 @@
+SENDGRID_API_KEY = "SG.jklçsdaoipwerfhklvds907akjlçsdfjklçsdf7890124-9=rjnklgdfçjklfgd890"
+
+
+# The sendgrid account should have a few custom fields pre added
+# there is one for each type of field
+# text, number, date
+# text_field (number)
```
bleonard, line 7: text?

plugins/@grouparoo/sendgrid/package.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "scripts": {
+    "prepare": "rm -rf dist && tsc --declaration",
+    "test": "jest",
+    "pretest": "npm run lint && npm run prepare",
+    "lint": "prettier --check src",
+    "watch": "tsc --watch"
+  },
+  "dependencies": {
+    "@grouparoo/app-templates": "^0.2.1-alpha.0",
```
bleonard, line 31: probably more updates to do here. sorry, we are moving fast on the versions.

plugins/@grouparoo/sendgrid/public/templates/destination/{{{id}}}.js.template
```diff
@@ -0,0 +1,28 @@
+exports.default = async function buildConfig() {
+  return [
+    {
+      id: {{{id}}},
+      name: {{{id}}},
+      class: "destination",
+      type: "{{{__pluginName}}}-export",
```
bleonard, line 7: "{{{__pluginName}}}-export-marketing"

plugins/@grouparoo/sendgrid/public/templates/destination/{{{id}}}.js.template
```diff
@@ -0,0 +1,28 @@
...
+      appId: "sendgridApp", // The id of the App this Source uses
+      groupId: "high_value_customers", // The id of the group whose members you want to export
+
+      options: {},
+
+      // Mappings are how you choose which properties to export to this destination.
+      // Keys are the name to display in the destination, values are the IDs of the Properties in Grouparoo.
+      mapping: {
+        email: "email",
```
bleonard, line 16: are these the real fields?

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,117 @@
...
+  const client = await connect(appOptions);
+  let email = newProfileProperties["email"]; // this is how we will identify profiles
+  let oldEmail = oldProfileProperties["email"];
+  if (!email) {
+    throw new Error(`newProfileProperties[email] is a required mapping`);
+  }
+  email = email.trim().toLowerCase();
+  if (oldEmail) {
+    oldEmail = oldEmail.trim().toLowerCase();
+  }
```
bleonard, line 41: bit of a funny case where it can now be empty.
maybe make a `cleanEmail` function and it at the beginning.
```
function cleanEmail(value) {
  if (!value) {
    return null;
  }
  value = value.toString().toLowerCase().trim();
  if (value.length === 0) {
    return null;
  }
  return value;
}
```

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,117 @@
...
+  const user = await client.getUser(email);
+
```
bleonard, line 43: My idea is to have something like this up here
```
let oldUser = null;
if (oldEmail && oldEmail !== email) {
  oldUser = await client.getUser(oldEmail);
}
```

Then user that in the delete case
```
const userToDelete = user || oldUser;
```

And further down:
```
      if (oldUser) {
        await client.deleteUsers([oldUser.id]);
      }
```

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,117 @@
...
+    let formattedDataFields = {};
+    for (const key of Object.keys(dataFields)) {
+      formattedDataFields[key] = formatVar(dataFields[key]);
+    }
+
+    // change email
+    if (!user && oldEmail && oldEmail !== email) {
+      const oldUser = await client.getUser(oldEmail);
+      if (oldUser) {
+        formattedDataFields = Object.assign({}, oldUser, formattedDataFields);
```
bleonard, line 77: why are we changing `formattedDataFields` here? I don't think it's our job to move anything over form the old user to the new one. The data we get in this function should be enough. Or am I missing something?

plugins/@grouparoo/sendgrid/src/lib/export/listMethods.ts
```diff
@@ -0,0 +1,80 @@
...
+    return matchingList.id;
+  }
+  return null;
+}
+
+export async function removeFromList(
+  client,
+  appId,
+  appOptions,
+  user,
```
bleonard, line 72: this user can be null, so you'll have to something like
```
const existingLists = user?.["list_ids"] || [];
```

plugins/@grouparoo/sendgrid/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,117 @@
...
+        delete formattedDataFields["id"];
+        delete formattedDataFields["_metadata"];
+        await client.deleteUsers([oldUser.id]);
+      }
+    }
+
+    let listsToAdd = [];
+    let existingLists = user?.["list_ids"] || [];
```
bleonard, line 85: do a pass on variables and see if they can be `const`

plugins/@grouparoo/sendgrid/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,483 @@
...
+const otherEmail = "sandro.arturo@mailinator.com";
+const first_name = "Caio";
+const last_name = "Silveira";
+const company = "grouparoo";
+const alternativeName = "Evan";
+const otherName = "Lucas";
+const city = "Campina Grande";
+const newEmail = "carlos.solimoes@mailinator.com";
+const newName = "Carlos";
+const signup_date = new Date(1597870204 * 1000);
```
bleonard, line 29: do a pass to make sure these are used.

plugins/@grouparoo/sendgrid/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,483 @@
...
+    // Phone must be valid.
+    await runExport({
+      oldProfileProperties: {
+        email,
+        first_name,
+        phone_number,
+      },
+      newProfileProperties: {
+        email,
+        first_name: alternativeName,
```
bleonard, line 204: we want to make sure to try one of each type here.
Date, Number, etc.
And then verify it worked.

PR #1281, opened 2021-02-10
Update Bigquery and Snowflake package.json to indicate node v14 is OK

evantahler:


evantahler:
Doesn't appear to be a deal-breaker when installing on node v14, but does cause some warnings

PR #1299, opened 2021-02-12
`grouparoo init` does not include any plugins

evantahler:
This PR removes plugins from new Grouparoo projects generated with `grouparoo init`, other than `ui-community`.  

This PR also adds messaging to the end of the `grouparoo init` and `grouparoo generate --list` commands to remind users they can add plugins to add additional functionality

> You can add plugins to to this project to connect to new Sources and Destinations and add additional commands with the `grouparoo install` command.

### DIFF
--- a/cli/templates/package.json
+++ b/cli/templates/package.json
@@ -9,40 +9,15 @@
     "node": ">=12.0.0 <16.0.0"
   },
   "dependencies": {
-    "@grouparoo/bigquery": "~~VERSION~~",
     "@grouparoo/core": "~~VERSION~~",
-    "@grouparoo/csv": "~~VERSION~~",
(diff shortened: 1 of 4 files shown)


cli/templates/package.json
```diff
@@ -9,40 +9,15 @@
...
-      "@grouparoo/redshift",
-      "@grouparoo/salesforce",
-      "@grouparoo/snowflake",
-      "@grouparoo/ui-community",
-      "@grouparoo/zendesk"
-    ],
-    "includedFiles": [
-      "assets"
+      "@grouparoo/ui-community"
     ]
```
evantahler, line 21: Also removing this, as no one is using it

PR #1301, opened 2021-02-12
Copy Code Config README templates into the work directory.

seancdavis:
This also results in a skeleton structure representing the directories in which Code Config files will be generated.

I added README files for:

- `/config`
- `/config/apps`
- `/config/destinations`
- `/config/groups`
- `/config/properties`
- `/config/sources`

After running `init` and then generating one object of each type, this is what the structure looks like:

Screen Shot 2021-02-12 at 4 00 41 PM

---

I didn't add any tests. There aren't many CLI tests. Would you like some here?

### DIFF
--- a/cli/src/lib/initialize.ts
+++ b/cli/src/lib/initialize.ts
@@ -24,13 +24,23 @@ export default async function Initialize(
     logger.succeed(`Created ${workDir}`);
   }
 
-  const configDir = path.join(workDir, "config");
-  if (!fs.existsSync(configDir) || opts.force) {
-    fs.mkdirpSync(configDir);
-    logger.succeed("Created config directory");
-  } else {
-    logger.warn("config directory already exists, not modifying");
-  }
+  /**
+   * Copy Code Config README templates into the work directory. This also
+   * results in a skeleton structure representing the directories in which Code
+   * Config files will be generated.
+   */
+  const templates = Templates.getConfigTemplates();
+  templates.map(({ absoluteFilePath, relativeFilePath }) => {
+    const destDir = path.dirname(relativeFilePath);
+    if (!fs.existsSync(destDir)) {
+      fs.mkdirSync(destDir, { recursive: true });
+    }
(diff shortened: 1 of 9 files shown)


cli/src/lib/initialize.ts
```diff
@@ -24,13 +24,23 @@ export default async function Initialize(
...
+   * Copy Code Config README templates into the work directory. This also
+   * results in a skeleton structure representing the directories in which Code
+   * Config files will be generated.
+   */
+  const templates = Templates.getConfigTemplates();
+  templates.map(({ absoluteFilePath, relativeFilePath }) => {
+    const destDir = path.dirname(relativeFilePath);
+    if (!fs.existsSync(destDir)) {
+      fs.mkdirSync(destDir, { recursive: true });
+    }
```
evantahler: Not a big deal, but since we are using `fs-extra`, this could be replaced with `fs.mkdirpsync()`

cli/templates/config/README.md
```diff
@@ -0,0 +1,11 @@
+# Code Configuration
+
+Welcome to your very own Grouparoo project! You are awesome for supporting open-source software!
```
evantahler: I'm not sure about "open source" being the main thing we call out here.  Maybe just stop after the welcome?

@andyjih what do you think?

evantahler (approved):
Love the content and the automatic copying of whatever the config template structure is already.  

I put some minor notes in just as an FYI.  We can tweak the copy later... approved by me!

evantahler:
Note that the CLI test is failing thought... something is off about the template copy paths

evantahler:
Don't forget to add PR labels!

evantahler:
You can test that your new readme files were generated in `test.sh` with something like:

```bash
if [ -d "$WORKDIR/config/README.md" ]; then
    echo "✅ config/README exists."
else
    echo "🚫 config/README does not exist."
    exit 1
fi
```

PR #1305, opened 2021-02-15
Validate config objects for duplicate ID values

seancdavis:
This will validate for duplicate ID values. This uniqueness validation is type-agnostic. While not strictly necessary, it has some benefits:

- Simpler logic.
- Potentially faster for users to debug.
- Seems like a good practices anyways — even though you _can_ have duplicate ID values across types, you probably _shouldn't_.
- One less thing for the object checker in the UI to do.

Screen Shot 2021-02-15 at 9 07 48 AM

This also moves existing core test util to a `__utils__` directory. It was being ignored. I wanted to use `utils` for a test because it matches the structure of `src`. I don't feel strongly about `__utils__` being the name or this being the ideal solution. I mostly just wanted it to get out of the way.

### DIFF
--- a/core/src/modules/configLoaders/index.ts
+++ b/core/src/modules/configLoaders/index.ts
@@ -5,6 +5,7 @@ import glob from "glob";
 import {
   ConfigurationObject,
   sortConfigurationObjects,
+  validateConfigObjects,
   IdsByClass,
 } from "../../classes/codeConfig";
 import { loadApp, deleteApps } from "./app";
@@ -110,6 +111,12 @@ export async function processConfigObjects(
   const errors: string[] = [];
 
   configObjects = sortConfigurationObjects(configObjects);
+
+  const { errors: validationErrors } = validateConfigObjects(configObjects);
+  validationErrors.map((err) =>
+    log(`[ config ] ${err}`, env === "test" ? "info" : "error")
+  );
+  errors.push(...validationErrors);
 
   for (const i in configObjects) {
     const configObject = configObjects[i];
--- a/core/src/modules/configLoaders/index.ts
+++ b/core/src/modules/configLoaders/index.ts
@@ -118,6 +118,8 @@ export async function processConfigObjects(
   );
   errors.push(...validationErrors);
 
+  if (errors.length > 0) return { seenIds, errors };
+
   for (const i in configObjects) {
     const configObject = configObjects[i];
     if (Object.keys(configObject).length === 0) continue;
--- a/core/__tests__/modules/codeConfig/errors.ts
+++ b/core/__tests__/modules/codeConfig/errors.ts
@@ -3,7 +3,10 @@ import { Setting, Team } from "../../../src";
 import path from "path";
 import { api } from "actionhero";
 import { loadConfigDirectory } from "../../../src/modules/configLoaders";
-import { validateConfigObjectKeys } from "../../../src/classes/codeConfig";
+import {
+  validateConfigObjectKeys,
+  validateConfigObjects,
+} from "../../../src/classes/codeConfig";
 
 describe("modules/codeConfig", () => {
   helper.grouparooTestServer({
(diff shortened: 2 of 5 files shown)


core/src/utils/validators.ts
```diff
@@ -0,0 +1,15 @@
...
+ * @param key (optional) If the array is an array of objects, use this property
+ * to locate duplicates.
+ */
+export function extractDuplicates(objects: any[], key?: string): string[] {
+  const values: string[] = key ? objects.map((obj) => obj[key]) : objects;
+  const onlyUnique = (value, index, self) => self.indexOf(value) === index;
+  const onlyDuplicates = (value, index, self) => self.indexOf(value) !== index;
+  const duplicates = values.filter(onlyDuplicates).filter(onlyUnique);
+  return duplicates;
+}
```
evantahler: Fancy!

evantahler (changes requested):
Great testing on that meta-model!

As a note, there are no tests in `__tests__/utils`, that's a helper that's used by other tests, which was why it's not being tested.  It's for the group rule tests https://github.com/grouparoo/grouparoo/blob/ab517ae159ed5ce3a32ad6528273af7874610ee5/core/__tests__/models/group/rules/urls.ts

seancdavis:
> As a note, there are no tests in __tests__/utils, that's a helper that's used by other tests, which was why it's not being tested. It's for the group rule tests

@evantahler [There is now](https://github.com/grouparoo/grouparoo/pull/1305/files#diff-8ead169be05a12fe8e27147807552362a9e65b062c4e40820be183582a4d0f6d)! That's why I moved the the test utils to a place that was different than the core's utils.

If you rather I go about this a different way, I'd be happy to adjust.

seancdavis:
Alrighty. I simplified this and now it just contains the new work.

core/src/modules/configLoaders/index.ts
```diff
@@ -111,6 +112,12 @@ export async function processConfigObjects(
 
   configObjects = sortConfigurationObjects(configObjects);
 
+  const { errors: validationErrors } = validateConfigObjects(configObjects);
+  validationErrors.map((err) =>
+    log(`[ config ] ${err}`, env === "test" ? "info" : "error")
+  );
+  errors.push(...validationErrors);
+
   for (const i in configObjects) {
```
evantahler, line 123: perhaps `for (const i in configObjects && errors.length === 0) {` to not continue if there were duplicates?  Otherwise, you might overwrite something you didn't intend.

evantahler (changes requested):
I think it would be safer to not to continue `processConfigObjects` if duplicates were found

core/src/modules/configLoaders/index.ts
```diff
@@ -118,6 +118,8 @@ export async function processConfigObjects(
   );
   errors.push(...validationErrors);
 
+  if (errors.length > 0) return { seenIds, errors };
```
seancdavis, line 121: @evantahler I put this here, but it seems like there may be a better way to go about all this.

With this in place, let's say I have a duplicate `id`, but I also have other issues. When I `validate` the first time, I see:

```
🦘 Grouparoo: validate

info: validating 3 objects...
error: [ config ] Duplicate ID values found: data_warehouse
error: ❌ Validation failed - 1 validation error
```

So I fix my `data_warehouse` duplicate issue and run `validate` again:

```
🦘 Grouparoo: validate

info: validating 3 objects...
notice: [ app ] testing app data_warehouse (data_warehouse) threw error: role "..." does not exist
error: [ config ] error with app `data_warehouse` (data_warehouse): Error: error testing app data_warehouse (data_warehouse) - role "..." does not exist
error: [ config ] error with property `data_warehouse` (data_warehouse_prop): Error: cannot find Source with id "..."
error: [ config ] error with source `gsheet` (gsheet): Error: cannot find App with id "..."
error: ❌ Validation failed - 3 validation errors
```

Maybe that's okay. Or unavoidable. What do you think?

evantahler: I think it's OK for now at least!

My main concern was that you, perhaps, made 2 sources called "users-table" (from different apps).  If they were both valid (had all the info they needed) you would replace one of them with the other.  I think that you need to eject quickly if you see that, before any of the `grouparoo apply` logic could be happening

evantahler (approved):

PR #1307, opened 2021-02-15
[Code Config] Add Schedule objects to Sources on generate

seancdavis:
Merges schedule templates with source templates. Now the output of searching for a "schedule" template yields a great big nothing:

```
$ grouparoo generate schedule --list

🦘 Grouparoo: generate schedule

Available Templates: matching "schedule"


You can add plugins to to this project to connect to new Sources and Destinations and add additional commands with the `grouparoo install` command.
```

I have all the schedules currently commented out. Wasn't sure if we'd prefer for them to be enabled by default or just a placeholder that the dev has to manually enable.


core/src/classes/configTemplate.ts
```diff
@@ -49,26 +52,28 @@ export abstract class ConfigTemplate {
   }): Promise<ConfigTemplateRunResponse>;
 
   prepareParams(params: ConfigTemplateParams) {
+    this.params = params;
```
evantahler, line 55: Nice.   I'll bet this can simplify some of the usage in the `bin/generate` command too

plugins/@grouparoo/app-templates/public/templates/table-source/{{{id}}}.js.template
```diff
@@ -42,5 +47,26 @@ exports.default = async function buildConfig() {
...
+    //   id: {{{schedule_id}}},
+    //   name: {{{schedule_id}}},
+    //   class: "schedule",
+    //   sourceId: {{{id}}}, // The ID of the Source above
+    //   recurring: true, // should this Schedule regularly run?
+    //   recurringFrequency: 1000 * 60 * 15, // 15 minutes, in ms
+    //   options: {
+    //     column: "updated_at", // the column to check for new records in table which this Schedule's Source is using
+    //   },
+    // },
```
evantahler, line 70: I think commenting these out for now is the right move.  We'll do something interesting in https://github.com/grouparoo/grouparoo/pull/1313

evantahler (approved):

core/src/classes/configTemplate.ts
seancdavis: 🎉  I was thinking it could come in handy down the road without affecting any of the existing code today!

PR #1310, opened 2021-02-17
handle boolean cases

bleonard:
Saw a column stored as "True"

### DIFF
--- a/core/src/modules/ops/profileProperty.ts
+++ b/core/src/modules/ops/profileProperty.ts
@@ -32,14 +32,14 @@ export namespace ProfilePropertyOps {
       case "url":
         return formatURL(value.toString());
       case "boolean":
-        if (![true, false, 0, 1, "true", "false"].includes(value)) {
-          throw new Error(`${value} is not a valid boolean value`);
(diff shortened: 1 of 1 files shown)


core/src/modules/ops/profileProperty.ts
```diff
@@ -32,14 +32,14 @@ export namespace ProfilePropertyOps {
       case "url":
         return formatURL(value.toString());
       case "boolean":
-        if (![true, false, 0, 1, "true", "false"].includes(value)) {
-          throw new Error(`${value} is not a valid boolean value`);
```
evantahler, line 36: That `"True"` is really odd - wouldn't this line have thrown?  And down below in `getValue` it would have been coerced to `false`...

Either way... good change!

evantahler (approved):

core/src/modules/ops/profileProperty.ts
bleonard: yes, it would have thrown, so I stopped them and had them lowercase it with custom SQL. But then made this so it's not an issue.

PR #1313, opened 2021-02-17
Complex Source + Properties generator

evantahler:
This PR makes `grouparoo generate *:table:source` able to generate a Source, Schedule, and N Properties all at once... and sort out bootstrapping.

## 🦖 Boring Way:

```bash
grouparoo generate postgres:table:source users_table \
  --parent data_warehouse

# Then, do this a bunch of times, filling stuff in manually
grouparoo generate postgres:table:property email \
  --parent users_table
```

Produces:

```js
// config/sources/users_table.js
  return [
    /**
     * --------------------
     * Source
     * --------------------
     */
    {
      class: "source",
      id: "users_table",
      name: "users_table",
      type: "postgres-table-import",
      appId: "data_warehouse", // Set this value to the ID of the App this Source uses - e.g. `appId: "data_warehouse"`
      options: {
        table: "...", // Name of the table in your DB - e.g. `table: "users"`
      },

      /**
       * A list of mappings for this source as:
       *
       *     "remote_column": "grouparoo_property_id"
       *
       * For example, if your users table has an `id` column, and you want to map that to the
       * `user_id` property in this Grouparoo source, your mapping would look like:
       *
       *     mapping: { id: "user_id" }
       *
       * If this is the first Source for this App, you'll want to create a property in the
       * bootstrappedProperty section below. Otherwise, you can create properties with the Grouparoo
       * CLI.
       */
      mapping: {
        "...": "user_id",
      },

      /**
       * If this is the first Source for this App, you must bootstrap it with an initial property.
       * To do so, uncomment the lines below and fill in the appropriate values. Otherwise, you can
       * create properties with the Grouparoo CLI.
       */
      // bootstrappedProperty: {
      //   id: "user_id",
      //   name: "User Id",
      //   type: "integer",
      //   options: {
      //     column: "id",
      //   },
      // },
    },

    /**
     * ---------------
     * Schedule
     * ---------------
     *
     * A Schedule tells Grouparoo how frequently to check the Source for updated
     * data and import it into the application database. If you would like to
     * set a Schedule for this Source, uncomment the following section.
     */
    // {
    //   id: "users_table_schedule",
    //   name: "users_table_schedule",
    //   class: "schedule",
    //   sourceId: "users_table", // The ID of the Source above
    //   recurring: true, // should this Schedule regularly run?
    //   recurringFrequency: 1000 * 60 * 15, // 15 minutes, in ms
    //   options: {
    //     column: "...", // the column to check for new records in table which this Schedule's Source is using (e.g. column: 'updated_at')
    //   },
    // },
  ];
};
```

## 🎩 Fancy Way:

```bash
# first, make an app
grouparoo generate postgres:app data_warehouse
# (fill out the app config)
gropuaroo apply

# then
grouparoo generate postgres:table:source users_table \
  --parent data_warehouse \
  --from users \ # The table to use
  --with id,first_name,email,last_name \  # The columns to make into properties (default *)
  --mapping 'id=user_id' \ # How the mapping should shake out + Bootstrapping if the column is in the `--with` list
  --high-water-mark updated_at
```

Produces:

```js
// config/sources/users_table.js
  return [
    /**
     * --------------------
     * Source
     * --------------------
     */
    {
      class: "source",
      id: "users_table",
      name: "users_table",
      type: "postgres-table-import",
      appId: "data_warehouse", // Set this value to the ID of the App this Source uses - e.g. `appId: "data_warehouse"`
      options: {
        table: "users", // Name of the table in your DB - e.g. `table: "users"`
      },

      /**
       * A list of mappings for this source as:
       *
       *     "remote_column": "grouparoo_property_id"
       *
       * For example, if your users table has an `id` column, and you want to map that to the
       * `user_id` property in this Grouparoo source, your mapping would look like:
       *
       *     mapping: { id: "user_id" }
       *
       * If this is the first Source for this App, you'll want to create a property in the
       * bootstrappedProperty section below. Otherwise, you can create properties with the Grouparoo
       * CLI.
       */
      mapping: {
        id: "user_id",
      },

      /**
       * If this is the first Source for this App, you must bootstrap it with an initial property.
       * To do so, uncomment the lines below and fill in the appropriate values. Otherwise, you can
       * create properties with the Grouparoo CLI.
       */
      bootstrappedProperty: {
        id: "user_id",
        name: "user_id",
        type: "integer",
        options: {
          column: "id",
        },
      },
    },

    /**
     * ---------------
     * Schedule
     * ---------------
     *
     * A Schedule tells Grouparoo how frequently to check the Source for updated
     * data and import it into the application database. If you would like to
     * set a Schedule for this Source, uncomment the following section.
     */
    {
      id: "users_table_schedule",
      name: "users_table_schedule",
      class: "schedule",
      sourceId: "users_table", // The ID of the Source above
      recurring: true, // should this Schedule regularly run?
      recurringFrequency: 1000 * 60 * 15, // 15 minutes, in ms
      options: {
        column: "updated_at", // the column to check for new records in table which this Schedule's Source is using (e.g. column: 'updated_at')
      },
    },
  ];
};
```

```js
// config/properties/email.js
  return [
    {
      id: "email",
      name: "email",
      class: "property",
      sourceId: "users_table", // The ID of the Source that this Property belongs to - e.g. `sourceId: "users_table"`
      type: "email", // The type of the Property.  Options are: "boolean", "date", "email", "float", "integer", "phoneNumber", "string", "url"
      unique: true, // Will Profiles have unique records for this Property?
      identifying: true, // Should we consider this property Identifying in the UI? Only one Property can be identifying.
      isArray: false, // Is this an Array Property?
      options: {
        column: "email", // The column to use for this Property - e.g. `column: "first_name"`
        aggregationMethod: "exact", // The aggregation method.  Options are: "exact", "average", "count", "sum", "min", "max", "most recent value", "least recent value"
        sort: null, // You can sort the results by another column in this table
      },

      // You can optionally filter the results of this Property.
      // For example, if you were collecting SUM(purchases.value) and wanted to exclude those refunded purchases, you could:
      //   filters: [{ key: "state", op: "equals", match: "successful" }],
      // Options for `op` are: equals, does not equal, greater than, greater than or equal to, less than, less than or equal to, contains, does not contain, in
      filters: [],
    },
  ];
};
```

```js
// config/properties/first_name.js
  return [
    {
      id: "first_name",
      name: "first_name",
      class: "property",
      sourceId: "users_table", // The ID of the Source that this Property belongs to - e.g. `sourceId: "users_table"`
      type: "string", // The type of the Property.  Options are: "boolean", "date", "email", "float", "integer", "phoneNumber", "string", "url"
      unique: false, // Will Profiles have unique records for this Property?
      identifying: true, // Should we consider this property Identifying in the UI? Only one Property can be identifying.
      isArray: false, // Is this an Array Property?
      options: {
        column: "first_name", // The column to use for this Property - e.g. `column: "first_name"`
        aggregationMethod: "exact", // The aggregation method.  Options are: "exact", "average", "count", "sum", "min", "max", "most recent value", "least recent value"
        sort: null, // You can sort the results by another column in this table
      },

      // You can optionally filter the results of this Property.
      // For example, if you were collecting SUM(purchases.value) and wanted to exclude those refunded purchases, you could:
      //   filters: [{ key: "state", op: "equals", match: "successful" }],
      // Options for `op` are: equals, does not equal, greater than, greater than or equal to, less than, less than or equal to, contains, does not contain, in
      filters: [],
    },
  ];
};
```

```js
// config/properties/last_name.js
  return [
    {
      id: "last_name",
      name: "last_name",
      class: "property",
      sourceId: "users_table", // The ID of the Source that this Property belongs to - e.g. `sourceId: "users_table"`
      type: "string", // The type of the Property.  Options are: "boolean", "date", "email", "float", "integer", "phoneNumber", "string", "url"
      unique: false, // Will Profiles have unique records for this Property?
      identifying: true, // Should we consider this property Identifying in the UI? Only one Property can be identifying.
      isArray: false, // Is this an Array Property?
      options: {
        column: "last_name", // The column to use for this Property - e.g. `column: "first_name"`
        aggregationMethod: "exact", // The aggregation method.  Options are: "exact", "average", "count", "sum", "min", "max", "most recent value", "least recent value"
        sort: null, // You can sort the results by another column in this table
      },

      // You can optionally filter the results of this Property.
      // For example, if you were collecting SUM(purchases.value) and wanted to exclude those refunded purchases, you could:
      //   filters: [{ key: "state", op: "equals", match: "successful" }],
      // Options for `op` are: equals, does not equal, greater than, greater than or equal to, less than, less than or equal to, contains, does not contain, in
      filters: [],
    },
  ];
};

```

---

- [x] Wait until https://github.com/grouparoo/grouparoo/pull/1307 is merged in to work on `--high-water-mark` and schedule bootstrapping
- [x] Right now everything is generating into one file... this is probably not right.
- [x] MySQL
- [x] BigQuery
- [x] Redshift
- [x] Snowflake
- [x] Postgres
- [x] Hack: column names that are likely to be unique
- [x] Hack: column names that are likely to be email properties (vs string)
- [x] Testing


plugins/@grouparoo/app-templates/public/templates/table-source/{{{id}}}.js.template
```diff
@@ -25,14 +25,25 @@ exports.default = async function buildConfig() {
        * CLI.
        */
       mapping: {
-        id: "user_id",
+        {{#__mappingKey}}{{{__mappingKey}}}: {{{__mappingValue}}}{{/__mappingKey}}{{^__mappingKey}}"...": "user_id"{{/__mappingKey}},
```
evantahler, line 33: You can do IFs in Mustache

plugins/@grouparoo/app-templates/src/source/table/templates.ts
```diff
@@ -22,11 +46,48 @@ export class TableSourceTemplate extends ConfigTemplate {
...
+        columnsMap = columnsMap.filter((o) => o.column !== parts[0]);
+      }
+    }
+
+    if (Object.keys(columnsMap).length > 0) {
+      params.columnsMap = columnsMap;
+      // mustache helpers to pluck a property from the object
+      params.__column = function () {
+        return JSON.stringify(this.column);
+      };
```
evantahler: Mustache can also support sync functions for lookup

plugins/@grouparoo/app-templates/src/source/table/templates.ts
```diff
@@ -72,3 +133,73 @@ export class TablePropertyTemplate extends ConfigTemplate {
...
+// TODO: Hack!
+const uniqueColumnNames = [
+  "email",
+  "email_address",
+  "id",
+  "user_id",
+  "userId",
+  "uuid",
+  "guid",
+];
```
evantahler: @bleonard any advice here?

plugins/@grouparoo/postgres/src/initializers/plugin.ts
```diff
@@ -14,6 +14,9 @@ import { getConnection as getQueryConnection } from "../lib/query-import/connect
 import { destinationOptions } from "../lib/export/destinationOptions";
 import { destinationMappingOptions } from "../lib/export/destinationMappingOptions";
 
+import { getTables } from "../lib/table-import/getTables";
+import { getColumns } from "../lib/table-import/getColumns";
```
evantahler, line 18: There are *super* helpful

plugins/@grouparoo/app-templates/public/templates/table-source/{{{id}}}.js.template
```diff
@@ -41,6 +52,26 @@ exports.default = async function buildConfig() {
       //     column: "id",
       //   },
       // },
+      {{/__bootstrappedId}}
     },
+
+    {{#columnsMap}}
```
evantahler: You can do loops in mustache too

plugins/@grouparoo/app-templates/src/source/table/tableSpeculation.ts
```diff
@@ -0,0 +1,40 @@
...
+      return databaseType;
+    }
+
+    if (columnName.includes("email")) {
+      return "email";
+    }
+
+    return "string";
+  }
+}
```
evantahler: @bleonard I created this helper to try to guess some additional information about the column defaults.  I'm not really sure where this belongs...

evantahler:
Documentation updates in https://github.com/grouparoo/www.grouparoo.com/pull/240

evantahler:
Tests for BiqQuery work OK, but there's something odd with Snowflake.  It doesn't appear that nock is really mockig/blocking all of the requests in the test environment.  https://github.com/grouparoo/grouparoo/pull/1313/commits/29f87e3eefb0dbeecb7fe29535b10355bfbe80b9

Marking this test as TODO for now and moving on.

PR #1314, opened 2021-02-17
Remove ability to disable SCHEDULER

seancdavis:
In digging through this, it seems like we're using `SCHEDULER` as `false` when running for the web, and `true` when using for web workers. This makes me wonder if this is actually a good/safe move? (I left inline comments.)

---

Other PRs to merge **after** this one:

- grouparoo/www.grouparoo.com#226 (doc references)
- grouparoo/app-example#108 (example app)

Though we may want to hold on merging those two until 0.2.2 is released since the absence of `SCHEDULER` will cause problems with previous versions.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
-web:    cd apps/$GROUPAROO_MONOREPO_APP && WEB_SERVER=true  SCHEDULER=false WORKERS=0 npm start
-worker: cd apps/$GROUPAROO_MONOREPO_APP && WEB_SERVER=false SCHEDULER=true  WORKERS=5 npm start
+web:    cd apps/$GROUPAROO_MONOREPO_APP && WEB_SERVER=true WORKERS=0 npm start
+worker: cd apps/$GROUPAROO_MONOREPO_APP && WEB_SERVER=false WORKERS=5 npm start
--- a/core/src/config/tasks.ts
+++ b/core/src/config/tasks.ts
@@ -5,7 +5,7 @@ export const DEFAULT = {
 
       // Run node as a scheduler to promote delayed tasks if there is at least
       // one worker.
-      scheduler: (parseInt(process.env.WORKERS) || 0) > 0,
+      scheduler: (process.env.WORKERS ? parseInt(process.env.WORKERS) : 0) > 0,
       // what queues should the taskProcessors work?
       queues: async () => {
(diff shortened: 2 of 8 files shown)


Procfile
```diff
@@ -1,2 +1,2 @@
-web:    cd apps/$GROUPAROO_MONOREPO_APP && WEB_SERVER=true  SCHEDULER=false WORKERS=0 npm start
```
seancdavis, line 1: This is the one that threw me. Now I'm digging through the website and see we have this pattern in our other deployment docs.

core/src/config/tasks.ts
```diff
@@ -4,7 +4,7 @@ export const DEFAULT = {
       _toExpand: false,
 
       // Should this node run a scheduler to promote delayed tasks?
-      scheduler: process.env.SCHEDULER === "true",
+      scheduler: true,
```
seancdavis: Maybe a better approach would be to have it default to `true`, but provide the option to set it to false? And then we could still leave it out of the basic examples and only include it where we want to disable it?

seancdavis: Or could we use `WEB_SERVER` to determine this value?

evantahler: I'd suggest using `process.env.WORKERS` (ie https://github.com/grouparoo/grouparoo/blob/master/core/src/config/tasks.ts#L51-L53 ) to set it.  If you have > 0 workers, turn on the scheduler too.

You generally don't want the web servers working on background tasks.

evantahler (changes requested):
I think there's something in the CLI project that generates the` .env` file which should be updated too

evantahler:
When this is merged in, we should also update https://github.com/grouparoo/app-example.  It's got .env files, docker things, and a Procfile

evantahler:
👍 from me!  FYI I changed the check for the scheduler env in the config file via https://github.com/grouparoo/grouparoo/pull/1314/commits/a0bdbb2c27d17d5d38184d0fff7280e92a63e9f7

It looks like `parseInt()` on `undefined` (if the env variable wasn't set) would be `NaN`, which is falsely would then move over to 0 `(NaN||0)`, which totally works.  I made the change mostly to match the way it was done elsewhere.

evantahler (approved):

PR #1319, opened 2021-02-18
Prevent the application from starting if there are code-config errors

evantahler:
```
$ grouparoo start

notice: Modified your runtime environment with /Users/evan/workspace/grouparoo/grouparoo/apps/staging-enterprise/.env

🦘 Grouparoo: start

notice: Starting Grouparoo 0.2.2 on node.js v14.12.0
notice: pid: 95535
notice: environment: development
info: *** Starting Actionhero ***
notice: [ app ] testing app data_warehouse (data_warehouse) threw error: getaddrinfo ENOTFOUND localhostX
error: [ config ] error with app `data_warehouse` (data_warehouse): Error: error testing app data_warehouse (data_warehouse) - getaddrinfo ENOTFOUND localhostX
emerg: Exception occurred in initializer `codeConfig` during start
emerg: Error with initializer step: "start"
notice: stopping process...
notice: Stopping server: web
notice: Stopping server: websocket
notice: *** Actionhero Stopped ***
```

### DIFF
--- a/core/src/initializers/codeConfig.ts
+++ b/core/src/initializers/codeConfig.ts
@@ -26,7 +26,10 @@ export class CodeConfig extends Initializer {
 
   async start() {
     const configDir = getConfigDir();
-    await CLS.wrap(async () => loadConfigDirectory(configDir));
+    await CLS.wrap(async () => {
+      const { errors } = await loadConfigDirectory(configDir);
+      if (errors.length > 0) throw new Error("code config error");
+    });
 
     // after this point in the Actionhero boot lifecycle, locked models cannot be changed
     api.codeConfig.allowLockedModelChanges = false;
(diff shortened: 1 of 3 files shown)


core/src/initializers/codeConfig.ts
```diff
@@ -26,7 +26,10 @@ export class CodeConfig extends Initializer {
 
   async start() {
     const configDir = getConfigDir();
-    await CLS.wrap(async () => loadConfigDirectory(configDir));
+    await CLS.wrap(async () => {
+      const { errors } = await loadConfigDirectory(configDir);
+      if (errors.length > 0) throw new Error("code config error");
```
bleonard, line 31: Can you throw the errors themselves so they can see what the problem is?

evantahler: They will already be logged!

evantahler: Example output added to description

PR #1351, opened 2021-02-23
`pacakge-fixer` - lock down package.json dependencies

evantahler:
After `lerna` bumps a new version, it relaxes the dependency versions of `@grouparoo/core` and `@grouparoo/app-example`, ie: `"@grouparoo/app-templates": "^0.2.3-alpha.2"`. This is bad because we may make internally breaking changes between versions, adding and removing methods, etc.

We need to re-write our `package.json` files after each version update to use only the exact current version


evantahler:
Here's the tool working in a github action https://github.com/grouparoo/grouparoo/pull/1351/commits/9d553c6ae6171df0c2ea5ebff008b8ad1834c621

PR #1358, opened 2021-02-24
[hubspot] add new tests and for hubspot.

pauloouriques:
- Add tests for destination-mapping;
- Add tests for export-profile;
- Add improvements on the change email mechanism.


plugins/@grouparoo/hubspot/src/lib/export/exportProfile.ts
```diff
@@ -21,24 +21,30 @@ export const exportProfile: ExportProfilePluginMethod = async ({
...
     let contact;
+    let oldContact;
     try {
-      contact = await client.contacts.getByEmail(email);
+      contact = await client.getContactByEmail(email);
+      if (oldEmail && oldEmail !== email) {
+        oldContact = await client.getContactByEmail(oldEmail);
+      }
     } catch (error) {
       if (!error.toString().match(/Request failed with status code 404/)) {
```
bleonard: is there a way to just check the status code?

plugins/@grouparoo/hubspot/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,503 @@
...
+async function deleteLists(suppressErrors) {
+  try {
+    for (const groupToDelete of [listOne, listTwo, listThree, listFour]) {
+      const listId = await getListId(groupToDelete);
+      if (listId) {
+        await apiClient.deleteList(listId);
+      }
+    }
+  } catch (err) {
+    console.log(err);
```
bleonard: remove consoles

plugins/@grouparoo/hubspot/src/lib/export/exportProfile.ts
bleonard: something like in the list above `if (error?.response?.data?.errorType === "LIST_EXISTS")`

bleonard (approved):

plugins/@grouparoo/hubspot/src/lib/export/exportProfile.ts
pauloouriques: actually, we don't need this anymore because I'm testing this on the client class.

plugins/@grouparoo/hubspot/src/lib/export/exportProfile.ts
```diff
@@ -78,3 +95,14 @@ export const exportProfile: ExportProfilePluginMethod = async ({
     }
   }
 };
+
+function formatVar(value) {
+  if (value === undefined || value === null) {
+    return "";
```
bleonard: sure this shouldn't be null?

pauloouriques: the results are the same with "" null or undefined. All returns "" as result.

bleonard: let's switch it to `null`. generally more correct for things that are integers, booleans, etc.

PR #1371, opened 2021-02-26
Throw better error if plugin not found

evantahler:
Partially related to https://github.com/grouparoo/grouparoo/discussions/1350


bleonard:
❤️

PR #1373, opened 2021-02-27
[mailchimp] Add export tests

pauloouriques:


plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
```diff
@@ -39,14 +39,30 @@ export const exportProfile: ExportProfilePluginMethod = async ({
     const oldMailchimpId = generateMailchimpId(
       oldProfileProperties["email_address"]
     );
-    await client.put(`/lists/${listId}/members/${oldMailchimpId}`, {
-      email_address,
-      status: "subscribed",
-      merge_fields: { email_address },
-    });
+    try {
+      await client.put(`/lists/${listId}/members/${oldMailchimpId}`, {
```
bleonard, line 44: does this remove them?

bleonard: if it succeeds, it would seem like it would add them.

bleonard: let's make sure we have a test where the first time we use a user (new email), there is also the `old` one too. That would simulate them not being there.
My guess is that we want to unsubscribe or delete them just like we do in the shared code, but ignore any issues.
that code should probably also catch 

```
    if (noDelete) {
      return { success: true };
    }
    const response = await client.delete(
      `/lists/${listId}/members/${mailchimpId}`
    );
   // TODO: probably catch some errors here. it's ok if they aren't there.
    return { success: response.statusCode === 200 };
```

bleonard: so another test would be to delete someone that wasn't there. that should be fine.

bleonard (approved):
Looks good!

PR #1391, opened 2021-03-01
Pardot destination

### LINKED ISSUE #1154: Feature Request: Pardot Destination
Is your feature request related to a problem? Please describe.
Add [Pardot](pardot.com) as a destination for Grouparoo profiles and groups.

pedroslopez:
Implements the Pardot destination 

closes #1154


plugins/@grouparoo/pardot/__tests__/.env.example
```diff
@@ -0,0 +1,5 @@
+# Note: These are only needed when recording new nock fixtures
+SALESFORCE_USERNAME = "demo@grouparoo.com"
+SALESFORCE_PASSWORD = "jhfd!kjdgd43h"
+SALESFORCE_SECURITY_TOKEN = "986273e87a3d9725fe3755c"
+PARDOT_BUSINESS_UNIT_ID = "0Uvabcd3fgh1jklmn0"
```
bleonard: it looks like there are some custom fields added. put that list here to setup env correctly. See other plugins like zendesk.

plugins/@grouparoo/pardot/public/templates/app/{{{id}}}.js.template
```diff
@@ -0,0 +1,14 @@
+exports.default = async function buildConfig() {
+  return [
+    {
+      class: "app",
+      id: {{{id}}},
+      name: {{{id}}},
+      type: "{{{__pluginName}}}",
+      options: {
+        siteId: "abc123", // Customer.io Site ID
```
bleonard: these should be the pardot options.

plugins/@grouparoo/pardot/src/lib/client.ts
```diff
@@ -0,0 +1,138 @@
...
+      "/prospect/version/4/do/query?format=json"
+    );
+    return res.data.result;
+  }
+
+  async getCustomFields(): Promise<PardotCustomField[]> {
+    // TODO: This is paginated, with a limit up to 200.
+    // &output=mobile apparently is NOT paginated, so we could try using that
+    const res = await this.request.get(
+      "/customField/version/4/do/query?format=json"
```
bleonard: should we add `page_size=200` or something there or is that the default?

plugins/@grouparoo/pardot/src/lib/connect.ts
```diff
@@ -0,0 +1,19 @@
+import { SimpleAppOptions } from "@grouparoo/core";
+import PardotClient from "./client";
+
+export async function connect(appOptions: SimpleAppOptions) {
+  const client = new PardotClient({
+    businessUnitId: appOptions.businessUnitId,
+    // you can change loginUrl and pardotUrl to connect to sandbox or prerelease env.
+    // loginUrl: "https://test.salesforce.com"
+    pardotUrl: "https://pi.demo.pardot.com",
```
bleonard: I wonder if we need a checkbox or something in the options for "demo" or "sandbox environment" or something. Some people might be using that when testing things out.

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,118 @@
...
+  { key: "state", type: "string" },
+  { key: "territory", type: "string" },
+  { key: "zip", type: "string" },
+  { key: "phone", type: "phoneNumber" },
+  { key: "fax", type: "phoneNumber" },
+  { key: "source", type: "string" },
+  { key: "annual_revenue", type: "string" },
+  { key: "employees", type: "string" },
+  { key: "industry", type: "string" },
+  { key: "years_in_business", type: "string" },
```
bleonard, line 37: sure about this one. looks like integer?

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,118 @@
...
+const requiredProspectFields: DestinationMappingOptionsResponseProperty[] = [
+  { key: "email", type: "email" },
+];
+
+const defaultProspectFields: DestinationMappingOptionsResponseProperty[] = [
+  { key: "campaign_id", type: "integer" }, // TODO check for validation
+  { key: "salutation", type: "string" },
+  { key: "first_name", type: "string", important: true },
+  { key: "last_name", type: "string", important: true },
+  { key: "password", type: "string" },
```
bleonard, line 19: is this for real?

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,118 @@
...
+import PardotClient from "../client";
+import { connect } from "./../connect";
+
+// Prospect field spec https://developer.pardot.com/kb/object-field-references/#prospect
+const requiredProspectFields: DestinationMappingOptionsResponseProperty[] = [
+  { key: "email", type: "email" },
+];
+
+const defaultProspectFields: DestinationMappingOptionsResponseProperty[] = [
+  { key: "campaign_id", type: "integer" }, // TODO check for validation
```
bleonard: this is probably internal and we don't set it, so can leave it out.

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,118 @@
...
+  { key: "salutation", type: "string" },
+  { key: "first_name", type: "string", important: true },
+  { key: "last_name", type: "string", important: true },
+  { key: "password", type: "string" },
+  { key: "company", type: "string" },
+  { key: "prospect_account_id", type: "integer" }, // TODO: check for validation
```
bleonard: this is probably internal and we don't set it.

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,118 @@
...
+  { key: "phone", type: "phoneNumber" },
+  { key: "fax", type: "phoneNumber" },
+  { key: "source", type: "string" },
+  { key: "annual_revenue", type: "string" },
+  { key: "employees", type: "string" },
+  { key: "industry", type: "string" },
+  { key: "years_in_business", type: "string" },
+  { key: "comments", type: "string" },
+  { key: "notes", type: "string" },
+  { key: "score", type: "integer" },
```
bleonard, line 40: make sure not `float`

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,118 @@
...
+    Date: "date",
+    Textarea: "string",
+    Dropdown: "any",
+    Hidden: "any",
+    "Radio Button": "any",
+    "CRM User": "string",
+
+    // TODO:
+    "Multi-Select": null,
+    Checkbox: null,
```
bleonard, line 85: boolean?

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,118 @@
...
+): DestinationMappingOptionsResponseTypes => {
+  // https://help.salesforce.com/articleView?id=sf.pardot_field_types.htm&type=5
+
+  const map = {
+    Text: "string",
+    Number: "number",
+    Date: "date",
+    Textarea: "string",
+    Dropdown: "any",
+    Hidden: "any",
```
bleonard: I'm not sure about these `any` values. they get us in trouble. is the actual type defined anywhere else? when you made it? Will `"string"` work?

plugins/@grouparoo/pardot/src/lib/client.ts
```diff
@@ -0,0 +1,138 @@
...
+  request?: AxiosInstance;
+
+  constructor(options?: PardotClientOptions) {
+    this.loginUrl = options?.loginUrl || "https://login.salesforce.com";
+    this.pardotUrl = options?.pardotUrl || "https://pi.pardot.com";
+    this.businessUnitId = options.businessUnitId;
+  }
+
+  // Salesforce SOAP Login
+  async login(username: string, password: string, securityToken?: string) {
```
bleonard, line 35: I'm not against using `jsforce` if it makes things easier.

bleonard (commented):
Looking great.

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
pedroslopez: Yeah there are some weird types there that don't make sense to me. They're documented here https://developer.pardot.com/kb/object-field-references/#prospect

image

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
pedroslopez: image
Yup!

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
pedroslopez: image
Documented as integer

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
pedroslopez: Thought so too initially, but it's actually a set of checkboxes that you can tick, not a single one.

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
pedroslopez: Yeah, `string` would work. The reason why I did `any` is because you can set it to integers or strings and it happily accepts them. I _would_ be more comfortable just setting it to a string though...

plugins/@grouparoo/pardot/src/lib/client.ts
pedroslopez: I don't have a strong opinion here. My reasoning here was basically since we're not using `jsforce` for anything else and this is what it does internally then we don't _need_ the dependency. Dealing with SOAP is kind of different than everything else though.

If you'd like me to replace this with `jsforce`, I'm not opposed to it.

plugins/@grouparoo/pardot/src/lib/connect.ts
pedroslopez: Thought about this too, but just went with what was being done in salesforce. It's interesting here because there can be multiple configs where you may use any combination of test.salesforce and pi.demo.pardot

See table in https://developer.pardot.com/kb/authentication/

image

If there's a way to do checkboxes in the config, we could do a checkbox for each (demo pardot, test salesforce) or just allow changing the url itself.

plugins/@grouparoo/pardot/__tests__/.env.example
pedroslopez: I was debating whether I should just document this like I saw other plugins were doing it or create/delete the test fields when running the test since there's an endpoint for that. The latter may be more complex than is necessary though...

plugins/@grouparoo/pardot/src/lib/client.ts
pedroslopez: 200 is the maximum and default limit, yeah

plugins/@grouparoo/pardot/__tests__/.env.example
bleonard: the comments are fine

plugins/@grouparoo/pardot/src/lib/client.ts
bleonard: I figure it will likely to be updated more than we would know to do here. Use the same version as the salesforce plugin so they stay together.

plugins/@grouparoo/pardot/src/lib/connect.ts
bleonard: gross.
i guess let's do 2 optional dropdowns and default to "Production Account" if not filled out in each case.

salesforce_domain: [login.salesforce.com, test.salesforce.com] (default to login)
pardot_domain: [pi.pardot.com, pi.demo.pardot.com] (default to non-demo)

plugins/@grouparoo/pardot/src/lib/export/destinationMappingOptions.ts
bleonard: string it is! grouparoo will allow integers to map to a string if they want to map other types.

plugins/@grouparoo/pardot/__tests__/utils/nockHelper.ts
```diff
@@ -0,0 +1,58 @@
...
+      realEnv.SALESFORCE_USERNAME,
+      nockEnv.SALESFORCE_USERNAME
+    );
+
+    nockCall = nockCall.replaceAll(
+      realEnv.SALESFORCE_PASSWORD,
+      nockEnv.SALESFORCE_PASSWORD
+    );
+
+    nockCall = nockCall.replace(
```
bleonard: Should these all just be `replaceAll` ?

bleonard: maybe update in the other plugins in another PR later.

plugins/@grouparoo/pardot/public/templates/destination/{{{id}}}.js.template
```diff
@@ -0,0 +1,30 @@
+exports.default = async function buildConfig() {
+  return [
+    {
+      id: {{{id}}},
+      name: {{{id}}},
+      class: "destination",
+      type: "{{{__pluginName}}}-export",
+      appId: "customerioApp", // The id of the App this Source uses
```
bleonard: will be for pardot

plugins/@grouparoo/pardot/src/initializers/plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+            },
+            {
+              key: "securityToken",
+              displayName: "Security Token",
+              required: false,
+              description:
+                'To get a new security token, click on "Reset My Security Token" in "personal settings"',
+            },
+            {
+              key: "salesforceDomain",
```
bleonard, line 70: 👍

plugins/@grouparoo/pardot/src/lib/client.ts
```diff
@@ -0,0 +1,215 @@
...
+
+class PardotClient {
+  private pardotUrl: string;
+  private businessUnitId: string;
+
+  conn: jsforce.Connection;
+  request?: AxiosInstance;
+
+  constructor(options?: PardotClientOptions) {
+    this.pardotUrl = options?.pardotUrl || "https://pi.pardot.com";
```
bleonard: 👍

plugins/@grouparoo/pardot/src/lib/client.ts
```diff
@@ -0,0 +1,215 @@
...
+    this.businessUnitId = options.businessUnitId;
```
bleonard, line 28: notably, you are optional in the above with `?` in the signature and url usage.
But then here's is required. Probably say it's required.

plugins/@grouparoo/pardot/src/lib/client.ts
```diff
@@ -0,0 +1,215 @@
...
+            format: "json",
+          },
+        }
+      );
+
+      if (Array.isArray(res.data.prospect)) {
+        return res.data.prospect[0];
+      }
+
+      return res.data.prospect;
```
bleonard, line 92: does this handle "not found" ?

plugins/@grouparoo/pardot/src/lib/client.ts
```diff
@@ -0,0 +1,215 @@
...
+        error.response?.data["@attributes"]?.err_code === 4
+      ) {
+        return null;
+      }
+
+      throw error;
+    }
+  }
+
+  async getProspectById(id: number) {
```
bleonard, line 105: not used?

plugins/@grouparoo/pardot/src/lib/client.ts
```diff
@@ -0,0 +1,215 @@
...
+        error.response?.data["@attributes"]?.err_code === 4
+      ) {
+        return null;
+      }
+
+      throw error;
+    }
+  }
+
+  async queryProspects(
```
bleonard, line 133: this is paginated right? don't want to blow out the memory.

plugins/@grouparoo/pardot/src/lib/client.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  async batchUpsertProspects(prospects: Record<string, any>[]) {
+    const res = await this.request.post(
+      "/prospect/version/4/do/batchUpsert",
+      {},
+      {
+        params: {
+          format: "json",
+          prospects: JSON.stringify({
```
bleonard: does`stringify` add any extra spaces? I don't think so but it might be worth checking because the length here can cause problems.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+};
+
+// delete the given destinationIds
+const deleteByDestinationIds: BatchMethodDeleteByDestinationIds = async ({
+  client,
+  users,
+}) => {
+  for (const user of users) {
+    // No batch delete available, do one by one
+    await client.deleteProspectById(parseInt(user.destinationId));
```
bleonard: it's likely they are there but if they aren't here, maybe we can catch. timing issues and it's fine as long as they are gone.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+// use the getByForeignKey to lookup results
+const findAndSetDestinationIds: BatchMethodFindAndSetDestinationIds = async ({
+  client,
+  foreignKeys,
+  getByForeignKey,
+}) => {
+  for (const email of foreignKeys) {
+    const prospect = await client.getProspectByEmail(email);
+    if (prospect) {
+      const user = getByForeignKey(email);
```
bleonard: 👍

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+  client,
+  users,
+  config,
+}) => {
+  const { cacheData } = config.data;
+
+  await updateProspectsHelper(client, users, cacheData);
+
+  for (let user of users) {
+    // We're not using destinationIds for created prospects, but grouparoo requires it to be set
```
bleonard: `app-templates/batch` requires (not grouparoo)

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+    if (error.response?.status === 414 || error.response?.status === 400) {
+      if (users.length === 1) {
+        users[0].error = new Error(
+          `${error.response.status} - ${error.response.statusText}`
+        );
+        return;
+      }
+
+      const half = Math.ceil(users.length / 2);
+      await updateProspectsHelper(client, users.slice(0, half), cacheData);
```
bleonard: what a world.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+) {
+  const input: any[] = [];
+  for (const exportedProfile of users) {
+    const payload = await buildPayload(client, exportedProfile, cacheData);
+    input.push(payload);
+  }
+
+  const errors = await client.batchUpsertProspects(input);
+  for (let errorIdx in errors) {
+    const resultIdx = parseInt(errorIdx);
```
bleonard, line 110: are there tests with a few users and a specific bad input to check the error gets on the right one?

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+}
+
+async function updateProspects(
+  client: PardotClient,
+  users: BatchExport[],
+  cacheData: PardotCacheData
+) {
+  const input: any[] = [];
+  for (const exportedProfile of users) {
+    const payload = await buildPayload(client, exportedProfile, cacheData);
```
bleonard, line 104: if there is an error (no email!)
this should catch and put the error jsut on that user.

bleonard: test for that too.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+    input.push(payload);
+  }
+
+  const errors = await client.batchUpsertProspects(input);
```
bleonard, line 108: if you are catching about because of an error. there might actually be zero inputs, so check the length here.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+  } = exportedProfile;
+
+  const payload: any = {};
+
+  payload.email = foreignKeyValue; // email
+  if (destinationId) {
+    payload.id = destinationId;
+  }
+
+  if (!payload.email) {
```
bleonard: I think we actually check for this and `throw` in `findAndSetDestinationIds` (because it using the email)

bleonard: actually maybe the helper does that. let's check. either way out form here.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+
+// fetch using the keys to set destinationId and result on BatchExports
+// use the getByForeignKey to lookup results
+const findAndSetDestinationIds: BatchMethodFindAndSetDestinationIds = async ({
+  client,
+  foreignKeys,
+  getByForeignKey,
+}) => {
+  for (const email of foreignKeys) {
+    const prospect = await client.getProspectByEmail(email);
```
bleonard: probably `try` around this and set error so a certain users doesn't mess everyone else up.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,273 @@
...
+      }
+    }
+  }
+
+  return payload;
+}
+
+function formatVar(value) {
+  if (!value) {
+    return ""; // empty string clears the value
```
bleonard, line 172: does null work here?

bleonard (commented):
looks great.

plugins/@grouparoo/pardot/public/templates/destination/{{{id}}}.js.template
pedroslopez: seems like this was reviewed outdated, should be good now

plugins/@grouparoo/pardot/src/lib/client.ts
pedroslopez: not found returns a `400` status with `err_code` set to 4. This is handled in the `catch` and returns `null`.

plugins/@grouparoo/pardot/src/lib/client.ts
pedroslopez: it's only used in tests, should I move it over there?

plugins/@grouparoo/pardot/src/lib/client.ts
pedroslopez: Yeah, but the `TestPluginMethod` was overriding it with `output=mobile`. Changed this so it respects the pagination when testing connection (93bd712)

plugins/@grouparoo/pardot/src/lib/client.ts
pedroslopez: Nope, it doesn't

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
pedroslopez: Yep, "can handle some of them working, but not others"

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
pedroslopez: You're right, the helper does that for the foreignKey. Added this here because it's a required field, in case we were going to support different foreignKeys other than email. Removing it for now.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
pedroslopez: Nope, it just leaves the old value and does nothing. `false` doesn't work either.

plugins/@grouparoo/pardot/src/lib/client.ts
pedroslopez: are you referring to `businessUnitId`? I believe it's always required here...

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
pedroslopez: `buildPayload` is not throwing any errors anymore based on previous comment, so this shouldn't be necessary. I did add a test to check for missing emails though.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
pedroslopez: shouldn't happen after removing the `throw` from `buildPayload`.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
pedroslopez: yeah, weird stuff. added some tests for this ("can handle updates with too long URI" and "can handle updates with even longer URI")

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
pedroslopez: Not entirely sure what you're suggesting to do on error. Wrap in try/catch and set error to the specific user and let grouparoo retry it?

plugins/@grouparoo/pardot/src/lib/client.ts
bleonard: ah. It's fine. that file wasn't expanded i guess.

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
pedroslopez: Update: removed this in favor of sending the payload as form data

plugins/@grouparoo/pardot/src/lib/client.ts
bleonard: Yeah, i was just saying in half this method you are acting like `options` might be null/undefined and in that spot not. So it will crash either way if it is null, so it is required, so it would be better to be more confident the other spots.
`options?.pardotUrl` -> `options.pardotUrl`

plugins/@grouparoo/pardot/src/lib/export/exportProfiles.ts
bleonard:
```
try {
  await client.deleteProspectById(parseInt(user.destinationId));
} catch(error) {
  user.error = error;
}
```

bleonard: Then Grouparoo will let the other ones work and try this one again later.

PR #1396, opened 2021-03-02
Mailchimp: remove tags before deleting a user.

pauloouriques:


bleonard (commented):
Actually, we need to remove only the ones that Grouparoo is handling.
```
const tagsToRemove = Array.from(
    new Set(newGroups.concat(oldGroups))
  );
```

We should probably only do that if the user is existing.
It's possible we should use some of the code below this to make that happen. It's kind of like resetting the groups:
```
oldGroups = Array.from(
    new Set(newGroups.concat(oldGroups))
  );
newGroups = [];
```

And then probably some tests for this, including keeping one the call didn't set.

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
```diff
@@ -49,7 +49,13 @@ export const exportProfile: ExportProfilePluginMethod = async ({
     } catch (error) {
       // Another user with the new email address (email_address) already exists,
       // so we need to delete the old one and let the updateProfile reuse the existing one.
-      await deleteMember(client, listId, oldMailchimpId);
+      await deleteMember(
+        client,
+        listId,
+        oldMailchimpId,
+        profileToExport.oldGroups,
```
bleonard: probably do this above to stick to the patterns.
```
  const {
    toDelete,
    newProfileProperties,
    oldProfileProperties,
    oldGroups,
    newGroups,
  } = profileToExport;
```

plugins/@grouparoo/mailchimp/src/lib/shared/updateProfile.ts
```diff
@@ -143,8 +143,21 @@ export const updateProfile: UpdateProfileMethod = async ({
...
+  listId,
+  mailchimpId,
+  oldGroups,
+  newGroups
+) {
   try {
+    let tagsToRemove = Array.from(new Set(oldGroups.concat(newGroups)));
+    await client.post(`/lists/${listId}/members/${mailchimpId}/tags`, {
+      tags: tagsToRemove.map((g) => ({
+        name: String(g).toLocaleLowerCase(),
```
bleonard: let's make a funciton in this file for the 3 spots called `normalizeGroupName`
several plugins have that.

bleonard (approved):
minor things, but looks good.

evantahler:
@bleonard don't forget to link these PRs to Pivotal!

PR #1418, opened 2021-03-03
Reset Cluster removes all data in Redis

evantahler:
Now that sessions have been moved to the database (https://github.com/grouparoo/grouparoo/pull/1369) and all runs can be recovered (https://github.com/grouparoo/grouparoo/pull/1353), there's no data that needs to persist in Redis.  We can safely erase all Redis data when reseting the cluster. 

This PR means that the `reset cluster` button will also clear up any problems with persisting resque locks.

### DIFF
--- a/core/src/actions/cluster.ts
+++ b/core/src/actions/cluster.ts
@@ -1,29 +1,27 @@
 import { AuthenticatedAction } from "../classes/actions/authenticatedAction";
 import { cache, api, task, config } from "actionhero";
 
-import {
-  App,
-  Destination,
-  DestinationGroupMembership,
-  Export,
-  Event,
-  EventData,
-  Group,
-  GroupMember,
-  GroupRule,
-  Import,
-  Log,
-  Mapping,
-  Option,
-  Profile,
-  ProfileProperty,
-  Property,
-  PropertyFilter,
-  Run,
-  Schedule,
-  SetupStep,
-  Source,
-} from "..";
+import { App } from "../models/App";
+import { Destination } from "../models/Destination";
+import { DestinationGroupMembership } from "../models/DestinationGroupMembership";
+import { Export } from "../models/Export";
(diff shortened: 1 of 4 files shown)


bleonard:
What do you think about (also) a button just for redis? Seems quite possible things have gone sideways there but they want to keep their profiles, etc.

evantahler:
> What do you think about (also) a button just for redis? Seems quite possible things have gone sideways there but they want to keep their profiles, etc.

👍 What I ended up doing was including the clearing of Resque in the existing `clear-cache` action (and updating the label).  I think we want our users to think of resque as part of an ephemeral cache anyway.  The more serious destroy cluster also does this now as well, in addition to truncating your tables.

evantahler:
Running `flushdb` was a little too serious, as it really messed up the websocket system, so that's been rolled back.

core/src/actions/cluster.ts
```diff
@@ -128,11 +104,34 @@ export class ClusterClearCache extends AuthenticatedAction {
...
+
+  // resque
+  await deleteKeys("*resque:queue:*"); // clear resque queues
+  await deleteKeys("*resque:delayed:*"); // clear resque delayed queues
+  await deleteKeys("*resque:timestamps:*"); // clear resque timestamps
+  await deleteKeys("*resque:*lock*:*"); // clear resque locks
+  await deleteKeys("*resque:stat:*"); // clear resque stats
+
+  // wait for any currently-running workers
+  await new Promise((resolve) => setTimeout(resolve, config.tasks.timeout * 2));
```
bleonard, line 123: This and the `keys` call can take a while. might need a UI accommodation of some sort.

evantahler: For me with 1K profiles, it's about 5s (most of the time being the sleep you've highlighted).  The UI buttons  have a loading state at least... 

reset

PR #1419, opened 2021-03-03
177077763/test snapshots

seancdavis:
⚠️ Resolve #1420 prior to merging this work.

Also note that grouparoo/www.grouparoo.com#279 contains some documentation to support the spec helper. We may want to hold on that until the next release with this change.

This does a few things:

- Update code config comments to make it more clear how bootstrapping a Property works.
- Truncate the database in the spec helper earlier (i.e. prior to the code config being loaded).
- Require that at least one bootstrapped property be included when calling `getProfile` from the spec helper.


plugins/@grouparoo/app-templates/public/templates/table-source/{{{id}}}.js.template
```diff
@@ -12,7 +12,8 @@ exports.default = async function buildConfig() {
       type: "{{{__pluginName}}}-table-import",
       appId: {{{appId}}}, // Set this value to the ID of the App this Source uses - e.g. `appId: "data_warehouse"`
       options: {
-        table: {{#from}}{{{from}}}{{/from}}{{^from}}"..."{{/from}}, // Name of the table in your DB - e.g. `table: "users"`
+        table: {{#from}}{{{from}}}
+{{/from}}{{^from}}"..."{{/from}}, // Name of the table in your DB - e.g. `table: "users"`
```
evantahler: The spacing was changed here... what does this look like when mustached?

evantahler: There are some other similar changes below

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -186,6 +186,9 @@ export namespace helper {
     beforeAll(async () => {
       const { Process } = await import("actionhero");
       actionhero = new Process();
+
+      await actionhero.initialize();
+      if (options.truncate) await helper.truncate();
       await actionhero.start();
```
evantahler, line 192: Awesome! The rest of the test suite is OK with this.

plugins/@grouparoo/app-templates/public/templates/table-source/{{{id}}}.js.template
seancdavis: Good catch! This is Prettier or VS Code formatter doing something weird on save.

PR #1420, opened 2021-03-03
Add app-example-config as an example app

seancdavis:
⚠️ Note: This is opened against the branch in #1419. I pulled this out as a separate PR because I think it'd be worthwhile to discuss the value of this being added to the monorepo.

This is essentially a duplicate of [grouparoo/app-example-config](https://github.com/grouparoo/app-example-config). We know there is value in that repo being separate. Is there value in this version also being inside the monorepo?

My intention here was that we'd have an example that was purely about code config and presented a stripped down set of plugins and config objects. It doesn't get deployed anywhere.

Or, should staging-community cover this example?


apps/app-example-config/.env.example
```diff
@@ -0,0 +1,69 @@
...
+S3_REGION=""
+S3_BUCKET=""
+
+############
+## GOOGLE ##
+############
+
+GOOGLE_API_KEY=""
+GOOGLE_CLIENT_ID=""
+
```
evantahler, line 63: We don't need the google stuff any more

apps/app-example-config/package-lock.json
```diff
@@ -0,0 +1,3385 @@
+{
```
evantahler, line 1: Don't check in `package-lock` files.  PNPM handles that for us.

evantahler (commented):
I don't feel too strongly about it, but I don't think this is needed:
* You aren't checking in the code config itself, so the test can't really be run on its own
* The data isn't here to run the test against either.  The dataset you are using is the one from `grouparoo demo-data-purchases` and the `@grouparoo/demo` plugin isn't a dependency here, but is in the other 2 staging apps.  You can create the demo data as needed in either of them.
* You can already add code-config to the staging apps of you want

bleonard:
I wonder how the demo tools and data relate to this and/or the staging-community.

seancdavis:
Not really sure what you means, @bleonard.

@evantahler: Fair points. I'm going to close this.

PR #1424, opened 2021-03-04
Don't allow `--high-water-mark` to be used without a column name

seancdavis:
Took a guess here that this should be done in the app-templates plugin and not the primary generate command. I was a little tripped up that the main generate command even mentions hwm as an input.


evantahler (approved):
Yep! It looks like `commander` makes flags without values into booleans... I guess that makes sense in most cases... but not here!

PR #1425, opened 2021-03-04
Truncate cluster name for names longer than 30 characters

seancdavis:
This appears in the main nav below the logo:

Screen Shot 2021-03-04 at 7 05 35 AM

I've added a truncate utility that will look at a string and append ellipses if it is shorter than the specified length.


evantahler:
The test failures may have something to do with what files jest is ignoring https://github.com/grouparoo/grouparoo/blob/main/ui/ui-components/jest.config.js

seancdavis:
@evantahler Thought I had made the appropriate change for that. I ended up making another change in the nav spec and it is working locally. It looked like the router was only being mocked once and that was causing a problem. It must be getting called a second time because of the `truncate` method I added. 🤷‍♂️ 

I guess we'll see if they pass now.

evantahler (approved):
Assuming the tests pass, 👍

PR #1426, opened 2021-03-04
Fix off-by-one error when limiting batch size

pedroslopez:
When using the batch helper, there was a bug that was returning batch sizes 1 greater than the specified limit. For example, if my batch size was 50, it would return 51 profiles to update. This PR addresses this issue and adds some tests to verify the behavior.


bleonard (commented):
Great!

Looks there is similiar code in app-templates/destionation/group/exportProfiles

PR #1436, opened 2021-03-04
Fix CTA for naming cluster missing in Community

seancdavis:
I did two different things here to address the issue:

1. Added an option for setup steps' CTA to be able to be added to Community edition.
2. Set `cluster-name` setting when initializing a team if it hasn't already been set. (This was an exciting discovery!) 

A couple issues with the approach in the latter change:

- I couldn't find where we are testing these sorts of actions that occur in initializers.
- It doesn't feel _super_ DRY because the same sort of thing is being done in `core/src/actions/team.ts`.

### DIFF
--- a/core/src/modules/configLoaders/team.ts
+++ b/core/src/modules/configLoaders/team.ts
@@ -1,3 +1,4 @@
+import { env } from "actionhero";
 import {
   ConfigurationObject,
   logModel,
@@ -6,6 +7,7 @@ import {
   validateConfigObjectKeys,
   IdsByClass,
 } from "../../classes/codeConfig";
+import { Setting } from "../../models/Setting";
 import { Team, Permission } from "../..";
 import { Op } from "sequelize";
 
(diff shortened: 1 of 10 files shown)


evantahler (commented):
Here's an idea - Perhaps this story should have been to ensure that the cluster name was set, which you did in the configLoader.    Maybe we don't need the CTA in the community UI as we can be sure that if they can log in, the've made a team, so the name will be set?

seancdavis:
@evantahler Okay, I think that's a fine approach. I could remove the need for the extra attribute on setup steps.

But then how do you feel about a lack of test and semi-duplicated code? Is it low-risk and okay here or should we abstract and strengthen it a bit?

core/src/modules/configLoaders/team.ts
```diff
@@ -60,6 +62,23 @@ export async function loadTeam(
...
+    clusterNameSetting &&
+    clusterNameSetting.value === clusterNameSetting.defaultValue
+  ) {
+    const nodeEnv = env || "development";
+    const clusterName = `${team.name} - ${
+      nodeEnv.charAt(0).toUpperCase() + nodeEnv.slice(1)
+    }`; // the cluster name would be `Grouparoo - Development`
+    await clusterNameSetting.update({ value: clusterName });
+  }
+
```
evantahler: I don't feel too badly about the duplication, but I guess the /correct/ thing would be to DRY this and the version in `actions/team.ts` up.  I think the most relevant place to put this would be in a new `modules/ops/settings.ts` file - Ops (short for "operations") are where we've been putting business logic that's related to a model but transforms it and/or other models.  Then, you could test the new `SettingOps.nameClusterFromTeam()` method (or whatever you choose to call it) as well.

seancdavis:
I'm going to close this. It will be quicker for me to pull out the commit into a new branch and open a new PR.

PR #1437, opened 2021-03-04
Pool and Retry Settings for Sequelize

evantahler:
This PR contains a collection of setting tweaks for Sequelize which should help make Grouparoo more resilient.
* Sequelize Docs: https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html#instance-constructor-constructor
* Retry Example: https://dev.to/anonyma/how-to-retry-transactions-in-sequelize-5h5c

For SQLite:
* Use the `EXCLUSIVE` transactions rather than `DEFERRED`.  Since SQLite can only handle one transaction at a time, explicitly set that the connection pool should only attempt on transaction at a time.  This will help remove errors of the type `SQLITE_BUSY`.  This will slow down parallel requests (i.e. in the UI) but should prevent errors.
* Set the min and max pool size to be 1 explicitly
* Retry errors of the type `SQLITE_BUSY`
* Since the above options seem to prevent deadlocks, re-allow more than one 1 SQLite worker **But you really shouldn't do it... it's slow**

For Postgres:
* Retry errors matching `SequelizeConnectionError`, `SequelizeDatabaseError`, `SequelizeConnectionTimedOutError` 3 times.  This will help avoid issues when the database has prematurely closed the socket, started auto-scaling, or similar shenanigans which will upset a persistent network connection.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -543,7 +543,7 @@ jobs:
           <<: *core-cache-options
       - run:
           name: test
-          command: cd core/ && WORKERS=1 ./node_modules/.bin/jest __tests__/models __tests__/actions --ci --maxWorkers 2
+          command: cd core/ && ./node_modules/.bin/jest __tests__/models __tests__/actions --ci --maxWorkers 2
           environment:
             DB_DIALECT: sqlite
   test-ui-components:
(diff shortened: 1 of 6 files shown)


seancdavis (approved):

evantahler:
@seancdavis could you try this branch out locally and let me know is the SQLite errors go away for you?  Please continue to use only 1 WORKER

PR #1456, opened 2021-03-09
Add option to show a setup step on Community Edition

seancdavis:
This is an extension of #1436.

The adjustment to the workflow here is if I create a team using code config, then sign into the UI, I should see that Setup Step 1 is not completed, and there should be a button that takes me to the settings to rename the cluster.


evantahler (approved):
Looks good! I guess this was all that was needed after all... sorry about the confusion!

seancdavis:
Eh, I pushed it in that direction, too. No worries.

evantahler:
I added the `enhancement` flag

PR #1459, opened 2021-03-10
Mongo source

pauloouriques:


plugins/@grouparoo/app-templates/src/source/table/getExamples.ts
```diff
@@ -24,6 +26,7 @@ export interface GetColumnExamplesMethod {
   (argument: {
     connection: any;
     appOptions: SimpleAppOptions;
+    sourceOptions?: SimpleSourceOptions;
```
bleonard, line 29: ideally this wouldn't be optional, but we can do that later.

plugins/@grouparoo/app-templates/src/source/table/meta.ts
```diff
@@ -98,19 +102,19 @@ export const buildConnection: BuildConnectionMethod = ({
     }
   );
 
+  const options = (additionalOptions?.options || []).concat({
+    key: tableNameKey,
+    displayName: "Table",
```
bleonard: maybe we need `tableOptionDisplayName` here so it can be "Collection"

plugins/@grouparoo/mongo/__tests__/table-import/get-changed-rows-count.ts
```diff
@@ -0,0 +1,31 @@
...
+import { getConfig, usersCollectionName } from "../utils/data";
+import { getChangedRowCount } from "../../src/lib/table-import/getChangedRowCount";
+import { connect } from "../../src/lib/connect";
+import { FilterOperation } from "@grouparoo/app-templates/dist/source/table";
+
+// these used and set by test
+const { appId, appOptions } = getConfig();
+
+async function callGetChangedRowsCount() {
+  const connection = await connect({ appOptions, app: null, appId: null });
```
bleonard: need to call `await client.close();`?

bleonard: it should probably be in a `beforeAll` and `afterAll` in all these tests. Something like
`beforeAll`: `connection = await connect({appOptions})`
`afterAll`: `await disconnect({connection})`

plugins/@grouparoo/mongo/__tests__/table-import/get-changed-rows.ts
```diff
@@ -0,0 +1,36 @@
+import "@grouparoo/spec-helper";
+
+import { getConfig, usersCollectionName } from "../utils/data";
+import { getChangedRows } from "../../src/lib/table-import/getChangedRows";
+import { connect } from "../../src/lib/connect";
+import { FilterOperation } from "@grouparoo/app-templates/dist/source/table";
+
+// these used and set by test
+const { appId, appOptions } = getConfig();
```
bleonard: `tableName` here?

plugins/@grouparoo/mongo/__tests__/table-import/get-columns.ts
```diff
@@ -0,0 +1,25 @@
...
+    connection,
+    appOptions,
+    tableName,
+  });
+}
+
+describe("mongodb/column/getColumnsTest", () => {
+  test("gets the list of columns", async () => {
+    const columns = await getMongoColumns();
+    expect(columns.length).toBe(5);
```
bleonard: check what they are and their types.

plugins/@grouparoo/mongo/__tests__/table-import/get-property-value.ts
```diff
@@ -0,0 +1,43 @@
...
+        filterOperation: FilterOperation.Equal,
+        value: "ourixilva@gmail.com",
+      },
+    ],
+    aggregationMethod: AggregationMethod.Exact,
+    isArray: false,
+  });
+}
+
+describe("mongodb/column/getColumnsTest", () => {
```
bleonard: title not the right one.
probably plenty more tests here for all the types, etc.

plugins/@grouparoo/mongo/src/lib/query-import/executeQuery.ts
```diff
@@ -0,0 +1,8 @@
+import { ExecuteQueryMethod } from "@grouparoo/app-templates/dist/source/query";
+
+export const executeQuery: ExecuteQueryMethod = async ({
+  connection,
+  query,
+}) => {
+  return connection.execute(query);
```
bleonard: `connect` returns `{ client, db }` so isn't this `connection.db.query()` or something?
Is `client` actually used or does it already have a point to the `client` ? Just `db` might be better.

plugins/@grouparoo/mongo/src/lib/table-import/connection.ts
```diff
@@ -0,0 +1,35 @@
...
+import { getPropertyValues } from "./getPropertyValues";
+import { getChangedRowCount } from "./getChangedRowCount";
+
+export function getConnection() {
+  return buildConnection({
+    app: "mongodb",
+    name: "mongodb-collection-import",
+    description:
+      "Import or update Profiles from a Mongo DB database collection.",
+    tableOptionDescription: "The table to scan",
```
bleonard, line 16: Probably "The collection to scan"

plugins/@grouparoo/mongo/src/lib/table-import/getColumns.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export const getColumns: GetColumnDefinitionsMethod = async ({
+  connection,
+  tableName,
+  appOptions,
+  sourceOptions,
+}) => {
+  const map: ColumnDefinitionMap = {};
+  if (sourceOptions) {
+    const fields = sourceOptions.fields.toString().split(",");
```
bleonard: let's also `trim`: `sourceOptions.fields.toString().split(",").map(f => f.trim())`

bleonard: And have a test with spaces.

plugins/@grouparoo/mongo/src/lib/table-import/getColumns.ts
```diff
@@ -0,0 +1,139 @@
...
+      break;
+
+    // semi-structured types
+    case "javascript":
+    case "array":
+      break;
+
+    // other?
+    default:
+      break;
```
bleonard, line 99: what if it's not found. How do other plugins handle this?

plugins/@grouparoo/mongo/src/lib/table-import/getSampleRows.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const getSampleRows: GetSampleRowsMethod = async ({
+  connection,
+  appOptions,
+  sourceOptions,
+  tableName,
+}) => {
+  const out: DataResponseRow[] = [];
+  if (sourceOptions) {
+    const fields = sourceOptions.fields.toString().split(",");
```
bleonard: `trim` here too. maybe `DRY` that up.

bleonard: There were spots before when you were replacing the tbale name and such. is that always the case?

plugins/@grouparoo/app-templates/src/source/table/pluginMethods.ts
```diff
@@ -29,6 +34,7 @@ export enum FilterOperation {
   Contain = "contains",
   NotContain = "does not contain",
   In = "in",
+  Exists = "exists",
```
bleonard: I don't think this is used.

plugins/@grouparoo/mongo/src/lib/table-import/getColumns.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  if (compare) {
+    ops.push(FilterOperation.GreaterThan);
+    ops.push(FilterOperation.GreaterThanOrEqual);
+    ops.push(FilterOperation.LessThan);
+    ops.push(FilterOperation.LessThanOrEqual);
+  }
+  if (contains) {
+    // ops.push(MongoFilterOperation.Contain);
```
bleonard: Is this supported by Mongo?

bleonard: If they are, these would be `FilterOperation` not `MongoFilterOperation`

plugins/@grouparoo/mongo/src/lib/table-import/getColumns.ts
bleonard: I notice postgres just says `null` so maybe core handles it. let's check.

plugins/@grouparoo/mongo/__tests__/utils/data.ts
```diff
@@ -0,0 +1,210 @@
...
+    validationLevel: "off",
+    validator: {
+      $jsonSchema: {
+        bsonType: "object",
+        required: ["profile_id"],
+        properties: {
+          profile_id: {
+            bsonType: "int",
+          },
+          purchase: {
```
bleonard, line 112: maybe add a nested object and array in here for testing the dot syntax

evantahler:
To install and run Mongo on OSX (via homebrew)

```
brew tap mongodb/brew
brew install mongodb-community
brew services start mongodb/brew/mongodb-community
```

From https://docs.mongodb.com/manual/tutorial/install-mongodb-on-os-x/

plugins/@grouparoo/mongo/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "scripts": {
+    "prepare": "rm -rf dist && tsc --declaration",
+    "test": "jest",
+    "pretest": "npm run lint && npm run prepare",
+    "lint": "prettier --check src __tests__",
+    "watch": "tsc --watch"
+  },
+  "dependencies": {
+    "@grouparoo/app-templates": "0.2.8",
+    "mongodb": "^3.6.4"
```
evantahler: Please don't use the `^` and choose a specific version

plugins/@grouparoo/mongo/public/templates/app/{{{id}}}.js.template
```diff
@@ -0,0 +1,14 @@
+exports.default = async function buildConfig() {
+  return [
+    {
+      class: "app",
+      id: {{{id}}},
+      name: {{{id}}},
+      type: "{{{__pluginName}}}",
+      options: {
+        uri: "mongodb://localhost:27017/mydb", // The MongoDB Connection String.
+        database: "mydb", // The MongoDB database name (normally, the Connection String have the database name too).
```
evantahler: Does this mean that this is optional because it will be in the uri?  If so, perhaps the default should be `null`?

plugins/@grouparoo/mongo/src/initializers/plugin.ts
```diff
@@ -0,0 +1,75 @@
...
+              description:
+                "The MongoDB Connection String. Normally, this is connection string is provided by the Mongo DB service. " +
+                "In case you're using the standalone Mongo DB, this connection string will looks line this: mongodb://localhost:27017.",
+              placeholder:
+                "e.g. mongodb+srv://user:password@cluster0.123.mongodb.net/mydb?retryWrites=true&w=majority",
+            },
+            {
+              key: "database",
+              displayName: "Database name",
+              required: true,
```
evantahler, line 61: I think I'm still confused if I should be providing database name in the URI or not.  This one says requried.

plugins/@grouparoo/mongo/src/initializers/plugin.ts
```diff
@@ -0,0 +1,75 @@
...
+          options: [
+            {
+              key: "uri",
+              displayName: "Connection String URI",
+              required: true,
+              description:
+                "The MongoDB Connection String. Normally, this is connection string is provided by the Mongo DB service. " +
+                "In case you're using the standalone Mongo DB, this connection string will looks line this: mongodb://localhost:27017.",
+              placeholder:
+                "e.g. mongodb+srv://user:password@cluster0.123.mongodb.net/mydb?retryWrites=true&w=majority",
```
evantahler: This differs from what you have in the template file.  Also, please remove the `e.g.`, because it will show literally this way in the UI

plugins/@grouparoo/mongo/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./dist",
+    "allowJs": true,
+    "module": "commonjs",
+    "target": "es2018",
```
evantahler: This is es2019 now

evantahler (changes requested):
Looking good! A few comments below:

plugins/@grouparoo/mongo/public/templates/app/{{{id}}}.js.template
pauloouriques: It is confusing. I'll remove the comment because it's not helping.

plugins/@grouparoo/mongo/src/initializers/plugin.ts
pauloouriques: Yes, it's required.

evantahler: Got it - so then shouldn't the connection string not include the database?

pauloouriques: It's optional on the connection string.

evantahler:
A good guide on loading CSV data into Mongo https://kb.objectrocket.com/mongo-db/how-to-import-a-csv-into-mongodb-327

evantahler:
Looks good in the UI!

In a future story, we'll likely want to remove the need for a comma-separated list of object properties, it's a little odd. 
Screen Shot 2021-03-24 at 5 18 02 PM
Screen Shot 2021-03-24 at 5 16 42 PM

evantahler (approved):

.circleci/config.yml
```diff
@@ -47,14 +47,14 @@ jobs:
...
             - plugins/@grouparoo/mailchimp/node_modules
-            - plugins/@grouparoo/mysql/node_modules
-            - plugins/@grouparoo/onesignal/node_modules
-            - plugins/@grouparoo/pipedrive/node_modules
-            - plugins/@grouparoo/redshift/node_modules
-            - plugins/@grouparoo/salesforce/node_modules
-            - plugins/@grouparoo/sentry/node_modules
-            - plugins/@grouparoo/spec-helper/node_modules
-            - plugins/@grouparoo/zendesk/node_modules
+            - plugins/@grouparoo/mongo/node_modules
```
bleonard, line 50: @evantahler  there would be less churn on these if, instead of `even` and `odd` we did something like alphabetize and then each has `length / 2`

Or maybe  set a max size for each (15?). and know the number of buckets (3). and then freak out when there are 4 buckets (>45)

bleonard: Another PR of course

evantahler:
Rebasing off of main and bumping version numbers...

PR #1462, opened 2021-03-10
OneSignal Destination

### LINKED ISSUE #1452: Feature Request: OneSignal Destination
https://onesignal.com/

pedroslopez:
closes #1452


plugins/@grouparoo/onesignal/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,25 @@
...
+      },
+      group: {
+        singular: "OneSignal Group Tag",
+        plural: "OneSignal Group Tags",
+      },
+    },
+    properties: {
+      required: [{ key: "external_user_id", type: "string" }],
+      known: [
+        { key: "first_name", type: "string", important: true },
```
bleonard, line 18: Are these known tags?

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const value = newProfileProperties[key];
+    payload.tags[key] = formatVar(value);
+  }
+
+  // Groups are managed as tags
+  // If user is in group "High Value", the attribute "in_high_value" will be set to 1
+  for (const group of newGroups) {
+    const groupAttribute = getGroupAttribute(group);
+    payload.tags[groupAttribute] = 1;
```
bleonard: any sense if others do "true" or "false" or 1/0 or other here?

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  for (const group of oldGroups) {
+    if (!newGroups.includes(group)) {
+      const groupAttribute = getGroupAttribute(group);
+      payload.tags[groupAttribute] = null;
+    }
+  }
+
+  try {
```
bleonard, line 75: For all the `newGroups` thing we should make a `Segment`?

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,101 @@
...
+  // Dates are stored as unix timestamps (int)
+  if (value instanceof Date) {
+    return Math.floor(value.getTime() / 1000);
+  }
+
+  return value;
+}
+
+function getGroupAttribute(groupName: string) {
+  const normalizedName = groupName.toLowerCase().replace(/ /g, "_");
```
bleonard: these names can probably have lots of other suff in them, emoticons, etc.
Should it be more like `groupName.toLowerCase().replace(/[^a-z0-9-_\/.]/g, "_");`?

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,101 @@
...
+  const oldKeys = Object.keys(oldProfileProperties);
+  const allKeys = new Set([...newKeys, ...oldKeys]);
+
+  if (toDelete) {
+    // When deleting, we just clear the old tags. No real deletion occurs.
+    newKeys = [];
+    newGroups = [];
+  }
+
+  for (const key of allKeys) {
```
bleonard, line 50: do we think keys need to be normalized too?

plugins/@grouparoo/onesignal/src/lib/export/destinationMappingOptions.ts
pedroslopez: These tags have a special meaning in OneSignal: https://documentation.onesignal.com/docs/add-user-data-tags#user-names--preferences

image

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
pedroslopez: Opted for 1/0 based on the OneSignal docs https://documentation.onesignal.com/docs/add-user-data-tags#recommended-tags-to-send

image

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
pedroslopez: We could create Segments, but I don't see an obvious way to get an existing segment based on its filters, which is a problem when avoiding duplicates. It seems like they would have to create the segments themselves based on these tags we've added.

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
pedroslopez: Yeah it seems like you can set any string as the tag key, but their docs recommend to... 

image
(https://documentation.onesignal.com/docs/add-user-data-tags#recommended-tags-to-send)

While it does work, I think normalizing to be better safe than sorry could be a good option

bleonard: cool. let's do that then with the same function as the groups.

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
pedroslopez: @bleonard following this, should we transform booleans to 1/0 in the formatVar?

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
bleonard: Let's see if this works with a `id` that is a reproducible UUID from the groupName

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
bleonard: I think they are just trying to save space, so let's not be too surprising for the user and just be "true"

pedroslopez: @bleonard Same for the in_group tag or keep 1/0 there?

bleonard: no strong opinion. I guess lets go with "true"

bleonard: Looks like Segment does true/false https://segment.com/docs/connections/destinations/catalog/onesignal/

PR #1467, opened 2021-03-12
Application Monitoring with Sentry

evantahler:
This PR allows the Grouparoo application to be monitored with [Sentry](sentry.io).  This includes issue and performance  tracking.   This PR also changes our sample apps to use Sentry rather than New Relic. 

Screen Shot 2021-03-12 at 11 25 31 AM 

Screen Shot 2021-03-12 at 11 25 41 AM

2 Environment variables are needed:

* `SENTRY_DSN` - the URL and auth for your Sentry project
* `SENTRY_TRACE_SAMPLE_RATE` - the percentage of transactions to trace and store with Sentry.  Grouparoo will be tracing both API requests and background tasks.  A number like `0.01` (1%) is normal.


apps/staging-community/package.json
```diff
@@ -60,7 +60,7 @@
       "@grouparoo/mailchimp",
       "@grouparoo/marketo",
       "@grouparoo/mysql",
-      "@grouparoo/newrelic",
+      "@grouparoo/sentry",
```
bleonard, line 65: I assume @dcramer likes this line the most.

PR #1468, opened 2021-03-12
Refactor `sortConfigObjectsWithIds` to use topological sort algorithm

evantahler:
This PR improves the `sortConfigObjectsWithIds` method which is used to determine the order in which we should apply code config objects, based on dependencies and what they provide.  This PR uses a [topological sort algorithm](https://en.wikipedia.org/wiki/Topological_sorting) to _properly_ solve the DAG and crash if there's a loop.

Thanks to https://jeremyckahn.github.io/javascript-algorithms for doing the hard part.

### DIFF
--- a/core/__tests__/classes/codeConfig.ts
+++ b/core/__tests__/classes/codeConfig.ts
@@ -1,34 +1,93 @@
 import path from "path";
 import { loadConfigObjects } from "../../src/modules/configLoaders";
-import { getParentIds } from "../../src/classes/codeConfig";
+import {
+  getParentIds,
+  getConfigObjectsWithIds,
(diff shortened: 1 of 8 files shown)


evantahler:
@bleonard there's going to be an error in the next CI build, about a stronger need for magicTable to be able to find its parents/children.... let's talk about it!

evantahler:
FYI @bleonard - Synctable does a recursive call to `processConfigObjects`, and the secondary calls need the parent objects for sorting (ie: you still need App when building up and sorting the properties) 

https://github.com/grouparoo/grouparoo/pull/1468/commits/7e009ebf2a63711000c8da0e9c899283502a090b

PR #1486, opened 2021-03-16
Pipedrive Destination

### LINKED ISSUE #1081: Feature Request: PipeDrive Destination
Add Pipedrive to Grouparoo as a Destination

Sync profiles and groups to Pipedrive

pedroslopez:
closes #1081


plugins/@grouparoo/pipedrive/README.md
```diff
@@ -0,0 +1,17 @@
+# @Grouparoo/Pipedrive
+
+## A note on permissions
```
evantahler, line 3: Nice addition!

plugins/@grouparoo/pipedrive/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,160 @@
...
+
+// these comments to use nock
+const newNock = false;
+require("./../fixtures/destination-mapping-options");
+// or these to make it true
+// const newNock = true;
+// helper.recordNock(nockFile, updater);
+
+const appOptions = loadAppOptions(newNock);
+const appId = "app_a1bb05e8-0a4e-49c5-ad42-545f2e8762f9";
```
evantahler, line 30: FYI for the future: these can be anything, like `"pipedrive_app"`

plugins/@grouparoo/pipedrive/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,160 @@
...
+  });
+
+  test("does not fetch unsupported custom Person fields", async () => {
+    const client = await connect(appOptions);
+    const fields = await fetchKnownPersonFields(client);
+
+    const multipleOption = fields.find(
+      (f) => f.key === "multiple_option_field"
+    );
+    expect(multipleOption).toBeUndefined();
```
evantahler, line 101: Note: We chose to skip array Properties for now.

plugins/@grouparoo/pipedrive/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,410 @@
...
+    await expect(
+      runExport({
+        oldProfileProperties: {},
+        newProfileProperties: { Name: "Bill" },
+        oldGroups: [],
+        newGroups: [],
+        toDelete: false,
+      })
+    ).rejects.toThrow(/Email/);
+  });
```
evantahler, line 162: This technically could be any field, but we chose to keep it simple for now

From Pedro:
> You can search by any field https://developers.pipedrive.com/docs/api/v1/#!/Persons/searchPersons so technically we could use any field as an external id. Since we would be setting emails and Grouparoo (at the moment) doesn’t support multiple values for a field, we could also just use email.

plugins/@grouparoo/pipedrive/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+By default, non-admin accounts do not have permission to delete Persons as well as creating custom Person Fields. These permissions will need to be granted to be able to delete people and manage group membership, respectively. To learn more about how group memberships are handled, refer to the next section.
+
+You can read more about Permission Sets in Pipedrive [here](https://support.pipedrive.com/en/article/permission-sets).
+
+## Groups
+
+[Filters](https://support.pipedrive.com/en/article/filtering) will automatically be created for configured groups. A custom field will also be created for each group to keep track of group membership. The filter and field will be prefixed by `In Group: `, so the name of the Filter and Person Field that will be created for the group "High Value" would be `In Group: High Value`.
```
evantahler, line 11: Nice!

plugins/@grouparoo/pipedrive/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,410 @@
...
+    expect(newPersonId).toBeTruthy();
+
+    const { data } = await client.PersonsController.getDetailsOfAPerson(
+      newPersonId
+    );
+    expect(data.name).toBe("Bobby");
+    expect(data.email).toHaveLength(1);
+    expect(data.email[0].value).toBe(email3);
+  });
+});
```
evantahler, line 487: Please also add a test that confirms that a new Profile that is in at least one Group can be created from scratch.  Sometimes the destination APIs get wonky, have a cache delay, or something dumb.

plugins/@grouparoo/pipedrive/jest.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  testEnvironment: "node",
+  testTimeout: 240000,
```
evantahler: This is huge! Do you need to wait that long because the API is slow?  Most plugins look like this:


```
const { helper } = require("@grouparoo/spec-helper");

module.exports = {
  testEnvironment: "node",
  testTimeout: helper.defaultTime,
  setupFiles: ["<rootDir>/jest.setup.js"],
  transform: {
    "^.+\\.ts?$": "ts-jest",
  },
  testPathIgnorePatterns: ["<rootDir>/src", "<rootDir>/dist"],
};
```

plugins/@grouparoo/pipedrive/src/lib/connect.ts
```diff
@@ -0,0 +1,13 @@
+import { SimpleAppOptions } from "@grouparoo/core";
+import pipedrive from "pipedrive";
+import { EnhancedPersonsController } from "./client";
+
+export async function connect(appOptions: SimpleAppOptions) {
+  pipedrive.Configuration.apiToken = appOptions.apiToken;
+
+  pipedrive.EnhancedPersonsController = new EnhancedPersonsController(
+    pipedrive
+  );
```
evantahler: This may be a bug.  The apiToken looks to be set globally on `import pipedrive from "pipedrive"`... which means that if there are 2 Pipedrive Apps the wrong api key could be used.  This should be an instance (`new pipedrive()`), or we need to duplicate the object per `connect`...

plugins/@grouparoo/pipedrive/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,172 @@
...
+    people: "integer",
+    phone: "phoneNumber",
+    date: "date",
+    address: "string",
+    daterange: "date",
+
+    // TODO
+    time: null, // time in the format of "00:00:00"
+    timerange: null,
+    set: null, // multiple options
```
evantahler, line 84: Ok for later!

plugins/@grouparoo/pipedrive/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,172 @@
...
+    if (field.is_subfield && !field.key.match(subfieldsRegex)) continue;
+
+    const type = mapTypesToGrouparoo(field.field_type, field.name);
+    if (type) {
+      out.push({
+        key: field.name,
+        pipedriveKey: field.key,
+        required: required.includes(field.key),
+        important: field.important_flag, // they have a built-in (paid) feature to mark fields as important
+        groupMembershipField: field.name.startsWith(GROUP_FIELD_PREFIX), // used to hide our fields from destinationMappingOptions
```
evantahler, line 164: That's a nice helper to simplify the UI (don't show our custom group fields)

plugins/@grouparoo/pipedrive/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,128 @@
...
+    return { success: true };
+  }
+
+  const client = await connect(appOptions);
+  const cacheData: PipedriveCacheData = { appId, appOptions };
+
+  const oldEmail = oldProfileProperties["Email"];
+  const foundId = await client.EnhancedPersonsController.findPersonIdByEmail(
+    oldEmail
+  );
```
evantahler: What happens if this is the first time we sync and want to delete?  Or, what if there was an email change (old -> new) and a toDelete?  

sendgrid does this:
```js
const user = await client.getUser(email);
  let oldUser = null;
  if (oldEmail && oldEmail !== email) {
    oldUser = await client.getUser(oldEmail);
  }

  if (toDelete) {
    const userToDelete = user || oldUser;
    if (userToDelete) {
      await client.deleteUsers([userToDelete["id"]]);
    }
    return { success: true };
  }
```

plugins/@grouparoo/pipedrive/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  // Set group fields
+  for (const group of newGroups) {
+    const groupKey = await getGroupFieldKey(client, cacheData, group);
+    payload[groupKey] = "true";
+  }
+
+  for (const group of oldGroups) {
+    if (!newGroups.includes(group)) {
+      const groupKey = await getGroupFieldKey(client, cacheData, group);
```
evantahler, line 117: Will this still create the filter if we just trying to remove people?  May some option to make the groupKey but not the filter.

plugins/@grouparoo/pipedrive/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./dist",
+    "allowJs": true,
+    "module": "commonjs",
+    "target": "es2018",
```
evantahler: we just changed this to es2019 on main

plugins/@grouparoo/pipedrive/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,128 @@
...
+    newGroups
+  );
+
+  if (foundId) {
+    // Update existing Person
+    await client.EnhancedPersonsController.updateAPerson(foundId, payload);
+  } else {
+    // Create new Person
+    await client.PersonsController.addAPerson({ body: payload });
+  }
```
evantahler: assuming we tried a bit harder to understand `oldFoundId` and `newFoundId`, we would use `const foundId = newFoundId || oldFOundId` here.
You are relying a bit much overall on core handling use `old` to know someone is in the system.
Someone could have been in the system ahead of when grouparoo syncs for the first time.

evantahler (changes requested):
Looking good!  There are a few things to update below.  The 2 main topics are:
* The "thread safety" of the API key (the pipedrive client seems to be a singleton) 
* Over-reliance on core's notion of "old" and "new" email properties (the primary keys).  Things will get wonky if the person exited in Pipedrive already, before Grouparoo started updating them.

plugins/@grouparoo/pipedrive/jest.config.js
pedroslopez: Nope! I must have copied this from somewhere. Fixed

plugins/@grouparoo/pipedrive/src/lib/export/destinationMappingOptions.ts
pedroslopez: Good catch! Yeah, the pipedrive library uses a singleton pattern. Hadn't considered the possibility of having two different apps. I tried cloning the pipedrive object in multiple ways, but it doesn't work since internally it works under the assumption that it is a singleton. The library didn't do much over the standard REST APIs so I implemented our own client.

plugins/@grouparoo/pipedrive/src/lib/export/exportProfile.ts
pedroslopez: You're right! Added some tests for this.

However, I believe sendgrid might have an issue... It seems like it should be `const userToDelete = oldUser || user`, no? If we're updating the email and deleting at the same time, the update would not have occurred so we need to delete the _old_ email, not the new. Correct me if I'm wrong please.

Also, I noticed sendgrid silently ignores deleting nonexistent users. I'm throwing an error in my case so that grouparoo retries if need be- what should be the right way to do it?

plugins/@grouparoo/pipedrive/src/lib/export/exportProfile.ts
pedroslopez: I believe it would be `oldFoundId || newFoundId` in that case, see my previous comment on the `toDelete` case

plugins/@grouparoo/pipedrive/src/lib/connect.ts
pedroslopez: Good catch! Yeah, the pipedrive library uses a singleton pattern. Hadn't considered the possibility of having two different apps. I tried cloning the pipedrive object in multiple ways, but it doesn't work since internally it works under the assumption that it is a singleton. The library didn't do much over the standard REST APIs so I implemented our own client.

plugins/@grouparoo/pipedrive/__tests__/utils/shared.ts
```diff
@@ -0,0 +1,8 @@
+import { helper } from "@grouparoo/spec-helper";
+
+export async function indexUsers(isNewNock: boolean) {
+  // wait for search index
+  if (isNewNock) {
+    await helper.sleep(helper.shortTime);
+  }
+}
```
evantahler, line 8: Such a shame...

plugins/@grouparoo/pipedrive/src/lib/client.ts
```diff
@@ -0,0 +1,112 @@
...
+  request: AxiosInstance;
+
+  constructor(apiToken: string) {
+    this.request = axios.create({
+      baseURL: "https://api.pipedrive.com/v1",
+      params: {
+        api_token: apiToken,
+      },
+    });
+  }
```
evantahler, line 13: That's a shame that their API client was unusable (singleton).  That's cool that Axios can automagically append the API Token for us!

plugins/@grouparoo/pipedrive/src/lib/client.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  async getCurrentUser() {
+    const { data } = await this.request.get("/users/me");
+    return data.data;
+  }
+
+  async findPersonIdByEmail(email: string): Promise<number | null> {
```
evantahler: Not a blocker, FYI for later - I don't think you ever need to ` | null` in typescript.  Null (or undefined) are always valid return types, e.g.: 

Screen Shot 2021-03-22 at 6 24 35 PM

plugins/@grouparoo/pipedrive/src/lib/client.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  async deletePerson(id: number) {
+    const { data } = await this.request.delete(`/persons/${id}`);
+    return data;
+  }
+
+  async getAllPersonFields(): Promise<any[]> {
+    const { data } = await this.request.get("/personFields");
+    return data.data;
+  }
```
evantahler: There's probably a /slightly/ better type which can be used here, at least indicating that you are getting back an array of objects: `async getAllPersonFields(): Promise<any[{[key: string]: any}>`.  The same goes for a few more of the `any` types in this file

plugins/@grouparoo/pipedrive/src/lib/export/exportProfile.ts
evantahler: I agree that you should be deleting the `oldEmail` in this case.  I think.  It's confusing :/

Whether or not to fail or silently when a user can't be found is a really good question, and one we need to address more globally with Grouparoo.  We are going to start making a distinction between `syncing data` (match my group exactly and delete people) and `enriching` data (update the people who are there, if you can, and don't delete them) which will make this more clear... hopefully.  Most destinations to-date are in the "sync" mode - which means we should delete people if we can. 

More to come!

evantahler (commented):
Good work writing a whole new API client!   I think this is getting _really_ close to ready.  Let's chat about the deletion case a bit to make sure we are on the same page.

plugins/@grouparoo/pipedrive/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,139 @@
...
+  let oldFoundId = null;
+  if (oldEmail && oldEmail !== newEmail) {
+    oldFoundId = await client.findPersonIdByEmail(oldEmail);
+  }
+
+  const foundId = oldFoundId || newFoundId;
+  if (toDelete) {
+    if (!foundId) {
+      throw new Error(
+        `Person with email ${oldEmail || newEmail} was not found to delete.`
```
bleonard: In other ones, this fails silently, so probably `return` here

plugins/@grouparoo/pipedrive/src/lib/export/exportProfile.ts
```diff
@@ -0,0 +1,139 @@
...
+  const oldEmail = oldProfileProperties["Email"];
+
+  const newFoundId = await client.findPersonIdByEmail(newEmail);
+
+  let oldFoundId = null;
+  if (oldEmail && oldEmail !== newEmail) {
+    oldFoundId = await client.findPersonIdByEmail(oldEmail);
+  }
+
+  const foundId = oldFoundId || newFoundId;
```
bleonard: The nuance in the other ones tends to be this:
* for deleting it's `oldFoundId || newFoundId`
* but for updating it's `newFoundId || oldFoundId`

pedroslopez: Hmm, I still think it should be `oldFoundId || newFoundId`. If you're updating the email/foreignkey, I believe you would want to search for the old email to be able to update it to a new one. 

Although in that case it still works since `newFoundId` would not exist, and would use `oldFoundId` instead. Maybe there's a case I'm not seeing?

pedroslopez: To document this here a bit, Brian is right in that there _is_ a difference between deleting and updating. The special case I wasn't seeing comes to light when dealing with profiles that may have been created from other sources (not grouparoo). 

Essentially, if you're **deleting** someone and changing their email at the same time _and_ a user with that new email already exists in the destination, we want to delete the old user and leave the one with the new email alone. As it stands now, the sendgrid plugin would not be doing this.

However, if you're **updating** someone and changing their email at the same time _and_ a user with that new email already exists in the destination, we would want to update the new one instead of the old one. This is mainly because if we were to update the old one at this point, we could create a unique key conflict by now having two people with the same (new) email. Note that in this case, the profile with the old email would remain untouched.

I'll write more about this and create a document to standardize the way Grouparoo handles primary id changes, but wanted to leave a note here since this is where it came up.

evantahler (approved):

PR #1492, opened 2021-03-17
Add an initializer that looks for multiple versions of core 

seancdavis:
It throws an error and gives a more informative message if it finds multiple versions.

### DIFF
--- a/core/src/initializers/coreVersionValidator.ts
+++ b/core/src/initializers/coreVersionValidator.ts
@@ -1,0 +1,20 @@
+import { Initializer } from "actionhero";
+
+export class CoreVersionValidator extends Initializer {
+  constructor() {
+    super();
+    this.name = "coreVersionValidator";
+    this.startPriority = 999999999999999;
+  }
+
+  async initialize() {
+    const modulePrefixes = Object.keys(require.cache)
+      .filter((file) => file.includes("grouparoo/core"))
+      .map((file) => file.split("grouparoo/core")[0]);
+    const uniqPrefixes = new Set(modulePrefixes);
+
+    if (uniqPrefixes.size > 1) {
+      throw new Error("Found multiple versions of @grouparoo/core.");
+    }
+  }
+}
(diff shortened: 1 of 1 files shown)


core/src/initializers/coreVersionValidator.ts
```diff
@@ -0,0 +1,20 @@
+import { Initializer } from "actionhero";
+
+export class CoreVersionValidator extends Initializer {
+  constructor() {
+    super();
+    this.name = "coreVersionValidator";
+    this.startPriority = 999999999999999;
```
seancdavis: Turns out Actionhero doesn't like `Infinity`.

core/src/initializers/coreVersionValidator.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  async initialize() {
+    const modulePrefixes = Object.keys(require.cache)
+      .filter((file) => file.includes("grouparoo/core"))
+      .map((file) => file.split("grouparoo/core")[0]);
+    const uniqPrefixes = new Set(modulePrefixes);
+
+    if (uniqPrefixes.size > 1) {
+      throw new Error("Found multiple versions of @grouparoo/core.");
```
seancdavis, line 17: This seemed appropriate enough for now.

core/src/initializers/coreVersionValidator.ts
```diff
@@ -0,0 +1,20 @@
+import { Initializer } from "actionhero";
+
+export class CoreVersionValidator extends Initializer {
```
seancdavis, line 3: Didn't seem like there was a great way to test this programmatically. I did it manually by switching the core version in the PG plugin and it worked great.

evantahler: Makes sense!

core/src/initializers/coreVersionValidator.ts
evantahler: I *think* that 10000 should be more than enough :D

evantahler (approved):

PR #1493, opened 2021-03-17
Grouparoo Init can run on Windows

evantahler:
Related to https://github.com/grouparoo/grouparoo/discussions/1491

Grouparoo is still not supported on on windows natively.   However, you can use the WSL system to run Ubuntu on Windows to then run Grouparoo.

Screen Shot 2021-03-18 at 2 51 21 PM

- [x] Use `npm.cmd` not `npm` when sub-shelling out to NPM for `grouparoo init` and `grouparoo install`

### DIFF
--- a/cli/src/utils/npm.ts
+++ b/cli/src/utils/npm.ts
@@ -3,6 +3,8 @@ import { spawnPromise } from "./spawnPromise";
 import Chalk from "chalk";
 
 export namespace NPM {
+  const NPM = /^win/.test(process.platform) ? "npm.cmd" : "npm";
+
   export async function install(
     logger: any,
     workDir: string,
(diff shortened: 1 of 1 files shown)


evantahler:
It looks like even on a vanilla install of node + npm on windows, there are problems

```
C:\Users\evant\grouparoo>npm install grouparoo
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm ERR! cb() never called!

npm ERR! This is an error with npm itself. Please report this error at:
npm ERR!     <https://npm.community>

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\evant\AppData\Roaming\npm-cache\_logs\2021-03-18T20_26_50_383Z-debug.log
```

You may need to ` npm install --global windows-build-tools` from an **administrator** instance of PowerShell

You may also need to `npm config set strict-ssl=false` (https://stackoverflow.com/questions/57444794/npm-err-cb-never-called-in-windows10)

evantahler:
Even after getting `grouparoo init` to properly call npm on windows, there are still issues building dependencies which compile, e.g. SQLite.  At this time, I think the most reliable path forward will be WSL (https://docs.microsoft.com/en-us/windows/wsl/install-win10) and to run Grouparoo on Windows via Ubuntu.

evantahler:
Another windows-specific issue - older versions of IE/Edge do not support `globalThis` and the UI will not function.  Newer versions of the Edge browser work OK.

PR #1464, opened 2021-03-11
Fix typos in destination config template files

seancdavis:
Found this while working on plugin documentation.

PR #1494, opened 2021-03-18
Clean up new plugin generator templates

seancdavis:
See discussion in #1492.


pedroslopez (commented):
I believe the initial discussion was actually on #1464 for future reference. This looks good! As long as we're making things consistent here, seems like there are a few other plugins that have the `abc123` thing on the app options template (iterable, sailthru, sendgrid)

evantahler (approved):
Thanks @seancdavis! Huzzah for consistency.

PR #1502, opened 2021-03-19
Error message when table source generator can't find specified columns

seancdavis:
I ran into an issue because I didn't quote the asterisk when looking for column names:

    $ grouparoo generate [thing]:table:source sourceId --from appId --with * ...

Now the generator looks for a column match if the `with` command is used and if it doesn't find any columns it will send a message that is hopefully helpful.

### DIFF
--- a/plugins/@grouparoo/app-templates/src/source/table/templates.ts
+++ b/plugins/@grouparoo/app-templates/src/source/table/templates.ts
@@ -239,6 +239,12 @@ async function loadTablesAndColumns(
           });
         }
       }
+
+      if (params.with && map.length === 0) {
+        throw new Error(
+          'Could not find any listed columns in source. If using * make sure you used quotes ("*")?'
+        );
+      }
     }
   } finally {
     await app.disconnect();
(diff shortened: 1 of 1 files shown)


seancdavis:
This could probably use some specs, but I didn't immediately find a place where we're testing using the command line (vs the command via Node). Is there an example of using the CLI from a child process? If so, please point me in that direction and I'll follow that convention. Otherwise, I'll take a whack at it.

evantahler:
I don't think we have any CLI tests that "shell out" in core.  There are some over [here](https://github.com/grouparoo/grouparoo/blob/main/apps/staging-enterprise/__tests__/cli#L55-L59) (test suite in bash), but I don't think that's the right way to do it.  

I think we probably should really run the command for this test.  Actionhero's test suite has a good utility to make a sub-process and capture outputs -> https://github.com/actionhero/actionhero/blob/master/__tests__/core/cli.ts#L30-L80

seancdavis:
@evantahler I think these CLI tests as a forked process would be super valuable, but there are some questions to answer and it's becoming something big enough to discuss.

Given that this is a smaller feature, I'd suggest we merge this as is, and then discuss and work CLI tests separately. I [created a story to discuss](https://www.pivotaltracker.com/story/show/177463643).

Thoughts?

evantahler (approved):
Makes sense - we can work on the CLI testing late

PR #1503, opened 2021-03-19
SQLite Plugin

seancdavis:
Adds a SQLite source and destination. This was built from the Postgres plugin and doesn't differ much in terms of functionality.

SQLite is an oddball, however. The few important notes are mentioned [in the working doc](https://docs.google.com/document/d/165U4GtQTnvZC-NT_4VWVDC6DsKtdw5F58XoRIa6axn8/edit#) and will eventually be added to the website documentation.


plugins/@grouparoo/app-templates/src/source/table/templates.ts
```diff
@@ -239,6 +239,12 @@ async function loadTablesAndColumns(
           });
         }
       }
+
+      if (params.with.replace(/['"]+/g, "") && map.length === 0) {
+        throw new Error(
+          'Could not find any listed columns in source. If using * make sure you used quotes ("*")?'
+        );
+      }
```
evantahler: Note: This change helps avoid an error in bash/zsh in which an un-quoted `*` (vs `"*"`) would be expanded to match the files in this directory, and not be the string `"*"` like we wanted.

plugins/@grouparoo/sqlite/__tests__/query-import/import-property.ts
```diff
@@ -0,0 +1,101 @@
...
+    const value = await getPropertyValue(sql);
+    expect(value).toEqual(["true"]);
+  });
+
+  test("returns undefined when data is not avilable", async () => {
+    const sql = `SELECT ios_app FROM ${usersTableName} WHERE email = '{{ badName }}'`;
+    const value = await getPropertyValue(sql);
+    expect(value).toEqual(undefined);
+  });
+});
```
evantahler, line 113: Because we know that SQLite has strange types, can you also add a test for a `date` here?  It may be helpful for the future to know the exact format we expect back, even if it is a string

plugins/@grouparoo/sqlite/__tests__/table-import/get-changed-row-count.ts
```diff
@@ -0,0 +1,69 @@
...
+      tableName: usersTableName,
+      highWaterMarkCondition: {
+        columnName: "stamp",
+        value: "2000/01/01 12:00:00",
+        filterOperation: FilterOperation.GreaterThan,
+      },
+    });
+
+    expect(rowCount).toBe(10);
+  });
```
evantahler, line 62: Note: we are getting the string response from SQLite so the date formats should be OK

plugins/@grouparoo/sqlite/__tests__/table-import/import-properties.ts
```diff
@@ -0,0 +1,881 @@
...
+          expect(values[otherProfile.id]).toEqual([5]);
+        });
+        test("min", async () => {
+          const values = await getPropertyValues({
+            column,
+            sourceMapping,
+            aggregationMethod: "min",
+          });
+          expect(values[profile.id]).toEqual(["2020/02/01"]);
+          expect(values[profile.id]).toEqual(["2020/02/01"]);
```
evantahler, line 289: They remain strings... so it works!

plugins/@grouparoo/sqlite/__tests__/table-import/import-property.ts
```diff
@@ -0,0 +1,806 @@
...
+        const value = await getPropertyValue(
+          {
+            column,
+            sourceMapping,
+            aggregationMethod,
+          },
+          [{ op, key: "date", match: "2020/02/15" }]
+        );
+        expect(value).toEqual(2);
+      });
```
evantahler, line 685: Anything wacky here with `>` and `<` on dates (that are actually strings).  Does SQLite do the right thing?

plugins/@grouparoo/sqlite/__tests__/table-import/profiles.ts
```diff
@@ -0,0 +1,180 @@
...
+      let importedIds;
+
+      const page1 = await runIt({
+        limit,
+        sourceOffset: 0,
+        highWaterMark,
+      });
+      expect(page1.importsCount).toBe(4);
+      expect(page1.sourceOffset).toBe(0);
+      expect(Object.values(page1.highWaterMark)[0]).toMatch("2020/02/04"); // the times changes based on the TZ of the test server, but the date seems to be OK...
```
evantahler, line 150: Good note!

plugins/@grouparoo/sqlite/__tests__/table-import/schedule-options.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  afterAll(async () => {
+    await afterData();
+  });
+  test("gets list of columns that can handle highwatermark", async () => {
+    const columns = await getColumns();
+    const columnNames = columns.map((r) => r.key).sort();
+    // In SQLite, data is essentially either text or a number, so they could all
+    // technically be HWM columns.
```
evantahler, line 43: It looks like you have chosen to filter out strings at least, which is a good step to making this list possible HWMs shorter.

plugins/@grouparoo/sqlite/__tests__/utils/data.ts
```diff
@@ -0,0 +1,142 @@
...
+  UNIQUE("userId", "group")
+)
+`,
+};
+
+export const appOptions = {
+  file: path.join(os.tmpdir(), `${config.sequelize.database}.sqlite`),
+};
+
+const appId = "app_31bb06e8-0a4e-49c3-ad42-545f2e8662e1";
```
evantahler: FYI this can be anything, like `test_app` in the future

plugins/@grouparoo/sqlite/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "author": "Grouparoo Inc <hello@grouparoo.com>",
+  "name": "@grouparoo/sqlite",
+  "description": "The Grouparoo SQLite Plugin",
+  "version": "0.2.8-alpha.1",
```
evantahler: Be sure to bump this, and the `@grouparoo/core` version

plugins/@grouparoo/sqlite/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "prepare": "rm -rf dist && tsc --declaration",
+    "test": "jest",
+    "pretest": "npm run lint && npm run prepare",
+    "lint": "prettier --check src __tests__",
+    "watch": "tsc --watch"
+  },
+  "dependencies": {
+    "@grouparoo/app-templates": "0.2.8-alpha.1",
+    "@types/sqlite3": "^3.1.7",
+    "sqlite3": "^5.0.2"
```
evantahler: Please remove the `^` and use exact versions

plugins/@grouparoo/sqlite/src/lib/connect.ts
```diff
@@ -0,0 +1,22 @@
...
+export const connect: ConnectPluginAppMethod = async ({ appOptions }) => {
+  const formattedOptions: any = Object.assign({}, appOptions);
+
+  let dbPath = formattedOptions.file;
+
+  if (!path.isAbsolute(dbPath)) dbPath = path.join(getParentPath(), dbPath);
+
+  // Don't allow connections unless the file already exists. This avoids
+  // in-memory storage and creating the file during instantiation.
+  if (!fs.existsSync(dbPath) || fs.lstatSync(dbPath).isDirectory()) return null;
```
evantahler: Maybe this should `throw` an explicit message about the database file not being found?

plugins/@grouparoo/sqlite/src/lib/sqlite.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.database = database;
+    this.connection = undefined;
+  }
+
+  async asyncConnect(): Promise<string> {
+    return new Promise((resolve, reject) => {
+      this.connection = new Database(this.database, (err: Error) => {
+        if (err) {
+          return reject(
+            new Error(`Error connecting to SQLite database: ${this.database}`)
```
evantahler: The original error may be useful here too (perhaps append it to the string)?

plugins/@grouparoo/sqlite/src/lib/sqlite.ts
```diff
@@ -0,0 +1,48 @@
...
+        );
+      });
+    });
+  }
+
+  async asyncQuery(query: string): Promise<Array<any>> {
+    return new Promise((resolve, reject) => {
+      this.connection.all(query, (err: Error, res: any[]) => {
+        if (err) {
+          return reject(new Error(`Error running query: ${err}`));
```
evantahler: I think you'll end up with `Error running query: Error: foo` I think.   Perhaps `new Error(`Error running query: ${err.message}`)`

plugins/@grouparoo/sqlite/src/lib/test.ts
```diff
@@ -0,0 +1,20 @@
...
+export const test: TestPluginMethod = async ({ connection }) => {
+  const genericError = {
+    success: false,
+    message: "Could not connect to database.",
+  };
+
+  // Return if we don't have a connection object.
+  if (!connection) return genericError;
+
+  const query = `SELECT count(*) FROM sqlite_master WHERE type = 'table' AND name != 'android_metadata' AND name != 'sqlite_sequence';`;
```
evantahler: `android_metadata`?  Are there other strange os-specific tables we should be ignoring?

plugins/@grouparoo/sqlite/src/lib/table-import/getColumns.ts
```diff
@@ -0,0 +1,56 @@
...
+  float: "float",
+  int: "integer",
+  integer: "integer",
+  num: "float",
+  phoneNumber: "phoneNumber",
+  string: "string",
+  text: "string",
+  url: "url",
+  varchar: "string",
+};
```
evantahler, line 36: We could be even fancier here, but for the demo, as long as the columns are named properly (eg: `email`), then the generated config files will be the right types.

plugins/@grouparoo/sqlite/src/lib/table-import/getSampleRows.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  GetSampleRowsMethod,
+  DataResponseRow,
+} from "@grouparoo/app-templates/dist/source/table";
+
+export const getSampleRows: GetSampleRowsMethod = async ({
+  connection,
+  tableName,
+}) => {
+  const query = `SELECT * FROM ${tableName} WHERE id IN (SELECT id FROM ${tableName} ORDER BY RANDOM() LIMIT 10)`;
```
evantahler, line 10: Other databases got really slow when `ORDER BY RANDOM()` was used.  Is SQLite OK?

plugins/@grouparoo/sqlite/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./dist",
+    "allowJs": true,
+    "module": "commonjs",
+    "target": "es2018",
```
evantahler: we bumped to `es2019` since this branch was started

evantahler (changes requested):
Looking really good, and strong testing!  
I've added a few things to check and update.

Looks like you'll also need to do a rebase since the version release.

plugins/@grouparoo/app-templates/src/source/table/templates.ts
seancdavis: This came in via #1502. Rebasing removed the change.

plugins/@grouparoo/sqlite/__tests__/table-import/get-changed-row-count.ts
seancdavis: I had updated the data formats to be consistent between the test data and the specs. Postgres was loading in dates like `2020/01/01`, but testing with a format like `2020-01-01`. Since we're working with strings, I chose to a consistent format. After making that change, everything worked as expected.

seancdavis: Added some notes [to the doc](https://docs.google.com/document/d/165U4GtQTnvZC-NT_4VWVDC6DsKtdw5F58XoRIa6axn8/edit#).

plugins/@grouparoo/sqlite/__tests__/table-import/import-property.ts
seancdavis: It does as long as the format matches. We don't have that explicit test here, but I'd say the HWM tests cover that condition.

plugins/@grouparoo/sqlite/__tests__/table-import/schedule-options.ts
seancdavis: It's actually not filtering anything out. It's just that this list is coming from the purchases data, so there aren't as many columns.

plugins/@grouparoo/sqlite/__tests__/utils/data.ts
seancdavis: I made the change anyways, just to simplify.

plugins/@grouparoo/sqlite/src/lib/test.ts
seancdavis: I couldn't find a good list of this. But upon further inspection, anything that begins with `sqlite_` is reserved for internal use. Therefore, I've made the change to only filter out those tables beginning with `sqlite_` since any other table could be used for meta purposes, theoretically.

plugins/@grouparoo/sqlite/src/lib/table-import/getColumns.ts
seancdavis: My thought exactly.

plugins/@grouparoo/sqlite/src/lib/sqlite.ts
seancdavis: I ended up printing the error first and then the query second. Thought that could come in handy, as it's likely to be a syntax goof or something  I did.

plugins/@grouparoo/sqlite/src/lib/table-import/getSampleRows.ts
seancdavis: I tested with 1k rows and the performance didn't bother me. Would you prefer I go the Postgres route and just pick the first 10?

seancdavis:
We now have better connection errors!

Screen Shot 2021-03-24 at 6 58 24 AM

Screen Shot 2021-03-24 at 6 58 47 AM

If using a relative path, the message will print the full path that it's looking for. It felt like it was more informative to do this, as opposed to being harmful and exposing the full working directory of the application. Especially since this tends to be an administrator-type of use case.

plugins/@grouparoo/sqlite/src/lib/table-import/getSampleRows.ts
evantahler: I don't think it matters at the moment - but it really was a bottleneck for large MySQL/Postgres databases, so I wanted to keep an eye out for it

PR #1528, opened 2021-03-24
Default option values

pedroslopez:
This allows setting a `defaultValue` when configuring app and destination options. This happens in the `OptionHelper` so that default values are applied both in the enterprise UI and when working with options in plugins (e.g. the received options on the `connection` method)

### DIFF
--- a/core/__tests__/models/app/app.ts
+++ b/core/__tests__/models/app/app.ts
@@ -1,5 +1,5 @@
 import { helper } from "@grouparoo/spec-helper";
-import { plugin, App, Option, Log, Destination } from "../../../src";
+import { plugin, App, Option, Log } from "../../../src";
 import { api, redis, utils } from "actionhero";
 
 describe("models/app", () => {
--- a/core/__tests__/models/app/app.ts
+++ b/core/__tests__/models/app/app.ts
@@ -285,14 +285,7 @@ describe("models/app", () => {
         apps: [
           {
             name: "test-template-app",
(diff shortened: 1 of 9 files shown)


core/__tests__/models/app/app.ts
```diff
@@ -350,6 +364,53 @@ describe("models/app", () => {
       expect(testCounter).toBe(1);
     });
 
+    describe("default option values", () => {
```
pedroslopez: I have no idea if this is the right file to do this in- it seems like at the moment there isn't a test for modules/optionHelper

core/src/modules/optionHelper.ts
```diff
@@ -324,4 +324,26 @@ export namespace OptionHelper {
 
     return options;
   }
+
+  async function getDefaultOptionValues(
+    instance: Source | Destination | Schedule | Property | App
```
pedroslopez, line 329: At the moment only App, Source and Destination are implemented as stated in the story. Should I look into these other types?

pedroslopez: From standup: not for now

core/__tests__/models/app/app.ts
pedroslopez: Moved to models/options

evantahler (approved):
Nice work! The optionHelper was totally the right place to put this - this means that we can use the same functionality elsewhere one day!

PR #1529, opened 2021-03-24
Imports and Exports that aren't complete after a time will be retried

evantahler:
This PR makes Import association and Export processing more resilient. 

Both Imports and Exports have been updated to follow this pattern:

- have a `startedAt` column  that will be set when the record is first enqueued to be processed
- when looking for records to process, consider those with a null `startedAt` or those with a `startedAt` that is far in the past & not yet complete
  - Imports: not yet complete = no profile Id
  - Exports: not yet complete = no completedAt
- add 2 new settings, `imports-retry-delay-seconds` and `exports-retry-delay-seconds` to know what "too far in the past" means; default to 5 minutes

Screen Shot 2021-03-24 at 3 14 39 PM
Screen Shot 2021-03-24 at 3 14 43 PM

### DIFF
--- a/core/src/migrations/000061-addImportStartedAt.ts
+++ b/core/src/migrations/000061-addImportStartedAt.ts
@@ -1,0 +1,12 @@
+export default {
+  up: async function (migration, DataTypes) {
+    await migration.addColumn("imports", "startedAt", {
+      type: DataTypes.DATE,
+      allowNull: true,
+    });
+  },
+
+  down: async function (migration) {
+    await migration.removeColumn("imports", "startedAt");
+  },
+};
--- a/core/src/modules/ops/export.ts
+++ b/core/src/modules/ops/export.ts
@@ -90,7 +90,8 @@ export namespace ExportOps {
 
   export async function processPendingExportsForDestination(
     destination: Destination,
-    limit = 100
+    limit = 100,
+    delayMs = 1000 * 60 * 5
   ) {
     const app = await destination.$get("app");
     const { pluginConnection } = await destination.getPlugin();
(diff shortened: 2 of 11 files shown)


core/src/migrations/000061-addImportStartedAt.ts
```diff
@@ -0,0 +1,12 @@
+export default {
+  up: async function (migration, DataTypes) {
+    await migration.addColumn("imports", "startedAt", {
+      type: DataTypes.DATE,
+      allowNull: true,
+    });
+  },
```
evantahler, line 7: We don't need to backfill this migration: either you are in the pending state and your `startedAt` is null, or you are already associated and have a `profileId` + related timestamp.

core/src/modules/ops/export.ts
```diff
@@ -99,7 +100,9 @@ export namespace ExportOps {
 
     _exports = await Export.findAll({
       where: {
-        startedAt: null,
+        startedAt: {
+          [Op.or]: [null, { [Op.lt]: new Date().getTime() - delayMs }],
```
bleonard, line 104: WIth this `or` it's notable that the code below on 117 isn't updated
```
        where: {
          id: { [Op.in]: _exports.map((e) => e.id) },
          startedAt: null,
        },
```
so in the "stale" case, we aren't really updating the `startedAt` to now.

bleonard: I'm wondering if it might be cleaner for a "Sweeper" job to set their startedAt to `null` instead of the `or` here.

In the ideal case, it would also make sure it wasn't still in the queue, I suppose. Though maybe a queueLock on that is better?

PR #1531, opened 2021-03-25
Better validation errors when using the wrong `id`

seancdavis:
When using the wrong `id` value for a config object, running `grouparoo validate` leads to some odd behavior. It looks like there are several errors when there is really only one.

This was a sorting error. Our topological sorting algorithm hit a snag and so the config objects were processed how they were read and were not sorted as they should have been. 

## Solutions

**Printing all errors:**

It seems like a problem to me that the errors being thrown by the sorting method are swallowed up. So the first thing I did was to make sure we're rendering those errors. Even if they aren't helpful to the developer, they may help us debug the issue and provide better support.

Here's an example of the sorter throwing some random issue:

```
🦘 Grouparoo: validate

Validating 13 objects...
error: You a bad dev.
info: [ config ] validated App `data_warehouse` (data_warehouse)
info: [ config ] validated App `mailchimp` (mailchimp)
warning: [ config ] error with group `all_emails` (all_emails): Error: cannot find Property email
warning: [ config ] error with group `high_value_spanish_speakers` (high_value_spanish_speakers): Error: cannot find Property lifetime_value
warning: [ config ] error with property `email` (email): Error: cannot find Source with id "users"
warning: [ config ] error with property `first_name` (first_name): Error: cannot find Source with id "users"
warning: [ config ] error with property `language` (language): Error: cannot find Source with id "users"
warning: [ config ] error with property `last_name` (last_name): Error: cannot find Source with id "users"
warning: [ config ] error with property `ltv` (ltv): Error: cannot find Source with id "purchases"
warning: [ config ] error with property `purchase_count` (purchase_count): Error: cannot find Source with id "purchases"
warning: [ config ] error with source `purchases` (purchases): Error: cannot find Property with id "user_id"
info: [ config ] validated Source `users` (users)
info: [ config ] validated Property `user_id` (user_id)
info: [ config ] validated Schedule `users_schedule` (users_schedule)

❌ Validation failed - 10 validation errors
```

**Better error message:**

The second thing I did was add a little logic that could determine whether or not the error message was the result of using an incorrect `id` value. It's a little rough and I'm guessing there's a better way to do this.

The downside to this second change is that I'm also then cutting off the rest of the validation. That is intentional since we know we're likely to run into issues with objects being in the wrong order and we're going to get bogus failures that aren't helpful in assessing the problem.

## Struggles

I feel like there's got to be a better way to serve our devs, though. Here are some random thoughts:

- It feels weird that I'm cutting off validation in one case (when the error message is "Cannot find ...") but not the other. Seems like I should be consistent, but I was on the fence with understanding what other issues might arise, so I left it as is.
- I wonder if there's a way we could be smarter about this? Instead of topo sorter throwing an error, could it return the error silently, ignore the thing, and move on? Mostly I'm wanting to help the developer locate where the issue is and the best we have here is the specific `id` they've used. That's a good start, but I feel like we could be more helpful. But I don't have a great idea.
- There's probably a better way to pinpoint the type of error we're receiving from the sorter helper without needing to parse the message's string. I didn't want to overthink it.

## References

[🐞  Tracker bug](https://www.pivotaltracker.com/story/show/177448952)


seancdavis:
I'm going to hold on fixing the specs until I get some feedback because it would be wiping away what looks like useful feedback. IOW it feels like a step back, even though it helps us in this one specific instance.

core/src/modules/configLoaders/index.ts
```diff
@@ -129,8 +128,24 @@ export async function processConfigObjects(
       configObjects = sortConfigurationObjects(configObjects);
     }
   } catch (error) {
-    // the config loaders below will produce better error messages
-    sortError = error;
+    errors.push(error);
+    // If the sorting issue was that we couldn't find an object, the assumption
+    // is that the developer is targeting an object of some ID that does not
+    // exist.
+    if (error.message.includes("Cannot find ")) {
```
bleonard: maybe `startsWith` since are assuming that in the next line.

core/src/modules/configLoaders/index.ts
```diff
@@ -129,8 +128,24 @@ export async function processConfigObjects(
       configObjects = sortConfigurationObjects(configObjects);
     }
   } catch (error) {
-    // the config loaders below will produce better error messages
-    sortError = error;
+    errors.push(error);
```
bleonard: I note below that `errors` has the `message` string and this is the whole object. should it be `error.message`?

core/src/modules/configLoaders/index.ts
bleonard: because we control this code, we could also set something on the error like `unknownNodeId` or whatever and not have funny parsing.

bleonard: We could also collect all the unknown nodes in the topological sort so we tell them all the missing nodes

core/src/modules/configLoaders/index.ts
```diff
@@ -212,8 +227,6 @@ export async function processConfigObjects(
     }
   }
 
-  if (sortError && errors.length === 0) errors.push(sortError.message);
```
bleonard, line 215: i think it would be ok to return here if there is any error in sorting and not do the specific parsing. For example, that code also does `throw new Error("The graph is not a DAG");`
We should stop then, too. (though probably with a better error message).

bleonard (commented):
I think my suggestion would be to have topologicalSort become able to return both a result and an array of error strings (or throw an Error with a `.errorMessages` property added).
Then if there are an errors with that sorting, return the `error.errorMessages` or `[error.message]`, whatever they are.

core/src/modules/configLoaders/index.ts
seancdavis: Yes! Found that when looking to clean up the specs.

core/src/modules/configLoaders/index.ts
seancdavis: I agree. But it leaves the bigger, generic question unanswered: Should we continue _trying_ to validate objects if we weren't able to properly sort them?

Or should we consider properly sorting to be a prerequisite of running through object validation? (i.e. If we find _any_ sorting error, we don't try to load _any_ objects?)

bleonard: The main question would be if we want to support referencing in code config objects that dont exist on code config. If so, then it still might be there and we need to do better messages downstream. If not, then this app is never going to run anyway, so it's probably easier to just throw here.

PR #1534, opened 2021-03-25
Properly handle changing email address for Pipedrive and Iterable

pedroslopez:
This PR is meant to apply the necessary changes to plugins so they conform with the new standard specified in the document [Standardizing Destination ForeignKey Value Changes](https://docs.google.com/spreadsheets/d/1fEVSaq6KJ71P4xkuxSalr1RL1tbqeUnjfxI4SxE-01s/edit?usp=sharing)

- [x] **Pipedrive**: On `toDelete: true`, deletes old user if both new and old are found.
- [x] **Iterable**: Does not search in destination and always deletes/updates the old FK.

I started with these on this PR, but for the others let's just make sure they're working as expected as we implement the SyncModes. The second sheet in the Google Sheets spreadsheet linked above contains a starting point for which plugins need to be addressed.


plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -529,6 +529,65 @@ describe("iterable/exportProfile", () => {
...
+      oldProfileProperties: {
+        email: brandNewEmail,
+        userId: brandNewUserId,
+        name: brandNewName,
+      },
+      newProfileProperties: {
+        email: newEmail,
+        userId: newUserId,
+        name: otherName,
+      },
```
evantahler: So... `newEmail` is newer than `brandNewEmail`?   This can probably be tightened up a bit :D

evantahler (changes requested):
Looks good, other than those variable names. The tests make it very clear what's supposed to happen... and... they pass!

Reminder - Please add a `label` so the PR ends up in our release notes (I added it).

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
pedroslopez: Hahahah true! I just reused names that were there. Cleaned it up a bit and changed the ordering so mental gymnastics are kept to a minimum

evantahler (approved):

PR #1535, opened 2021-03-26
Properly bump `startedAt` for old Imports or Exports when retried

evantahler:
This PR fixes a bug introduced in https://github.com/grouparoo/grouparoo/pull/1529 which would cause Imports or Exports which had previously gotten stuck to be forever retried.  This PR now properly bumps the `startedAt` timestamp when they are retried.

This PR also applies a minimum delay before retrying, even if the Setting was updated to a shorter window.

I didn't go with the sweeper pattern because I think it's important to bump the `startedAt` idempotently (in a transaction), so there's no risk of the Import/Export being enqueued twice.

### DIFF
--- a/core/src/modules/ops/export.ts
+++ b/core/src/modules/ops/export.ts
@@ -9,6 +9,8 @@ import { CLS } from "../../modules/cls";
 import { Op } from "sequelize";
 
 export namespace ExportOps {
+  const defaultExportProcessingDelay = 1000 * 60 * 5;
+
   /** Count up the exports in each state, optionally filtered for a profile or destination */
   export async function totals(
     where: { profileId?: string; destinationId?: string } = {}
@@ -91,8 +93,12 @@ export namespace ExportOps {
   export async function processPendingExportsForDestination(
     destination: Destination,
     limit = 100,
-    delayMs = 1000 * 60 * 5
+    delayMs = defaultExportProcessingDelay
   ) {
+    if (!delayMs || delayMs < defaultExportProcessingDelay) {
+      delayMs = defaultExportProcessingDelay;
+    }
+
     const app = await destination.$get("app");
     const { pluginConnection } = await destination.getPlugin();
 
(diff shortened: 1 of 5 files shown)


core/src/modules/ops/export.ts
```diff
@@ -9,6 +9,8 @@ import { CLS } from "../../modules/cls";
 import { Op } from "sequelize";
 
 export namespace ExportOps {
+  const defaultExportProcessingDelay = 1000 * 60 * 5;
```
bleonard, line 12: What's the behavior of this approach if an export is enqueued and it doesn't get processed within 5 minutes? Maybe there aren't enough workers or another long run comes up or the workers are stoped for devops reasons.

So the first one is still in the queue. Will this put another one in the queue or is there a queue lock in place and so this just, in effect, bumps the start time?

evantahler: We've got QueueLocks!  

It gets weird though with export batches - if the batch collection is different, the lock won't prevent the export from being included twice.  

I think this means the default delay should be higher - I'll bump that in the settings to be 6 hours

bleonard: We can fix that on the other side too then. In the beginning of the export jobs, do we check that they haven't been successfully completed? Then, if they are in the twice, the second one is a no-op. That narrows the window down to just the runtime.

evantahler: Nice idea!

evantahler: oh hey! We already do - https://github.com/grouparoo/grouparoo/blob/main/core/src/tasks/export/send.ts#L22-L25 and https://github.com/grouparoo/grouparoo/blob/main/core/src/tasks/export/sendBatch.ts#L25-L31

evantahler: With ^ in mind, I went back down to a 30 min delay

bleonard: Nice. if this still poses an issue, we'd en up full delayed_job and get a `processingAt` column

PR #1537, opened 2021-03-26
Add custom fields tests for Hubspot & Support Dates

pauloouriques:
- Add tests to increase coverage of number, date and text custom fields;
- Add date type mapping.


plugins/@grouparoo/hubspot/src/lib/export/exportProfile.ts
```diff
@@ -101,6 +101,7 @@ function formatVar(value) {
     return null;
   }
   if (value instanceof Date) {
+    value.setUTCHours(0, 0, 0, 0); //Must be midnight.
```
evantahler, line 104: I'm just clarifying what's happening here:
* All grouparoo Properties of type `date` (which are JS Dates) will be converted to be UTC Dates at 00:00.  
* This implies that there is no "timestamp" type property in Hubspot (or at least we don't support that yet)

pauloouriques: Correct.

PR #1543, opened 2021-03-27
Transaction lock for SQLite

bleonard:
We keep getting the `SQL_BUSY` issue when the background and UI are using the db. 

I tried several approaches including combinations of:
- journal_mode=WAL
- different types of transaction in UI and background
- knowing what actions were read and write
- redis locks to mutex the database when sqlite

The best approach seems to be:
- always make transactions (we were already doing that)
- know which are read-only and which write
- for the ones that write, use redis to get a lock on the database
- use an `IMMEDIATE` transaction so if it does fail, it does so right away (`DEFERRED` would wait until there was an `UPDATE`)
- prioritize UI updates over background tasks so the world seems responsive

A few things I'm curious about:
- how to write tests to make sure that we don't miss a `{ write: true }` on any action.
- can we do better than saying all Tasks are `{ write: true }`? The Export Tasks for example have plenty of time in the middle when they are connecting to Salesforce or whatever. Does that have to be in one big transaction or could it be two smaller ones before and after (or just one after?)


core/.npmignore
```diff
@@ -11,7 +11,7 @@ newrelic_agent.log
 *.retry
 coverage
 *.sqlite
-*.sqlite-journal
+*.sqlite-*
```
bleonard, line 14: there are other files in different modes

core/src/actions/plugins.ts
```diff
@@ -7,6 +7,7 @@ export class PluginsList extends CLSAction {
     this.name = "plugins:list";
     this.description =
       "I will return information about the active plugins on this server";
+    this.permission = { topic: "system", mode: "read" };
```
bleonard, line 10: re-using this notion for "read" even though it's not authenticated.

core/src/actions/setupSteps.ts
```diff
@@ -12,6 +12,11 @@ export class SetupStepsList extends AuthenticatedAction {
     this.outputExample = {};
   }
 
+  isWriteTransaction() {
+    // setupStep.performCheck() can do an update
+    return true;
```
bleonard, line 17: should the `permission` be `write` - I wasn't sure.

core/src/classes/tasks/clsTask.ts
```diff
@@ -9,7 +9,17 @@ export abstract class CLSTask extends Task {
...
   async run(inputs: TaskInputs, worker: any) {
     if (typeof this.runWithinTransaction === "function") {
-      return CLS.wrap(async () => this.runWithinTransaction(inputs, worker));
+      try {
+        return CLS.wrap(async () => this.runWithinTransaction(inputs, worker), {
+          write: true,
+        });
+      } catch (err) {
+        const message = (err?.message || "").toString();
+        if (message.match(/SQLITE_BUSY/)) {
```
bleonard, line 18: some resilience. 
is this the right way to reenqueue?

core/src/modules/cls.ts
```diff
@@ -31,23 +32,94 @@ export namespace CLS {
...
+
+  const wrapInternal: CLSWrapMethod = async (f, options = {}) => {
+    const { write } = options;
+    let runResponse: any;
+    let afterCommitJobs: Array<Function> = [];
+
+    const transOptions: any = {};
+    const dialect = api.sequelize.options.dialect;
+    if (dialect === "sqlite" && write) {
+      transOptions.type = "IMMEDIATE";
```
bleonard, line 78: this will fail right away if the db is busy

core/src/modules/cls.ts
```diff
@@ -31,23 +32,94 @@ export namespace CLS {
...
+  };
+
+  async function getLock(priority) {
+    // check more often in the ui to get the lock
+    const longEnough = 30 * 1000; // 30 seconds
+    const sleepTime = priority ? 1 : 100; // check more often if it's a priority
+    const ttl = longEnough + 1;
+    const maxAttempts = 300; // hard-coded in the locks.ts code
+    // might have to check more times, so start the count negative to get to 30 seconds
+    const attempts = 0 - longEnough / sleepTime + maxAttempts;
```
bleonard, line 99: less then ideal API for `waitForLock` but out of scope for this PR, so made this affordance.

core/src/actions/setupSteps.ts
evantahler: Yeah - `await setupSteps[i].performCheck()` will potentially mark a setup step as complete in the DB.  Alternatively, checking steps could be moved to a task maybe?  Doing it here made the UI more responsive to changes...

core/src/config/sequelize.ts
```diff
@@ -101,7 +101,7 @@ export const DEFAULT = {
       migrations: [join(__dirname, "..", "migrations")],
       storage, // only used for sqlite
       dialectOptions: { ssl },
-      transactionType: dialect === "sqlite" ? "EXCLUSIVE" : "DEFERRED",
+      transactionType: "DEFERRED",
```
evantahler, line 104: Back to the default!

evantahler (commented):
This is awesome!

To clarify: 
* We shouldn't use EXCLUSIVE transactions everywhere because that makes things really slow.  If we know that a task or action is read-only, DEFERRED is faster and non blocking

Questions:
* Did the WAL settings speed things up?  If coupled with this new approach, are things even speedier?
* Now that we know if an Action/Task is `write` or `read`, could you use an EXCLUSIVE transaction rather than the redis lock?  That might accomplish the same goal - one transaction active at a time.
  * > EXCLUSIVE is similar to IMMEDIATE in that a write transaction is started immediately. EXCLUSIVE and IMMEDIATE are the same in WAL mode, but in other journaling modes, EXCLUSIVE prevents other database connections from reading the database while the transaction is underway.


  * From https://sqlite.org/lang_transaction.html
* For testing, I wonder if there's a way in `NODE_ENV=test` to look at the stored up statements in a transaction before they are committed in the CLS helper. If there is, we could look for UPDATE/INSERT statements in and throw an error if the wrong permission was used

I think work coupled with https://www.pivotaltracker.com/story/show/177206079 to reduce the amount of requests the UI makes will get us in a really good spot.

evantahler:
Oh, and I think re-using the permission system for this purpose is OK for now!

bleonard:
There's the hope you and I seem to share that `EXCLUSIVE` will queue things up just like redis might do. That doesn't seem to play out in practice, so I added Redis to the mix. I haven't actually read anywhere where that happens. It fails immediately if it can't get the write token. I think `sequelize` just has some built-in retry behavior, so it felt like that (sleep and try again). So some degree that's what the Redis locks are doing but using more context to prioritize the UI ones.

`WAL` seemed to help, but was not fully sufficient to remove the errors. It still will not handle two writers at a time. `DEFERRED` seems ok to read while writing and once I added the locks, WAL didn't seem necessary. I was trying not to add N things at once, but I'm neutral to putting WAL back in.

evantahler:
Having a dependency on Redis to write to the database /feels/ wrong to me.   

Would it be possible to split this PR into 2?  
* Part 1: Use the permission system to indicate whether or not each action/task needs an IMMEDIATE (write) transaction or the read-only DEFERRED type of transaction
* Part 2: Use Redis to queue up IMMEDIATE transactions, really ensuring only one happens at a time

I'd want to merge in part 1 and  https://www.pivotaltracker.com/story/show/177206079 (reducing the number of requests the UI makes) and see if part 2 is needed.

bleonard:
Ok, that's fine if that's what you like. Based on my experience (currently on `npm start > 41.log`), that story will not help. Doing any write (change from settings page) from the UI while the single worker is also writing will cause `SQLITE_BUSY`.

bleonard:
Made this one that, in lieu of a Redis lock to maintain access, leverages retry much more to come as close as we can get. https://github.com/grouparoo/grouparoo/pull/1563

PR #1544, opened 2021-03-28
Fancier caches for merger

bleonard:


tools/merger/data/ci/publish/internal.yml.template
```diff
@@ -1,10 +0,0 @@
-  publish:
-    docker:
-      - image: circleci/node:14
-        auth:
-          username: $DOCKERHUB_USERNAME
-          password: $DOCKERHUB_PASSWORD
-    steps:
-      - run:
-          name: not publishing
-          command: echo "Not automatically publishing"
```
evantahler, line 10: Oops! Probably left over from an old template

evantahler (approved):
This is fancier!  Alphabetical chunks.

Question - why 4 buckets and not 2?  Room to grow?

tools/merger/data/ci/publish/internal.yml.template
bleonard: I also remove "internal" things.

bleonard:
I'm not sure the size that's correct, so i made them all about 10 each and room to grow ( I think maybe up to 15) so we won't have to think about it for awhile.

PR #1566, opened 2021-03-29
Use `ProfileProperty.startedAt` to more quickly find properties needing to be imported

evantahler:
Using the patterns introduced in https://github.com/grouparoo/grouparoo/pull/1529, this PR adds `ProfileProperty.startedAt` to make it faster to find which Profile Properties need to be imported and skip potentially re-importing the same Profile Property.  

Specifically, the `profileProperties:enqueue` task can now run faster, by re-enqueuing itself immediately to be sure that it finds all the pending Profile Properties made by the latest tick of the Run.  This has the effect of reducing the idle time between batch processing.  **_In testing with one worker, I'm seeing a ~2x speed increase of the initial run_**!


core/src/modules/ops/profileProperty.ts
```diff
@@ -73,6 +79,64 @@ export namespace ProfilePropertyOps {
...
+    property: Property,
+    limit = 100,
+    delayMs = defaultProfilePropertyProcessingDelay
+  ) {
+    if (!delayMs || delayMs < defaultProfilePropertyProcessingDelay) {
+      delayMs = defaultProfilePropertyProcessingDelay;
+    }
+
+    const source = await property.$get("source", { scope: null });
+    if (source.state !== "ready") return;
```
bleonard: maybe `return []` so this always returns an array.

core/src/modules/ops/profileProperty.ts
```diff
@@ -2,8 +2,14 @@ import { parsePhoneNumberFromString, CountryCode } from "libphonenumber-js/max";
 import { plugin } from "../../modules/plugin";
 import isEmail from "../validators/isEmail";
 import isURL from "validator/lib/isURL";
+import { ProfileProperty } from "../../models/ProfileProperty";
+import { Property } from "../../models/Property";
+import { CLS } from "../../modules/cls";
+import { Op } from "sequelize";
 
 export namespace ProfilePropertyOps {
+  const defaultProfilePropertyProcessingDelay = 1000 * 60 * 5;
```
evantahler, line 11: I wonder if this default should be exposed as a setting.  I think it's ok to be only internal for now as the batch size is already configurable, which is the biggest component to any delays that occur

core/src/modules/ops/profileProperty.ts
evantahler: https://github.com/grouparoo/grouparoo/pull/1566/commits/525cb2c2cf9a67abf8e699f9b380611f34cf5ab2

PR #1568, opened 2021-03-30
Add info level to Mailchimp email validation errors.

pauloouriques:
- in cases of email strict validation (e.g. foo@example.com) the error level is set to info.


plugins/@grouparoo/mailchimp/package.json
```diff
@@ -37,6 +37,7 @@
     "@types/node": "*",
     "actionhero": "25.0.8",
     "dotenv": "8.2.0",
+    "fs-extra": "9.1.0",
```
evantahler, line 40: Why was `fs-extra` added?  I don't see it in use in his PR.

plugins/@grouparoo/mailchimp/src/lib/shared/updateProfile.ts
```diff
@@ -130,8 +130,13 @@ export const updateProfile: UpdateProfileMethod = async ({
...
     });
   } catch (error) {
-    // sometimes there are multiple errors that mailchimp puts in error.errors (which are json objects)
+    const errorMessage = error?.response?.body?.detail || "";
+    // in case of strict validation cases the error must be set to info level.
+    if (errorMessage.includes("looks fake or invalid")) {
+      error.errorLevel = "info";
+    }
     if (error.errors) {
+      // sometimes there are multiple errors that mailchimp puts in error.errors (which are json objects)
```
evantahler, line 139: Is it possible that in the case where there are more than one errors (`error.errors`), one of them could also be the `looks fake or invalid` message from the Mailchimp API?

plugins/@grouparoo/mailchimp/__tests__/export/export-profile.ts
```diff
@@ -570,4 +571,21 @@ describe("mailchimp/exportProfile", () => {
       })
     ).rejects.toThrow();
   });
+
+  test("can add a user passing another invalid email", async () => {
```
evantahler: Can you please clarify this test a little more - What you are testing here is specifically if mailchimp thinks the email address is invalid, grouparoo's error level will be info. 

It might also help to check that the error's message matches what mailchimp returns, e.g.: `/looks fake or invalid/`

evantahler (changes requested):
Looking good!  It's a shame that there's nothing more certain (like an error code) from Mailchimp for this class of error, but I think the string message you are checking for is a good way to do this.  I *think* Mailchimp's API responses are always in english, so we should be OK. 

Just some minor changes requested for clarity and I think there's an un-used dependency.

plugins/@grouparoo/mailchimp/package.json
pauloouriques: It's been used inside nockHelper. I saw the error when I ran the tests with newNock = true.

evantahler (approved):
Thanks @pauloouriques!

PR #1569, opened 2021-03-30
Destination Sync Modes

pedroslopez:
This PR seeks to establish Sync Modes more globally so destinations can easily implement them. Some initial plugins have been upgraded to support these new sync modes to test viability: Pipedrive and Intercom.


core/src/actions/destinations.ts
```diff
@@ -118,6 +118,7 @@ export class DestinationCreate extends AuthenticatedAction {
       appId: { required: true },
       options: { required: false },
       mapping: { required: false, default: {} },
+      syncMode: { required: false },
```
evantahler, line 121: Maybe this is required here, when creating the destination

core/src/models/Destination.ts
```diff
@@ -43,6 +43,39 @@ export interface SimpleDestinationGroupMembership {
...
 
+const SYNC_MODES = ["Sync", "Additive", "Enrich"] as const;
+export type DestinationSyncMode = typeof SYNC_MODES[number];
+
+export interface DestinationSyncActions {
+  create: boolean;
+  update: boolean;
+  delete: boolean;
+  description: string;
+}
```
evantahler: I'd suggest not using the word "action" here (because actions ~= controllers to Grouparoo).  Perhaps "DestinationSyncMethods" or something like that?

core/src/models/Destination.ts
```diff
@@ -43,6 +43,39 @@ export interface SimpleDestinationGroupMembership {
 }
 export interface SimpleDestinationOptions extends OptionHelper.SimpleOptions {}
 
+const SYNC_MODES = ["Sync", "Additive", "Enrich"] as const;
```
evantahler: We generally tend to use lower-case words for this kind of thing

core/src/migrations/000062-addSyncModeToDestination.ts
```diff
@@ -0,0 +1,17 @@
+export default {
```
evantahler: Note before we merge this - I think another PR has migration `#000062` already.  We'll need to update this.

core/src/migrations/000062-addSyncModeToDestination.ts
```diff
@@ -0,0 +1,17 @@
...
+  up: async function (migration, DataTypes) {
```
evantahler: Check the notes in https://www.pivotaltracker.com/story/show/177553580 to see how to wrap this migration in a transaction

ui/ui-enterprise/pages/destination/[id]/edit.tsx
```diff
@@ -155,6 +155,26 @@ export default function Page(props) {
                 </Form.Control.Feedback>
               </Form.Group>
 
+              <Form.Group controlId="syncMode">
+                <Form.Label>Sync Mode</Form.Label>
+                <Form.Control
+                  as="select"
+                  required={true}
```
Comment (evantahler) [inline ui/ui-enterprise/pages/destination/[id]/edit.tsx]: Since it's required in the form, that's a good hint that it's required in the action

evantahler (changes requested):
Really strong start!  You've navigated Core well!  

I added some notes about some of the conventions we already are using in the app.  It doesn't mean that those conventions are correct to use going forward... so let's talk about any changes you might want to make!

core/src/actions/destinations.ts
evantahler: You can also add a test for the action to make sure it fails without `syncMode `

core/src/migrations/000062-addSyncModeToDestination.ts
evantahler: Lol... that's not a color github!

plugins/@grouparoo/intercom/src/lib/export-contacts/exportProfile.ts
```diff
@@ -386,20 +395,24 @@ function addCreationPayload(
...
   switch (creationMode) {
     case CreationMode.User:
     case CreationMode.Lead:
-    case CreationMode.Enrich:
       // no changes
       break;
     case CreationMode.Lifecycle:
       // make sure it's the right one
-      if (external_id) {
+      if (external_id && syncOperations.create) {
```
pedroslopez, line 407: When `creationMode` was set to `Enrich`, this would not run. Why is this the case? Does setting the role to user with an external id create a new contact? There's also a test that intentionally checks for this behavior:
https://github.com/grouparoo/grouparoo/blob/51c7f9e431a5d00e934aed9aa811cb7b4b593cb5/plugins/%40grouparoo/intercom/__tests__/integration/export-profile-no-create.ts#L340

@bleonard do you remember why this is the case?

core/src/actions/destinations.ts
pedroslopez: I tried making this required, but in the UI when the destination is created there's no input for these things like name and sync mode. This shows up in the edit page.

Instead, I'm adding a check when it transitions to ready state and checking if it's valid there.

Should it be required in the edit action? Name isn't required here either.

ui/ui-enterprise/pages/destination/[id]/edit.tsx
```diff
@@ -155,6 +155,31 @@ export default function Page(props) {
...
+                  {destination.syncModes.map((mode) => (
+                    <option key={mode.key} value={mode.key}>
+                      {mode.displayName} | {mode.description}
+                    </option>
+                  ))}
+                </Form.Control>
+                <Form.Text className="text-muted">
+                  {destination.syncModes.length > 0
+                    ? "How should this destination sync profiles?"
+                    : "This destination has not been updated to support sync modes."}
```
Comment (pedroslopez) [inline ui/ui-enterprise/pages/destination/[id]/edit.tsx]: Should we just completely hide the field in the UI?

Comment (evantahler) [reply ui/ui-enterprise/pages/destination/[id]/edit.tsx]: I think so, or at the very least, disable it.

core/src/actions/destinations.ts
pedroslopez: This is now just being checked on the BeforeSave hook if state is ready and the plugin has supported sync modes defined, mainly so it plays nicely with plugins that have not yet been upgraded.

core/src/migrations/000063-addSyncModeToDestination.ts
```diff
@@ -8,12 +8,12 @@ export default {
 
       // Pipedrive
       await migration.sequelize.query(
-        `UPDATE "destinations" SET "syncMode"='sync' WHERE "state"='ready' AND "type"='pipedrive-export'`
+        `UPDATE "destinations" SET "syncMode"='sync' WHERE "state"='ready' AND "type"='pipedrive-export' AND "locked" IS NULL`
```
pedroslopez: As it stands now, this does not do anything for code config. Since the syncMode is now required for this destination, when code config is validated (`grouparoo validate`) or when the server starts up, you will get an error saying "Sync mode is required for destination". The user then needs to update their code config accordingly.

Do we need to do anything else for code config?

core/__tests__/actions/destinations.ts
```diff
@@ -49,6 +49,7 @@ describe("actions/destinations", () => {
         name: "test destination",
         type: "test-plugin-export",
         appId: app.id,
+        syncMode: "sync",
       };
       const { error, destination } = await specHelper.runAction(
         "destination:create",
```
evantahler, line 55: You might want to also test that the data you get back in `destination:create` and `destination:view` include the sync mode.  This will in turn test that `Destination.apiData()` has the new syncMode included.

core/src/actions/destinations.ts
evantahler: Nice.  Can you then please add a pivotal tracker story (or a TODO in the code) to toggle this to `required` once all the plugins are upgraded.

core/src/bin/generate.ts
```diff
@@ -74,6 +74,12 @@ Commands:
         description:
           "For batch generators, what should we use for the high-water-mark?",
       },
+      "sync-mode": {
+        required: false,
+        letter: "s",
+        description:
+          "For destination generators, how will we sync to the destination?",
+      },
```
evantahler, line 82: Same TODO here with the action `destination:create` - this should one day become required, once all the plugins are updated.

core/src/migrations/000063-addSyncModeToDestination.ts
evantahler: Let's discuss this one a little more.  Here are some thoughts about the migration that I want to understand:

* Does this mean we'll need a migration for every plugin as they gain SyncModes? If so, that doesn't seem great...
* Should the default value be `skip`? 
* Can you do something fancy with defaults for each type of plugin based on what the old behavior was closest too?  We could add the syncMode column now for all plugins even though it won't be used yet.  
* Or, should we call this a breaking change (and move to v0.3) and default all the existing destinations to the least dangerous option (`additive`), with a note for our users to check if that's what they want?

core/src/migrations/000063-addSyncModeToDestination.ts
```diff
@@ -0,0 +1,64 @@
...
+        }
+        dest[opt.id][opt.key] = opt.value;
+        return dest;
+      }, {});
+
+      for (const destId in intercomDestinations) {
+        const dest = intercomDestinations[destId];
+        let syncMode;
+
+        if (dest.creationMode !== "Enrich" && dest.removalMode !== "Skip") {
```
evantahler: Should these be lower case?

core/src/models/Destination.ts
```diff
@@ -240,6 +300,46 @@ export class Destination extends LoggedModel<Destination> {
...
+    if (supportedSyncModes.length === 1) {
+      // If destination only supports one sync mode, use it
+      return supportedSyncModes[0];
+    } else if (supportedSyncModes.length === 0) {
+      // If destination does not support sync modes, let it create, update, and delete
+      return "sync";
+    }
+
+    return null;
+  }
```
evantahler: You can overload a Sequelize Model's getter and setting, which I think this is what you are trying to do.  In this way, you can then reference `destination.syncMode` directly.  

Check out how we modify `Import.data` with a custom getter and setter here to do some JSON parsing.  https://github.com/grouparoo/grouparoo/blob/main/core/src/models/Import.ts#L58-L67

core/src/models/Destination.ts
```diff
@@ -240,6 +300,46 @@ export class Destination extends LoggedModel<Destination> {
...
+
+  async validateSyncMode() {
+    if (this.state !== "ready") return;
```
evantahler: I'm reading that we we want to validate this as a part of becoming ready, with the `@BeforeSave` below.   I'd suggest that you remove this line so the validation method can be used more generically and then in the hook below we'll put the logic to check only if the instance is ready (or trying to become ready).

core/src/models/Destination.ts
```diff
@@ -440,6 +540,11 @@ export class Destination extends LoggedModel<Destination> {
     }
   }
 
+  @BeforeSave
+  static async validateSyncMode(instance: Destination) {
+    await instance.validateSyncMode();
```
evantahler, line 551: I think here is where you do the ready check:

```ts
if (instance.state === 'ready') await instance.validateSyncMode();
```

plugins/@grouparoo/intercom/src/lib/export-contacts/exportProfile.ts
```diff
@@ -287,14 +292,16 @@ async function deleteUser(
   newGroups = [];
   await updateTags(client, cacheData, user, oldGroups, newGroups);
 
+  if (!syncOperations.delete) {
+    throw new InfoError("Destination is not removing contacts.");
```
evantahler: 👍 on keeping this an `InfoError` - we don't want these types of errors to scare users / be retried.

ui/ui-enterprise/pages/destination/[id]/edit.tsx
```diff
@@ -155,6 +155,31 @@ export default function Page(props) {
                 </Form.Control.Feedback>
               </Form.Group>
 
+              {destination.syncModes.length > 0 ? (
```
Comment (evantahler) [inline ui/ui-enterprise/pages/destination/[id]/edit.tsx:158]: Nice - this won't show up in the UI unless the plugin supports syncModes

evantahler (commented):
Sorry for the delay!

Great work here - Lets' talk a little more about the migration.  There might be some assumptions we can make about our users to simplify it.

core/src/migrations/000063-addSyncModeToDestination.ts
pedroslopez: Nope! These are intercom's options, which were upper case

evantahler:
Discussion on April 5:
* Move migrations that are specific to each plugin into the plugin (sequelize supports this)
  * `{number}-plugin-{pluginName}-{description}` eg `000001-plugin-intercom-syncMode.ts`
* Code Config after the migration happens, will see an error (eg: `syncMode is required`) - this is OK for now
* `getSyncMode` is async, so it cannot be used as a sequelize getter
* make a new `core/modules/errors.ts` to contain `InfoError` and export it so all the plugins can use it
  * move `AuthenticationError` and `AutorizationError` there as well

core/src/models/Destination.ts
evantahler: Turns out you can't - async

core/src/modules/errors.ts
```diff
@@ -0,0 +1,27 @@
+export class InfoError extends Error {
```
evantahler: (idea, not blocking) If you wrap these in a namespace, you can then export all the Errors from `index.ts` - which would allow plugins in the future to use any of the errors:

```ts
exports namespace Errors {
  export class InfoError {}
}

```

Then in `index.ts` `export {Errors}  from '/path/to/file'`

Finally in plugins: 

```ts
import {Errors} from `@grouparoo/core`
throw new Errors.InfoError('Oh no!')
```

plugins/@grouparoo/app-templates/src/destination/templates.ts
```diff
@@ -1,7 +1,11 @@
-import { ConfigTemplate } from "@grouparoo/core";
+import { ConfigTemplate, DestinationSyncMode } from "@grouparoo/core";
 
 export class DestinationTemplate extends ConfigTemplate {
-  constructor(name: string, files: string[]) {
+  constructor(
+    name: string,
+    files: string[],
+    syncModes?: DestinationSyncMode[]
```
evantahler: You had mentioned you might want to leave `["sync"]` as the default.   I think an empty array is better - it's clear that this isn't implemented yet.

plugins/@grouparoo/pipedrive/src/migrations/000001-plugin-pipedrive-syncMode.ts
```diff
@@ -0,0 +1,14 @@
+export default {
+  up: async function (migration) {
+    await migration.sequelize.transaction(async () => {
+      // Set all existing instances to "sync" to maintain same behavior
+      await migration.sequelize.query(
+        `UPDATE "destinations" SET "syncMode"='sync' WHERE "state"='ready' AND "type"='pipedrive-export' AND "locked" IS NULL`
+      );
```
evantahler: Note to our future selves:
* We didn't modify those Apps configured via code-config, because as the application restarts immediately after this migration Grouparoo will then try to apply the code-config, and an error will be thrown about a missing `syncMode`.
* We don't want to modify people's code-config files - the error about a (now) required `syncMode` is the best we can do. 

To discuss: 
* Is this a breaking change (ie: starts the `v0.3.x` line of Grouparoo?

evantahler (approved):
This was a huge PR... and it looks good! 

Nice!  Plugins have migrations now.  This is going to be a very useful tool for later.

As a note: CI runs tests on both Postgres and SQLite versions of Grouparoo, so the migrations are working OK on both.

PR #1572, opened 2021-03-31
Disable logging to file without `GROUPAROO_LOGS_PATH` set

evantahler:
This PR disables logging to files unless `GROUPAROO_LOGS_PATH` is set.  It can be relative to your project root (`logs`) or absolute (`/var/log/grouparoo`).  Prior to this PR, log files were being stored within `node_modules/@grouparoo/core`, which wasn't great.

Google App Engine (https://github.com/grouparoo/app-example-gcp/pull/1) requires that no logs be written, as they have a strictly enforces read-only file system once the app is deployed. 

Logging to `stdout`/`stderr` will still continue by default


evantahler:
This is technically a breaking change, but I don't think that many users view the on-disk logs, as they are hard to find...

PR #1574, opened 2021-03-31
Mongo Query result issue

pauloouriques:
- Remove field _id when the object has more than one field.


evantahler (changes requested):
I think this makes sense for a first pass!  If you really wanted `_id` (maybe it's the primary key for your users), you can select just the `_id`, otherwise we assume you don't want it.

Can you add a test that shows this working?

evantahler:
Thanks!

PR #1577, opened 2021-04-01
UI Stats via Websocket

evantahler:
This PR replaces all the timer-based polling the UI used to use for system-wide heartbeat whose data is sent over websockets.  This will dramatically reduce the number of API requests that the UI makes to the API.  For example, every page was polling the API every 5 seconds, just to get the number resque errors.  The dashboard was making ~6 requests every 5 seconds, etc. To power this feature, there's a single Task (`system:status`) that calculates all the stats/counts the front-end may need on a rather fast frequency.  Each sample is stored in Redis for a while and broadcast over the websocket when calculated. You can see an example of the payload sent over the websocket [in the test](https://github.com/grouparoo/grouparoo/blob/e9951ab128fe671dc3c8ecea2700906094199d4f/core/__tests__/modules/status.ts#L77-L177).

In this way, we now will only calculate these stats once-per-cluster, no matter how many tabs/users are active on the site. Because we are caching these status samples in redis, the UI can also now pre-hydrate its pages by reading the most N recent value in the stats cache, and keep them locally.  In this way, when you first visit the Dashboard or Resque Overview, you can see previous samples. 

We are re-using the old `TelemetryReporters` (now renamed to `StatusReporters`) with new additions to calculate these metrics.

A new Setting `status-calculation-frequency-seconds` was added to control frequency of this status task, with a default of 10 seconds.  The status task does do quite a lot of SQL aggregation, and may be slow on busy/big clusters, so this speed may need to be turned down at some point.

---

The `status:private` action has been re-used as a way to get metrics on request.  This action is now protected by authentication. Updates in https://github.com/grouparoo/www.grouparoo.com/pull/369

---

Sub-Tasks:
- [x] Build status cache
- [x] create metric reporters
- [x] Task to calculate metrics
- [x] update UI to rely on WS only / remove all timed polling requests
  - [x] Navigation
  - [x] Dashboard
  - [x] Resque/Overview
- [x] pre-hydrate dashboard and resque pages from the stats cache
  - [x] Hide/Show sync issues with navigation pulls (likely solved with the above)
- [x] Allow the task's frequency to slow down if there are no active UI users / recent API requests (went with a setting instead)
- [x] Remove out-of-component chart storage and use samples instead
- [x] Ensure that WS chat-rooms are authenticated against the team member
- [x] tests


evantahler:
Note to future @evantahler - the way to change a tasks frequency in an already running Actionhero server would be to modify `api.tasks.tasks['taskName'].frequency`

PR #1600, opened 2021-04-06
Get MongoDB fields names automatically.

pauloouriques:
- Get all first-level fields straight from mongo;
- Get nested fields using sampling.

### DIFF
--- a/plugins/@grouparoo/mongo/src/lib/table-import/connection.ts
+++ b/plugins/@grouparoo/mongo/src/lib/table-import/connection.ts
@@ -15,16 +15,6 @@ export function getConnection() {
       "Import or update Profiles from a Mongo DB database collection.",
     tableOptionDescription: "The table to scan",
     tableOptionDisplayName: "Collection",
-    sourceOptions: {
-      options: [
-        {
-          key: "fields",
-          displayName: "Fields",
-          required: true,
-          description: "Comma separated list of fields",
-        },
-      ],
-    },
     getSampleRows,
     getColumns,
     getTables,
--- a/plugins/@grouparoo/mongo/src/lib/table-import/util.ts
+++ b/plugins/@grouparoo/mongo/src/lib/table-import/util.ts
@@ -4,7 +4,6 @@ import {
   FilterOperation,
   MatchCondition,
 } from "@grouparoo/app-templates/dist/source/table";
-import { SimpleSourceOptions } from "@grouparoo/core";
 
 export enum MongoAggregationMethod {
   Exact = "$eq",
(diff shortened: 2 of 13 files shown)


plugins/@grouparoo/mongo/src/lib/table-import/connection.ts
```diff
@@ -15,16 +15,6 @@ export function getConnection() {
...
-    sourceOptions: {
-      options: [
-        {
-          key: "fields",
-          displayName: "Fields",
-          required: true,
-          description: "Comma separated list of fields",
-        },
-      ],
-    },
```
evantahler, line 27: Since the mongo plugin is so new, I think it's OK to not worry about what a migration looks like here.  

* In the future, we would need to remove old options in the database or else the app is no longer valid
* We would also need to instruct people how to change the code-config they have

plugins/@grouparoo/mongo/src/lib/table-import/util.ts
```diff
@@ -193,3 +184,80 @@ function getNestedValue(entity, fieldName) {
...
+  connection: any,
+  tableName: string
+): Promise<Array<string>> {
+  const firstLevelFieldsResult = await connection.db
+    .collection(tableName)
+    .aggregate([
+      { $project: { keyValue: { $objectToArray: "$$ROOT" } } },
+      { $unwind: "$keyValue" },
+      { $group: { _id: null, keys: { $addToSet: "$keyValue.k" } } },
+    ])
```
evantahler, line 198: Here's the mongo magic!

plugins/@grouparoo/mongo/src/lib/table-import/util.ts
```diff
@@ -193,3 +184,80 @@ function getNestedValue(entity, fieldName) {
...
+async function getAllFields(
+  connection: any,
+  tableName: string,
+  firstLevelFields: Array<string>
+): Promise<Set<string>> {
+  let allFields = new Set<string>();
+  const docs = await connection.db
+    .collection(tableName)
+    .aggregate([{ $sample: { size: 10 } }, { $project: { _id: 0 } }])
+    .toArray();
```
evantahler, line 221: Will this always return the same 10 records or is it random?  
Postgres and MySQL return the same 10 each time so either is OK... just curious.

evantahler (approved):
This looks good!  The limitation of only using the first key-depth of objects in Mongo is OK for now... and the UX/DX is much better now! 

Screen Shot 2021-04-06 at 12 22 10 PM

plugins/@grouparoo/mongo/src/lib/table-import/util.ts
pauloouriques: it obtains N random documents of the given collection.

PR #1608, opened 2021-04-06
Fix type on Google Sheets Source template

seancdavis:


evantahler:
I'm using this branch to add tools to allow for plugins to test the the CLI commands, eg `grouparoo generate` and `grouparoo validate`

plugins/@grouparoo/google-sheets/__tests__/cli/integration.ts
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  test("the config files can be locally validated", async () => {
+    const { stdout, exitCode } = await runCliCommand(
+      "validate --local".split(" ")
+    );
+
+    expect(exitCode).toBe(0);
+    expect(stdout).toContain("Validation succeeded");
+  });
```
evantahler, line 58: This test works because we can `validate --local`.  This will skip testing against "real" data, meaning we don't need to use real usernames and passwords, but it will test that the plugins, sources, and properties are configured correctly, have the right names, and that the options required are present in the template.  

This test confirms the bug has been fixed.  Testing against real data is done in the other tests.

bleonard: Let's think about a way to do this for all plugins and templates in another PR sometime.

evantahler: Already updated https://www.pivotaltracker.com/story/show/177463643!

PR #1612, opened 2021-04-07
Refactor Salesforce Sync Modes to use standardized sync modes

pedroslopez:
Switches salesforce over from a custom destination option to the new standardized sync modes (#1569). 

Also, as part of this work we're ensuring foreignkey changes (email or id changes) work as intended in batched destinations (salesforce, marketo, pardot):
- The contact that would be updated is randomized based on the order that the destination API decides to return results. This PR fixes this on the batch-helper by first fetching the old set of FKs, and _then_ fetching the new set of FKs.
- If deleting a contact that does not exist in the destination, an error was thrown to indicate grouparoo to try again later. This has been changed to only throw an `InfoError` so it will not be retried, to match behavior across other plugins.


plugins/@grouparoo/app-templates/src/destination/batch/exportProfiles.ts
```diff
@@ -450,10 +456,21 @@ async function lookupDestinationIds(
...
     }
 
-    const { foreignKeyValue, oldForeignKeyValue } = exportedProfile;
+    let foreignKeyValue: string;
+    switch (fkType) {
+      case "new":
+        foreignKeyValue = exportedProfile.foreignKeyValue;
+        break;
+      case "old":
+        foreignKeyValue = exportedProfile.oldForeignKeyValue;
```
bleonard, line 465: If this case if they are the same, you don't have to do it.
then note that the total could be zero.

plugins/@grouparoo/salesforce/src/migrations/000001-plugin-salesforce-syncMode.ts
```diff
@@ -0,0 +1,49 @@
+export default {
+  up: async function (migration) {
```
bleonard, line 4: Let's be sure to test this well. It's critical that Salesforce works as expected here. People have them configured just how they want them.

plugins/@grouparoo/app-templates/src/destination/batch/exportProfiles.ts
pedroslopez: `oldForeignKeyValue` is only set if they are different. See `assignForeignKeys()`:
https://github.com/grouparoo/grouparoo/blob/ce6dc7b70ccd6a77fa654f373718b29cf394b730/plugins/%40grouparoo/app-templates/src/destination/batch/exportProfiles.ts#L571-L577

bleonard: ah ha. perfect.

plugins/@grouparoo/salesforce/src/migrations/000001-plugin-salesforce-syncMode.ts
```diff
@@ -0,0 +1,100 @@
...
+            break;
+          case "additive":
+            syncMode = "Additive";
+            break;
+          default:
+            continue;
+        }
+
+        newOptions.push({
+          id: `opt_${crypto.randomBytes(16).toString("hex")}`,
```
evantahler, line 75: rather likely to be unique! 👍

evantahler (approved):

PR #1619, opened 2021-04-07
Fix `@grouparoo/postgres` date parsing

evantahler:
This PR fixes a bug that fail to parse Postgres timestamps (without zone) that had decimal second precision.


plugins/@grouparoo/postgres/src/lib/connect.ts
```diff
@@ -46,25 +46,15 @@ export const connect: ConnectPluginAppMethod = async ({ appOptions }) => {
...
-  }
-  var zone: any = TIME_ZONE.exec(isoDate.split(" ")[1]);
-  if (!zone) {
-    return false;
-  }
-  return true;
+const TIME_ZONE = /([Z+-])(\d{2})?:?(\d{2})?:?(\d{2})?/;
+export function formatInUtcDefault(text: string) {
+  let hasTZ = false;
+  if (text.length > 19 && text.match(TIME_ZONE)) hasTZ = true;
```
evantahler: The regex is mad if the string is normal? `YYYY-MM-DD hh:mm:ss`

plugins/@grouparoo/postgres/src/lib/connect.ts
```diff
@@ -46,23 +46,28 @@ export const connect: ConnectPluginAppMethod = async ({ appOptions }) => {
...
+
+export function formatInUtcDefault(text: string) {
+  if (!text) return null;
   const zone = timeZoneOffset(text);
-  if (!zone) return parseDate(text + ".000Z");
-  return parseDate(text);
+  let date: Date = parseDate(text);
+
+  if (!zone) {
+    const dateInUTC: Date = parseDate(text + "+00");
```
evantahler, line 59: This reverts a change made in #1541 and goes back to using `+00` to add a zone, rather than `.000` - which likely caused the problem if there already was a `.` in the timestamp.  

Now our test suite has some timestamps with a `.` so we'll notice this breakage in the future.

evantahler: Also added a fallback to parse the un-modified string if our appending of the timezone didn't work for some reason - you may get a date in the wrong zone, but at least you'll get something back in this unlikely case

PR #1623, opened 2021-04-08
Add sync mode to Mailchimp.

pauloouriques:


plugins/@grouparoo/mailchimp/__tests__/data/emails.txt
```diff
@@ -1 +1 @@
-{"email":"pedyv8x5cipvk1t4qmtf@grouparoo.com","alternativeEmail":"5pg2xobzqxfuhuumi8dryf@grouparoo.com","otherEmail":"s49jwf4h7cqfh1mxr3cvpm@grouparoo.com","newEmail":"w5era252i4l9ug3v5xzw04@grouparoo.com","brandNewEmail":"o3vmerpx8ntha4nwi2wt6s@grouparoo.com","nonexistentEmail":"2wjkro7skvg5jxypxfm9uo@grouparoo.com"}
```
pedroslopez, line 1: Remember to run prettier here to format the fixtures

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
```diff
@@ -51,11 +57,12 @@ export const exportProfile: ExportProfilePluginMethod = async ({
     } catch (error) {
```
pedroslopez, line 64: We should probably check the error for a code or string to make sure this is why the error was thrown. There may be a couple cases here that are slightly different, such as when the oldEmail does not exist in the destination.

plugins/@grouparoo/mailchimp/src/migrations/000001-plugin-mailchimp-syncModes.ts
```diff
@@ -0,0 +1,14 @@
+export default {
+  up: async function (migration) {
+    await migration.sequelize.transaction(async () => {
+      // Set all existing instances to "sync" to maintain same behavior
+      await migration.sequelize.query(
+        `UPDATE "destinations" SET "syncMode"='sync' WHERE "state"='ready' AND ("type"='mailchimp-export' OR "type"='mailchimp-export-id') AND "locked" IS NULL`
```
pedroslopez: This migration is setting the mode to "sync" for both `mailchimp-export` and `mailchimp-export-id`, but `mailchimp-export-id` only supports "enrich" mode.

plugins/@grouparoo/mailchimp/src/migrations/000001-plugin-mailchimp-syncModes.ts
```diff
@@ -0,0 +1,14 @@
...
+      );
+    });
+  },
+
+  down: async function (migration) {
+    // nothing to do
```
pedroslopez: I'm going to change this on the others that say `// nothing to do`, we should actually set the syncMode column to `null` on these destinations.

plugins/@grouparoo/mailchimp/src/initializers/plugin.ts
```diff
@@ -34,8 +34,16 @@ export class Plugins extends Initializer {
         MailchimpAppTemplate,
         MailchimpSourceTemplate,
         MailchimpPropertyTemplate,
-        MailchimpEmailDestinationTemplate,
-        MailchimpIdDestinationTemplate,
+        new DestinationTemplate(
+          "mailchimp:email",
+          [path.join(templateRoot, "destination", "email", "*.template")],
+          ["sync", "additive", "enrich"]
```
pedroslopez: Just in case we change these later, you could export the supported sync modes from where you define the connections and use them here. See facebook for an example:

https://github.com/grouparoo/grouparoo/blob/facebook-sync-modes/plugins/%40grouparoo/facebook/src/initializers/plugin.ts#L7-L10

https://github.com/grouparoo/grouparoo/blob/facebook-sync-modes/plugins/%40grouparoo/facebook/src/initializers/plugin.ts#L32-L36

plugins/@grouparoo/mailchimp/src/initializers/plugin.ts
```diff
@@ -34,8 +34,16 @@ export class Plugins extends Initializer {
...
+        ),
```
pedroslopez, line 49: We can now pass another argument to the `DestinationTemplate` constructor that specifies the default sync mode. You'll probably need to rebase on `feat/sync-modes` to be able to use it.

https://github.com/grouparoo/grouparoo/blob/feat/sync-modes/plugins/%40grouparoo/pipedrive/src/initializers/plugin.ts#L40

plugins/@grouparoo/mailchimp/src/initializers/plugin.ts
```diff
@@ -10,13 +10,13 @@ import emailDestination from "../lib/export/connection";
 import idDestination from "../lib/export-id/connection";
 import importSource from "../lib/import/connection";
 
+const templateRoot = path.join(__dirname, "..", "..", "public", "templates");
```
pedroslopez: To keep things tidy we might want to move variable definitions after all the imports

plugins/@grouparoo/mailchimp/src/initializers/plugin.ts
```diff
@@ -1,6 +1,6 @@
 import path from "path";
 import { Initializer } from "actionhero";
-import { plugin } from "@grouparoo/core";
+import { DestinationSyncMode, plugin } from "@grouparoo/core";
```
pedroslopez, line 3: `DestinationSyncMode` doesn't look like is being used in this file

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
```diff
@@ -41,6 +42,11 @@ export const exportProfile: ExportProfilePluginMethod = async ({
...
       oldProfileProperties["email_address"]
     );
+    if (!syncOperations.update) {
+      throw new Errors.InfoError(
+        "Destination sync mode does not allow updating existing profiles."
+      );
+    }
     try {
       // this will update the email of existing member only if the new email (email_address) do not exists yet.
       await client.put(`/lists/${listId}/members/${oldMailchimpId}`, {
```
pedroslopez, line 59: I'd like to understand how this behaves when OLD email does not exist. For example, if the old email is `email1@demo.com` and new email is `email2@demo.com`... What does the api return if `oldMailchimpId` does not exist in the destination? Will it error/404?

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
```diff
@@ -41,6 +42,11 @@ export const exportProfile: ExportProfilePluginMethod = async ({
     const oldMailchimpId = generateMailchimpId(
       oldProfileProperties["email_address"]
     );
+    if (!syncOperations.update) {
```
pedroslopez: It's important to know that even though oldEmail !== newEmail, it doesn't mean that they actually exist in the destination. At this point I don't think we know for sure that it's going to be an `update`. If oldEmail and newEmail don't exist, it would cause a creation instead.

plugins/@grouparoo/mailchimp/src/lib/shared/updateProfile.ts
```diff
@@ -17,6 +22,7 @@ export interface UpdateProfileMethod {
 }
 
 export const updateProfile: UpdateProfileMethod = async ({
+  syncOperations,
```
pedroslopez, line 23: I think because we're passing `syncOperations` here we can remove `noDelete` and `noCreate`. `syncOperations` should be enough to determine what we should do.

plugins/@grouparoo/mailchimp/src/lib/shared/updateProfile.ts
```diff
@@ -34,12 +40,19 @@ export const updateProfile: UpdateProfileMethod = async ({
...
-
   if (toDelete) {
     if (noDelete) {
       return { success: true };
     }
-    await deleteMember(client, listId, mailchimpId, oldGroups, newGroups);
+    await deleteContactOrClearGroups(
+      client,
+      listId,
+      mailchimpId,
```
pedroslopez, line 43: What happens if there's an email change _and_ a `toDelete`? The code in exportProfile that deals with changing IDs only runs if `toDelete` is `false`, which means that when you get to this point you might be deleting based on the new email instead of the old one.

plugins/@grouparoo/mailchimp/src/lib/export-id/exportProfile.ts
```diff
@@ -24,6 +25,7 @@ export const exportProfile: ExportProfilePluginMethod = async ({
   }
 
   return updateProfile({
+    syncOperations,
     client,
     listId,
     mailchimpId,
```
pedroslopez, line 31: What happens if the profile's ID changes? I'm assuming that changing the ID isn't supported, correct?

pedroslopez (changes requested):
Nice start! There are a couple things to check out, mainly with regards to handling foreign key changes. We should probably add some tests for these. Some interesting cases come up when changing email when either old or new emails don't already exist in the destination. It's also important to check if changing emails at the same time as a `toDelete` works as expected.

plugins/@grouparoo/mailchimp/__tests__/export/export-profile.ts
```diff
@@ -428,9 +464,27 @@ describe("mailchimp/exportProfile", () => {
         expect.objectContaining({ name: listThree.toLowerCase() }),
       ])
     );
+  });
 
-    const oldUser = await getUser(emails["alternativeEmail"]);
-    expect(oldUser).toBe(null);
+  test("can not delete a user if sync mode does not allow it but should remove the tags", async () => {
```
evantahler: For this test, we should assert that we are only removing the tags we know about (if possible).  eg: give the profile some other random tag, and ensure that it still exists after the sync

pedroslopez:
Note: we have now merged #1569 into main, so you'll have to rebase to fix these conflicts. You'll need to rebase onto main and only apply your commits, since the others have been squashed down:
1. `git rebase --onto main ef655d0 d6dfd0f`. The first hash refers to the commit right before where you started to implement sync modes and the second refers to your last commit.
2. `git push origin HEAD:mailchimp-sync-modes -f` to force push the detached HEAD to this branch

Then, to make your local branch match this one:
1. `git checkout mailchimp-sync-modes`
2. `git fetch origin`
3. `git reset --hard origin/mailchimp-sync-modes`

**Very important note**: If you have unsaved changes you should `stash` them before doing these steps as they will be lost.

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
pauloouriques: nope, it returns 200 since the method allows you to add or update a list member.

plugins/@grouparoo/mailchimp/src/lib/export-id/exportProfile.ts
pauloouriques: correct.

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
pauloouriques: nice catch!

plugins/@grouparoo/mailchimp/src/initializers/plugin.ts
```diff
@@ -34,8 +40,16 @@ export class Plugins extends Initializer {
         MailchimpAppTemplate,
         MailchimpSourceTemplate,
         MailchimpPropertyTemplate,
-        MailchimpEmailDestinationTemplate,
-        MailchimpIdDestinationTemplate,
+        new DestinationTemplate(
+          "mailchimp:email",
+          [path.join(templateRoot, "destination", "email", "*.template")],
+          emailSupportedSyncModes
```
pedroslopez: Check out some of the other plugins to see how to pass default sync mode here as well:

https://github.com/grouparoo/grouparoo/blob/def64e8a82a673a04d7a84e4e88614c699c39e78/plugins/%40grouparoo/hubspot/src/initializers/plugin.ts#L40

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
```diff
@@ -49,19 +59,35 @@ export const exportProfile: ExportProfilePluginMethod = async ({
         merge_fields: { email_address },
       });
     } catch (error) {
-      // Another user with the new email address (email_address) already exists,
-      // so we need to delete the old one and let the updateProfile reuse the existing one.
-      await deleteMember(client, listId, oldMailchimpId, oldGroups, newGroups);
+      if (error.errors!.length > 0) {
```
pedroslopez: We should probably `throw` if the error is not the one we expect

plugins/@grouparoo/mailchimp/src/migrations/000001-plugin-mailchimp-syncModes.ts
```diff
@@ -0,0 +1,37 @@
...
+  },
+
+  down: async function (migration) {
+    await migration.bulkUpdate(
+      "destinations",
+      {
+        syncMode: null,
+      },
+      {
+        type: "mailchimp-export-id",
```
pedroslopez: not blocking, but just so you know you can also pass an array to `type` and it will be converted to an "IN" statement in sql

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
pedroslopez: In that case, I think there may be some interesting behavior going on with email changes + `toDelete`.

- If the old email and new email do not exist, someone would be created by this PUT request and then immediately deleted by the updateProfile call. This should probably just not do anything instead of creating/deleting someone.
- If the old email does not exist but the new email does exist, I'm not sure what would happen. I'm assuming this would fail because the new email already exists, and then the code in the `catch` block would try to clear groups on the `oldMailchimpId`, which is nonexistent and could cause an error, not sure what the api returns.

Could you add tests for these cases with both `toDelete: true` and `toDelete: false`?

pedroslopez (changes requested):

plugins/@grouparoo/mailchimp/src/initializers/plugin.ts
```diff
@@ -27,15 +33,27 @@ export class Plugins extends Initializer {
...
+          "mailchimp:email",
+          [path.join(templateRoot, "destination", "email", "*.template")],
+          emailSupportedSyncModes,
+          defaultSyncMode
+        ),
+        new DestinationTemplate(
+          "mailchimp:id",
+          [path.join(templateRoot, "destination", "id", "*.template")],
+          idSupportedSyncModes,
+          defaultSyncMode
```
pedroslopez: `defaultSyncMode` value is `sync`, but `id` does not support it (only enrich). You can remove `defaultSyncMode` in this case. If there is only one supported sync mode, it will use that as default.

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
```diff
@@ -49,19 +61,48 @@ export const exportProfile: ExportProfilePluginMethod = async ({
...
+      } else if (
+        error.message &&
+        error.message.match(
+          /is already a list member. Use PUT to insert or update list members./i
+        )
+      ) {
+        toThrow = false;
+      }
+      if (toThrow) {
+        console.log(error);
```
pedroslopez: i think we can remove this log

plugins/@grouparoo/mailchimp/src/lib/export/exportProfile.ts
```diff
@@ -34,13 +33,26 @@ export const exportProfile: ExportProfilePluginMethod = async ({
...
   ) {
     const oldMailchimpId = generateMailchimpId(
       oldProfileProperties["email_address"]
     );
+    const oldUserResponse = await getUser(client, listId, oldMailchimpId);
+    const userResponse = await getUser(client, listId, mailchimpId);
+    if (toDelete && !oldUserResponse && !userResponse) {
+      return { success: true };
+    }
+    if (!userResponse && !syncOperations.create) {
```
pedroslopez: it might not be a create if `oldUserResponse` is found and `userResponse` is not found. (In this case the old user is updated). I think a similar situation could happen with line 51 (checking if it's an update)

I would suggest doing something like `const foundUser = userResponse || oldUserResponse` and then using this new variable to do the comparison.

pedroslopez (changes requested):
Looking good!  Thanks for adding the tests.
Just a couple small comments and we should be good to go :)

pedroslopez (approved):

plugins/@grouparoo/mailchimp/src/lib/shared/updateProfile.ts
```diff
@@ -142,7 +149,7 @@ export const updateProfile: UpdateProfileMethod = async ({
         .join(", ")}`;
 
       // just in case of one of the multiple errors message is the one related to the email strict validation.
-      if (error.message.includes("looks fake or invalid")) {
+      if (error.message!.includes("looks fake or invalid")) {
```
evantahler: Perhaps `error.message?.includes`?

evantahler (changes requested):

plugins/@grouparoo/mailchimp/src/lib/shared/updateProfile.ts
pauloouriques: yes. ty

evantahler (approved):

plugins/@grouparoo/mailchimp/__tests__/export/export-profile.ts
```diff
@@ -420,17 +457,41 @@ describe("mailchimp/exportProfile", () => {
...
+        toDelete: true,
+      })
+    ).rejects.toThrow(/sync mode does not allow removing/);
+    user = await getUser(emails["otherEmail"]);
+    expect(user).not.toBe(null);
+    expect(user["email_address"]).toBe(emails["otherEmail"]);
+    expect(user["tags"].length).toEqual(1);
+    expect(user["tags"]).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({ name: listFour.toLowerCase() }),
```
pedroslopez, line 495: hmm, after the latest changes I would expect this test to fail, but it seems to be still passing- it should just throw and not remove anything anymore. we should probably first check out why this isn't failing, and then update it as needed to reflect the new intended behavior.

pauloouriques: hmm that's true

plugins/@grouparoo/mailchimp/__tests__/export/export-profile.ts
```diff
@@ -468,7 +468,7 @@ describe("mailchimp/exportProfile", () => {
     );
   });
 
-  test("can not delete a user if sync mode does not allow it but should remove the tags", async () => {
+  test("cannot delete a user if sync mode does not allow it but should remove the tags", async () => {
```
pedroslopez:
```suggestion
  test("cannot delete a user if sync mode does not allow it", async () => {
```

pedroslopez (changes requested):
Last change!

plugins/@grouparoo/mailchimp/__tests__/export/export-profile.ts
pauloouriques: my bad, sorry

pedroslopez (approved):

PR #1625, opened 2021-04-08
Clean up Mongo package-lock.

pauloouriques:


evantahler:
Thanks @pauloouriques!

PR #1627, opened 2021-04-09
Add sync modes for Marketo

pedroslopez:
Don't merge until #1612 is merged. It includes some changes to the batch helper that this relies on.

### DIFF
--- a/plugins/@grouparoo/marketo/src/migrations/000001-plugin-marketo-syncMode.ts
+++ b/plugins/@grouparoo/marketo/src/migrations/000001-plugin-marketo-syncMode.ts
@@ -2,8 +2,16 @@ export default {
   up: async function (migration) {
     await migration.sequelize.transaction(async () => {
       // Set all existing instances to "sync" to maintain same behavior
-      await migration.sequelize.query(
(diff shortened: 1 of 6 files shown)


plugins/@grouparoo/marketo/src/migrations/000001-plugin-marketo-syncMode.ts
```diff
@@ -0,0 +1,23 @@
+export default {
+  up: async function (migration) {
+    await migration.sequelize.transaction(async () => {
+      // Set all existing instances to "sync" to maintain same behavior
+      await migration.sequelize.query(
```
evantahler: General question: Why does the `up` use a custom query and the `down` use `migration.bulkUpdate`?

pedroslopez: Because I didn't know that the [queryInterface](https://sequelize.org/master/manual/query-interface.html) was a thing when I wrote the `up` part hahaha

I made it use `bulkUpdate` on both to be a bit more consistent 🤷🏽‍♂️

PR #1628, opened 2021-04-09
Do not fail tasks if primary objects cannot be found

evantahler:
This PR relaxes many of Grouparoo's tasks so that if the primary object the task is operating on cannot be found, the task exits gracefully.  For example, before this PR if `export:send` was passed an export that was no longer present (because the destination was deleted) the task would fail and present as a failure within the resque dashboard.

### DIFF
--- a/core/src/tasks/profile/destroyEvents.ts
+++ b/core/src/tasks/profile/destroyEvents.ts
@@ -9,19 +9,18 @@ export class ProfileDestroyEvents extends CLSTask {
     this.frequency = 0;
     this.queue = "events";
     this.inputs = {
-      id: { required: true },
+      profileId: { required: true },
     };
   }
 
   async runWithinTransaction(params) {
-    const { id } = params;
     const limit = 1000;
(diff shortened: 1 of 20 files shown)


core/src/tasks/profile/destroyEvents.ts
```diff
@@ -9,19 +9,18 @@ export class ProfileDestroyEvents extends CLSTask {
     this.frequency = 0;
     this.queue = "events";
     this.inputs = {
-      id: { required: true },
```
bleonard, line 12: is there anything we need to put in place to transition here?

bleonard: For example, there are things queue'd up with the old inputs.

evantahler: I think this is safe - There's no automated task that will destroy a profile besides merging them - from events.  It's rather unlikely folks will have one of these tasks queued up when they upgrade.

PR #1630, opened 2021-04-09
`export:enqueue` should not re-attempt to send complete exports

evantahler:
This PR prevents a bug in which old, already-sent exports were constantly being re-considered to be sent by the `export:enqueue` task.   The effect of this was an ever-changing `export.startedAt` timestamps for all complete exports.

### DIFF
--- a/core/src/modules/ops/export.ts
+++ b/core/src/modules/ops/export.ts
@@ -106,6 +106,8 @@ export namespace ExportOps {
 
     _exports = await Export.findAll({
       where: {
+        completedAt: null,
+        errorMessage: null,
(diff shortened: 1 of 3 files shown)


core/src/modules/ops/export.ts
```diff
@@ -106,6 +106,8 @@ export namespace ExportOps {
 
     _exports = await Export.findAll({
       where: {
+        completedAt: null,
+        errorMessage: null,
```
bleonard, line 110: @andyjih  this is a fix we jsut did that would prevent old, failed exports from re-processing.

PR #1631, opened 2021-04-09
`migrator` command

evantahler:
This PR adds the `./bin/migrator` command (internal tool).  With this tool you can step up and down by 1 migration.

> Run me as if I were a bash program eg: `./bin/migrator up [name]` or `./bin/migrator down <name>`

```
$ ./bin/migrator up

== 000062-addProfilePropertiesStartedAt: migrating =======
000062-addProfilePropertiesStartedAt migrating
Executing (784d5dbb-c403-47b2-bb9d-868f15f0dfd3): START TRANSACTION;
Executing (784d5dbb-c403-47b2-bb9d-868f15f0dfd3): ALTER TABLE "public"."profileProperties" ADD COLUMN "startedAt" TIMESTAMP WITH TIME ZONE;
Executing (784d5dbb-c403-47b2-bb9d-868f15f0dfd3): COMMIT;
== 000062-addProfilePropertiesStartedAt: migrated (0.047s)

000062-addProfilePropertiesStartedAt migrated
✅ Done
```

```
$ ./bin/migrator down 000062-addProfilePropertiesStartedAt

== 000062-addProfilePropertiesStartedAt: reverting =======
000062-addProfilePropertiesStartedAt reverting
Executing (91abd347-541e-441a-bb82-14971cfb3582): START TRANSACTION;
Executing (91abd347-541e-441a-bb82-14971cfb3582): ALTER TABLE "public"."profileProperties" DROP COLUMN "startedAt";
Executing (91abd347-541e-441a-bb82-14971cfb3582): COMMIT;
== 000062-addProfilePropertiesStartedAt: reverted (0.045s)

000062-addProfilePropertiesStartedAt reverted
✅ Done
```

No special config is needed, and it works against the normal Grouparoo configuration ENV variables and `/core/config/sequelize`.

This tool is configured to run core's `src` migrations (eg: TS) vs the `dist` migrations (JS), however it will run the `dist` directory of plugin migrations. 

Notes:
* Because this is a migration tool, we log the SQL our migrations generate!
* `./bin/migrator up` needs to be run from the working directory of `grouparoo/core` (not within bin), as I'm using a relative shebang for ts_node
* As we are soon going to have multiple migration directories (from plugins - https://github.com/grouparoo/grouparoo/pull/1569) if you are going `down`, you need to provide a migration name so we know which migration you want to revert.  You can optionally provide a migration to apply when going `up`, otherwise the first pending migration will be applied.
  * migration names are the filename, minus the extension, eg `000062-addProfilePropertiesStartedAt`


evantahler:
Now with 100% more SQL!

evantahler:
@pedroslopez I think this is in a good place to support multiple migration directories now.  Check out the notes in the description about the TS/JS split though.  This happens because I'm following the configuration in `config/seqeulize.ts`, which is "local" and then will always be the "dist" directories of the plugins.

pedroslopez:
I'm having a few issues with this:
1. Running it from the core directory (`./bin/migrator`) causes an endless spam of `./bin/migrator: line 3: /Applications: is a directory`:
```
core plopez$ ./bin/migrator up test
./bin/migrator: line 3: /Applications: is a directory
./bin/migrator: line 3: /Applications: is a directory
./bin/migrator: line 3: /Applications: is a directory
./bin/migrator: line 3: /Applications: is a directory
./bin/migrator: line 3: /Applications: is a directory
...
```
I was able to get it to run by using `./node_modules/.bin/ts-node-script ./bin/migrator up test`. Might be something with my local setup 🤷🏽‍♂️

2. Stepping up or down migrations from core works, but this does not seem to be the case for plugins. Running `up` with the name of a migration defined in a plugin returns `✅ Done`, but nothing is excecuted. This doesn't seem to actually be running anything, as passing any random value as the migration name returns the same thing. `down`, on the other hand, does give me an error: `❌ could not find migration `000001-plugin-pipedrive-syncMode` in the migration directories`.

Looking into this a bit more, I found that the `getPluginManifest()` function we're calling in the sequelize config returns an empty array when running the migrator command (when running grouparoo from `/apps/staging-enterprise` this is not the case and properly returns all plugins). I'm not sure if this means we need to revisit how we're loading plugin migrations or if it's due to how the config is being initialized.

evantahler:
Thanks @pedroslopez! I think I've got fixes here https://github.com/grouparoo/grouparoo/pull/1642

PR #1632, opened 2021-04-10
`export:send` tasks should not rollback transaction if there's a returned error

evantahler:
The two tasks that send the exports to destinations, `export:send` and `exports:sendBatch` previously were wrapped entirely in a transaction, and would eventually `throw` any error we got back from the destination to retry the task. 
However, with our new transaction safety, that `throw` would rollback any changes that occurred in the task - most specifically the error messages would not be appended to the Export! 

This had the negative side-effect of exports which should be marked as failed being retried over and over again as they were found again by `export:enqueue` after N minutes (introduced in https://github.com/grouparoo/grouparoo/pull/1529).

### DIFF
--- a/core/src/tasks/export/send.ts
+++ b/core/src/tasks/export/send.ts
@@ -1,9 +1,12 @@
-import { RetryableTask } from "../../classes/tasks/retryableTask";
+import { Task } from "actionhero";
 import { Destination } from "../../models/Destination";
 import { Export } from "../../models/Export";
+import { App } from "../../models/App";
+import { DestinationOps } from "../../modules/ops/destination";
 import { CLS } from "../../modules/cls";
+import { AsyncReturnType } from "type-fest";
 
-export class ExportSend extends RetryableTask {
+export class ExportSend extends Task {
   constructor() {
     super();
     this.name = "export:send";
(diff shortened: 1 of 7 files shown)


core/src/tasks/export/send.ts
```diff
@@ -16,31 +19,37 @@ export class ExportSend extends RetryableTask {
...
         return CLS.enqueueTaskIn(
-          retryDelay,
+          response.retryDelay,
           "export:send",
           params,
           `exports:${app.type}`
         );
       } else {
-        // auto retry
-        throw error;
```
evantahler, line 43: This was the problem - if the plugin threw an error (vs returning errors), we'd get here and roll everything back.  The solution was to move this outside of the transaction.

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -117,11 +117,3 @@ function formatVar(value) {
...
   }
 }
-
-async function getUser(client, email): Promise<any> {
-  const userResponse = await client.users.get({ email });
-  if ("user" in userResponse) {
-    return userResponse.user;
-  }
-  return null;
-}
```
evantahler, line 127: Method was unused

PR #1644, opened 2021-04-12
Refactor Facebook destination to use standardized sync modes

pedroslopez:


plugins/@grouparoo/facebook/src/migrations/000001-plugin-facebook-syncMode.ts
```diff
@@ -0,0 +1,94 @@
...
+      const newOptions = [];
+      for (const dest of destinations) {
+        let syncMode: string;
+        switch (dest.syncMode) {
+          case "sync":
+            syncMode = "Sync";
+            break;
+          case "additive":
+            syncMode = "Additive";
+            break;
```
evantahler, line 63: Good catch with the case-change!

evantahler (approved):

PR #1653, opened 2021-04-12
Move Auth middlewares into transactions

evantahler:


bleonard:
The problem with this middleware is that it can be in any action. Assuming it writes, though I can't quite see where that is happening.
The SQLite situation works by knowing which ones only read and blocking on the ones that write. It's actually less about the transaction, per se.
This middleware makes it so that any action might just write in that rare case. We might think about how to do whatever writes are involved in a task. Or maybe the `catchError` thing makes it ok. I don't know much about that.

core/src/modules/middleware/authentication.ts
```diff
@@ -82,31 +62,68 @@ async function authenticateTeamMember(
...
+    { catchError: true }
+  );
 
-    if (!teamMember) {
-      throw new AuthenticationError("Team member not found");
+  if (error) {
+    if (error["code"] === "AUTHENTICATION_ERROR") {
+      await CLS.wrap(async () => {
+        await api.session.destroy(data.connection);
+      });
```
evantahler, line 110: Here's the write

evantahler:
@bleonard I think because I didn't give `CLS.wrap` a priority or `write:true`, aren't I saying that these new transactions are not writing (eg not blocking for sqlite)?  That was my intent anyway... even in the case where we do that write, which isn't that important so it can lag.

bleonard:
The issue is if it does write, it will try to be in charge and maybe hang because another thread is doing so.
It will retry a few times and maybe work out.
All the `write` flag does is tell SQLite that it needs write permission from the beginning so it doesn't waste time. There is no queuing or smart behavior in SQLite. The `priority` makes it retry faster so the UI isn't affected. Since this is the UI, I think I'd make this little one with the destroy `write: true, priority: true`

evantahler:
👍 https://github.com/grouparoo/grouparoo/pull/1655

PR #1657, opened 2021-04-13
Mongo sum operation issue

pauloouriques:


plugins/@grouparoo/demo/src/mongo/connection.ts
```diff
@@ -60,6 +60,9 @@ const TYPES = {
           user_id: {
             bsonType: "int",
           },
+          item: {
```
bleonard, line 63: what's item?

plugins/@grouparoo/demo/src/mongo/connection.ts
```diff
@@ -90,6 +93,47 @@ function findConfig() {
...
+  collectionName,
+  field,
+  value
+): string | number | boolean | Date {
+  if (value === null || value === undefined) {
+    return null;
+  }
+  const type = TYPES[collectionName]?.validator?.$jsonSchema?.properties[field];
+  if (!type) {
+    return null;
```
bleonard: maybe `throw` here because it's unexpected.

plugins/@grouparoo/demo/src/mongo/connection.ts
```diff
@@ -90,6 +93,47 @@ function findConfig() {
...
+    case "int":
+      return parseInt(value);
+    case "double":
+      return parseFloat(value);
+    case "date":
+      return new Date(value);
+    case "timestamp":
+      return new Date(value).toISOString();
+    default:
+      return null;
```
bleonard: maybe throw here because it's unexpected.

PR #1658, opened 2021-04-13
Dynamically test and validate all plugin templates

evantahler:
This PR creates a test suite that can be applied to almost every plugin to test the `grouparoo generate` and `grouparoo validate` commands dynamically.

```
 PASS  ../plugins/@grouparoo/sqlite/__tests__/integration/cli-generate-validate.ts (13.252 s)
  @grouparoo/sqlite CLI
    sqlite:app
      ✓ generate app sqlite:app (1229 ms)
      ✓ generate source sqlite:query:source (1268 ms)
      ✓ generate property sqlite:query:property (1269 ms)
      ✓ generate source sqlite:table:source (1224 ms)
      ✓ generate property sqlite:table:property (1259 ms)
      ✓ generate destination sqlite:destination (2469 ms)
    validation
      ✓ the generated config files can be locally validated (1380 ms)
    everything was tested
      ✓ sqlite:app was tested (1 ms)
      ✓ sqlite:destination was tested
      ✓ sqlite:query:property was tested (1 ms)
      ✓ sqlite:query:source was tested
      ✓ sqlite:table:property was tested (5 ms)
      ✓ sqlite:table:source was tested
```

The suite:
1. runs `grouparoo generate --list` to find matching generators for this plugin
2. runs `grouparoo generate {thing}` for everything the plugin can generate.  We do things in order: app, source, property, destination.
   * As needed, we replace some of the default mappings with links to `user_id` to build the DAG
3. If needed, we generate groups and other properties to link to
4. run `grouparoo validate --local` to ensure that all the plugin types link up, and the DAG can be built.

Of note:
* I *did* find some problems with existing templates! It worked!
* I updated the placeholder values for some plugins to match the property `user_id` rather than `userId` to make the DAG work generically
* Added templates for `@grouparoo/logger`
* The specHelper introduced here uses spawnSync to inspect the output of `grouparoo generate --list` to learn what templates it should be testing.  This mirrors the customer experience the best, but it is rather brittle.


evantahler:
The final plugins, `sqlite` and `snowflake` depend on https://github.com/grouparoo/grouparoo/pull/1665

plugins/@grouparoo/mongo/public/templates/query-source/{{{id}}}.js.template
```diff
@@ -0,0 +1,42 @@
...
+     */
+    // {
+    //   id: {{{schedule_id}}},
+    //   name: {{{schedule_id}}},
+    //   class: "schedule",
+    //   sourceId: {{{id}}}, // The ID of the Source above
+    //   recurring: true, // should this Schedule regularly run?
+    //   recurringFrequency: 1000 * 60 * 15, // 15 minutes, in ms
+    //   options: {
+    //     query: "SELECT id FROM users WHERE updated_at >= (NOW() - INTERVAL '2 day')", // A SQL query to return that includes only the IDs of properties to check.
```
evantahler, line 37: @pauloouriques do you have an example query we might use in this template for a mongo query source?

plugins/@grouparoo/mongo/src/initializers/plugin.ts
```diff
@@ -41,7 +41,9 @@ export class Plugins extends Initializer {
         new TableSourceTemplate("mongo", { getTables, getColumns }),
         new TablePropertyTemplate("mongo"),
         new QueryPropertyTemplate("mongo"),
-        new QuerySourceTemplate("mongo"),
+        new QuerySourceTemplate("mongo", [
+          path.join(templateRoot, "query-source", "*.template"),
+        ]),
```
evantahler, line 46: The Mongo Query source needed a custom template as it requires "table"  ("collection") that the others don't

plugins/@grouparoo/mailchimp/public/templates/destination/id/{{{id}}}.js.template
```diff
@@ -4,7 +4,7 @@ exports.default = async function buildConfig() {
       id: {{{id}}},
       name: {{{id}}},
       class: "destination",
-      type: "{{{__pluginName}}}-export",
+      type: "mailchimp-export-id",
```
evantahler, line 7: Mailchimp template snafu!

plugins/@grouparoo/intercom/public/templates/destination/{{{id}}}.js.template
```diff
@@ -4,7 +4,7 @@ exports.default = async function buildConfig() {
       id: {{{id}}},
       name: {{{id}}},
       class: "destination",
-      type: "{{{__pluginName}}}-export",
+      type: "intercom-export-contacts",
```
evantahler, line 7: Intercom template snafu!

plugins/@grouparoo/facebook/public/templates/destination/{{{id}}}.js.template
```diff
@@ -4,7 +4,7 @@ exports.default = async function buildConfig() {
       id: {{{id}}},
       name: {{{id}}},
       class: "destination",
-      type: "{{{__pluginName}}}-export",
+      type: "facebook-audiences-custom",
```
evantahler, line 7: Facebook template snafu!

plugins/@grouparoo/mongo/public/templates/query-source/{{{id}}}.js.template
evantahler: Moving to a new story

seancdavis (commented):
Overall, I _really_ like what you did here. A few thoughts:

1. Brittle, sure, but it follows the user's path pretty closely. That's a true integration test. In fact, if the test which mimics user behavior is brittle, it suggests that the user journey is brittle. Of course, we know that now, which is why we're moving toward a more interactive configuration experience. But that's a great way to find these rough spots.

2. It bugs me that every plugin's `cli-generate-validate.ts` spec is _exactly_ the same. The pattern of explicitly implementing it within each plugin's test suite feels right, but the output of that approach feels [insert adjective]. IOW if I want to add this to a new plugin I'm creating, I have to copy and paste 27 lines of code just for it to do exactly what it should do. I'd prefer it to be more like a line or two.

    ```js
    import { CLISpecHelper } from "@grouparoo/spec-helper";
    CLISpecHelper.validateGenerators("sqlite")
    ```

evantahler:
I got close @seancdavis! check out https://github.com/grouparoo/grouparoo/pull/1692

PR #1660, opened 2021-04-14
Fix customer.io app template generation

pedroslopez:
Currently, trying to generate a customer.io app through `grouparoo generate customerio:app cioapp` causes an error on startup:
```
[ config ] error with app `cioapp` (cioapp): Cannot find a "customerio" plugin.  Did you install it? 
```

This is because the app name is actually `customer.io` instead of `customerio` (note the `.`). 

There are two main options here:
1. Rename the app to `customerio` so that it matches the plugin name and other templates. I believe this would be the right thing to do from the start, but this likely means we would need to add a migration so existing apps aren't affected.
2. Update the template so it uses `customer.io` as app type. Doing it this way is slightly inconsistent, but would not constitute a breaking change / require migrations.

### DIFF
--- a/plugins/@grouparoo/customerio/src/initializers/plugin.ts
+++ b/plugins/@grouparoo/customerio/src/initializers/plugin.ts
@@ -36,7 +36,7 @@ export class Plugins extends Initializer {
       ],
       apps: [
         {
-          name: "customer.io",
+          name: "customerio",
           options: [
             {
               key: "siteId",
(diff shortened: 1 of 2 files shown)


evantahler:
Assuming the migration is easy (I think you would only need to update `app.type`?), option 1 is OK with me!

evantahler:
Tracking in this bug https://www.pivotaltracker.com/story/show/177767908

pedroslopez:
> Assuming the migration is easy (I think you would only need to update `app.type`?), option 1 is OK with me!

Yeah, I believe so! Should we wait until we merge #1569 and add the migration specifically to this plugin? 


Note that if someone figured out that the type was wrong in the template and are using code config here, it could be breaking for them, though this is unlikely.

evantahler (approved):
Looks good!

@pedroslopez  We may want to note this in the v0.3 migration docs as well

evantahler:
cc @seancdavis FYI - plugins have migrations now!

PR #1664, opened 2021-04-14
CodeConfig to consider properties in mustache queries when sorting objects

evantahler:
Prior to this PR, query-mode properties which had queries like `SELECT email from users where id = {{ userId }}` were not being considered properly in the topological sort when applying code-config to the sever.  This PR now considers both Properties which are already stored in the server's database and not-yet-applied but in the same code-config batch.


evantahler:
Discovered as part of https://github.com/grouparoo/grouparoo/pull/1658, in which some of the `*:source:query` properties would fail validation

PR #1666, opened 2021-04-16
Get changed rows count issue.

pauloouriques:
- Add treatment to when the number of changed rows is 0.


plugins/@grouparoo/mongo/src/lib/table-import/getChangedRowCount.ts
```diff
@@ -24,5 +24,8 @@ export const getChangedRowCount: GetChangedRowCountMethod = async ({
     .collection(tableName)
     .aggregate(aggPipeline)
     .toArray();
-  return parseInt(rows[0]["__COUNT"]);
+  if (rows.length > 0) {
```
bleonard, line 27: Out of curiosity, do you have a sense of when this was zero?
It feels like it should still always have a number here.

PR #1685, opened 2021-04-19
Fix issue adding to groups after being cleared by no-delete syncMode

pedroslopez:
Because deleting a profile from a destination when in a sync mode that doesn't delete (enrich, additive) does not actually delete them and only clears their groups, currently we are returning an info level error and clearing the groups in the destination. This is causing issues when the profile is re-added. When Grouparoo checks to see if anything has changed, it will not see any changes since the export that cleared groups was marked as an error. This means that its group memberships are not updated in the destination.

This moves the group clearing logic to core so that the export is properly generated with a `toDelete: false` and `newGroups: []`. The destination just interprets this like any other export that changes a profile's group memberships. As a bonus to this, exports in the UI will also properly show what's going on (clearing groups instead of marking it as `toDelete`).

Before:
image

After:
image

Then, when it's re-added:
image


core/__tests__/models/destination/plugins/exportProfile.ts
```diff
@@ -438,6 +438,110 @@ describe("models/destination", () => {
       await profile.destroy();
     });
 
+    test("if profile is removed from destination's tracked group, toDelete is true", async () => {
+      await destination.setMapping({
```
evantahler, line 444: You may want to confirm that the syncMode of the destination is "sync".

core/__tests__/models/destination/plugins/exportProfile.ts
```diff
@@ -438,6 +438,110 @@ describe("models/destination", () => {
...
+      expect(exportArgs.newGroups).toEqual(
+        [groupA, groupB].map((g) => g.name).sort()
+      );
+      expect(exportArgs.toDelete).toEqual(true);
+
+      await profile.destroy();
+    });
+
+    test("if profile is removed from destination's tracked group with a no-delete syncMode, toDelete is false and groups are cleared", async () => {
+      await destination.update({ syncMode: "additive" });
```
evantahler: I'd suggest running this test in a loop, testing both additive and enrich:

```js
["additive", "enrich"].forEach(syncMode => {
  test('...')
})
```

core/__tests__/models/destination/plugins/exportProfiles.ts
```diff
@@ -339,6 +339,130 @@ describe("models/destination", () => {
       expect(exportArgs.exports[0].toDelete).toEqual(true);
     });
 
+    test("if profile is removed from destination's tracked group, toDelete is true", async () => {
```
evantahler, line 344: "... and syncMode = sync"

core/src/modules/ops/destination.ts
```diff
@@ -350,6 +350,16 @@ export namespace DestinationOps {
           )[0].remoteKey
       );
 
+    const syncMode = await destination.getSyncMode();
+    const canDelete = syncMode
+      ? DestinationSyncModeData[syncMode].operations.delete
```
evantahler, line 355: Confirming that `DestinationSyncModeData[syncMode].operations.delete` can never be null.  If so, this wold be true if undefined

evantahler (commented):
Looking good! I love how simple the implementation ended up.  I also love deleting plugin code!

Just some minor suggestions and some robustness ideas for the tests.

Should we write an additional story to simplify any of the other syncMode plugins?

core/src/modules/ops/destination.ts
pedroslopez: Hmm, not sure I understand what you mean. What would be `true` if what is `undefined`?

evantahler: sorry about that! I read this logic to say that `canDelete` would be `true` if `DestinationSyncModeData[syncMode].operations === null || undefined`.  I don't think that's possible, but i wanted to confirm.

pedroslopez: `undefined` isn't possible because of the type on DestinationSyncModeData, but `null` is technically valid. If the operations are `null` then this would error here saying you can't access `delete` of `null`. We're the ones defining what each syncMode does in core though so this shouldn't be possible.

PR #1696, opened 2021-04-21
cli: add extra help text if no local commands have been loaded

pedroslopez:


evantahler (approved):
Nice! 

Can you add a test for this? Perhaps running `grouparoo` both before and then after `grouparoo generate` to see that the new help text appears.  The test is in bash, so let me know if I can help!

apps/staging-enterprise/__tests__/cli
```diff
@@ -2,22 +2,37 @@
 
 set -e
 
-cd "`dirname "$0"`/.."
+PROJECT_PATH="`pwd`"
```
pedroslopez, line 6: I had to change this to get an absolute path to the CLI script since we're changing directories.

@evantahler do you know of a better way to get the absolute path? This current solution works, but requires calling the script from within the project directory.

evantahler: Can you combine the approaches?

```
cd "`dirname "$0"`/.."
PROJECT_PATH="`pwd`"
```

This way it doesn't matter if you start the program via `./path/to/file.sh` or `cd path/to && ./file.sh`, which at the moment would have different outcomes.

Bash is hard.

pedroslopez: That works!

evantahler (approved):

PR #1697, opened 2021-04-21
Add sync modes to Zendesk plugin

pauloouriques:


plugins/@grouparoo/zendesk/src/lib/export/exportProfile.ts
```diff
@@ -37,10 +42,14 @@ export const exportProfile: ExportProfilePluginMethod = async ({
...
 
+    // tags need to be formatted
+    const newTags = newGroups.map((groupName) => makeTagName(groupName));
+    const oldTags = oldGroups.map((groupName) => makeTagName(groupName));
+
     if (toDelete) {
       if (user) {
         // this does a soft delete
-        await client.users.delete(user.id);
+        await deleteContact(client, syncOperations, user, true);
```
pedroslopez: Was this added because we were doing something more complex here before? I believe we can just check the syncOp before calling the client function like you're doing in the update/create cases.

pedroslopez (commented):
Looking good! Just one non-blocking comment.

Also, did we check that the foreignkey change situation is working as intended? If so, please add some tests for that.

pedroslopez (commented):
Looks good, though I'm not sure about the split between `/__tests__/export/export-profile` and `/__tests__/integration/export-profile`, seems like much of the same stuff is being tested?

Let's wait for @evantahler 's review

plugins/@grouparoo/zendesk/__tests__/integration/export-profile.ts
```diff
@@ -21,14 +22,14 @@ const email = "brian@bleonard.com";
 const name = "Brian";
 const exampleDate = new Date(1597870204 * 1000);
 
-const nockFile = path.join(__dirname, "../", "fixtures", "export-profile.js");
+const nockFile = path.join(__dirname, "../", "fixtures", "zendesk-profile.js");
```
evantahler: Ideally, the fixture name matches the file name of the test

evantahler (commented):
Yep, there does seem to be some overlap between those tests.  In the other plugins, the "integration" tests called the actions directly (eg https://github.com/grouparoo/grouparoo/blob/main/plugins/%40grouparoo/postgres/__tests__/integration/postgres-table-import.ts) or used the CLI.

evantahler (approved):

plugins/@grouparoo/zendesk/__tests__/export/export-profile.ts
```diff
@@ -0,0 +1,691 @@
...
+      oldGroups: [groupOne, outsideGroup, groupThree],
+      newGroups: [groupOne],
+      toDelete: false,
+    });
+
+    const user = await searchForUser(client, { external_id: externalId1 });
+    expect(user.email).toBe(email2);
+  });
+
+  // test("it can make one user with the original id and no email", async () => {
```
pedroslopez: Why is this one commented out?

pedroslopez (approved):

PR #1700, opened 2021-04-22
fix: don't clear 0-value attributes on intercom and salesforce

pedroslopez:


plugins/@grouparoo/intercom/src/lib/export-contacts/exportProfile.ts
```diff
@@ -300,7 +300,7 @@ async function deleteUser(
 }
 
 function formatVar(value) {
-  if (!value) {
```
evantahler, line 303: Yeah... these get me in trouble :/

what about `null`?

evantahler: Yeah, Salesforce is checking for both `null` and `undefined`

evantahler (changes requested):

plugins/@grouparoo/intercom/src/lib/export-contacts/exportProfile.ts
pedroslopez: since `value` is `null`, it will just be returned in line 310 (`return value = return null`). I did it explicitly for salesforce because it has some extra logic in there for getting custom default values.

evantahler: fair enough!

evantahler (approved):

PR #1701, opened 2021-04-22
Selective local config validation with grouparoo cli

pedroslopez:
- [x] Also do local validation for dependents
- [x] better handling for variadic options

Usage:
- `roo validate` does external validation for all
- `roo validate --local` disables external validation for all
- `roo validate --local id_here another_id_here yet_another_id_here` disables external validation for all listed object ids

Note: if external validation is disabled for a prerequisite of another object, external validation will also be disabled for that object. For example, if you have `app_id` and `destination_id` and run `roo validate --local app_id`, it will be equivalent to running `roo validate --local app_id destination_id`

### DIFF
--- a/core/src/modules/configLoaders/index.ts
+++ b/core/src/modules/configLoaders/index.ts
@@ -90,6 +90,7 @@ async function loadConfigFile(file: string): Promise<ConfigurationObject> {
 export async function processConfigObjects(
   configObjects: Array<ConfigurationObject>,
   externallyValidate: boolean,
+  locallyValidateIds: Array<string> = [],
   validate = false,
   extraSortingConfigObjects?: Array<ConfigurationObject>
 ): Promise<{ seenIds: IdsByClass; errors: string[] }> {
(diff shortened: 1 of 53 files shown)


plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -238,9 +238,17 @@ export namespace helper {
               key: "password",
               required: false,
             },
+            {
+              key: "failRemoteValidation",
```
evantahler: Because `failRemoteValidation` is something special only for the apps created in the spec-helper, it might be helpful to signal that this isn't a real/core thing by calling it `_failRemoteValidation`. or something. 

I had thought this was a new concept you had introduced to `@grouparoo/core`

core/src/modules/configLoaders/index.ts
```diff
@@ -87,9 +88,38 @@ async function loadConfigFile(file: string): Promise<ConfigurationObject> {
...
+  }
+
+  return true;
+}
+
 export async function processConfigObjects(
   configObjects: Array<ConfigurationObject>,
-  externallyValidate: boolean,
+  canExternallyValidate: boolean,
+  locallyValidateIds?: Set<string>,
```
evantahler, line 122: Comment (no need to change it): This might be the first use of `Set` in the codebase!  Given the rest of the code, is this a surprising data structure, or since we have TS helping us out, is this OK?  Set is certainly the best, being that we don't want duplicates.

core/src/modules/configLoaders/index.ts
```diff
@@ -150,6 +180,21 @@ export async function processConfigObjects(
...
+      configObjects,
+      locallyValidateIds
+    );
+
+    if (!externallyValidate) {
+      log(
+        `[ config ] skipping external validation for ${configObject.class} \`${configObject.id}\``,
+        "notice"
+      );
+    }
```
evantahler, line 196: This message is interesting! If I `grouparoo validate --local data_warehouse` (assuming data_warehouse is an App), I think I would see this message a lot of times - for the sources, schedules, and properties.   I guess that's OK for now, as it's certainly clear, but perhaps we only want to show it for the parent objects?

evantahler (changes requested):
This is looking really good.

I like that `--local` passes down from parent to child! 

I believe that there should also be a companion PR in the WWW project to talk about these changes.

core/src/modules/configLoaders/index.ts
pedroslopez: We've used it a few times in some of the plugins, but yeah I dont think it's been used in core

core/src/modules/configLoaders/index.ts
pedroslopez: We've used it a few times in some of the plugins, but yeah I dont think it's been used in core. Since we're adding the children IDs I wanted to avoid duplicates. An alternative is also to take in an array like usual and make the set within the funciton

core/src/modules/configLoaders/index.ts
pedroslopez: Would collecting everything that's being locally validated and logging it in a single message at the end be better? That way there arent a bunch of these in between the validation messages?

evantahler: I think the way you have it is OK for now!   Let's get this merged in and see how annoying it gets

PR #1702, opened 2021-04-23
Return runs sorted by updatedAt

evantahler:
Sort tuns by `updatedAt` in list view.

### DIFF
--- a/core/src/actions/runs.ts
+++ b/core/src/actions/runs.ts
@@ -2,7 +2,6 @@ import { AuthenticatedAction } from "../classes/actions/authenticatedAction";
 import { Run } from "../models/Run";
 import { Op } from "sequelize";
 import { Schedule } from "../models/Schedule";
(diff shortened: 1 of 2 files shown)


evantahler:
@seancdavis There's something puzzling me here that I can't figure out.  I've confirmed that the action is properly returning the runs sorted by `updatedAt desc`.  If I log the runs in the UI hydrate (getInitialProps), they are also in the property order. 

If the component renders hydrated from the server (eg: command + r), the runs array is rendered reversed!  If the runs array is rendered by the browser (navigating around and coming back to the page), the runs array on the page is in the right order.

Any ideas?  I added an additional sort in the component to account for this, but I hate it.

evantahler:
I'm going to merge this even though I cannot figure out why the Runs array order changes between server-side and browser-side rendering.  TODO left in the code :(

PR #1703, opened 2021-04-23
Do not update Mailchimp status when updating others variables.

pauloouriques:
- add tests to this use case.


plugins/@grouparoo/mailchimp/src/lib/shared/updateProfile.ts
```diff
@@ -73,26 +73,23 @@ export const updateProfile: UpdateProfileMethod = async ({
     }
   }
 
-  try {
-    const userResponse = await getUser(client, listId, mailchimpId);
-    exists = userResponse !== null;
+  const userResponse = await getUser(client, listId, mailchimpId);
+  exists = userResponse !== null;
```
bleonard, line 77: I'm not sure this `exists` flag is serving us anymore that we aren't catching. can probably just use
> const existingUser = await getUser(client, listId, mailchimpId)

bleonard: and `if(existingUser)`

PR #1707, opened 2021-04-26
Build step on CI requires more RAM

evantahler:
Sometimes the `pnpm install` step fails, running out of RAM.  This moves just the build step to a larger server. 

Interestingly, there doesn't seem to be a way to tell `pnpm` not to do 4 installs at once

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,7 @@ references:
 # Define Jobs to run
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


evantahler:
An alternative approach would have been to limit the parallelism of `postinstall` processes after `pnpm install`, but that doesn't appear to be possible at the moment.  https://github.com/pnpm/pnpm/issues/3399

PR #1708, opened 2021-04-26
`ui-config` technical spike

evantahler:
Screen Shot 2021-04-26 at 4 08 00 PM

Notes
* There were some shenanigans about using ENV to determine which mode we were booting into.  I needed to clear the require.cache in the CLI (which needed to load actionhero, to find plugins), and then to re-load config with the env updated after the process
* You can run ui-config with next in development mode via `NEXT_DEVELOPMENT_MODE=true roo config`.  It's odd because it's a CLI command (compiled grouparoo) but you want to use the dev mode UI


cli/src/utils/loadLocalCommands.ts
```diff
@@ -106,6 +106,12 @@ function getActionhero() {
   return actionheroPackage;
 }
 
+function clearRequireCache() {
+  for (const path in require.cache) {
+    if (path.endsWith(".js")) delete require.cache[path];
+  }
+}
```
evantahler: TODO: This is a bit heavy-handed...

PR #1709, opened 2021-04-26
support specific local validation on `apply` command

pedroslopez:
updates the `--local` flag to support same features from #1701 on `grouparoo apply`


evantahler (changes requested):
Whoa.  It doesn't look like we have any bin test for `grouparoo apply` (https://github.com/grouparoo/grouparoo/tree/main/core/__tests__/bin).   Please add one!

evantahler (approved):
Thanks for adding the test... and thanks for the typo fix too!

PR #1712, opened 2021-04-27
auto determine bootstrapped property

pedroslopez:
* table: users
  * mapping: id (col) = user_id (property)
* table: purchases
  * mapping: user_id (col) = user_id (property)
 
 * property: user_id
      * sourceId: users
* property: email
  * sourceId: users
* property: ltv
   * sourceId: purchases

Auto bootstrap property:
- users.mapping.sourceId == this => bootstrap
  -  (and unique and not array)

---

TODO: 

- [x] Remove bootstraped Propety from all generators
- [x] Ensure that source `--with` generators make previorusly the bootstraped property
- [x] Ensure that old bootstrapped properties still work (maybe?)

Quetsions:
- what is identifying -> I think this is just for the UI.  
   - no properties, guess the first bootsrapped one is identifying -> user_id
   - user can pick a better propertiy (from ui/settings) -> email   
- what is `proeprty.directlyMapped` -> Is this the same thing?  Should bootstrapped properties be this?


core/src/modules/configLoaders/index.ts
```diff
@@ -204,7 +204,12 @@ export async function processConfigObjects(
           ids = await loadApp(configObject, externallyValidate, validate);
           break;
         case "source":
-          ids = await loadSource(configObject, externallyValidate, validate);
+          ids = await loadSource(
+            configObject,
+            configObjects.filter((o) => o.id !== configObject.id),
```
bleonard: it feels like `loadSource` could know to skip this one.

bleonard: in other words, suggestion to just pass `configObjects` here.

core/src/classes/codeConfig.ts
```diff
@@ -147,6 +147,32 @@ export function extractNonNullParts(
...
 
+export function getAutoBootstrappedProperty(
+  sourceConfigObject: ConfigurationObject,
+  otherConfigObjects: ConfigurationObject[]
+) {
+  if (sourceConfigObject.class?.toLowerCase() !== "source") return null;
+  if (!sourceConfigObject.mapping) return null;
+
+  const mappingValues = Object.values(sourceConfigObject["mapping"]);
+  for (const value of mappingValues) {
```
bleonard, line 158: should we do the mapping story as part of this? at least the source one? https://www.pivotaltracker.com/story/show/177734437

bleonard: hard to say but wanted to bring up we have something in this same area.

core/__tests__/fixtures/codeConfig/manual-bootstrapped-property/config.js
```diff
@@ -0,0 +1,177 @@
+module.exports = async function getConfig() {
```
evantahler: maybe the name of this file & test case is something like `bootstrap-included-with-source` or something like that

core/src/classes/codeConfig.ts
evantahler: We'll do that story soon after

core/src/modules/configLoaders/source.ts
```diff
@@ -64,21 +67,28 @@ export async function loadSource(
     }
   }
 
+  const bootstrappedPropertyConfig =
+    configObject.bootstrappedProperty ||
+    getAutoBootstrappedProperty(
+      configObject,
+      configObjects.filter((o) => o.id !== configObject.id)
+    );
```
evantahler, line 75: Note: Here's how both the old way and the new way are supported.

core/src/modules/configLoaders/source.ts
```diff
@@ -101,7 +111,11 @@ export async function loadSource(
   }
 
   logModel(source, validate ? "validated" : isNew ? "created" : "updated");
-  if (bootstrappedProperty) {
+  if (configObject.bootstrappedProperty) {
+    log(
+      `source.bootstrappedProperty is deprecated. You can generate the property by using \`grouparoo generate\` as usual and the bootstrapped property will be automatically determined. (${configObject.id})`,
```
evantahler: 👍 for the warning message. 

I don't think it will "automatically determined" though, unless you do the `--with` ?  How about:

> source.bootstrappedProperty is deprecated. Please generate a the config file for the Property by using \`grouparoo generate\`. (${configObject.id})

plugins/@grouparoo/app-templates/public/templates/table-source/{{{id}}}.js.template
```diff
@@ -24,40 +24,10 @@ exports.default = async function buildConfig() {
...
-      {{^__bootstrappedId}}
-      // bootstrappedProperty: {
-      //   id: "user_id",
-      //   name: "User Id",
-      //   type: "integer",
-      //   options: {
-      //     column: "id",
-      //   },
-      // },
-      {{/__bootstrappedId}}
```
evantahler, line 60: Yeah!

plugins/@grouparoo/spec-helper/src/lib/cliSpecHelper.ts
```diff
@@ -247,6 +227,48 @@ DATABASE_URL="sqlite://grouparoo_test.sqlite"
...
+  export async function generatePropertyToBootstrap(
+    runCliCommand: Function,
+    projectPath: string,
+    propertyName: string,
+    sourceId: string,
+    options?: CLIValidationOptions
+  ) {
+    await runCliCommand(
+      `generate manual:property ${propertyName} --parent ${sourceId}`
+    );
```
evantahler: Does this mean that you are saying a `manual:property` belongs to a `postgres:source:table`?  That doesn't seem right (and will fail as of https://www.pivotaltracker.com/story/show/177875103).   You might need to read the source file to learn it's type to know what kind of property to make from it.

pedroslopez: I don't think that's the case here because property templates are pretty much the same throughout all sources and don't have a `type` on them that's specific to the parent (source). Instead, the options between them may be different. 

Initially I solved this by passing an object with the options it should set on the bootstrapped property when calling `validateGenerators`, but after reading this I went with something a bit smarter and am now generating the right property for each generator.

evantahler (approved):
💥

PR #1719, opened 2021-04-29
Handle Pipedrive rate-limit errors

pauloouriques:
- Add the same strategy of other plugins which is to pick a random retry delay from the interval [1 - 11]


evantahler (approved):
This matches the behavior of how we handle rate limits with Zendesk, Sendgrid, etc.  This is probably also very hard to test (you'd need to get lucky with a Nockfile that caught a rate-limit).

In the future, we'll do something smarter with these cases (storing information in postgres on the Export model about retries rather than relying on redis)

PR #1723, opened 2021-04-30
 Add sync modes to Iterable plugin.

pauloouriques:


plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -81,6 +87,11 @@ export const sendProfile: ExportProfilePluginMethod = async ({
         emailPayload["currentUserId"] = newProfileProperties.userId;
       }
       try {
+        if (!syncOperations.update) {
+          throw new Errors.InfoError(
```
pedroslopez, line 91: This looks good and I think it would pass, but just in case since this is another place where the error is being thrown we should probably add a test to check that the "no update" warning comes up when trying to change the email address.

pedroslopez (commented):
Looks great! Only one comment about a possibly missing test

pedroslopez (approved):

evantahler (approved):

PR #1726, opened 2021-04-30
Set all options when creating bootstrappedProperty

pedroslopez:
This is mainly an issue for the old way of defining the `bootstrappedProperty`. If the plugin didn't define its own `uniquePropertyBootstrapOptions` method, no options would be set on the property. It also wasn't possible to override the defaults set by the plugin.

All options defined in the `bootstrappedProperty` section are now set on the property.


evantahler (approved):
Nice!

PR #1727, opened 2021-04-30
Remove `[ run ] starting...` log messages

evantahler:
These are annoying, and we have other, better ways in the UI and CLI to see what runs are running.


evantahler:
Ping @bleonard and @andyjih - any worries on removing these log messages?

PR #1728, opened 2021-04-30
Rescue stuck Groups who lost their runs

evantahler:
If a calculated Group was updating its members, it's possible for that calculation to never complete if the Run was stopped manually or had an error.   This PR will start a new calculation run for those Groups if they haven't been checked in some time and if there is not a running Run for them already.

### DIFF
--- a/core/src/tasks/group/updateCalculatedGroups.ts
+++ b/core/src/tasks/group/updateCalculatedGroups.ts
@@ -1,6 +1,6 @@
 import { Group } from "../../models/Group";
+import { Run } from "../../models/Run";
 import { plugin } from "../../modules/plugin";
-import { Op } from "sequelize";
 import Moment from "moment";
 import { CLSTask } from "../../classes/tasks/clsTask";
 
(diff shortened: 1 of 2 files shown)


core/src/tasks/group/updateCalculatedGroups.ts
```diff
@@ -21,21 +21,37 @@ export class GroupsUpdateCalculatedGroups extends CLSTask {
...
-            [Op.lt]: lastCheckTime,
-          },
-        },
-      },
+      where: { type: "calculated" },
     });
 
-    await Promise.all(calculatedGroups.map((group) => group.run()));
+    for (const group of calculatedGroups) {
+      if (group.state === "ready" && !group.calculatedAt) {
```
bleonard, line 32: acceeptance testing this. I think there are some more cases, especially on first time running.

For example, if the state is "updating" and `!group.calculatedAt` is not handled.

Stack for reference
```
2021-05-03T22:32:18.066Z - error: error from task name=task:group:updateCalculatedGroups queue=groups worker=1 arguments={} stacktrace=TypeError: Cannot read property 'getTime' of null
    at GroupsUpdateCalculatedGroups.runWithinTransaction (/Users/brian/grouparoo/grouparoo/core/dist/tasks/group/updateCalculatedGroups.js:42:36)
    at async /Users/brian/grouparoo/grouparoo/core/dist/modules/cls.js:75:27
    at async /Users/brian/grouparoo/grouparoo/node_modules/.pnpm/sequelize@6.6.2_4805d29a365866739fa3d4ea3781136c/node_modules/sequelize/lib/sequelize.js:1090:24
    at async wrapInternal (/Users/brian/grouparoo/grouparoo/core/dist/modules/cls.js:74:9)
    at async Object.CLS.wrap (/Users/brian/grouparoo/grouparoo/core/dist/modules/cls.js:38:33)
    at async Worker.perform (/Users/brian/grouparoo/grouparoo/node_modules/.pnpm/actionhero@25.0.15/node_modules/actionhero/dist/initializers/tasks.js:97:36)
    at async Worker.perform (/Users/brian/grouparoo/grouparoo/node_modules/.pnpm/node-resque@8.2.8/node_modules/node-resque/dist/core/worker.js:165:27) 
```

bleonard: since most of these assume the `group.calculatedAt`, maybe deal with that first
```
const lastRunAt = group?.calculatedAt?.getTime();
if(!lastRunAt) {
  // handle ready and updating cases as relevant.
}
else if (lastRunAt < lastCheckTime.getTime()) {
  // it's been awhile. handle based on state
}
```

PR #1729, opened 2021-05-01
Support SSL redis (redis v6)

evantahler:
Heroku Redis now comes with SSL... and you can't opt out :D

### DIFF
--- a/core/src/config/redis.ts
+++ b/core/src/config/redis.ts
@@ -15,9 +15,11 @@ function RealRedisConfig() {
   let db = process.env.REDIS_DB;
   let username = process.env.REDIS_USER;
   let password = process.env.REDIS_PASS;
+  let protocol = "redis";
 
   if (process.env.REDIS_URL) {
     const parsed = new URL(process.env.REDIS_URL);
+    if (parsed.protocol) protocol = parsed.protocol.split(":")[0];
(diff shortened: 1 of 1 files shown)


core/src/config/redis.ts
```diff
@@ -35,6 +37,8 @@ function RealRedisConfig() {
     password,
     // If we have a REDIS_URL, but no database, assume we mean db=0
     db: parseInt(db || "0"),
+    // ssl options
+    tls: protocol === "rediss" ? { rejectUnauthorized: false } : undefined,
```
bleonard, line 41: is this `rediss` a standard of some sort? just looks like a typo.

evantahler: The extra `s` is for secure, lol

> As of Redis version 6, Hobby plans support both TLS and unencrypted connections, while production plans require TLS connections. For hobby plans you can find the unencrypted (plaintext) and TLS URIs in your app’s config vars REDIS_URL and REDIS_TLS_URL respectively. For Premium, Private, and Shield plans on version 6 your REDIS_URL config var will use the rediss: scheme, and you must enable TLS in your Redis client’s configuration in order to connect to a Redis 6 database.

https://devcenter.heroku.com/articles/heroku-redis

PR #1730, opened 2021-05-01
Telemetry errors should log and not throw

evantahler:


core/src/modules/telemetry.ts
```diff
@@ -40,7 +40,7 @@ export namespace Telemetry {
         });
       } catch (newError) {}
 
-      throw error;
+      log(`[ telemetry ] ${error}`, "error");
```
bleonard: this might be good for `grouparoo run` but mahybe it should retry in the `grouparoo start` background case?

evantahler: 👍 The tasks weren't retryable... they are now!

PR #1737, opened 2021-05-02
Use `force` to determine if Profiles in the Group should be re-imported fully

evantahler:
When a Profile is newly added to a Destination because the Group is tracked, we still re-import all the ProfileProperties.  But:

* When a Destination un-tracks a Group, the Profiles leaving the Destination do not need to be fully re-imported.  
* When a Profile is newly added to a calculated Group, do not re-import all the Properties of that Profile
* Add the plumbing to let Destinations and Groups choose if this run should mark all of a Profile's Properties as pending or not based on the `force` option.  This PR does not change the existing behavior for a Destination tracking a group - all the Profiles in that group will have all Properties re-imported before they are exported.


evantahler:
@pedroslopez can you please check this for me?  I just rebased it to work with your changes from #1769

core/src/modules/ops/group.ts
```diff
@@ -51,10 +51,11 @@ export namespace GroupOps {
   /**
    * Given a Profile, create an import to recalculate its Group Membership
```
pedroslopez: I think this comment doesn't quite match up anymore

core/src/models/Destination.ts
```diff
@@ -294,12 +294,12 @@ export class Destination extends LoggedModel<Destination> {
...
 
-  async trackGroup(group: Group) {
-    return DestinationOps.trackGroup(this, group);
+  async trackGroup(group: Group, force = true) {
+    return DestinationOps.trackGroup(this, group, force);
   }
 
-  async unTrackGroup() {
-    return DestinationOps.unTrackGroup(this);
+  async unTrackGroup(force = false) {
```
pedroslopez, line 301: Confirming behavior:
- `trackGroup` has a default of `force=true` for the run
- `unTrackGroup` has a default of `force=false` for the run

core/src/modules/ops/runs.ts
```diff
@@ -144,9 +144,10 @@ export namespace RunOps {
 
       // there are 3 phases to group runs, but only 2 really could have work, so we attribute 1/2 to each phase
       let percentComplete = Math.floor(
-        100 *
-          (membersAlreadyUpdated /
-            (totalGroupMembers > 0 ? totalGroupMembers : 1))
+        membersAlreadyUpdated === 0
+          ? 1
+          : membersAlreadyUpdated /
```
pedroslopez: It looks like this was a percentage (out of 100) before, but it's no longer being multiplied by 100. I believe the `100 *` should still apply?

core/src/modules/ops/group.ts
evantahler: Actually, it still does!  As long as the Profile has `state=pending`, the task to move it to `ready` will also check it's Groups.  I'll clarify the comment.

core/src/modules/ops/runs.ts
evantahler: Good catch! Fixed via aa13824e4a36fec8689354da6b2e271362a936fc

PR #1738, opened 2021-05-03
only auto-locally-validate direct dependencies, invalid id warning

pedroslopez:
- When using `roo validate --local id_here` or similar, only automatically apply local validation of direct dependents of `id_here`. Previous behavior would include any dependents down the tree. For example, locally validating a source would also do local validation of a destination that has a property from that source because the destination depended on the property which depended on the source
- Log warning when trying to locally validate unknown ids


evantahler (approved):
Nice catch!

PR #1741, opened 2021-05-03
throw error if bootstrapped property is an array or not unique

pedroslopez:
This prevents the ambiguous error of "Maximum call stack size exceeded" if the property was set as an array or is not unique.

### DIFF
--- a/core/src/classes/codeConfig.ts
+++ b/core/src/classes/codeConfig.ts
@@ -157,16 +157,26 @@ export function getAutoBootstrappedProperty(
   const mappingValues = Object.values(sourceConfigObject["mapping"]);
   for (const value of mappingValues) {
     // If this source id == its mapped property's sourceId, we should bootstrap the property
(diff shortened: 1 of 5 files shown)


core/src/classes/codeConfig.ts
```diff
@@ -157,16 +157,26 @@ export function getAutoBootstrappedProperty(
...
       (o) =>
         o.class.toLowerCase() === "property" &&
         o.id === value &&
-        o.sourceId === sourceConfigObject.id &&
-        o.unique &&
-        !o.isArray
+        o.sourceId === sourceConfigObject.id
     );
-    if (autoBootstrappedProperty.length > 0) {
-      return autoBootstrappedProperty[0];
```
evantahler, line 169: Assuming there can't be more than 1.  I think this is safe for now.

evantahler (approved):
Nice!

PR #1749, opened 2021-05-04
Add sync modes to Pardot plugin.

pauloouriques:


pedroslopez (changes requested):
Oh no! I just realized we've been forgetting to do something very important! We need to add the syncMode to the destination template. See example here: https://github.com/grouparoo/grouparoo/blob/d844d0bb9c405b5218b0ab9c58d909722cce4dd7/plugins/%40grouparoo/pipedrive/public/templates/destination/%7B%7B%7Bid%7D%7D%7D.js.template#L10

Please check on the other plugins we've added sync modes to and add them in another PR if they don't have it!

The rest of the implementation looks good :)

pedroslopez (approved):

evantahler (approved):

PR #1750, opened 2021-05-04
Add sync modes to OneSignal destination plugin.

pauloouriques:


plugins/@grouparoo/onesignal/src/migrations/000001-plugin-onesignal-syncMode.ts
```diff
@@ -0,0 +1,31 @@
+export default {
+  up: async function (migration) {
+    await migration.sequelize.transaction(async () => {
+      // Set all existing instances to "sync" to maintain same behavior
```
pedroslopez:
```suggestion
      // Set all existing instances to "enrich" to maintain same behavior
```

pedroslopez (changes requested):
At the moment, the `toDelete` case is still being handled by the plugin- since core should never send `toDelete: true` because of the syncMode, we should probably just `throw` here.

plugins/@grouparoo/onesignal/src/lib/export/exportProfile.ts
```diff
@@ -34,6 +34,11 @@ export const exportProfile: ExportProfilePluginMethod = async ({
   const allKeys = new Set([...newKeys, ...oldKeys]);
 
   if (toDelete) {
+    if (!syncOperations.delete) {
+      throw new Errors.InfoError(
+        "Destination sync mode does not delete existing device tags."
+      );
+    }
     // When deleting, we just clear the old tags. No real deletion occurs.
```
pedroslopez: `// When deleting, we just clear the old tags. No real deletion occurs.`

We don't need to do this anymore. Core should never send a `toDelete: true`. We were always doing this before because OneSignal only supports enriching. I think we can always throw the error if `toDelete: true`.

```js

if (toDelete) {
  throw new Errors.InfoError(
    "OneSignal does not delete existing devices."
  );
}
```

(or smething like this)

The corresponding tests for this should also be removed, and just check that it throws the error if `toDelete: true`

pauloouriques: ohhh I see

pedroslopez (commented):
Looks good! 

Not a blocking issue, but maybe we should standardize the error messages here to follow the same structure? Currently we have:
- Destination sync mode does not update existing devices.
- Destination sync mode does not delete.
- OneSignal destination does not create new devices.

pedroslopez (approved):

evantahler (approved):

PR #1752, opened 2021-05-04
fix: set options before updating property when loading from codeConfig

pedroslopez:
If a property was somehow initially created without a required option, it's impossible to validate afterwards.
This happened because of a bug with the old method of bootstrapping properties (fixed by #1726), which did not set the options on the property. When we create the bootstrapped property, the hook that causes this validation does not run, so it was able to be created in an invalid state. Because of the previous fix, this now works on a brand new database, but we still have an issue if you use the existing db with the invalid property (see pivotal #178013312) and switch over to the new auto bootstrapped property.
This errors because when we try to update the existing, invalid property, it runs validation before setting the new options on the property. The PR fixes this by setting options before updating.


evantahler (approved):
Great find and fix!

PR #1754, opened 2021-05-04
Add error message on `grouparoo run` when there are no schedules

seancdavis:
This adds the following error message when there is no schedule on run:

```
error: No schedules found. The run command uses schedules to know what profiles to import.
See this link for more info: https://www.grouparoo.com/docs/getting-started/product-concepts#schedule
If you have a schedule and are seeing this message, try `grouparoo apply` first.
```

It seemed like a nice and simple path to check if there are any schedules in the database when booting. That's why I added the line about running `apply`. Not sure if I should be looking for more than just count — filtering the schedules in some way.

I also didn't add any tests. This is one of those tricky ones in which we would want to test the failure of a process that is exited. Do we have an example of that elsewhere?

---

Other changes:

- Ran `pnpm install`, which updated the lock file.
- Removes a log that seemed unnecessary in the demo command.

### DIFF
--- a/core/src/bin/run.ts
+++ b/core/src/bin/run.ts
@@ -56,13 +56,6 @@ export class RunCLI extends CLI {
     GrouparooCLI.logCLI(this.name, false);
     this.checkWorkers();
 
-    const scheduleCount = await Schedule.count();
-    if (scheduleCount === 0) {
-      GrouparooCLI.logger.fatal(
-        `No schedules found. The run command uses schedules to know what profiles to import.\nSee this link for more info: https://www.grouparoo.com/docs/getting-started/product-concepts#schedule\nIf you have a schedule and are seeing this message, try \`grouparoo apply\` first.`
-      );
-    }
-
     if (!params.web) GrouparooCLI.disableWebServer();
     if (params.reset) await Reset.data(process.env.GROUPAROO_RUN_MODE);
     if (params.resetHighWatermarks) await Reset.resetHighWatermarks();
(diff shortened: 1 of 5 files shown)


core/src/bin/run.ts
```diff
@@ -56,6 +56,15 @@ export class RunCLI extends CLI {
...
     this.checkWorkers();
 
+    const scheduleCount = await Schedule.count();
+    if (scheduleCount === 0) {
+      log(
+        `No schedules found. The run command uses schedules to know what profiles to import.\nSee this link for more info: https://www.grouparoo.com/docs/getting-started/product-concepts#schedule\nIf you have a schedule and are seeing this message, try \`grouparoo apply\` first.`,
+        "error"
+      );
+      process.exit(1);
+    }
```
evantahler: Check out `GrouparooCLI.logger.fatal` for this kind of thing

plugins/@grouparoo/demo/src/bin/grouparoo/demo/demo.ts
```diff
@@ -23,7 +23,6 @@ export class Console extends CLI {
   }
 
   async run({ params }) {
-    console.log(JSON.stringify(params));
```
evantahler, line 26: Oops!

evantahler (changes requested):
Looks simple enough!  Check out the fatal logger helper.  Also, I wouldn't expect a `pnpm-lock` change?

evantahler:
For testing, if you use `logger.fatal`, the process won't actually exit.  Then, you can look for your message in a mocked `console.log`.  This test is like that: https://github.com/grouparoo/grouparoo/blob/main/core/__tests__/bin/generate.ts#L83-L93

core/src/bin/run.ts
```diff
@@ -70,14 +77,14 @@ export class RunCLI extends CLI {
 
     if (params.runAllSchedules) await this.runNonRecurringSchedules();
 
-    return false;
+    return true;
```
evantahler: The return method of a CLI command is `toStop` - you want to keep it as `false` or else the process will exit after this method

core/src/bin/run.ts
```diff
@@ -56,11 +56,18 @@ export class RunCLI extends CLI {
     GrouparooCLI.logCLI(this.name, false);
     this.checkWorkers();
 
+    const scheduleCount = await Schedule.count();
+    if (scheduleCount === 0) {
+      GrouparooCLI.logger.fatal(
+        `No schedules found. The run command uses schedules to know what profiles to import.\nSee this link for more info: https://www.grouparoo.com/docs/getting-started/product-concepts#schedule\nIf you have a schedule and are seeing this message, try \`grouparoo apply\` first.`
```
evantahler: I'll convert this to a multi-line string.

core/src/bin/run.ts
```diff
@@ -56,11 +56,18 @@ export class RunCLI extends CLI {
     GrouparooCLI.logCLI(this.name, false);
     this.checkWorkers();
 
+    const scheduleCount = await Schedule.count();
+    if (scheduleCount === 0) {
```
evantahler: Ideally, this would be after `main()`, so the new config files would be loaded in first before checking.  As is (as you call out), you would need to `grouparoo apply` the first time or whenever you add your first schedule

evantahler (commented):
Added some notes.  I'll fix them up and I think I have a path forward for the test suite hanging too.

evantahler:
Ha! I can't @seancdavis for a review because it's his branch.  Let me know what you think about the changes in https://github.com/grouparoo/grouparoo/pull/1754/commits/80a7ead29b19f63a8f0660c22e9fe3a5f0d8ce6d

I tried a few mocking approaches for the test that didn't work, so I ended up testing the sub-methods instead.

evantahler (commented):
🥇

seancdavis:
Cool, cool. Thank you for your help, @evantahler!!

PR #1755, opened 2021-05-04
177936242 - added CTA to join slack in app nav

teallarson:
Added button to top level nav to join our Slack community


evantahler:
Screenshots attached so we can discuss

<table><tr>
<td>
Screen Shot 2021-05-04 at 4 06 35 PM
</td>
<td>
Screen Shot 2021-05-04 at 4 06 38 PM
</td>
</tr></table>

* When the page is taller than the nav, the button floats in the middle.  Perhaps it should be linked to the bottom nav?
* Can the button be extended to fill the whole horizontal space of the navigation area?  It looks off-center as it is now
* Maybe we want to say `join our slack community`? @andyjih do you have any opinions?
* There's a button style called `outline` which might be better here to be less visually strong. 

Was the slack icon not available in the font-awesome kit?  It seems like it [might be](https://fontawesome.com/icons/slack).

teallarson:
Those all sound good.

The font awesome brands library was not functioning correctly with the font awesome library on this repo, but I can take another look.  The icon I've included is from the Slack media kit.

andyjih:
"Join our Slack community" sounds good to me, though that might cause the text to wrap to a second line. If so, "Join our Slack" is fine to me.

andyjih (changes requested):
Agreed with the changes that Evan suggested as well!

teallarson:
> "Join our Slack community" sounds good to me, though that might cause the text to wrap to a second line. If so, "Join our Slack" is fine to me.

Yep!  Word wrap woes.  Same with "Join us on Slack".  I also was considering "?  Need help?" and linking to the "/help" (which could use a face lift itself)

andyjih (approved):
👍  Nice!

core/public/images/slack-logo.svg
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
```
evantahler: It looks like you got the Slack icon from font awesome.  Is this file still needed?

ui/ui-enterprise/package.json
```diff
@@ -30,6 +30,7 @@
   },
   "dependencies": {
     "@fortawesome/fontawesome-svg-core": "1.2.35",
+    "@fortawesome/free-brands-svg-icons": "5.15.3",
```
evantahler, line 33: 👍 adding it to all 3 UI packages was required.

ui/ui-components/pages/index.tsx
```diff
@@ -53,6 +53,7 @@ export default function Page(props) {
               <Button
                 variant="primary"
                 size="lg"
+                className="signIn"
```
evantahler: For the other buttons you added an `id`, but this one you added a `class`, why the change?  

Usually, classes are for style or type (eg: this is a "submit" or "cancel" type button that looks a certain way) an IDs are to identity the specific instance of the button (eg: add-to-cart vs sign-in) both of which might be of class "submit".

teallarson: True.  There are two sign in buttons in the app and I had thought it poor form to use an id for something with more than one instance.

Also, the classes and IDs aren't explicitly needed for testing purposes anymore as the CTA button is now button-outline instead of button-primary, so they could be removed if that seems better.

core/public/images/slack-logo.svg
teallarson: Nope! removed

ui/ui-components/pages/index.tsx
teallarson: removed the unneeded css ids/classes.

evantahler:
Screenshots

Screen Shot 2021-05-07 at 6 47 29 PM
Screen Shot 2021-05-07 at 6 47 40 PM

evantahler (commented):
It's looking good! I would suggest dropping the "!" at the end, and then it's done!

evantahler (approved):

evantahler:
Looks something went wrong with one of the test on CI... re-running.

PR #1760, opened 2021-05-04
fix: update tracked group when going back to previously tracked group

pedroslopez:
Switching back to the group that was tracked when loading the page wouldn't cause any changes.

### DIFF
--- a/core/__tests__/actions/destinations.ts
+++ b/core/__tests__/actions/destinations.ts
@@ -301,22 +301,14 @@ describe("actions/destinations", () => {
         connection.params = {
           csrfToken,
           id,
-          groupId: group.id,
+          trackedGroupId: group.id,
         };
-        const { error } = await specHelper.runAction(
-          "destination:trackGroup",
-          connection
-        );
-        expect(error).toBeUndefined();
 
-        connection.params = {
-          csrfToken,
-          id,
-        };
-        const { destination } = await specHelper.runAction(
-          "destination:view",
+        const { destination, error } = await specHelper.runAction(
+          "destination:edit",
           connection
         );
+        expect(error).toBeFalsy();
         expect(destination.destinationGroup.id).toBe(group.id);
       });
 
@@ -334,6 +326,7 @@ describe("actions/destinations", () => {
           connection
         );
         expect(error).toBeFalsy();
+        expect(destination.destinationGroup.id).toBe(group.id);
         expect(destination.destinationGroupMemberships).toEqual([
           {
             groupId: group.id,
@@ -472,17 +465,23 @@ describe("actions/destinations", () => {
         connection.params = {
           csrfToken,
           id,
-          groupId: destination.destinationGroup.id,
+          trackedGroupId: null,
         };
-        const { error } = await specHelper.runAction(
-          "destination:unTrackGroup",
-          connection
-        );
-        expect(error).toBeUndefined();
+        const {
+          destination: updatedDestination,
+          error,
+        } = await specHelper.runAction("destination:edit", connection);
+        expect(error).toBeFalsy();
+        expect(updatedDestination.destinationGroup).toBe(null);
       });
 
       test("an administrator can export the members of a destination with a forced group run", async () => {
-        await specHelper.runAction("destination:trackGroup", connection);
+        connection.params = {
+          csrfToken,
+          id,
+          trackedGroupId: group.id,
+        };
+        await specHelper.runAction("destination:edit", connection);
 
         connection.params = {
           csrfToken,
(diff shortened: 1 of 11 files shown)


pedroslopez:
This works, though I'm not sure if this is something that should be decided in the UI (whether to track/untrack a group)

ui/ui-enterprise/pages/destination/[id]/data.tsx
```diff
@@ -77,7 +80,7 @@ export default function Page(props) {
```
Comment (evantahler) [inline ui/ui-enterprise/pages/destination/[id]/data.tsx:79]: This is a lot of logic to have in the UI.  It's my fault for setting it up this way.

I think the path forward is to have one API action (eg: `destination:edit`, `PUT /destination/id`) that can also accept a `trackedGroupId`, and we handle this logic in the Action.  This way the form could just submit whatever it has.

* if `trackedGroupId === destination.groupId` do nothing
* if `trackedGroupId === null || ""` call `desintaion.untrack()`
* if `trackedGroupId !== destination.groupId` call `destinatio.track(trackedGroupId)`

evantahler (changes requested):
I'm glad you found what the problem was... and I think you are right... the UI is probably overloaded at this point 😦

ui/ui-enterprise/pages/destination/[id]/data.tsx
Comment (evantahler) [reply ui/ui-enterprise/pages/destination/[id]/data.tsx:79]: This may also mean we can remove the Actions for destination:track and destination:untrack

Comment (evantahler) [reply ui/ui-enterprise/pages/destination/[id]/data.tsx:79]: I think it's also confusing that the save button on this page /sometimes/ also means "re-export everyone" (that's the case that does `await execApi("post", "/destination/${id}/export", { force: true });`.  

That should probably be moved to it's own button called "re-export group members" or similar.

pedroslopez:
Updated the edit action to handle tracking/untracking groups. I also made the force export a separate button, but the only thing I'm worried about here is it may give the impression that you _need_ to click this after updating the tracked group for an export to trigger

image

core/src/actions/destinations.ts
```diff
@@ -176,6 +177,20 @@ export class DestinationEdit extends AuthenticatedAction {
 
     await destination.update(params);
 
+    if (
+      params.trackedGroupId &&
+      params.trackedGroupId !== "_none" &&
```
evantahler, line 182: This `'_none'` thing is kind of janky, but the alternative would be to have another param like `unTrackedGroupId`, and the whole point is to remove this logic from the client... so I'm 👍 with it!

ui/ui-components/utils/apiData.ts
```diff
@@ -343,12 +341,6 @@ export namespace Actions {
   export type DestinationProfilePreview = AsyncReturnType<
     typeof DestinationProfilePreview.prototype.runWithinTransaction
   >;
-  export type DestinationTrackGroup = AsyncReturnType<
-    typeof DestinationTrackGroup.prototype.runWithinTransaction
-  >;
-  export type DestinationUnTrackGroup = AsyncReturnType<
-    typeof DestinationUnTrackGroup.prototype.runWithinTransaction
-  >;
```
evantahler, line 351: Glad you found these!  This file is how the UI's know the TS Types of the Action's responses.

core/__tests__/actions/destinations.ts
```diff
@@ -514,7 +513,12 @@ describe("actions/destinations", () => {
           })
         );
 
-        await specHelper.runAction("destination:unTrackGroup", connection);
+        connection.params = {
+          csrfToken,
+          id,
+          trackedGroupId: "_none",
+        };
+        await specHelper.runAction("destination:edit", connection);
```
evantahler: Might want to confirm here that the destination you get back from `runAction` (or the model directly) isn't tracking a group any more

core/__tests__/actions/destinations.ts
```diff
@@ -472,17 +465,23 @@ describe("actions/destinations", () => {
...
       });
 
       test("an administrator can export the members of a destination with a forced group run", async () => {
-        await specHelper.runAction("destination:trackGroup", connection);
+        connection.params = {
+          csrfToken,
+          id,
+          trackedGroupId: group.id,
+        };
+        await specHelper.runAction("destination:edit", connection);
```
evantahler: might want to confirm here that the destination now has a groupId tracked.

evantahler (approved):
Nice.
Screen Shot 2021-05-05 at 2 00 31 PM

I added some minor testing comments.  As for the second button, I think making it `outline` does hint that it's it's own thing. You could also put it next to vs below the main CTA which may help show that it's a choice?

core/src/actions/destinations.ts
pedroslopez: Yeeah, those were my thoughts as well

PR #1764, opened 2021-05-05
Allow logged models to have id values that are 191 characters in length

seancdavis:
Does the following:

- Adds a migration to change all `id`-like columns that were 40 characters to have a max length of `191`. This includes `apiKeys.apiKey`, but I was unsure if that was a relevant change.
- Adjusts Logged Model to validate against the new `191` max length. There are a couple tests for the pass and fail varieties of this new max length.

I also checked the UI. From what I could tell, we seem to be using name/title values in most places and are only really surfacing the `id` in the route. That leads to some potentially goofy routes (see screenshot below). But, otherwise, things looked find. Of course, I may not be looking in the right places.

Screen Shot 2021-05-05 at 4 04 59 PM

### DIFF
--- a/core/__tests__/classes/loggedModel.ts
+++ b/core/__tests__/classes/loggedModel.ts
@@ -30,7 +30,7 @@ describe("classes/loggedModel", () => {
         "foo-dash",
         "fooUpperLower",
         "foo-with-numb3r",
-        "foo-that-is-oh-so-much-longer-than-40-characters",
+        "a".repeat(191),
       ].forEach((id) => {
         const app = new App();
         app.id = id;
@@ -49,7 +49,7 @@ describe("classes/loggedModel", () => {
         "foo:stuff",
         "foo;stuff",
         "foo$stuff",
-        "foo-is-long-really-long-really-really-long-really-really-really-long-really-really-really-really-long-really-really-really-really-really-long-really-really-really-long-like-192-characters-long",
+        "a".repeat(192),
       ].forEach((id) => {
         const app = new App();
         app.id = id;
--- a/core/__tests__/classes/loggedModel.ts
+++ b/core/__tests__/classes/loggedModel.ts
@@ -30,6 +30,7 @@ describe("classes/loggedModel", () => {
         "foo-dash",
         "fooUpperLower",
         "foo-with-numb3r",
+        "foo-that-is-oh-so-much-longer-than-40-characters",
       ].forEach((id) => {
         const app = new App();
         app.id = id;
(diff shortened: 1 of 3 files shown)


evantahler:
Don't forget about all the migrations! 

Screen Shot 2021-05-05 at 1 12 18 PM

seancdavis:
@evantahler This was very much in progress. I didn't open as a draft and only now am seeing the tiny text to convert it to a draft.

Of course, the thing is so tiny, that in the process of converting it back to a draft, I also requested a review from you. 🤦

seancdavis:
Now we're ready for a review. See updated notes above.

core/__tests__/classes/loggedModel.ts
```diff
@@ -48,6 +49,7 @@ describe("classes/loggedModel", () => {
         "foo:stuff",
         "foo;stuff",
         "foo$stuff",
+        "foo-is-long-really-long-really-really-long-really-really-really-long-really-really-really-really-long-really-really-really-really-really-long-really-really-really-long-like-192-characters-long",
```
evantahler: 🤣

bleonard: might I suggest testing `"a".repeat(191)` and `"a".repeat(192)`

seancdavis: That's no fun.

core/src/migrations/000065-lengthen-id-columns.ts
```diff
@@ -0,0 +1,65 @@
...
+const runMigration = async ({ maxIdLength, migration, DataTypes }) => {
+  const changeColumn = async (tableName, columnName) => {
+    if (config.sequelize.dialect === "postgres") {
+      const query = `ALTER TABLE "${tableName}" ALTER COLUMN "${columnName}" SET DATA TYPE varchar(${maxIdLength}); `;
+      await migration.sequelize.query(query);
+    } else {
+      await migration.changeColumn(tableName, columnName, {
+        type: DataTypes.STRING(191),
+      });
+    }
```
seancdavis, line 44: Super interesting (and another SQLite quirk) is that SQLite does not support altering a column. Its `ALTER TABLE` functions [are super limited](https://sqlite.org/lang_altertable.html).

But that's where Sequelize comes in to save the day. So ... this ends up still looking a little goofy, but at least we don't have to mess with constraints.

evantahler (approved):
Whew! That constraint thing was scaring me. This is still a little wacky, but much better.

PR #1766, opened 2021-05-05
Exports State + DB Retry

evantahler:
Rather than rely on resque for retrying Exports, we should move this information to the DB.  This allows the `export:enqueue` task to be in charge of how many exports are sent and not be in conflict with previously enqueued & delayed send tasks

- [x] Add new columns: `export.state`, `export.sendAt` and `export.retryAttempts`
- [x] Remove Columns: `export.mostRecent`
- [x] Update `export:send/batch` tasks to write to the DB if there's a delay vs enqueuing a new task
- [x] Update `export:enqueue` to not enqueue Exports that have a retry time that hasn't come to pass yet
- [x] Display the retry information in the UI
- [x] Create a new setting for the maximum number of export retry attempts

Interesting Notes:
* If there's an export that's in the middle of a retry during this migration, and it has an error, it will become failed and not be retried. 
* An Export failing with an info-level error should still be `state = failed`.
* The Export sweeper is now one SQL statement and we can keep the latest complete export without storing `export.mostRecent` (which was the main reason we had that column in the first place).  The SQL in gnarly (see below), bit it works.

This PR also cleans up the Import/Export DIFF components, making them easier to read and appear earlier in the sport lifecycle.
Screen Shot 2021-05-07 at 3 57 23 PM


core/src/models/Export.ts
```diff
@@ -234,21 +289,29 @@ export class Export extends Model {
...
+          FROM exports e2
+          WHERE
+            e2."profileId" = exports."profileId"
+            AND state = 'complete'
+        )
+        LIMIT ${limit}
+      )
+      `,
+      { type: QueryTypes.DELETE }
+    );
```
evantahler: This is interesting - in SQL alone we can delete all old exports, but leave the latest complete one.

core/src/models/Export.ts
```diff
@@ -234,21 +289,52 @@ export class Export extends Model {
...
+                AND state = 'complete'
+            )
+          )
+        )
+        LIMIT ${limit}
+      )
+     ${config.sequelize.dialect === "postgres" ? "RETURNING id" : ""}
+      ;`,
+      { type: QueryTypes.SELECT }
+    );
```
evantahler: Removing `Export.mostRecent` was simple... except for the sweeper.  This Query will keep the the most recent `complete` export per profile.   There's different logic between SQLite and Postgres to know how many Exports were delete.

core/src/initializers/settings.ts
```diff
@@ -117,6 +117,14 @@ export class Plugins extends CLSInitializer {
...
         type: "number",
       },
+      {
+        key: "exports-max-retries-count",
+        title: "Exports Maximum Retry Count",
+        defaultValue: 5,
+        description:
+          "How many times should we attempt to retry an Export before considering it failed?",
+        type: "number",
+      },
```
evantahler, line 121: What's a good default value here?  5 might be too low.

core/src/migrations/000066-addExportStatesAndRetry.ts
```diff
@@ -0,0 +1,72 @@
...
+      await migration.addColumn("exports", "state", {
+        type: DataTypes.STRING(191),
+        allowNull: true,
+      });
+
+      await migration.bulkUpdate(
+        "exports",
+        { state: "failed" },
+        { state: null, errorMessage: { [Op.ne]: null } }
+      );
```
evantahler, line 15: TODO: If there are Exports mid-retry with a first error, we should put them into the `pending` state, not `failed`.  Any suggestions for a good query to determine this?

core/src/migrations/000066-addExportStatesAndRetry.ts
```diff
@@ -0,0 +1,72 @@
...
+      await migration.bulkUpdate(
+        "exports",
+        { state: "pending" },
+        { state: null }
+      );
+
+      await migration.changeColumn("exports", "state", {
+        type: DataTypes.STRING(191),
+        allowNull: false,
+      });
```
evantahler, line 32: TODO: should default to "NOW()" for those in the `pending` state

core/__tests__/models/destination/plugins/exportProfile.ts
```diff
@@ -749,21 +749,28 @@ describe("models/destination", () => {
...
-      // the task should be re-enqueued with no error
       await specHelper.runTask("export:send", foundSendTasks[0].args[0]);
-      foundSendTasks = await specHelper.findEnqueuedTasks("export:send");
-      expect(foundSendTasks.length).toBe(1 + 1);
+
       await _export.reload();
+      expect(_export.state).toBe("pending");
       expect(_export.completedAt).toBeFalsy();
+      expect(_export.errorMessage).toBeFalsy();
+      expect(_export.sendAt.getTime()).toBeGreaterThan(new Date().getTime());
```
pedroslopez, line 758: maybe also check that `retryCount` works as expected here?

core/src/initializers/settings.ts
pedroslopez: Do we have any sort of historical data here to inform this decision? (average time something is retried before it's successful or something like that). 5 does seem kind of low if it includes rate limits.

core/src/modules/ops/destination.ts
```diff
@@ -291,9 +291,9 @@ export namespace DestinationOps {
       where: {
         destinationId: destination.id,
         profileId: profile.id,
-        mostRecent: true,
-        errorMessage: null,
+        state: "complete",
```
pedroslopez, line 299: Note that because we're marking exports that failed with an info error as `complete`, this will consider the latest one with an info error as the most recent export, which is different from the previous behavior. I think the right thing to do is _not_ consider `info` ones as valid because it indicates that the export should have not had any effect on the destination.

ui/ui-components/components/export/diff.tsx
```diff
@@ -2,62 +2,49 @@ import { Badge } from "react-bootstrap";
...
-                _export.newProfileProperties[k] !== undefined ? (
-                  <>
-                    {" "}
-                    | <Badge variant="success">+</Badge>&nbsp;
-                    {_export.newProfileProperties[k].toString()}
-                  </>
-                ) : null}
-              </>
+            {hasChanges ? (
+              <Badge variant="warning">-/+</Badge>
```
pedroslopez: I like this `-/+` badge for the changed properties, but I feel like maybe it would be clearer/cleaner to keep `+` when a property is added (`null` -> `[value]` ) and `-` when it's removed (`[value]` -> `null`).

core/src/modules/ops/destination.ts
pedroslopez: Hmm... reading the PR's description it suggest `info` would be `complete` but I don't think I see this happening in the code, so maybe it's fine? Can you clarify what the behavior is?

core/src/initializers/settings.ts
evantahler: No... 
Alternatively, we could use a 'time' limit vs a 'count' limit.  For example, if it's been "more than 6 hours" since the Export was created, stop retrying.

ui/ui-components/components/export/diff.tsx
evantahler: Good suggestion - added in https://github.com/grouparoo/grouparoo/pull/1766/commits/4809c7deb9db2cedc62911ad1aff1a4bbafe6d61

core/src/modules/ops/destination.ts
evantahler: I think you are right! An inforError is still a state of `failed`.  Clarified and updated via ea1c95d3dc50014b2fdf09a6a2b040f38b956e53

core/__tests__/models/destination/plugins/exportProfile.ts
evantahler: 👍 done in 2b2c3ea5

core/src/migrations/000066-addExportStatesAndRetry.ts
evantahler: I don't think this can be done... that's one of the previous problems of relying on resque for retries :/

core/src/migrations/000066-addExportStatesAndRetry.ts
evantahler: Added in 7255fd88

core/src/models/Export.ts
```diff
@@ -234,21 +295,80 @@ export class Export extends Model {
...
+      `
+      DELETE FROM exports
+      WHERE id IN (
+        SELECT id FROM exports
+        WHERE "createdAt" < '${whereDate}'
+        AND "createdAt" < (
+          SELECT MAX("createdAt")
+          FROM exports e2
+          WHERE
+            e2."profileId" = exports."profileId"
```
bleonard, line 317: not clearing failed?

bleonard: This might be a `n+1` kind of query.
I wonder if if should be something like DELETE from exports where created < ? AND mostRecent = FALSE` and leave around mostRecent.

evantahler: `mostRecent` is gone now! that's what this all trying to work around

bleonard: yeah, this workaround made me think we should talk about keeping it or otherwise making a `sweepable` boolean pattern or something.

PR #1768, opened 2021-05-06
Add sync modes to Sendgrid destination plugin.

pauloouriques:


plugins/@grouparoo/sendgrid/__tests__/export/export-profile.ts
```diff
@@ -398,6 +455,36 @@ describe("sendgrid/exportProfile", () => {
     expect(oldUser).toBe(null);
   });
 
+  test("it can change the email address and orphan the old user if sync mode is not deleting", async () => {
+    // hubspot requires deleting the old user on FK change
```
evantahler: Sendgrid

evantahler (approved):

plugins/@grouparoo/sendgrid/__tests__/export/export-profile.ts
```diff
@@ -398,6 +455,36 @@ describe("sendgrid/exportProfile", () => {
...
+      toDelete: false,
+    });
+
+    await indexContacts(newNock);
+
+    // old user still there
+    const oldUser = await apiClient.getUser(alternativeEmail);
+    expect(oldUser).not.toBe(null);
+    expect(oldUser.email).toBe(alternativeEmail);
+    expect(oldUser["list_ids"]).toHaveLength(0); // but has been removed from lists
```
pedroslopez, line 480: The user with alternativeEmail did not have any lists to start with, so this would have length 0 anyway. You may want to add these groups to the previous test where this user is created to ensure it has the lists in the first place and it's now being removed from those lists. My initial impression is that this would not happen.

pedroslopez (changes requested):

pedroslopez (approved):

PR #1769, opened 2021-05-06
fix: trigger export when saving destination data

pedroslopez:


core/src/actions/destinations.ts
```diff
@@ -189,6 +190,8 @@ export class DestinationEdit extends AuthenticatedAction {
       destination.groupId
     ) {
       await destination.unTrackGroup();
+    } else if (params.triggerExport) {
+      await destination.exportGroupMembers();
```
pedroslopez: Should this be forced?

pedroslopez: (yes, nothing is actually exported if it's not forced)

evantahler: Yep - because it will look like all the exports have no changes otherwise.

evantahler (approved):
Thanks for changing this @pedroslopez, and I'm sorry I didn't catch this yesterday.

Confirming the behavior here:
* Back to one button on the page
* That button will always create a run, regardless of if the group being tracked has changed or not.

In another story (https://www.pivotaltracker.com/story/show/178055151) we can decide if we want to not enqueue the run if there are no changes to the destination

PR #1770, opened 2021-05-06
Add sync modes to Customer.io destination plugin.

pauloouriques:


plugins/@grouparoo/customerio/src/lib/client.ts
```diff
@@ -0,0 +1,57 @@
+import CIO from "customerio-node";
+import axios from "axios";
+
+class CustormerioClient {
```
pedroslopez: Note: typo
```suggestion
class CustomerioClient {
```

plugins/@grouparoo/customerio/src/initializers/plugin.ts
```diff
@@ -50,6 +55,12 @@ export class Plugins extends Initializer {
               required: true,
               description: "Customer.io Tracking API key",
             },
+            {
+              key: "appApiKey",
```
pedroslopez, line 59: We may want to consider renaming `apiKey` and its display name `API Key` to reflect that it's the track api (`trackApiKey`, `Tracking API Key`). Migration would be necesary.

@evantahler Note that this is adding a new required option to this plugin, which could be considered a breaking change. I _think_ you mentioned we didn't have people using this so it's fine? Still something to keep in mind.

plugins/@grouparoo/customerio/src/lib/client.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor(siteId, apiKey, appApiKey) {
+    this.client = new CIO(siteId, apiKey);
+    this.appApiKey = appApiKey;
+  }
+
+  async destroy(customerId: string) {
+    return this.client.destroy(customerId);
+  }
+
+  async identify(customerId: string, payload: string) {
```
pedroslopez: the payload type seems wrong here, should be an object?

plugins/@grouparoo/customerio/src/lib/client.ts
```diff
@@ -0,0 +1,57 @@
+import CIO from "customerio-node";
+import axios from "axios";
+
+class CustormerioClient {
+  client: any;
+  appApiKey: any;
```
pedroslopez: this should probably be a string

plugins/@grouparoo/customerio/__tests__/export/export-profile.ts
```diff
@@ -294,16 +344,38 @@ describe("customer.io/exportProfile", () => {
...
+    oldCustomer = await cioClient.getCustomer(otherCustomerId);
     expect(oldCustomer).toBe(null);
 
-    newCustomer = await getCustomer(customerId);
+    newCustomer = await cioClient.getCustomer(customerId);
     expect(newCustomer.id).toBe(customerId);
     expect(newCustomer.attributes.email).toBe(otherEmail);
   });
 
+  test("cannot delete an user if sync mode does not allow it", async () => {
```
pedroslopez:
```suggestion
  test("cannot delete a user if sync mode does not allow it", async () => {
```

plugins/@grouparoo/customerio/public/templates/app/{{{id}}}.js.template
```diff
@@ -7,7 +7,8 @@ exports.default = async function buildConfig() {
       type: "{{{__pluginName}}}",
       options: {
         siteId: "...", // Customer.io Site ID
-        apiKey: "..." // Customer.io Track API Key
+        apiKey: "...", // Customer.io Track API Key
+        appApiKey: "..." // Customer.io App API Keys
```
pedroslopez:
```suggestion
        appApiKey: "..." // Customer.io App API Key
```

plugins/@grouparoo/customerio/public/templates/app/{{{id}}}.js.template
```diff
@@ -7,7 +7,8 @@ exports.default = async function buildConfig() {
       type: "{{{__pluginName}}}",
       options: {
         siteId: "...", // Customer.io Site ID
-        apiKey: "..." // Customer.io Track API Key
+        apiKey: "...", // Customer.io Track API Key
```
pedroslopez:
```suggestion
        apiKey: "...", // Customer.io Tracking API Key
```

pedroslopez (changes requested):
Sync modes look good!

Did we check that the foreign key changes are working as expected? It seems like we are only getting the new user, which suggests we're missing `newUser || oldUser` somewhere... Please add some tests for the common cases (email change + toDelete where both exist, only old exists, only new exists; email change + update where both exist, only old exists, only new exists)

pauloouriques (commented):
login > logic

plugins/@grouparoo/customerio/src/initializers/plugin.ts
evantahler: Decided to release this new change without a minor semver bump.  We will be sure to mention the breaking change in Slack and in a 'whats new' post

pedroslopez: I think now's the time to change `apiKey` to `trackApiKey` so it's clearer in code config as well then

plugins/@grouparoo/customerio/public/templates/app/{{{id}}}.js.template
```diff
@@ -7,7 +7,8 @@ exports.default = async function buildConfig() {
       type: "{{{__pluginName}}}",
       options: {
         siteId: "...", // Customer.io Site ID
-        apiKey: "..." // Customer.io Track API Key
+        apiKey: "...", // Customer.io Tracking API Key
```
pedroslopez: Non-blocking comment, but to match `appApiKey` and be clearer, now is probably a good time to rename this to `trackApiKey` or `trackingApiKey`

plugins/@grouparoo/customerio/src/lib/export/exportProfile.ts
```diff
@@ -19,21 +20,27 @@ export const exportProfile: ExportProfilePluginMethod = async ({
...
-
-  if (oldCustomerId !== undefined && customerId !== oldCustomerId) {
-    // Must delete old customer if ID has changed
-    await client.destroy(oldCustomerId);
-  }
+  newCustomer = await client.getCustomer(customerId);
 
   if (toDelete) {
-    await client.destroy(customerId);
+    const oldCustomer = await client.getCustomer(oldCustomerId);
```
pedroslopez, line 31: Just to be sure, their API doesn't throw an error here if `oldCustomerId` is undefined, right?

plugins/@grouparoo/customerio/src/lib/test.ts
```diff
@@ -8,6 +8,8 @@ export const test: TestPluginMethod = async ({ appOptions }) => {
     name: "grouparoo_test_connection",
   });
 
+  await client.getCampaigns();
```
evantahler: Is there any information we get from this response we can share with the user?  A message like `Connected.  Found 10 campaigns.` would be helpful

plugins/@grouparoo/customerio/public/templates/app/{{{id}}}.js.template
evantahler: In the comments, since there are now 2, it might be helpful to provide a URL they can get each key from (if they are different)

plugins/@grouparoo/customerio/src/initializers/plugin.ts
```diff
@@ -50,6 +55,12 @@ export class Plugins extends Initializer {
...
+              displayName: "App API Key",
+              required: true,
+              description: "Customer.io App API key",
```
evantahler, line 62: How are the APIkeys different?

pauloouriques: is each for a separated API, the existing one for tracking and the other for application.

plugins/@grouparoo/customerio/src/lib/export/exportProfile.ts
pauloouriques: nope, It returns 404 which is treated by the client class.

evantahler (approved):
Thanks!

plugins/@grouparoo/customerio/src/migrations/000001-plugin-customerio-syncMode.ts
```diff
@@ -0,0 +1,31 @@
+export default {
+  up: async function (migration) {
```
pedroslopez: We should also migrate the existing `apiKey` option to `trackingApiKey` here (and do the inverse in the `down` section)

pedroslopez (changes requested):

evantahler:
From standup: We should migrate the old apiKey name to the new one

pedroslopez (commented):
Both migrations start with `000001-`, one of these should probably be `2`

pedroslopez (changes requested):
We shouldn't change existing migrations, as they may have already run. If you want to combine migrations into one, you can combine the apikey change and sync modes into one (000002) because they're being added by this PR. In any case the exisiting one (000001-plugin-customerio-renameApp) should be left untouched.

pedroslopez (approved):

evantahler (approved):

PR #1773, opened 2021-05-07
enqueue deletion of various models in code config

pedroslopez:
Currently an error occurs if you try to delete an object and an object that depends on it at the same time. We're moving some models to also be deleted async so we can mark them as deleted and complete the deletion when possible.

We may want to later add another story for `roo apply` to have a parameter that waits until deletion is done, or a parameter to force deletion without draining. At the moment the command returns once the async-deleted models have been marked as `deleted`, which means if you try to add them again before the deletion actually happens, you'll get some errors due to duplicate keys. 

At the moment this is only considering async deletes for code config to be able to wait for other stuff to be deleted, but we could also consider using this from the UI (specially for apps- maybe we want the app deletion to trigger deleting the destination)


core/src/config/tasks.ts
```diff
@@ -23,7 +23,13 @@ export const DEFAULT = {
           api?.plugins?.plugins
             .filter((plugin) => plugin.apps?.length > 0)
             .map((plugin) => plugin.apps.map((app) => `exports:${app.name}`)),
-          ["profiles", "profileProperties", "destinations", "default"]
+          [
+            "profiles",
+            "profileProperties",
+            "destinations",
+            "properties",
```
pedroslopez, line 30: Does this order matter?

bleonard:
One thing I'm curious about here is sensitivity to redis having issues or getting cleared.
Do we need a sweeper task to make sure we are are working on "deleting" items?

core/src/config/tasks.ts
evantahler: Yes! The order is worked per the array, eg: all "system" tasks happen before any "groups" tasks.  This is how we to batching - ensuring that any properties are imported before the run come around again making more tasks.

pedroslopez:
> One thing I'm curious about here is sensitivity to redis having issues or getting cleared.
> Do we need a sweeper task to make sure we are are working on "deleting" items?

I think that's a good idea, else they could potentially get stuck in the deleting state. It doesn't seem like the current models that are deleted in this way do this either, so it would probably be good to consider those as well in the recurring task.

core/__tests__/modules/codeConfig/codeConfig.ts
```diff
@@ -477,11 +477,72 @@ describe("modules/codeConfig", () => {
...
+
+      // clean up the queue for other tests
+      await specHelper.deleteEnqueuedTasks(
+        "property:destroy",
+        foundTasks[0].args[0]
+      );
+      await specHelper.deleteEnqueuedTasks(
+        "property:destroy",
+        foundTasks[1].args[0]
+      );
```
evantahler: FYI you can always ` await api.resque.queue.connection.redis.flushdb();` to nuke the queue between tests

core/__tests__/tasks/group/destroy.ts
```diff
@@ -45,7 +45,6 @@ describe("tasks/group:destroy", () => {
     });
```
evantahler, line 61: It might be helpful to add:
* The group will not be deleted if a destination is tracking this group
* The group will not be deleted if a destination is sending this group's memberships

core/__tests__/tasks/source/destroy.ts
```diff
@@ -0,0 +1,70 @@
...
+      expect(reloadedSource.state).toBe("deleted");
+
+      // remove property
+      await property.destroy();
+
+      // run it again
+      await specHelper.runTask("source:destroy", foundTasks[0].args[0]);
+      await expect(source.reload()).rejects.toThrow(/does not exist anymore/);
+    });
+  });
```
evantahler, line 69: Note: There's a specific story for deleting the bootstrapped property as well as the source https://www.pivotaltracker.com/story/show/178112482.  That's not expected to work as of yet.

core/src/actions/properties.ts
```diff
@@ -340,7 +340,7 @@ export class PropertyProfilePreview extends AuthenticatedAction {
 
     apiData.properties[property.key] = {
       id: property.id,
-      state: property.state,
+      state: null,
```
evantahler, line 343: Why was this changed?

core/src/modules/configLoaders/app.ts
```diff
@@ -57,7 +57,7 @@ export async function deleteApps(ids: string[]) {
   });
 
   for (const i in apps) {
-    await apps[i].destroy();
+    await apps[i].update({ state: "deleted", locked: null });
```
evantahler, line 60: To delete a locked model, we need to unlock it - makes sense!

core/src/tasks/group/destroy.ts
```diff
@@ -1,79 +1,53 @@
...
+      // kick off a run in the `runRemoveGroupMembers` stage
+      // this will be processed by run:tick -> group:run
+      return Run.create({
         creatorId: group.id,
         creatorType: "group",
         state: "running",
+        groupMemberOffset: 0,
+        groupMemberLimit: limit,
+        groupMethod: "runRemoveGroupMembers",
       });
```
evantahler: I think you may want to add a test for this explicitly, since this is special behavior:
> test("If the group is in the deleted state, and there is not a run, `task:group:delete` will make a new run")

core/src/tasks/property/destroy.ts
```diff
@@ -0,0 +1,39 @@
...
+    const property = await Property.scope(null).findOne({
+      where: { id: params.propertyId, state: "deleted" },
+    });
+
+    // the property may have been force-deleted
+    if (!property) return;
+
+    // check if we're still being used by something
+    try {
+      await Property.ensureNotInUse(property);
```
evantahler, line 27: There are different names to some of these validations: `.ensureNotInUse()`, `. checkDependents ()`, `. ensureNoProperties ()`, etc.  Perhaps it is time to unify them.

core/src/modules/configLoaders/index.ts
```diff
@@ -322,5 +323,9 @@ export async function deleteLockedObjects(seenIds) {
   // back to normal
   deletedIds["app"] = await deleteApps(seenIds.app);
 
+  // Enqueue deletion tasks
+  const task: CLSTask = api.tasks.tasks["destroy"];
+  await task.runWithinTransaction({}, {});
+
```
evantahler: I'm guessing this is for the case where we are running via `grouparoo apply`?  In some of the other cases (`grouparoo start` and `grouparoo run`) the task will be handled for us, and in `grouparoo validate` we probably don't want to run this.

If this is added for the `grouparoo apply` case, perhaps we run it from within `src/bin/apply` to indicate that it's special for that command.

evantahler (changes requested):
This is great - I like the task-driven architecture that keeps all the state in the DB.  Resilient! 

Some comments below, mostly about ensuring we have thorough testing, but overall this looks to be a good path forward!

core/__tests__/tasks/group/destroy.ts
pedroslopez: Added point 1, but for 2 at least at the moment this isn't being checked and instead an `AfterDestroy` hook takes care of deleting those. Should this be changed?

core/src/modules/configLoaders/index.ts
pedroslopez: Yeah it was for the apply case, but I just removed it entirely. Even if you run `apply` I don't think there's much of a benefit to enqueuing the tasks here. They won't actually be processed until you run the server, where in 30s when the recurring task runs they'll get enqueued anyway.

core/src/actions/properties.ts
pedroslopez: The type for this object is `ProfilePropertyType` and it should actually be referring to the `ProfileProperty`'s state, but here it was being set to the `Property`'s state. Previously the types were compatible, but since we added a `deleted` state to `Property`, typescript now gives an error here. I actually fixed it previously doing by accepting Property state there too (see https://github.com/grouparoo/grouparoo/pull/1773/commits/bb25e02533c8a564561ec4528c77ca780cc5ac01#diff-20d91e3e2a7bfa969dce006af2bd727a09d0d97f85fff330ac30e936d0ffa727L18-L19) but I didn't think that was the right thing and it doesnt seem like that state isn't really being used.

Let me know if there's a better approach here.

core/__tests__/tasks/source/destroy.ts
pedroslopez: Note on the note: it's not expected to work yet when working in _enterprise ui_, but it was handled previously for code config.

core/src/modules/configLoaders/index.ts
evantahler: It looks like the `task.runWithinTransaction()` is still here?

evantahler (commented):
Looking good all around! Just one comment about enqueueing the task when code-configuring

core/src/modules/configLoaders/index.ts
pedroslopez: derp, missed the commit for that. nice catch!
should be good now.

PR #1776, opened 2021-05-10
Support batch imports with non-unique primary keys

evantahler:
This PR fixes a bug for batch table sources that were requesting non-unique data.  For example, if may Profiles had an `account_id` and `account_name`, the query to load `account_name` would sometimes produce `null` results if there were many Profiles in the import batch with the same `account_id`.

This bug effects Postgres, Mysql, BigQuery, Redshift, and Snowflake sources. 

TODO:
- [x] find the problem
- [x] add tests for each source
  - [x] Postgres
  - [x] MySQL
  - [x] BigQuery
  - [x] Snowflake


evantahler:
@bleonard - any tips on getting the nock for BigQuery to work?  New data is uploaded to BigQuery, and when re-recording, the test passes OK...

plugins/@grouparoo/bigquery/__tests__/utils/nockHelper.ts
```diff
@@ -47,8 +48,20 @@ export const updater = {
...
+      let response = "";
+      const lines = nockCall.split(os.EOL);
+      for (let line of lines) {
+        if (line.includes(".post('/oauth2/v4/token', \"grant_type=urn")) {
+          line = ` .post("/oauth2/v4/token", { grant_type: "urn:ietf:params:oauth:grant-type:jwt-bearer", assertion: /.+/g })`;
+        }
+        response += `${line}${os.EOL}`;
+      }
+      return response;
+    }
```
evantahler, line 63: Shady, but this manual mocking of oAuth requests got bigquery working again

PR #1778, opened 2021-05-11
If a Profile Property's dependencies are not ready, update `startedAt` try again soon

evantahler:
This PR updates our retry logic when a Profile Property is to be imported, but it has a dependency that isn't ready yet.  For example in this chain: `user_id -> account_id -> account_name` if account_id is not yet ready and we are trying to import account_name.

Prior to this PR, account_name's `startedAt` would be untouched, so it would not be retried until (default) 5 minutes later.  This PR updates `startedAt` so that it can be retied again as soon as ~5 seconds later (as a function of `config.tasks.timeout`)

### DIFF
--- a/core/src/tasks/profileProperty/importProfileProperty.ts
+++ b/core/src/tasks/profileProperty/importProfileProperty.ts
@@ -1,9 +1,10 @@
 import { RetryableTask } from "../../classes/tasks/retryableTask";
 import { Profile } from "../../models/Profile";
 import { Property } from "../../models/Property";
-import { log } from "actionhero";
+import { config } from "actionhero";
 import { ProfileProperty } from "../../models/ProfileProperty";
 import { PropertyOps } from "../../modules/ops/property";
+import { ImportOps } from "../../modules/ops/import";
 
 export class ImportProfileProperty extends RetryableTask {
   constructor() {
@@ -46,7 +47,19 @@ export class ImportProfileProperty extends RetryableTask {
       if (properties[dep.key].state !== "ready") ok = false;
     });
 
-    if (!ok) return; // there's a dependency we don't have yet
+    if (!ok) {
+      // there's a dependency we don't have yet, sleep a little and will be retried later
+      return ProfileProperty.update(
+        { startedAt: ImportOps.retrySendAt() },
+        {
+          where: {
+            propertyId: property.id,
+            profileId: profile.id,
+            state: "pending",
+          },
+        }
+      );
+    }
 
     const propertyValues = await source.importProfileProperty(
       profile,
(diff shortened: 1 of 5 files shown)


core/src/tasks/profileProperty/importProfileProperty.ts
```diff
@@ -46,7 +47,19 @@ export class ImportProfileProperty extends RetryableTask {
       if (properties[dep.key].state !== "ready") ok = false;
     });
 
-    if (!ok) return; // there's a dependency we don't have yet
+    if (!ok) {
+      // there's a dependency we don't have yet, sleep a little and will be retried later
+      return ProfileProperty.update(
+        { startedAt: ImportOps.retryStartedAt() },
```
bleonard, line 53: it seems potentially problematic to move this time backwards. would it work to set it to `null` ? There's something weird in that about how `processPendingProfileProperties` orders things always getting the same ones over and over again, though, so I understand the need to "bump it ahead".

Conceptually you want to rerun these ones after all the dependencies are filled. That could be a little time or quite a bit.

Things worth considering:
* reenqueue in resque (it will still get picked up after 5 minutes if redis goes away)
* use some stuff from how we are thinking about exports and have a `fetchAt` - that more or less what you are trying to get at with manipulating the `startAt` I think.
* understand the DAG more in `processPendingProfileProperties` or wherever we mark them "pending" - if properties had a `depth` field to order by, this case would come up much less commonly.

bleonard: I'd be ok merging too, but noting this is technical debt and a weird spot.

evantahler: Yeah - it felt weird.  I like option 1, keeping the 5 minute retry as a fallback if redis goes away

bleonard: The trick would be to only do that a few times (so there isn't one in resque while it's getting re-enqueued) because it's been five minutes.
So that might look like checking when that would be (by getting the setting and adding it to startedAt). if that's "soon" (within 20 seconds), drop it on the floor. If it's more than that. re-enqeue for 10 seconds later. You could also check the length of how many are pending to get a sense.

The main downside of this approach is that this this re-enqueued batch would be smaller and you'd end up with lots of little ones. Whereas the messing with the db ones (including the merged startedAt approach) keeps big batch sizes.

evantahler: Yeah... that's why I started with this PR to keep it all in the DB and maintain large batches.  We could just retry once, and then if that fails, wait the 5 minutes.    Or... leave it as-is!

bleonard: as-is is ok with me. let's see if more things go the way of `exports` with their enqueuing logic and we can revisit.

PR #1780, opened 2021-05-12
final summary after `roo run`

teallarson:
Generates a summary after each instance of `roo run` giving details about Profiles, Sources, and Destinations affected by the run.
Screen Shot 2021-05-19 at 2 28 28 PM

### DIFF
--- a/core/__tests__/modules/status.ts
+++ b/core/__tests__/modules/status.ts
@@ -50,7 +50,6 @@ describe("modules/status", () => {
   describe("with many metrics", () => {
     beforeAll(async () => {
       await api.resque.queue.connection.redis.flushdb();
-
       await Status.set([metric]);
       await utils.sleep(10);
       await Status.set([metric]);
@@ -285,7 +284,6 @@ describe("modules/status", () => {
           undefined
         );
         await helper.changeTimestamps(_export3, true);
-
 
         const now = new Date();
         await _export.update({
--- a/core/__tests__/modules/status.ts
+++ b/core/__tests__/modules/status.ts
@@ -207,6 +207,7 @@ describe("modules/status", () => {
       await helper.factories.properties();
       oldProfile = await helper.factories.profile();
       await helper.changeTimestamps(oldProfile, true); // 'true' will set both updatedAt and createdAt
+
       newProfile = await helper.factories.profile();
     });
 
@@ -285,6 +286,7 @@ describe("modules/status", () => {
         );
         await helper.changeTimestamps(_export3, true);
 
+
         const now = new Date();
         await _export.update({
           completedAt: new Date(now),
@@ -300,3 +302,13 @@ describe("modules/status", () => {
     });
   });
 });
+
+async function changeTimestamps(
+  instance: Profile | Export,
+  timestamp = "1900-01-01 12:13:14"
+) {
+  const tableName = instance.constructor["tableName"];
+  return api.sequelize.query(
+    `UPDATE ${tableName} SET "updatedAt" = '${timestamp}', "createdAt" = '${timestamp}' WHERE id = '${instance.id}'`
+  );
+}
--- a/core/__tests__/modules/status.ts
+++ b/core/__tests__/modules/status.ts
@@ -206,7 +206,7 @@ describe("modules/status", () => {
       await helper.truncate();
       await helper.factories.properties();
       oldProfile = await helper.factories.profile();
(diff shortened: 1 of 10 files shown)


core/src/modules/statusReporters.ts
```diff
@@ -414,3 +413,143 @@ function mergeMetrics(metrics: StatusMetric[]) {
...
+  const lastRunStart = new Date(api.bootTime);
+  export namespace Sources {
+    export interface SourceData {
+      name: string;
+      profilesCreated: number;
+      profilesImported: number;
+      importsCreated: number;
+      error: string;
+    }
+    export async function getData(): Promise<Array<SourceData>> {
```
teallarson, line 426: @evantahler We'd discussed typing returns inside a function vs. when the function is declared.  How might I do that in this case?

evantahler: You might be close already as what you are returning is a function of the already-typed:

```
      const sources: {        [id: string]: {          name: string;          profilesCreated: number;          profilesImported: number;          importsCreated: number;          error: string;        };      } = {};
```

You could re-work that to be something like:

```ts
const sources: {[id:string]: SourceData } = {}
```

which may help

core/__tests__/modules/status.ts
```diff
@@ -178,4 +182,116 @@ describe("modules/status", () => {
...
+
+async function changeTimestamps(
+  instance: Profile | Export,
+  timestamp = "1900-01-01 12:13:14"
+) {
+  const tableName = instance.constructor["tableName"];
+  return api.sequelize.query(
+    `UPDATE ${tableName} SET "updatedAt" = '${timestamp}', "createdAt" = '${timestamp}' WHERE id = '${instance.id}'`
+  );
+}
```
evantahler: This function is because there is no good way with Sequelize to manually change `updatedAt`.  @tealjulia I think we have a method like this in a few places now.  Can you please move this into `@grouparoo/spec-helper`?   Then this method will be available via something like `helper.changeTimesamps(model)`

core/src/modules/statusReporters.ts
evantahler: Were you unable to type the return values to the function didn't need to be typed?

evantahler: Noting that this summary will only include sources which had a run.

core/src/modules/statusReporters.ts
```diff
@@ -414,3 +412,135 @@ function mergeMetrics(metrics: StatusMetric[]) {
...
+
+  export namespace Destinations {
+    export interface DestinationData {
+      name: string;
+      exportsCreated: number;
+      exportsFailed: number;
+      exportsComplete: number;
+    }
+
+    export async function getData() {
```
evantahler, line 499: Similarly, this summary will only include Destinations which had at least one Export

core/src/tasks/system/status.ts
```diff
@@ -24,11 +24,20 @@ export class StatusTask extends Task {
...
+      await this.stopServer(toStop);
+    }
 
     await this.updateTaskFrequency();
   }
 
+  async logFinalSummary() {
+    const finalSummaryLog = await FinalSummary.getFinalSummary();
+    GrouparooCLI.logger.finalSummary(finalSummaryLog);
+    return false;
```
evantahler: You aren't using the return value, to it's best to return nothing.

evantahler (changes requested):
Looks really good!  

* Please move the changeTimestamps utility to the `@grouparoo/specHelper`
* Please don't return a value if you don't use it 

Then this is good to go!

teallarson:
> Looks really good!
> 
> * Please move the changeTimestamps utility to the `@grouparoo/specHelper`
> * Please don't return a value if you don't use it
> 
> Then this is good to go!

@evantahler not entirely clear what's going on with these failures installing Postgres.  I'm not able to reproduce them on my machine.  Versioning?  Related to CircleCI's partial outage?

core/__tests__/modules/status.ts
```diff
@@ -193,4 +196,117 @@ describe("modules/status", () => {
...
```
evantahler: I don't think you need this any more, as it's in the helper?

evantahler (changes requested):
One minor thing to remove and this is good to go!

evantahler (approved):
Ship it!

PR #1784, opened 2021-05-13
New Publishing Workflow

evantahler:
**THE NEW GROUPAROO PUBLISHING PROCESS**

When the `main` branch is in a good place and you want to publish it, run the Github `publish` workflow 

The `./bin/publish` command:
  1. Choose your release mode:
    * If you want to make an alpha release (for testing), run `./bin/publish alpha`
    * If you want to make a stable release (for customers), run `./bin/publish stable`  
  2. This command will:
       a. Update the versions of all packages in the monorepo
       b. Create a tag for the new version on Github
       c. Generate release notes for that tag and publish them
       d. Publish new packages to NPM

The new `publish.yaml` workflow:

* Will be run weekly on `main` every Thursday to create the alpha releases
* Can be run manually as a Github Action and told to cut a stable release as needed.

Changes:
  * No more "stable" branch.  The tags are what matters.
  * Publication is divorced from CI.  
  * CircleCI does not do any publishing anymore
  * We don't need to rebuild the lockfile after publish, the `update-lockfile` action does that for us now


evantahler:
An alternative approach would be to somehow 'lock' the main branch during a deploy, but didn't find a good API for that (something about branch protections and removing all access)

evantahler:
Talking with @seancdavis Here are some better ideas:
* why do we even have a stable branch?  This isn't SaaS! Perhaps all we need to trigger a mainline release is a PR that changes the version number to a non-alpha release?
* we shouldn't be writing to code in CI and pushing back.  That's weird
* maybe we can make a github action to help us bump the versions (workflow_dispatch + inputs)

PR #1785, opened 2021-05-13
Add sync modes to Sailthru destination plugin.

pauloouriques:


pedroslopez (changes requested):
Looking good! I don't see any tests for the case where the sync mode doesn't allow deleting. We should probably add some for that.

pedroslopez (approved):

plugins/@grouparoo/sailthru/src/lib/client.ts
```diff
@@ -40,16 +38,10 @@ export default class SailthruClient {
...
-
     try {
       const response: any = await this.get("user", payload);
       return response.keys.sid;
     } catch (err) {
-      // {
-      //   statusCode: 400,
-      //   error: 99,
-      //   errormsg: 'User not found with email: brian@bleonard.com'
-      // }
```
evantahler, line 52: Please remove this comment

evantahler (changes requested):
Good work overall!

There's some scrubbing in the nockfiles to do, which we can talk about in Slack

evantahler (approved):

PR #1786, opened 2021-05-13
Base scaffold and file-writing capabilities for UI Config

seancdavis:
This does the following:

- Scaffolds the app by sharing code from ui-enterprise. You can now create, edit (if created during the current session — see below), and destroy apps, sources, destinations, properties, schedules, and groups. That shared code is now available in ui-components.
- Builds an idempotent `ConfigWriter` module that handles writes config objects to file. It is also responsible for deleting files from objects that no longer exist.

This does not cover the follow cases, each made into their own Tracker stories. I did this because what I have here feels like a good baseline to get merged in and I didn't want this already-big PR to become unmanageable.

- [Allow editing of config objects loaded into the app](https://www.pivotaltracker.com/story/show/178160078) (i.e. ones that existed prior to booting Config UI).
- ~[Storing a reference to the sourced config file on boot](https://www.pivotaltracker.com/story/show/178143612), which means you can create duplicate files today.~ I am now doing this by storing in a variable local to the ConfigWriter.
- ~[Account for config files having more than one object](https://www.pivotaltracker.com/story/show/178143690).~ This _should_ be working based on my new approach. However, I'm not writing more than one object to any file, though we'll want to write schedules to their source files. That is covered in [this story](https://www.pivotaltracker.com/story/show/178160116).
- ~[Deleting a config file when a record is deleted](https://www.pivotaltracker.com/story/show/178143741). It will run the ConfigWriter on delete, but won't remove anything.~ This works with the new approach, assuming the object was created in the same session.
- ~[Using an in-memory SQLite database](https://www.pivotaltracker.com/story/show/178143765). In fact, I changed this to a physical file during development, but the story has an idea on how to make it easier to manage long-term.~ This is now included in this work.


core/__tests__/modules/configWriter.ts
```diff
@@ -0,0 +1,354 @@
+import fs from "fs";
```
seancdavis, line 1: I chose to put everything related to writing files here. The bottom section should probably be in the model specs, but it felt more efficient to cover it while building out a series of objects all related to the same tasks. I chose efficiency of the test suite over spreading these things out.

core/src/actions/schedules.ts
```diff
@@ -1,5 +1,6 @@
 import { Schedule } from "../models/Schedule";
 import { AuthenticatedAction } from "../classes/actions/authenticatedAction";
+import { ConfigWriter } from "../modules/configWriter";
```
seancdavis, line 3: This is the pattern you're going to see throughout the app. After first putting the responsibility on the model to trigger the ConfigWriter, I moved it to the actions. It feels more like a controller method because the controller can more appropriately accept parameters from the client to specify whether or not we should do a write.

core/src/bin/config.ts
```diff
@@ -15,7 +15,8 @@ export class Apply extends CLI {
     GrouparooCLI.setGrouparooRunMode(this);
 
     // use temporary databases
-    process.env.DATABASE_URL = "sqlite://memory";
+    // process.env.DATABASE_URL = "sqlite://memory";
+    process.env.DATABASE_URL = "sqlite://grouparoo_config.sqlite";
```
seancdavis: There is a story to make this easier to manage.

core/src/migrations/000067-addConfigFilePath.ts
```diff
@@ -0,0 +1,29 @@
+const tables = [
```
seancdavis: At first I didn't really want to put these references in the database. But the more I worked through the problem, it made sense to me to be there. It felt nice to have this reference to give the instance a one-to-one tie with its config source.

In production we won't use this. However, with [this story](https://www.pivotaltracker.com/story/show/178143612), we'll have a reference to the files even when they are loaded from config. That might actually help with some debugging or some future feature in either enterprise or community.

core/src/models/App.ts
```diff
@@ -207,6 +211,23 @@ export class App extends LoggedModel<App> {
     return { source, destination };
   }
 
+  async setConfigFilePath(newPath?: string) {
+    this.configFilePath = newPath ? newPath : `apps/${this.id}.json`;
+    await this.save();
+  }
```
seancdavis: I went back and forth on whether I should abstract this elsewhere. In the end, it's so simple that I kept it on the instance and am testing it in ConfigWriter specs.

core/src/models/App.ts
```diff
@@ -207,6 +211,23 @@ export class App extends LoggedModel<App> {
...
+
+  async getConfigObject() {
```
seancdavis, line 212: Every writable model now has a `getConfigObject` method that will build a writable config object.

As I dug deep into here, I realize we have _several_ N+1 problems throughout these methods where we do things like `getOptions()` when we could have already had the options eager loaded. To solve it was going to be a huge rewrite, so I kept the pattern the same here. I wasn't super worried about it in this context, primarily because we're not going to be working with a huge number of records at any given time, since we just writing configuration files.

ui/ui-components/components/tabs/destination.tsx
```diff
@@ -6,10 +6,19 @@ export default function DestinationTabs({
 }: {
   destination: Models.DestinationType;
 }) {
-  const tabs =
-    process.env.GROUPAROO_UI_EDITION === "enterprise"
-      ? ["edit", "data", "exports"]
-      : ["exports"];
+  let tabs = [];
```
seancdavis, line 9: I made decisions on tabs throughout this PR. I'm not sure that they're right, but I figured we could easily add them when the time came.

core/src/models/Group.ts
```diff
@@ -37,7 +37,7 @@ export const GROUP_RULE_LIMIT = 10;
 const numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
 
 export interface GroupRuleWithKey {
-  key: string;
+  key?: string;
```
seancdavis, line 40: This doesn't feel like a good decision, but I also didn't want to make a new type. However, when we're building an object of rules to be written to file, we don't need the key to come along, and I didn't want to bring anything that the developer isn't going to care about.

seancdavis (commented):
Inline comments for some context here since this is a biggie.

seancdavis:
The builds are failing, but I can't seem to get any good information from the linter. I have prettier installed and have run it. Any ideas?

core/src/bin/config.ts
```diff
@@ -15,7 +15,8 @@ export class Apply extends CLI {
     GrouparooCLI.setGrouparooRunMode(this);
 
     // use temporary databases
-    process.env.DATABASE_URL = "sqlite://memory";
+    process.env.DATABASE_URL =
+      process.env.CONFIG_DATABASE_URL || "sqlite://memory";
```
seancdavis, line 19: This makes it trivial to run against a physical file db for easier debugging.

core/src/modules/configWriter.ts
```diff
@@ -0,0 +1,85 @@
...
+import { ConfigurationObject } from "../classes/codeConfig";
+
+import { getConfigDir } from "./configLoaders";
+
+type WritableConfigObject = {
+  filePath: string;
+  object: ConfigurationObject;
+};
+
+let FILES_TO_DELETE: string[] = [];
```
seancdavis, line 21: This is how I'm tracking which files to delete before running next time.

core/src/modules/mappingHelper.ts
```diff
@@ -9,7 +9,10 @@ export namespace MappingHelper {
     [remoteKey: string]: any;
   }
 
-  export async function getMapping(instance: Source | Destination) {
+  export async function getMapping(
+    instance: Source | Destination,
+    propertyColumn?: "key" | "id"
+  ) {
```
seancdavis, line 15: The value here is the property key by default, but that's not what we want for the code config shape, so I made this more flexible. Also created a spec for it.

seancdavis (commented):
Few more inline comments since my latest update.

core/src/models/App.ts
evantahler: Oh! Make a story for that and tag it `#speed` please!

core/src/actions/schedules.ts
evantahler: You moved the responsibility to call `ConfigWriter.run()` to the Actions (vs the models). I thin that's OK.  I wonder if each action needs a `writeConfig` param, or would they always want to write if `GROUAROO_RUN_MODE="cli:config"`?  If so, than you can always call the method from the Action and return early if not needed.

ui/ui-components/components/tabs/destination.tsx
evantahler: Seems OK for now (as it matches the UI patterns we already have)

evantahler (approved):
Looking Good Overall!

It's a shame that git sees all those UI file moves as "deleted" vs "moved".  Oh well.

Minor questions below, nothing blocking.

core/src/actions/schedules.ts
seancdavis: That's a great idea. I think I put this together before I knew about the mode variable. I like the pattern you're suggesting better. I'll put that in place.

PR #1788, opened 2021-05-14
fix: don't crash on `null` option for bootstrapped property

pedroslopez:
Fixes pivotal-178152658


evantahler (approved):
oh no! But good catch!

PR #1790, opened 2021-05-16
`runs:updateCounts` is resilient to one Run throwing

evantahler:


evantahler:
Interestingly - the behavior implemented is roughly equivalent to `Promise.allSettled()`.  But, since we are still targeting node.js v12, we cant use it.

PR #1793, opened 2021-05-17
Inject APM instrumentation into CLSAction and CLSTask

evantahler:
In order to create the `domain` that Sentry requires to isolate transaction traces, we cannot use Actionhero's middleware approach.  This PR moves to a "wrapping" pattern which is part of CLSAction and CLSTask, if used.  

The downsides of this approach are that middleware execution will not be traced by APM.


evantahler:
Let's get this up on staging and see how it performs.

PR #1794, opened 2021-05-18
In-Memory `Property` cache

evantahler:
The most heavily (re)-loaded model is `Property`.  This PR tries out the pattern that rather than preLoading known associations, we could keep all the Properties in memory.  Rather than use a redis cache (eg something like https://www.npmjs.com/package/sequelize-transparent-cache), this PR opts to keep the Property models in local memory as they are used so often.

This PR also doesn't investigate eager loading because in many cases the object in question (say an Import) doesn't know the properties related at the time of loading.

This PR implements a two-fold cache invalidation strategy: 
1) Timed (30 seconds)
2) Pub/Sub RPC on any Property model change.

This PR has the bonus side-effect of removing the preloading and passing of Properties added in https://github.com/grouparoo/grouparoo/pull/1782 and https://github.com/grouparoo/grouparoo/pull/1791

In `NODE_ENV=test`, the cache duration is -1, so the cache is instantly invalidated.

---

Test results (`./bin/timed_test staging-enterprise`)
* `main`: `🎉 Complete in 0hrs 2min 34sec [154 seconds]`
* `property-cache`: `🎉 Complete in 0hrs 1min 29sec [89 seconds]`
... 2x faster?!

---

Things to test:
- [x] That CLS Wrapping the RPC call works as expected
- [x] That a missed cache hit invalidates

### DIFF
--- a/core/src/models/Property.ts
+++ b/core/src/models/Property.ts
@@ -438,7 +438,7 @@ export class Property extends LoggedModel<Property> {
   @AfterDestroy
   static async invalidateCache() {
     Property.invalidateLocalCache();
-    await CLS.wrap(async () =>
+    await CLS.afterCommit(async () =>
       redis.doCluster("api.rpc.property.invalidateCache")
     );
   }
--- a/core/src/models/Property.ts
+++ b/core/src/models/Property.ts
@@ -15,9 +15,11 @@ import {
   DefaultScope,
   BeforeUpdate,
   BeforeCreate,
+  AfterSave,
+  AfterUpdate,
 } from "sequelize-typescript";
 import { Op } from "sequelize";
-import { env, api, config } from "actionhero";
+import { env, api, redis, config } from "actionhero";
 import { plugin } from "../modules/plugin";
 import { LoggedModel } from "../classes/loggedModel";
 import { Profile } from "./Profile";
@@ -117,6 +119,11 @@ export interface PropertyFiltersWithKey {
   relativeMatchUnit?: string;
   relativeMatchDirection?: string;
 }
+
+const _cachedProperties: { expires: number; properties: Property[] } = {
+  expires: 0,
+  properties: [],
+};
 
 @DefaultScope(() => ({
   where: { state: "ready" },
@@ -391,6 +398,38 @@ export class Property extends LoggedModel<Property> {
       options,
       filters,
     };
+  }
+
+  // --- Cache Methods --- //
+
+  static async findAllWithCache(): Promise<Property[]> {
+    const now = new Date().getTime();
+    if (
+      _cachedProperties.expires > now &&
+      _cachedProperties.properties.length > 0
+    ) {
+      return _cachedProperties.properties;
+    }
+
+    _cachedProperties.properties = await Property.findAll();
+    _cachedProperties.expires = now + CACHE_TTL;
+    return _cachedProperties.properties;
+  }
+
+  static async findOneWithCache(key: string, value: string) {
+    const properties = await Property.findAllWithCache();
+    return properties.find((p) => p[key] === value);
+  }
+
+  static invalidateLocalCache() {
+    _cachedProperties.expires = 0;
+  }
+
+  @AfterSave
+  @AfterUpdate
+  @AfterDestroy
+  static async invalidateCache() {
+    await redis.doCluster("api.rpc.property.invalidateCache");
   }
 
   // --- Class Methods --- //
--- a/core/src/models/Property.ts
+++ b/core/src/models/Property.ts
@@ -417,7 +417,7 @@ export class Property extends LoggedModel<Property> {
     return _cachedProperties.properties;
   }
 
-  static async findOneWithCache(key: string, value: string) {
+  static async findOneWithCache(value: string, key = "id") {
     const properties = await Property.findAllWithCache();
     let property = properties.find((p) => p[key] === value);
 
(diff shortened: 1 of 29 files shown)


core/src/models/Property.ts
```diff
@@ -373,6 +380,46 @@ export class Property extends LoggedModel<Property> {
...
+  }
+
+  static invalidateLocalCache() {
+    _cachedProperties.expires = 0;
+  }
+
+  @AfterSave
+  @AfterUpdate
+  @AfterDestroy
+  static async invalidateCache() {
```
bleonard, line 423: technically, if this is in transaction, you want to do this after commit. otherwise, you'll clear on those other clusters opening up a window to refetch the wrong info just before this transaction runs,

evantahler:
@seancdavis depending on what you learn about `$get`, this is a parallel approach to consider

core/src/models/Property.ts
```diff
@@ -373,6 +380,46 @@ export class Property extends LoggedModel<Property> {
     };
   }
 
+  // --- Cache Methods --- //
+
+  static async findAllWithCache(): Promise<Property[]> {
```
bleonard, line 390: maybe something interesting to think about here with regard to "deleted" properties.

core/src/models/Property.ts
bleonard: Should it do `invalidateLocalCache()` right here just to be sure or will `await` wait for that?

My gut says that `AfterSave`, `AfterDestroy` you would `invalidateLocalCache()` and then do a CLS sort of thing so that  `await redis.doCluster();` happens at the end of the transaction (and invliadate again just for good measure)

bleonard: Random note: At least in Rails, `AfterSave` is either of `AfterCreate` and `AfterUpdate` so the `AfterUpdate` here might be rendundant.

bleonard (commented):
In my brief investigation, it wasn't `findAll` properties that was the issue. That's a relatively fast single query. It was where we are fetching single ones like in the `getValue` method of `ProfileProperties` that is called in places while iterating.

I'm generally nervous about caching, but I think what it would take to make it impactful is
1. this "all" one, ideally with the sequelize situation where args query and more transparent `filter`ing from the caller.
2. more use of `findOneWithCache` especially by id.

Random thoughts or questions:
1. maybe `expires` should be reset each CLS wrapper (beginning and end) just to be sure.
2. does the `rpc` approach happen immediately on the same system or is there something else to do? 
3. does the `rpc` wait for all the other nodes? Race conditions if not. Single point of failure if it does.
4. It's interesting that on long running jobs, this can actually be _more_ correct than the other way to fix perf issues (fetching more at the beginning).
5. It's clearing avoiding a Redis approach even though it needs Redis to do the rpc. Why not use that? Because it's still an extra connection?

core/src/models/Property.ts
evantahler: Nice! I'll add the CLS wrap

core/src/models/Property.ts
evantahler: This should be obeying the default scope, which is `ready`

core/src/models/Property.ts
```diff
@@ -393,6 +401,48 @@ export class Property extends LoggedModel<Property> {
...
+    ) {
+      return _cachedProperties.properties;
+    }
+
+    _cachedProperties.properties = await Property.findAll();
+    _cachedProperties.expires = now + CACHE_TTL;
+    return _cachedProperties.properties;
+  }
+
+  static async findOneWithCache(key: string, value: string) {
```
bleonard: If this is always "id", it might be nice to make a method for that `findByPkWithCache()` or something

core/src/modules/reset.ts
```diff
@@ -66,6 +66,8 @@ export namespace Reset {
       counts[model.name] = count;
     }
 
+    Property.invalidateCache();
```
bleonard: If we do more of these, I'd make some method like `System.invalidateCaches()`

core/src/models/Property.ts
evantahler: or we can switch the oder and make a default - see 98c3ac24fd41a019890622023bc0994982f4c530

bleonard: Works for me.

core/src/modules/reset.ts
evantahler: Yeah - for now I think `Property` is special. lets see!

evantahler:
@bleonard answering some of your questions above:

1. the RPC methods do invoke on the calling node, and all other node.  They work over redis pub/sub, and as configured, do wait for at least one server to respond (could be self).  I've added a local cache invalidation too before waiting for the RPC methods. 
2. I didn't want to cache these objects on redis for 2 main reasons: 
   i. They are so in-use, that waiting for *any* network seemed slow, even redis
   ii. If we can keep these records in their native, Seqeulize-object form, we can reliably use all their methods.  Otherwise, we'd need to (re) hydrate them somehow
3. Clearing the cache at the end of every CLSAction/Task is something we could do, but it gets weird when there are multiple tasks in parallel.  That means that the cache could be cleared a bunch of times in the middle of a task... that seems like it would add back a lot of the slowness we are trying to solve

seancdavis (approved):
I didn't see anything to flag here. Looks like you and Brian worked through a lot already.

I like the pattern this sets and feels like it could be extended elsewhere. Like @bleonard mentioned, I am super curious to see what kind of additional speedups we get by exploring where we have N+1 issues. I wonder if it'll be more or less dramatic that finding properties in bulk.

PR #1795, opened 2021-05-18
Write schedules to their source config files

seancdavis:
Note that this can be merged separate of #1796, but will likely run into some conflicts before we can merge. I pulled them apart so you could review separately. But I can handle merging and conflict resolution.

### DIFF
--- a/core/src/models/Source.ts
+++ b/core/src/models/Source.ts
@@ -294,11 +294,14 @@ export class Source extends LoggedModel<Source> {
       options,
     };
 
-    const schedule = await this.$get("schedule");
-    if (schedule) {
-      const scheduleConfigObject = await schedule.getConfigObject();
+    const setSchedule = async () => {
+      if (!this.schedule) return;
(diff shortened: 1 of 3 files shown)


core/src/models/Source.ts
```diff
@@ -283,7 +284,7 @@ export class Source extends LoggedModel<Source> {
     const options = await this.getOptions();
     const mapping = await MappingHelper.getMapping(this, "id");
 
-    return {
+    let configObject: any = {
```
evantahler, line 287: Perhaps type `ConfigurationObject` might be a little tighter than `any`?  Or, was the `any` added to allow you to change its makeup below with the merge?

evantahler (approved):
Looks great! 

Some non-blocking comments below

core/src/models/Source.ts
seancdavis: Yep. The merge thing.

It also wouldn't let me do something like `ConfigurationObject | ConfigurationObject[]` because they don't share properties in common with one another.

PR #1796, opened 2021-05-18
Add ability to edit config files

seancdavis:
This adjusts the current locking mechanism for code config when in Config UI mode (i.e. `process.env.GROUPAROO_RUN_MODE = "cli:config"`). Objects that are loaded from JSON files are unlocked, while JS files are locked. This should not adjust the locking pattern when in start/run mode for either Community or Enterprise.

Note that we're now assuming that there is parity between the Loader(s) and the Writer — that all appropriate loaders cache the proper `locked` value in the database. The Writer then does not cached files it expects to be locked, such that it simply ignores them altogether.

I added some comments to that effect in the individual config loaders. I also added a big comment with updated logic on how the Loader and Writer work together with one another.

### DIFF
--- a/core/src/modules/configLoaders/app.ts
+++ b/core/src/modules/configLoaders/app.ts
@@ -19,34 +19,22 @@ export async function loadApp(
   let isNew = false;
   validateConfigObjectKeys(App, configObject);
 
-  let locked: string;
-  if (process.env.GROUPAROO_RUN_MODE !== "cli:config") {
-    locked = getCodeConfigLockKey();
-  } else if (ConfigWriter.isLockable(configObject)) {
-    locked = ConfigWriter.getLockKey();
-  }
-
-  let getParams: { id: string; locked?: string } = { id: configObject.id };
-  if (locked) getParams.locked = locked;
-
-  let createParams: {
-    id: string;
-    name: string;
-    type: string;
-    locked?: string;
-  } = {
-    id: configObject.id,
-    name: configObject.name,
-    type: configObject.type,
-  };
-  if (locked) createParams.locked = locked;
-
+  // We assume we will always have to create a new object when in config mode,
+  // so it is safe to leave locked in the find query.
   let app = await App.scope(null).findOne({
-    where: getParams,
+    where: {
+      id: configObject.id,
+      locked: getCodeConfigLockKey(),
+    },
   });
   if (!app) {
     isNew = true;
-    app = await App.create(createParams);
+    app = await App.create({
+      id: configObject.id,
+      name: configObject.name,
+      type: configObject.type,
+      locked: ConfigWriter.getLockKey(configObject),
+    });
   }
 
   await app.update({ type: configObject.type, name: configObject.name });
@@ -71,6 +59,9 @@ export async function loadApp(
 }
 
 export async function deleteApps(ids: string[]) {
+  // Since this method is only used when config is loaded and because we assume
+  // the db is ephemeral, we can target locked objects, even though this will
+  // always return zero objects when in config mode.
   const apps = await App.scope(null).findAll({
     where: { locked: getCodeConfigLockKey(), id: { [Op.notIn]: ids } },
   });
--- a/core/src/modules/configLoaders/app.ts
+++ b/core/src/modules/configLoaders/app.ts
@@ -9,6 +9,8 @@ import {
 import { App } from "../..";
 import { Op } from "sequelize";
 
+import { ConfigWriter } from "../configWriter";
+
 export async function loadApp(
   configObject: ConfigurationObject,
   externallyValidate: boolean,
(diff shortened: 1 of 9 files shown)


core/src/modules/configLoaders/app.ts
```diff
@@ -17,16 +19,21 @@ export async function loadApp(
   let isNew = false;
   validateConfigObjectKeys(App, configObject);
 
+  // We assume we will always have to create a new object when in config mode,
+  // so it is safe to leave locked in the find query.
```
seancdavis: Dropping these in every loader. Happy to remove if you feel like it is unnecessary. The locked logic was starting to feel a little tangled, so I did this to help future devs.

core/src/modules/configLoaders/app.ts
```diff
@@ -52,6 +59,9 @@ export async function loadApp(
 }
 
 export async function deleteApps(ids: string[]) {
+  // Since this method is only used when config is loaded and because we assume
+  // the db is ephemeral, we can target locked objects, even though this will
+  // always return zero objects when in config mode.
```
seancdavis: Same here. This is duplicated in each of the relevant loaders.

core/src/modules/configWriter.ts
```diff
@@ -18,7 +21,32 @@ type WritableConfigObject = {
...
+ *
+ * Writing
+ * ----------------------------------------
+ * 1. Delete all files in the cache.
+ * 2. Query the database for all writable objects. These are UNLOCKED Apps,
+ *    Sources (and their Schedules), Properties, Groups, and Destinations.
+ * 3. Each config object is retrieved from the model, written to file, and then
+ *    cached.
+ *
+ */
```
seancdavis, line 47: This felt like a lot to have in this file. It really helped me work through it and pull it together. I didn't come up with a better place to put this, but happy to move or remove it.

seancdavis (commented):
A few comments for context here. Happy to talk through it if it's confusing what is happening here.

core/src/modules/configLoaders/app.ts
evantahler: I think the comment here is a bit confusing:  This method is likely to be run on a "real" database - `grouparoo start,run,apply` all do it.  Apps that are no longer in the config will be deleted.

core/src/modules/configWriter.ts
```diff
@@ -36,15 +64,78 @@ export namespace ConfigWriter {
...
+    if (isLockable(configObject)) {
+      return "config:writer";
+    }
+    return null;
+  }
+
+  function fileIsLockable(absFilePath: string): boolean {
+    // Otherwise, it is lockable if it is a JS file.
+    const ext = path.extname(absFilePath);
+    return ext === ".js";
```
evantahler: This is a good check - TS files will also be non-lockable

core/src/modules/configWriter.ts
```diff
@@ -36,15 +64,78 @@ export namespace ConfigWriter {
...
+  }
+
+  function isLockable(object: ConfigurationObject) {
+    const isMatch = (o) => o.id === object.id && o.class === object.class;
+    const cachedFileObj: CachedConfigFile = CONFIG_FILE_CACHE.find(
+      (cache) => cache.objects.filter(isMatch).length > 0
+    );
```
evantahler, line 121: Will this get weird when https://github.com/grouparoo/grouparoo/pull/1795 is merged in (2 objects in one file)?

evantahler (commented):
I like the approach of re-using `model.locked`.

You asked about project layout - I think where you have it is OK.  You could also make a few sub-files in `core/src/modules/configWriter/*` if you wanted to split up some of the big files.

(idea; might be terrible) I wonder if there's a way to store even more data in `model.locked`, so you don't need to rely on CONFIG_FILE_CACHE at all. maybe `app.locked = JSON.stringify({new: true, originalFile: '/path/to/file', writable: false})`

core/src/modules/configLoaders/app.ts
bleonard: I'm not sure these comments help me. maybe something in the `ConfigWriter.getLockKey` function instead as to why it is null.

core/src/modules/configWriter.ts
bleonard: maybe it's `!== "json"`

core/src/modules/configWriter.ts
seancdavis: It _shouldn't_. This checks against the object itself. So if both a source and a schedule are loaded from the same file, they will both either be locked or not, based on the file extension.

We don't have to worry about the object being an array based on how this is being used today.

seancdavis:
@evantahler @bleonard Updated with cleaner comments.

>  I wonder if there's a way to store even more data in model.locked, so you don't need to rely on CONFIG_FILE_CACHE at all. maybe app.locked = JSON.stringify({new: true, originalFile: '/path/to/file', writable: false})

That's certainly doable. Not sure if it gives us an advantage in some way. Do you feel like it's worth exploring?

evantahler:
re: https://github.com/grouparoo/grouparoo/pull/1796#issuecomment-844402422

Nah, if what you have is working, go with it!  That's a potentially maybe refactor for later, if at all

PR #1798, opened 2021-05-18
`properties:list` action should not always include examples

evantahler:
This PR makes the `properties:list` action optionally include examples via a new `includeExamples` param

### DIFF
--- a/core/src/actions/properties.ts
+++ b/core/src/actions/properties.ts
@@ -21,6 +21,7 @@ export class PropertiesList extends AuthenticatedAction {
       unique: { required: false },
       state: { required: false },
       sourceId: { required: false },
+      includeExamples: { required: true, default: "false" },
       order: {
         required: false,
         default: [
@@ -32,6 +33,9 @@ export class PropertiesList extends AuthenticatedAction {
   }
 
   async runWithinTransaction({ params }) {
+    const includeExamples =
+      params.includeExamples === "true" || params.includeExamples === true;
+
     const where = {};
     if (params.state) where["state"] = params.state;
     if (params.sourceId) where["sourceId"] = params.sourceId;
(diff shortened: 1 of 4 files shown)


core/src/actions/properties.ts
```diff
@@ -50,30 +54,31 @@ export class PropertiesList extends AuthenticatedAction {
...
-        },
-        order: [["id", "asc"]],
-        limit: 5,
-      });
-      const exampleValues = examples.map((e) => e.rawValue);
-
       responseProperties.push(apiData);
-      responseExamples[property.id] = exampleValues;
+
+      if (includeExamples) {
```
bleonard, line 61: can probably do an N+1 Eager load fix later if you want.

evantahler: Can you share more about how that applies in this case?  We don't want to use the same 5 profiles for each Property, because they may have null values... so we search for 5 non-null ProfileProperties for each Property.

bleonard: I believe it's conceptaully similar to `User` that `hasMany` `Tools` like in this example.
https://sequelize.org/master/manual/eager-loading.html

In our case
In the model: `Property.hasMany(ProfileProperties)`

In this file:
```
const properties = await Property.findAll({
  include: {
    model: ProfileProperty,
    where: {
      rawValue: { [Op.not]: null },
    },
  }
});

for (const property of property) {
  const examples = property.profileProperties;
  // use that
}
```

I suppose the issue is the `limit` situation.
There is this `separate` thing.
```
const properties = await Property.findAll({
  include: {
    model: ProfileProperty,
    where: {
      rawValue: { [Op.not]: null },
    },
    order: [["id", "asc"]],
    limit: 5,
    separate: true,
  }
});

for (const property of property) {
  const examples = property.profileProperties;
  // use that
}
```
I would check to see if it's smarter than what we are doing manually here (or maybe it's the same).

PR #1799, opened 2021-05-19
Add sync modes to sqlite destination plugin.

pauloouriques:


evantahler (approved):
Thank you for adding these tests!  The other database sources will likely need similar help.

PR #1801, opened 2021-05-19
Force delete destinations and groups in config mode

seancdavis:
I did a lot of digging through the weeds here, but I could really only find that destinations and groups had this deleted state. And they already had a simple way to force it. I feel like I'm missing something.

### DIFF
--- a/ui/ui-components/pages/group/[id]/edit.tsx
+++ b/ui/ui-components/pages/group/[id]/edit.tsx
@@ -158,7 +158,7 @@ export default function Page(props) {
                   disabled={loading}
                   size="sm"
                   onClick={() => {
-                    handleDelete();
+                    handleDelete(process.env.GROUPAROO_UI_EDITION === "config");
                   }}
                 >
                   Delete
--- a/core/src/models/Destination.ts
+++ b/core/src/models/Destination.ts
@@ -617,6 +617,7 @@ export class Destination extends LoggedModel<Destination> {
 
   @BeforeDestroy
   static async cannotDeleteDestinationWithTrackedGroup(instance: Destination) {
+    if (process.env.GROUPAROO_RUN_MODE === "cli:config") return;
     if (instance.groupId) {
       const group = await Group.findOne({
         where: { id: instance.groupId },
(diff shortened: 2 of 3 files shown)


evantahler:
I'm going to pass on my review to @pedroslopez.  After https://github.com/grouparoo/grouparoo/pull/1773, I think a lot more models have `state: deleted`.   I'm curious if in cli:config mode you'll run into conflicts if those models aren't allowed to 'drain'.  It might not be a problem at all, or you may need to add more `force:true` to "delete now" in the actions.

ui/ui-components/pages/group/[id]/edit.tsx
```diff
@@ -158,7 +158,7 @@ export default function Page(props) {
                   disabled={loading}
                   size="sm"
                   onClick={() => {
-                    handleDelete();
+                    handleDelete(process.env.GROUPAROO_UI_EDITION === "config");
```
Comment (pedroslopez) [inline ui/ui-components/pages/group/[id]/edit.tsx:161]: I _feel_ like just for consistency maybe both `handleDelete` functions (in group and in destination edit) should work the same way, e.g. both taking a `force` arg or both checking the env var within the function.

pedroslopez (commented):
There _are_ more models that have a deleted state as of #1773:
- App
- Property
- Source

However, in that story they're only being soft deleted (marked with `state: "deleted"`) when deleting the files in code config because in that scenario they need wait for other models that depend on them (groups, destinations) before they're actually deleted. There's no "draining", and all they need to do is sit around and wait until their dependents are deleted.

When doing it from the UI it's a bit different since you delete things one by one by pressing a button, and we check that all dependents are really gone before deleting something. If we want to also allow you to enqueue deletion of things that are still being used then I think that's another story, but I would say it could be confusing.

I think what this means for config UI is we still don't need to worry about the models affected by #1773 and forcing groups/destinations is enough. If something changes about having to do some "draining"-like actions for these other models, then at that point we would need to also consider it for the UI.

ui/ui-components/pages/group/[id]/edit.tsx
Comment (seancdavis) [reply ui/ui-components/pages/group/[id]/edit.tsx:161]: Thanks for the thoughtful review, @pedroslopez!

This shape is slightly different because there is an enterprise workflow for groups in which you can force delete a group after deleting it. Therefore, there is a call above this one that says `handleDelete(true)`. So groups are going to differ from destinations in some way. This seemed like an easy path to get there.

What do you think?

Comment (pedroslopez) [reply ui/ui-components/pages/group/[id]/edit.tsx:161]: Ah, I see! In that case the destination's `handleDelete` could take an argument in the same way, but it's fine either way. I don't feel too strongly about it 🙂

PR #1807, opened 2021-05-20
Webpack artifact improvements 

evantahler:
A few improvements regarding how we handle webpack artifacts:

1. Do not include `.next/cache/*.pack` files in our npm packages.  These are not needed at run time and are HUGE.  This will make our UI NPM packages ~200mb smaller!
2. We don't need to clear the whole `.next` directory on `npm prepare`. We can keep the cache around, and sped up subsequent build (triggered by `pnpm install`) by ~2x!
   a. **fresh build** - npm run prepare  130.36s user 7.85s system 220% cpu 1:02.77 total
   b. **with cache** - npm run prepare  28.56s user 2.78s system 154% cpu 20.251 total
3. We do want to ensure that we include fresh buildfiles in our releases, so the nuke and prepare scripts will now remove the .next cache and rebuild from scratch before publishing.

### DIFF
--- a/bin/publish
+++ b/bin/publish
@@ -95,6 +95,13 @@ echo ""
 echo "-----> INSTALL DEPENDENCIES & BUILD PACKAGES"
 echo ""
 
+# Clear build caches
+rm -rf core/dist
+rm -rf clients/*/*/dist
+rm -rf cli/dist
+rm -rf plugins/*/*/dist
+rm -rf ui/*/.next
+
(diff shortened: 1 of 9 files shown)


evantahler:
Looks like with this one weird trick, you can shed over 300mb! 

Screen Shot 2021-05-20 at 10 14 25 AM

Screen Shot 2021-05-20 at 10 22 43 AM

PR #1809, opened 2021-05-20
Support upper-case MySQL INFORMATION_SCHEMA

bleonard:
It seems that the information_schema in MySQL can return uppercase column names in some users' environments.
Normalize, so things work as expected.


evantahler:
@bleonard so this appears in the changelog, I've updated the title and added the "bug" label.

PR #1810, opened 2021-05-20
Filter App.options from logs

evantahler:
Using App Option `type=password` to filter things out of the logs

```
2021-05-21T00:22:34.027Z - info: [ action @ web ] to=127.0.0.1 action=app:test params={"action":"app:test","apiVersion":"1","options":{"port":5432,"schema":"demo","host":"localhost","database":"grouparoo_development","user":"evan","password":"[FILTERED]"},"csrfToken":"[FILTERED]","id":"app_869379ee-7347-4784-b1c9-b8cc97700bef"} duration=35
```

This PR made it so that App Option's `type` could be defined both statically on the App Option, and overriden by the dynamic `appOptions()` method from the plugin.  For many plugins `appOptions()` was only returning static values, so this PR tightened those up.


bleonard:
there might still be `optional` but secret things. I'd be ok adding a `type` to the `plugin.ts` where required or doing a boolean there like `sensitive=true`

core/src/config/api.ts
```diff
@@ -30,7 +33,29 @@ export const DEFAULT = {
...
       // params you would like hidden from any logs
-      filteredParams: ["password", "passwordHash", "csrfToken", "private_key"],
+      // api.plugins.plugins.map(p => p.apps).filter(a => a).flat().map(a => a.options).flat().filter(o => o.required).map(o => o.key)
+      filteredParams: function () {
+        const filteredParams = [
+          "password",
+          "passwordHash",
+          "csrfToken",
+          "private_key",
+        ];
```
bleonard, line 43: we could add very common ones here like `apiKey`, `token`, `apiToken`, `secret`, `apiSecret`, `securityToken`, `accessToken`

evantahler: Made some changes so we can use `type=password` directly

PR #1811, opened 2021-05-20
Add sync modes to Postgres SQL destination plugin.

pauloouriques:
- Minor fix to redundant sync modes check on sqlite plugin.

### DIFF
--- a/plugins/@grouparoo/postgres/package.json
+++ b/plugins/@grouparoo/postgres/package.json
@@ -42,6 +42,7 @@
     "@types/pg": "*",
     "actionhero": "25.0.16",
     "csv-parse": "4.15.4",
+    "fs-extra": "10.0.0",
     "jest": "26.6.3",
     "prettier": "2.3.0",
     "ts-jest": "26.5.6",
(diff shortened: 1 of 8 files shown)


plugins/@grouparoo/postgres/package.json
```diff
@@ -42,6 +42,7 @@
     "@types/pg": "*",
     "actionhero": "25.0.16",
     "csv-parse": "4.15.4",
+    "fs-extra": "10.0.0",
```
evantahler, line 45: Doesn't look like this is used?

evantahler (changes requested):
Looks good except for that dependency

plugins/@grouparoo/postgres/package.json
pauloouriques: It's used [here](https://github.com/grouparoo/grouparoo/blob/2426f644938144c9a5399388268afa105b8df4ba/plugins/%40grouparoo/postgres/__tests__/integration/cli-generate-from.ts#L2)

evantahler: whoa!  So we had just been lucky that it was brought in as a transient dependency (probably via @grouparoo/core) until now.  Good find!

evantahler (approved):

PR #1813, opened 2021-05-20
replaced browser.sleep with until(elementLocated)

teallarson:
current max timeout is 20 seconds, though can be adjusted


evantahler (approved):
Nice!

evantahler:
20s should be more than enough

PR #1819, opened 2021-05-23
Speed up `profile:checkReady` with a custom query

evantahler:
The previous query `ProfileOps.makeReady()` used failed with larger datasets:

```SQL
SELECT * FROM "profiles" AS "Profile" 
WHERE "Profile"."state" = 'pending' 
AND "Profile"."id" NOT IN (
	SELECT DISTINCT("profileId") FROM "profileProperties" WHERE "profileProperties"."state" = 'pending') 
ORDER BY "Profile"."id" ASC 
LIMIT 100;
```

This PR adds a better query, one without a sub-select.

```sql
SELECT
  profiles.id as id
FROM profiles
JOIN
  "profileProperties" on "profileProperties"."profileId" = profiles.id
WHERE
  profiles.state = 'pending'
GROUP BY profiles.id
HAVING
  COUNT(DISTINCT "profileProperties".state) = 1
  AND MAX("profileProperties".state) = 'ready'
LIMIT 100;
```

--- 

This PR also adds a test suite running all Tasks against a SQLite database

### DIFF
--- a/core/src/modules/ops/profile.ts
+++ b/core/src/modules/ops/profile.ts
@@ -625,12 +625,10 @@ export namespace ProfileOps {
           api.sequelize.where(
             api.sequelize.fn(
               "COUNT",
-              // api.sequelize.col(
               api.sequelize.fn(
                 "DISTINCT",
                 api.sequelize.col("profileProperties.state")
               )
-              // )
             ),
             "=",
             1
--- a/core/src/modules/ops/profile.ts
+++ b/core/src/modules/ops/profile.ts
@@ -588,22 +588,57 @@ export namespace ProfileOps {
    * Task `profile:completeImport` will be enqueued for each Profile.
    */
   export async function makeReady(limit = 100, toExport = true) {
-    let profiles: Profile[] = await api.sequelize.query(
-      `SELECT
-          profiles.id as id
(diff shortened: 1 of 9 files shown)


bleonard (commented):
any sqlite concerns? I wonder if it's possible in `sequelize` format.

evantahler:
https://github.com/grouparoo/grouparoo/pull/1819/commits/1a8f4a7a9cfb14102244ed43fb50813ebb0fe15a is how to build this query with Sequelize.  https://github.com/grouparoo/grouparoo/pull/1819/commits/24613f333cd90ddecf1657dbbd21161503e78a8b will now test our tasks against SQLite in our test suite (we already test Actions and Models against sqlite)

core/src/modules/ops/profile.ts
```diff
@@ -588,24 +588,55 @@ export namespace ProfileOps {
...
+              "MAX",
+              api.sequelize.col("profileProperties.state")
+            ),
+            "=",
+            "ready"
+          ),
+          api.sequelize.where(
+            api.sequelize.fn(
+              "COUNT",
+              // api.sequelize.col(
```
bleonard: remove this surrounding comment?

PR #1824, opened 2021-05-24
remove csv:file from roo generate list

teallarson:


evantahler (approved):
Well... that was certainly the simplest solution... and the tests pass!

evantahler (changes requested):
Please remove the .template file we aren't using any more

evantahler (approved):

PR #1826, opened 2021-05-25
bring back deleted models in code config

pedroslopez:


core/__tests__/modules/codeConfig/codeConfig.ts
```diff
@@ -577,25 +586,267 @@ describe("modules/codeConfig", () => {
...
+      expect(app.locked).toBeNull();
     });
 
     test("settings remain", async () => {
       expect(await Setting.count()).toBeGreaterThan(1);
     });
   });
 
+  describe("bring it all back", () => {
+    let previousRun;
```
evantahler: It might be good to have a run for a few types of things that got deleted, in addition to the destination - App, Property, etc.

core/src/modules/configLoaders/app.ts
```diff
@@ -22,7 +22,10 @@ export async function loadApp(
   let app = await App.scope(null).findOne({
     where: {
       id: configObject.id,
-      locked: getCodeConfigLockKey(),
+      [Op.or]: {
+        locked: getCodeConfigLockKey(),
+        state: "deleted",
+      },
```
evantahler, line 28: Question - we made the choice a while back that code-config'd models which were deleted become unlocked.  This is so the destroy tasks could eventually delete them.  Just checking that that was still the right choice.

core/src/modules/configLoaders/destination.ts
```diff
@@ -77,6 +84,10 @@ export async function loadDestination(
     await destination.trackGroup(group);
   }
 
+  if (destination.state === "deleted") {
+    await destination.exportGroupMembers(true);
+  }
```
evantahler: Were we not doing a final export before?  If not... this should be added in a test.

evantahler (commented):
Good work!  

* Do you think we need to add any more safety to the state transitions to leave "deleted"?  e.g. should this only be done before the server is fully up-and-running?  Or, if someone were to try to un-delete something via an Action, would it be OK?
* I'm curious to learn more about the delete behavior of the Destination - Why weren't we kicking off a run to export everyone one last time until now?

PR #1828, opened 2021-05-25
Allow date strings in group rules for code config

teallarson:
Previously, code config would only allow dates as numbers (epoch time) for calculated groups. 

Now, date strings can be passed and will be parsed and saved as epoch time for calculated rules.


teallarson:
Removed support for date time without timezone given because Javascript parses such strings as local time.  

Accepted formats are a date string (YYYY-MM-DD which is parsed as 0:00:00.000 UTC) or a full 8601 date-time with timezone. (YYYY-MM-DDTHH:mm:SS.sss+ZZ:ZZ)

core/__tests__/fixtures/codeConfig/group-rule-with-date/config.js
```diff
@@ -0,0 +1,102 @@
...
+          match: "2021-04-05", //date only with gt rule
+          type: "date",
+        },
+        {
+          propertyId: "last_appointment_date",
+          operation: { op: "gte" },
+          match: "2019-10-10T14:48:00.000+09:00", //date with time and timezone
+          type: "date",
+        },
+      ],
```
evantahler: 👍 testing a few types of date strings

core/src/modules/configLoaders/group.ts
```diff
@@ -36,12 +36,31 @@ export async function loadGroup(
...
+        let epochTime: number;
+        //will parse as exact timestamp, or will default to 00:00:00.000 UTC if none
+        epochTime = Date.parse(rules[i]["match"].toString());
+        //if rule uses <= or > and no time is given, base rule on 11:59:59.999
+        if (
+          calculatesWithDate.indexOf(rules[i]["operation"]["op"]) <= 1 &&
+          rules[i]["match"].toString().length === 10
+        ) {
+          epochTime += 86399999;
+        }
```
evantahler: I think you are trying to determine if the user gave a timezone or not, and if not, coerce the time into UTC, to the end of the day.  I'm a little worried about this approach because there are other ways they might provide a time that are > 10 chars that will validly pass `Date.parse`, eg: `Date.parse("January 10, 2010")`.  

Perhaps you can skip this step, and instead we'll instruct people in the docs to provide a timezone + hh:mm:ss if they want to be specific, otherwise, we'll do our best.

evantahler (changes requested):
Good start! 

Can you please also add a test for the the relative group rules, eg: "last_purchase_date is greater than 7 days ago" (eg: they purchased something this week).  Those Group Rules look like:

```
          {
            propertyId: "last_purchase_date",
            operation: { op: "relative_gt" },
            relativeMatchNumber: 7,
            relativeMatchUnit: "days",
          },
```

teallarson:
the ui test failure shouldn't be an issue once this is merged with main.  This branch doesn't have the updated version.

evantahler (approved):
Nice!

PR #1829, opened 2021-05-25
Make plugin installation part of the new app process in Config UI

seancdavis:
This provides the ability to add apps from plugins that aren't installed while staying in the flow of the Config UI.

- I created a new staging-config app. I wanted to do this so we could have a place with a minimal number of plugins installed. Something that would be closer to a _starting point_ could be useful in dev.
- It added a plugin manifest to www via grouparoo/www.grouparoo.com#461 and uses that manifest to return the list of available plugins from the API.
- Adds an `installed` badge to those apps that are ready to go and don't have to be installed. These also will appear at the top of the list.
- Adds a `timeout` feature to our `useApi` hook. We found out that Axios will hang when a request is pending. Here we wanted it to fail because that means the server is down.


core/src/actions/status.ts
```diff
@@ -18,6 +18,7 @@ export class PublicStatus extends OptionallyAuthenticatedAction {
 
   async runWithinTransaction({ response }) {
     response.status = "ok";
+    return response;
   }
```
evantahler: I think this is from an old version of actionhero.  Best to just `return {status: 'ok'}`

core/__tests__/actions/plugins.ts
```diff
@@ -42,20 +42,25 @@ describe("actions/plugins", () => {
   });
 
   test("can list available plugins", async () => {
-    // cannot actually test this without injecting some plugins at the package.json level...
     const { plugins, error } = await specHelper.runAction(
       "plugins:available:list"
     );
```
evantahler, line 53: Since this test is now making an API request, it probably should be recorded with nock.

core/src/modules/plugins.ts
```diff
@@ -77,52 +79,18 @@ export namespace Plugins {
...
 
-  export const ignoredPlugins = [
-    "@grouparoo/ui",
-    "@grouparoo/ui-components",
-    "@grouparoo/client-web",
-    "@grouparoo/app-templates",
-    "@grouparoo/email-authentication",
-    "@grouparoo/logger",
-    "@grouparoo/node-marketo-rest",
-  ];
```
evantahler, line 88: 👋 Bye!

ui/ui-components/client/client.ts
```diff
@@ -105,7 +105,8 @@ export class Client {
     useCache = true,
     uploadHandler?,
     req?,
-    res?
+    res?,
+    timeout?: number
```
evantahler: @seancdavis I don't think we ended up needing the timeout maybe?  that might have been a red herring before we realized that the api server wasn't rebooting properly.

ui/ui-components/pages/app/new.tsx
```diff
@@ -1,55 +0,0 @@
-import Head from "next/head";
```
evantahler, line 1: 👍 making an app/new page for ui-config that's different from ui-enterprise

evantahler (commented):
Looking good overall! Some minor changes below

core/src/actions/status.ts
seancdavis: I think I ran into an issue with TypeScript on this one. I'll give it a shot.

core/__tests__/actions/plugins.ts
seancdavis: That sounds like a fun adventure. On it!

core/src/modules/plugins.ts
seancdavis: 👋

ui/ui-components/client/client.ts
seancdavis: Worth a test, anyways.

PR #1831, opened 2021-05-26
profile list speedup

bleonard:
The query was particularly slow, but in the common case, we don't have to do a join.
Also leaned heavier on the cached properties.

Went from 15 seconds to less than 1 with db with 1.5M rows.

### DIFF
--- a/core/src/actions/profiles.ts
+++ b/core/src/actions/profiles.ts
@@ -7,7 +7,6 @@ import { GroupMember } from "../models/GroupMember";
 import { Property } from "../models/Property";
 import { internalRun } from "../modules/internalRun";
 import { Op } from "sequelize";
-import { profile } from "winston";
 
 export class ProfilesList extends AuthenticatedAction {
   constructor() {
--- a/core/src/actions/profiles.ts
+++ b/core/src/actions/profiles.ts
@@ -131,8 +131,7 @@ export class ProfilesList extends AuthenticatedAction {
         total,
         profiles: await Promise.all(
           profiles.map(async (p) => {
-            const profile = await Profile.findById(p.id);
-            return profile.apiData();
+            return p.apiData();
           })
         ),
       };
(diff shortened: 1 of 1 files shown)


core/src/actions/profiles.ts
```diff
@@ -105,29 +105,36 @@ export class ProfilesList extends AuthenticatedAction {
...
 
       const total = await Profile.count({
         distinct: true,
-        where: { state: params.state ? params.state : { [Op.ne]: null } },
-        include: [{ model: ProfileProperty, where, required: requiredJoin }],
+        where: profileWhere,
+        include: countInclude,
       });
 
+      const profileData = [];
```
bleonard, line 163: Fascinatingly, it's actually faster to go serial because `Promise.all` was in parallel and didn't take advantage of the `Property` cache.

core/src/actions/profiles.ts
```diff
@@ -7,6 +7,7 @@ import { GroupMember } from "../models/GroupMember";
 import { Property } from "../models/Property";
 import { internalRun } from "../modules/internalRun";
 import { Op } from "sequelize";
+import { profile } from "winston";
```
evantahler: Probably not the profile you were looking for :D

evantahler (commented):
Nice!

As a note, @pedroslopez recently discovered a [`Model.findAndCountAll()`](https://sequelize.org/master/class/lib/model.js~Model.html#static-method-findAndCountAll) which could save another query here as well.

bleonard:
I'd be interested to know if `findAndCountAll` is actually faster. I bet it's still two queries because one has the offset.
And it mihgt not know how to work around what is slow here. Specifically, it might still do all the joining the business for the `count` query, even though it's not needed. The `join` was fine with 100 limit but took forever with the full 2M row count.


Investigation in sequelize source. Still two queries and wit hthe same options, so something to look out for @evantahler @pedroslopez 
```
    const [count, rows] = await Promise.all([
      this.count(countOptions),
      this.findAll(options)
    ]);
```
It's nice they are in parallel, but mostly helper is really just so you don't have to make shared variables like I did in this PR so they stay in sync, but it still might have performance issues if the `count` query is rough.

PR #1832, opened 2021-05-26
Bulk write ProfileProperties 

evantahler:
This PR aims to speed up all instances of writing ProfileProperties by writing them in bulk.  This will have a material speed increase for the `profile:completeImport` task, and every update ProfileProperties should be helped a little as well.

The main change is that we've come up with a conditional/partial index for the database which can do the unique ProfileProperty validation so we don't need to do it in code (hooks).  This required copying over the value of Property.unique (boolean) to the ProfileProperties table so the index could use it.   This allowed the logic for updating ProfileProperties to be moved to a single bulk method.

Notes:
* This will trigger v0.4 of Grouparoo because of the slow migrations.  Release notes here https://github.com/grouparoo/www.grouparoo.com/pull/485

TODO:
- [x] Sort out what's going on with the Event integration tests (array ProfileProperties)
- [x] Move the `@AfterSave` for Properties to update all thier ProfileProperties.unique values to a task, as this could be very slow.

### DIFF
--- a/core/src/migrations/000068-profilePropertyCompositeUniqueIndex.ts
+++ b/core/src/migrations/000068-profilePropertyCompositeUniqueIndex.ts
@@ -1,3 +1,4 @@
+import { config } from "actionhero";
 import { Op } from "sequelize";
 
 export default {
@@ -35,6 +36,14 @@ export default {
           where: { unique: { [Op.eq]: true } },
         }
       );
+
+      // All previous SQLite indexes had been removed (migration 000053), but we need to manually apply the per-profile position/value lock
+      // We cannot use Sequelize to apply the second index as it clobbers the first (https://github.com/sequelize/sequelize/issues/12823)
+      if (config.sequelize.dialect === "sqlite") {
+        await migration.sequelize.query(
+          "CREATE UNIQUE INDEX `profile_properties_profile_guid_profile_property_rule_guid_posi` ON `profileProperties` (`profileId`, `propertyId`, `position`)"
+        );
+      }
     });
   },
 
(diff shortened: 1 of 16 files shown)


core/src/migrations/000068-profilePropertyCompositeUniqueIndex.ts
```diff
@@ -0,0 +1,65 @@
...
+        }
+      );
+
+      // All previous SQLite indexes had been removed (migration 000053), but we need to manually apply the per-profile position/value lock
+      // We cannot use Sequelize to apply the second index as it clobbers the first (https://github.com/sequelize/sequelize/issues/12823)
+      if (config.sequelize.dialect === "sqlite") {
+        await migration.sequelize.query(
+          "CREATE UNIQUE INDEX `profile_properties_profile_guid_profile_property_rule_guid_posi` ON `profileProperties` (`profileId`, `propertyId`, `position`)"
+        );
+      }
```
evantahler, line 59: This is a real shame

core/src/migrations/000068-profilePropertyCompositeUniqueIndex.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      if (uniqueProperties.length > 0) {
+        await migration.sequelize.query(
+          `UPDATE "profileProperties" SET "unique"='${
+            uniqueProperties[0].unique
+          }' WHERE "propertyId" IN (${uniqueProperties
+            .map((p) => `'${p.id}'`)
+            .join(", ")})`
+        );
+      }
```
evantahler: This migration could be very slow for large datasets - it needs to update every row (profiles x number of properties)

evantahler:
This is going to be a new semantic version.  It's a slow migration.

PR #1833, opened 2021-05-26
Fix for ui-enterprise test

teallarson:
Created a css id in the main DOM of the desired page and anchored the test to that.  This should be a more stable and consistent solution.

(Previously this test was anchored to the Alert element on the setup page.  This element sometimes renders before the redirect is complete and the rest of the DOM is loaded, causing the test to fail.)


evantahler (approved):
Great research into why this is failing!

PR #1835, opened 2021-05-26
Setup steps page on /setup

teallarson:
Visiting `/`:
- If user is not logged in, displays button to login screen
- If user is logged in, displays button to complete setup steps (step one auto-completes from cluster name
- If setup steps are complete, displays button to configure profiles

After logging in:
- If user has not finished steps, takes them to `/setup`
- If user has finished steps, takes them to `/profiles`

Also shows setup steps progress on sidebar as in the other ui-* projects.


ui/ui-components/components/navigation.tsx
```diff
@@ -173,7 +174,7 @@ export default function Navigation(props) {
           </span>
         </div>
 
-        {navigationMode === "authenticated" && (
+        {navigationMode !== "unauthenticated" && (
```
evantahler: What's the state of the user when they first visit the page before registering?  I think they would be "unauthenticated", but can't access the setup steps until they register

Screen Shot 2021-06-03 at 5 36 00 PM

ui/ui-community/pages/session/sign-in.tsx
```diff
@@ -7,16 +7,37 @@ export default function SignInPage(props) {
...
+            </p>
+            <p>
+              Learn more at{" "}
+              <a target="_blank" href="https://www.grouparoo.com">
+                www.grouparoo.com
+              </a>
+            </p>
+            <br />
+          </Col>
+        </Row>
```
evantahler: If the user will see the common (ui-components) index page, they will already have seen this information. Why show it again on the sign in page?

ui/ui-components/components/navigation.tsx
evantahler: If there's a ever an attempt to load an action in any of the UIs which get an authentication error (which this is), we immediately take the user to the sign in page with a GET param indicating the original page they tried to view which caused the error.  I'm seeing `http://localhost:3000/session/sign-in?nextPage=/` which means that I'm currently unable to see the `/` page.

evantahler (changes requested):
I think the main issue is that the setup steps should not be loaded until the user is registered.  That's causing the pre-registered user to be unable to visit the `/` page, and the error.

seancdavis:
@evantahler I think what @tealjulia has done here has essentially combined the home page and the sign in page, making the home page essentially unnecessary. It sounded right when we talked it through, but thinking about it now, it does feel like we have this duplicate home page that the user is unlikely to see. But the move to load the setup steps was intentional, as we want to prompt registration even on the home page. That eliminates a click when we know first time users are going to have to register.

However, I wanted to reuse the client from ui-components and the auth flow we're using in the other applications, which assumes a redirect to `/session/sign-in`. Thoughts on some options:

1. Keep it as it is, but add a shared component for the duplicate content, knowing the user should never actually see that home page content after initial registration.
2. Allow the home page to be unauthenticated with a button to go to the registration form.
3. Add a conditional to the client that redirects to the home page in configland. And then the home page handles the entire redirect and registration process.

What do you think?

ui/ui-community/pages/session/sign-in.tsx
teallarson: @seancdavis explained it well below.  The way things redirect now, it automatically pushes them to /session/sign-in currently... so they don't see that initial landing page in config (as things are currently)

evantahler:
Given the choices from @seancdavis, I'd vote to stick with #2 (keep the `/` page and not do a redirect), which is how the other ui-* packages work - you can always got to `/`, logged in or logged out, and then see a CTA button about what to do next.  I agree that the page may be redundant, but we can revisit that holistically in another story (it's also potentially redundant in the same way if you are logged into ui-enterprise, for example). 

Regardless of the above though, it still looks like `/session/sign-in` is trying to load the setup steps when you aren't logged in.. which is causing the error.

ui/ui-components/pages/index.tsx
```diff
@@ -1,16 +1,29 @@
...
   let CTAMessage = "Sign In";
   let CTATarget: string;
 
-  if (navigationMode === "authenticated") {
+  if (navigationMode === "config" && currentStep) {
+    CTAMessage = "Set Up Grouparoo";
+    CTALink = "/setup";
+  } else if (navigationMode === "config" && !currentStep) {
+    CTAMessage = "Configure Your Profiles";
+    CTALink = "/profile";
```
pedroslopez: I believe this should be `/profiles`
```suggestion
    CTALink = "/profiles";
```

teallarson:
@seancdavis and @evantahler I've got it functioning in the way outlined above/discussed in standup.

I have a few mixed thoughts on how I've gone about it: 

1. It feels clunky how I've dealt with pausing the rendering/lifecycle, _but_ I'm not able to make the call as part of `getInitialProps()` because we don't know the `navigationMode` yet when that runs).  It doesn't cause a noticeable change in load time, but it doesn't feel right and I don't think I've seen this pattern used elsewhere in the ui.  If there is a better way to do this, I would love to learn more!

2. I wonder if it makes more sense to make a separate `ui-config/pages/index.tsx` with this logic versus using the shared `ui-components/pages/index.tsx` (would be a fast/easy change if so).  It doesn't quite feel right to be altering the lifecycle of the component for all ui editions if it's really only needed for one.

3. Still need to tweak testing.  Perhaps `ui-components` (where I first started them) is not the right place for this and the tests belong instead in the `ui-config` test suite?  The code is in `ui-components`, but it only really changes things in the `ui-config` world.

ui/ui-community/pages/session/sign-in.tsx
seancdavis: Is this a concern now that you aren't using redirects?

ui/ui-components/components/navigation.tsx
```diff
@@ -158,7 +159,7 @@ export default function Navigation(props) {
               </a>
             </Link>
             <br />
-            <Link href="/settings/[tab]" as="/settings/core">
+            <Link href={settingsLink} as={settingsLinkAs}>
```
seancdavis: I recall you mentioned you were going to remove the settings page. Did you do that? If so, might want to disable this link. Based on our workflow, I'd say it's safe to assume that it will be complete by the time the user is viewing it, since they have to register before they can view this page.

That is, unless they call their company "My Grouparoo Cluster"

seancdavis (commented):
My two cents:

> 1. It feels clunky how I've dealt with pausing the rendering/lifecycle, _but_ I'm not able to make the call as part of `getInitialProps()` because we don't know the `navigationMode` yet when that runs).  It doesn't cause a noticeable change in load time, but it doesn't feel right and I don't think I've seen this pattern used elsewhere in the ui.  If there is a better way to do this, I would love to learn more!

My gut is that we're fine for MVP. Let's at least get folks using it and clicking the buttons and then talk to them afterwards and see how we can improve.

> 2. I wonder if it makes more sense to make a separate `ui-config/pages/index.tsx` with this logic versus using the shared `ui-components/pages/index.tsx` (would be a fast/easy change if so).  It doesn't quite feel right to be altering the lifecycle of the component for all ui editions if it's really only needed for one.

Agreed. Feels weird that there is so many lines of logic in a shared space that is only relevant to one of the UIs. Sounds like we will eventually be thinking about this entire workflow more holistically and be able to find a common ground between all the UIs, but it feels unique for now.

> 3. Still need to tweak testing.  Perhaps `ui-components` (where I first started them) is not the right place for this and the tests belong instead in the `ui-config` test suite?  The code is in `ui-components`, but it only really changes things in the `ui-config` world.

Makes sense to me.

core/src/actions/navigation.ts
```diff
@@ -192,47 +192,50 @@ export class NavigationList extends OptionallyAuthenticatedAction {
       });
     }
 
-    if (navigationMode === "config") {
+    if (navigationMode === "config:unauthenticated") {
```
evantahler, line 195: This means that the user is able to navigate to `/apps`, `/profiles`, etc before they register.  Is this intentional?
Screen Shot 2021-06-09 at 8 41 41 AM

ui/ui-community/pages/session/sign-in.tsx
evantahler: All 3 sign-in pages `ui-enterprise/pages/session/sign-in.tsx`, `ui-community/pages/session/sign-in.tsx` and `ui-config/pages/session/sign-in.tsx` have been modified in this PR to copy over the content from `/`.  Perhaps now that there aren't redirects, that's not the case?

ui/ui-config/pages/session/sign-in.tsx
```diff
@@ -26,71 +26,112 @@ export default function SignInPage(props) {
...
+            </p>
+            <p>
+              Learn more at{" "}
+              <a target="_blank" href="https://www.grouparoo.com">
+                www.grouparoo.com
+              </a>
+            </p>
+            <br />
+          </Col>
+        </Row>
```
evantahler: Linking the thread from https://github.com/grouparoo/grouparoo/pull/1835/files#r648408807 - now that there aren't redirects to sign in from `/`, is this content needed?

ui/ui-components/components/navigation.tsx
teallarson: Good call.  I had it linking to '`'/'` instead of `'/settings/core'` on config mode... but I think disabling it makes more sense.

teallarson:
@evantahler:
@seancdavis  and I were able to get things _near_ where we're happy about it.  

Currently the tests run and pass as written.  However, we weren't able to test the behavior if a user has completed their setup steps.  

A few thoughts on what's next/how we _could_ do that:
1. Because that page has more things going on than just paths/authorization, if we decide to include it, it should probably be its own test file and not a part of `happyPath.ts`
2. How to go about it:
    a. Generate config files in the temp directory where the test lays down the `.local`.  Then it would pass when `index` checks if the steps are complete
    b. Alternately, is there a way to access the sequelize instance in an integration test (ie: using `integrationSpecHelper` and not `grouparooTestServer`?  We could manually mark them as complete and return that object.
3. Is this all just a lot for testing the behavior of one button and what happens when we click on it?  Feels like it might be.

seancdavis:
@evantahler Background on this ...

> b. Alternately, is there a way to access the sequelize instance in an integration test (ie: using integrationSpecHelper and not grouparooTestServer? We could manually mark them as complete and return that object.

Is that my first idea was to access the `SetupStep` model directly and set all the steps to `complete: true`, which should have mimicked the behavior Teal wanted. But that threw an error saying we were detached from a Sequelize instance.

We also went through the process of creating an action specifically for this. Ran into some issues along the way, but it felt really silly to have a whole action just to serve test behavior. So it ended up feeling like we were spending too much time to test a button and a redirect.

core/src/actions/navigation.ts
seancdavis: @evantahler The logic here is saying that if a config user is _not_ authenticated, all they see is a link to `/` (which would have the button to sign in). If they are authenticated, then they get the full nav.

core/src/actions/navigation.ts
```diff
@@ -192,9 +192,13 @@ export class NavigationList extends OptionallyAuthenticatedAction {
...
+      navigationItems = [{ type: "link", title: "Home", href: "/" }];
```
seancdavis, line 196: @tealjulia Two questions:

1. Should we add an icon for home, just to be consistent?
2. Do you think we should also have a sign in link?

plugins/@grouparoo/spec-helper/src/lib/integrationSpecHelper.ts
```diff
@@ -66,6 +66,23 @@ export namespace IntegrationSpecHelper {
...
+
+  export async function createConfigUser(projectPath: string) {
+    // const testUserFilePath = projectPath + "../.local/user.json";
+    // const testFileDir = path.dirname(testUserFilePath);
+    // if (!fs.existsSync(testFileDir)) {
+    //   fs.mkdirSync(testFileDir, { recursive: true });
+    // }
+    // const fileContent = { email: true };
+    // fs.writeFileSync(testUserFilePath, JSON.stringify(fileContent, null, 2));
+  }
```
seancdavis: Seems like you're not using this any more?

ui/ui-community/pages/session/sign-in.tsx
```diff
@@ -1,4 +1,4 @@
-import { Row, Col } from "react-bootstrap";
+import { Row, Col, Image } from "react-bootstrap";
```
seancdavis: Is `Image` being used?

ui/ui-components/components/navigation.tsx
```diff
@@ -60,6 +60,8 @@ export default function Navigation(props) {
   const [expandAccountMenu, setExpandAccountMenu] = useState(false);
   const subMenuGreeting = `Hello ${teamMember ? teamMember.firstName : ""} »`;
   const logoLink = teamMember?.id ? "/dashboard" : "/";
+  const settingsLink = navigationMode === "config" ? "/" : "/settings/[tab]";
+  const settingsLinkAs = navigationMode === "config" ? "/" : "/settings/core";
```
seancdavis: Is this logic still necessary?

ui/ui-components/pages/index.tsx
```diff
@@ -8,7 +8,7 @@ export default function Page(props) {
 
   let CTALink = "/session/sign-in";
   let CTAMessage = "Sign In";
-  let CTATarget: string;
+  let CTATarget = null;
```
seancdavis: You may not have to lose the type definition here. Could you do something like this:

```ts
let CTATarget?: string;
```

Or this:

```ts
let CTATarget: string | null = null;
```

Not sure exactly what you're protecting against with this change.

ui/ui-config/__tests__/integration/happyPath.ts
```diff
@@ -33,11 +39,68 @@ describe("integration", () => {
   });
 
   test(
-    "it renders the registration page",
+    "it renders the home page",
```
seancdavis, line 42: These are great happy path tests!

ui/ui-config/pages/index.tsx
```diff
@@ -1,47 +1,104 @@
...
+    return <Loader />;
+  }
 
-export default function IndexPage() {
   return (
     <>
       <Head>
         <title>Grouparoo</title>
+        <p>NAV MODE: {navigationMode}</p>
+        <p>{JSON.stringify(CTAs, null, 2)}</p>
```
seancdavis: This one is on me for adding these while we were debugging.

ui/ui-enterprise/pages/session/sign-in.tsx
```diff
@@ -1,4 +1,4 @@
-import { Row, Col } from "react-bootstrap";
+import { Row, Col, Image } from "react-bootstrap";
```
seancdavis: Image being used here?

ui/ui-config/pages/session/sign-in.tsx
```diff
@@ -1,11 +1,10 @@
 import { useState } from "react";
-import { Row, Col, Form } from "react-bootstrap";
+import { Row, Col, Form, Image } from "react-bootstrap";
```
seancdavis: Image being used here?

seancdavis (changes requested):
Hey @tealjulia — this feels super close! Left a handful of minor comments, almost entirely about removing what seem like old references or were for debugging.

ui/ui-config/pages/index.tsx
evantahler: 🐙

evantahler:
Not to over-crowd this thread, but I'll just add that I'm OK with not testing some of this work for now.  It seems like we've hit a wall where further testing would would be a diminishing return.  I've added a [chore](https://www.pivotaltracker.com/story/show/178500990) to the icebox to expose models to the integration test helper

ui/ui-config/pages/index.tsx
```diff
@@ -1,47 +1,104 @@
...
+
+  useEffect(() => {
+    if (navigationMode === "config:authenticated") {
+      getSetupSteps(props);
+    } else {
+      setShouldRender(true);
+    }
+
+    async function getSetupSteps(props) {
+      console.log(`fetching steps`);
```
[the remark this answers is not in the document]
teallarson: 😅 IN MY DEFENSE.... I was pushing this to get feedback on testing, not to merge.  But yeah.

core/src/actions/navigation.ts
teallarson: I left this as is because it mirrors the other ui-* apps.  If we want to change that, I think we should change them all to maintain consistency?

seancdavis (approved):
Woohoo!

PR #1836, opened 2021-05-26
`buildNullProperties` defaults to Profile Properties in the 'pending' state

evantahler:


core/__tests__/tasks/system/status.ts
```diff
@@ -59,6 +63,10 @@ describe("tasks/status", () => {
 
     test("will not be complete with a pending export", async () => {
       const profile = await helper.factories.profile();
+      await ProfileProperty.update(
```
pedroslopez, line 66: when the profile is created, the null profileproperties that are created now start "pending", so we manually mark them as ready, makes sense!

pedroslopez (approved):
I think this looks good!

core/__tests__/tasks/system/status.ts
evantahler: Yeah - I was worried that /a lot/ of tests would need to be updated, but it wasn't that many!

PR #1838, opened 2021-05-27
dont lock bootstrapped property in config ui mode

pedroslopez:


core/__tests__/modules/codeConfig/codeConfig.ts
```diff
@@ -901,6 +901,119 @@ describe("modules/codeConfig", () => {
...
+
+    afterAll(async () => {
+      await helper.truncate();
+      process.env.GROUPAROO_RUN_MODE = undefined;
+    });
+
+    test('settings are locked with "config:code"', async () => {
+      const setting = await plugin.readSetting("core", "cluster-name");
+      expect(setting.value).toBe("Test Cluster");
+      expect(setting.locked).toBe("config:code");
```
pedroslopez, line 951: There are still some models (settings, teams, apiKeys) that are always locked independent of running in config mode. Is the right behavior?

core/src/modules/configLoaders/property.ts
```diff
@@ -12,6 +12,15 @@ import { Op } from "sequelize";
...
 
+async function isBootstrappedProperty(
+  configObject: ConfigurationObject,
+  source: Source
+) {
+  const sourceMapping = await source.getMapping();
+  const mappedPropertyKeys = Object.values(sourceMapping);
+  return mappedPropertyKeys.includes(configObject.key || configObject.name);
+}
+
```
evantahler: Can you use `property.directlyMapped` here?

core/__tests__/modules/codeConfig/codeConfig.ts
evantahler: Those are models that shouldn't be used in config mode... so that's an interesting question.  I think for now keeping them locked is OK.  If someone tries to write to them they will get an error which hints that we don't expect those models to be used right now

evantahler (approved):
👍 But with a question about `isBootstrappedProperty` - i'd like to clarify how that differs from `property.directlyMapped` and if we can start using the same methods/names to talk about them.   I for one am confused!

core/src/modules/configLoaders/property.ts
pedroslopez: you're right!

PR #1839, opened 2021-05-27
Sample Profiles in Config UI

pedroslopez:


core/__tests__/modules/configWriter.ts
```diff
@@ -10,6 +10,7 @@ import { Group } from "../../src/models/Group";
 import { Property } from "../../src/models/Property";
```
evantahler, line 10: Are there any tests to check that a Profile Config object:
* won't be written in an environment other than `cli:config`
* a reasonable error will be shown if a poorly defined profile config object is provided, e.g. (`{firstName: 'evan'}`)

core/src/actions/profiles.ts
```diff
@@ -255,6 +256,8 @@ export class ProfileCreate extends AuthenticatedAction {
 
     const groups = await profile.$get("groups");
 
+    await ConfigWriter.run();
```
evantahler, line 259: 👍 the configWriter does nothing in environments that are not `cli:config`

core/src/classes/codeConfig.ts
```diff
@@ -117,15 +118,16 @@ export function validateConfigObjectKeys(
...
+  name?: string
 ) {
   let logLevel = "info";
   if (mode === "created") logLevel = "notice";
   if (mode === "deleted") logLevel = "warning";
 
   log(
     `[ config ] ${mode} ${instance.constructor.name} \`${
-      instance.key || instance.email || instance.name
+      name || instance.key || instance.email || instance.name
```
evantahler, line 130: I think I missed where did the new `name` property get used?  It looks like the new Profile config objects still have `id`?

```
    {
      id: "profile_john",
      class: "Profile",
      properties: {
        user_id: 20,
      },
    },
```

core/src/models/Profile.ts
```diff
@@ -109,7 +109,7 @@ export class Profile extends LoggedModel<Profile> {
 
   async addOrUpdateProperties(
     properties: {
-      [key: string]: Array<string | number | boolean | Date>;
+      [key: string]: Array<string | number | boolean | Date> | any;
```
evantahler: That's a shame... it's been really helpful to enforce the Array type.   What required this change?

ui/ui-components/components/tabs/profile.tsx
```diff
@@ -8,5 +8,10 @@ export default function ProfileTabs({
   profile: Models.ProfileType;
 }) {
   const tabs = ["edit", "imports", "events", "exports", "logs"];
+
+  if (process.env.GROUPAROO_UI_EDITION === "config") {
+    tabs.splice(tabs.indexOf("imports"), 2);
+  }
```
evantahler, line 14: 👍 "events" and "exports" tabs are not shown in ui-config

ui/ui-config/pages/profile/new.tsx
```diff
@@ -0,0 +1,85 @@
...
+    if (response?.profile) {
+      router.push(`/profile/[id]/edit`, `/profile/${response.profile.id}/edit`);
+    } else {
+      setLoading(false);
+    }
+  }
+
+  const uniqueDirectlyMappedProfileProperties = props.properties.filter(
+    (rule) => rule.unique && rule.directlyMapped
+  );
```
evantahler: It might be helpful to give ` props.properties` a type of `Actions.PropertiesList`

ui/ui-config/pages/profiles.tsx
```diff
@@ -0,0 +1,24 @@
+import { Button } from "react-bootstrap";
+import { useRouter } from "next/router";
+import ProfilesPage from "@grouparoo/ui-components/pages/profiles";
+
+export default function Page(props) {
+  const router = useRouter();
+  return (
+    <>
+      <ProfilesPage {...props} />
```
evantahler, line 9: Nice! Taking the existing page and wrapping the new components around it.

evantahler (changes requested):
You are on the right track for sure.  I think there are some more tests to add, and perhaps there are some improvements which can be made to the Types of a few things - specifically, keeping `Profile.addOrUpdateProperties()` as only accepting arrays.

core/src/classes/codeConfig.ts
pedroslopez: profiles don't have a name, so logging it produced `undefined`:

```
notice: [ config ] created Profile `undefined` (pro_440aee4d-a936-46f2-a3ee-ca822b405d3c)
```

I'm kind of faking a name by passing the stringified properties:
```
notice: [ config ] created Profile `{"user_id":21}` (pro_440aee4d-a936-46f2-a3ee-ca822b405d3c)
```

core/src/models/Profile.ts
pedroslopez: Re-checked and it wasn't really necessary. Moved back to using arrays!

core/__tests__/modules/configWriter.ts
pedroslopez: For the first one, I think this is more a test for the configwriter.run itself (which is tested in `modules/configWriter`). However, I added a test to check that config writer is being run on create/destroy actions.

core/src/classes/codeConfig.ts
evantahler: Ah! That's a good log message too.  Carry on then!

evantahler (approved):
Awesome!

PR #1841, opened 2021-05-28
Auth workflow for Config UI

seancdavis:
Adds  an authentication workflow to Config UI.

## Workflow

This piggy-backs off the current auth middleware, adding a conditional for being in configland. I could have chosen to use a separate middleware, but decided to keep it together for the time being.

1. Authenticated requests look for a `.local/user.json` file with an object containing `email: true`. If it find that, auth is passed.
2. Otherwise, redirect to the registration form.
3. Registration form will load company name from cluster name setting if it has been changed. Otherwise, it's a blank form field.
4. When submitting, company gets saved to cluster name setting (not written to file yet, see below), subscribed will trigger a call to Telemetry, and the email address triggers creation of a `.local/user.json` file with contents `{ email: true }`.
5. Redirect to index page.

## Other Notable Items

- I added some icons to the navigation, but didn't pick one for Apps, thinking @tealjulia could help me out there. Also didn't make a story for that, considering the change is so small.
- There are no navigation items when not authenticated in configland.
- Created a new action that returns the cluster name without authentication. This was to be able to pre-populate the company name in the form.
- Adjusted GrouparooSubscription to allow for sending a subscribe request when we have only an email, not a full team member.

## Not Addressed

To get this merged and so @tealjulia can continue her work, I've broken out the following out into other stories:

- [Write cluster name to file idempotently when ConfigWriter runs](https://www.pivotaltracker.com/story/show/178336436), and run it after updating a setting.
- [Refactor navigation action](https://www.pivotaltracker.com/story/show/178336448). It's messy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ coverage
 *.sqlite-*
 apps/*/config/*
 apps/*/*.log
+apps/*/.local
 .pnpm-debug.log
 
 # API
--- a/core/__tests__/actions/settings.ts
+++ b/core/__tests__/actions/settings.ts
@@ -3,6 +3,8 @@ import { specHelper } from "actionhero";
 import { Setting, plugin } from "../../src";
 
 describe("actions/settings", () => {
+  let connection;
+
   helper.grouparooTestServer({ truncate: true, resetSettings: true });
 
   beforeAll(async () => {
@@ -12,10 +14,33 @@ describe("actions/settings", () => {
       password: "P@ssw0rd!",
       email: "mario@example.com",
     });
+
+    connection = await specHelper.buildConnection();
+
+    await Setting.truncate();
+  });
+
+  test("can publicly read the cluster name", async () => {
+    const setting = await plugin.registerSetting(
+      "testPlugin",
+      "cluster-name",
+      "My Grouparoo Cluster",
+      "My Grouparoo Cluster",
+      "Name of the cluster",
+      "string"
+    );
+    const {
+      error,
+      clusterName,
+      default: defaultValue,
+    } = await specHelper.runAction("settings:clusterName:view", connection);
+    expect(error).toBeUndefined();
+    expect(clusterName).not.toBeUndefined();
+    expect(clusterName).toEqual(setting.value);
+    expect(defaultValue).toEqual(true);
   });
 
   describe("reader signed in", () => {
-    let connection;
     let csrfToken;
     let id;
 
--- a/core/__tests__/actions/settings.ts
+++ b/core/__tests__/actions/settings.ts
@@ -20,26 +20,6 @@ describe("actions/settings", () => {
     await Setting.truncate();
   });
 
-  test("can publicly read the cluster name", async () => {
-    const setting = await plugin.registerSetting(
-      "testPlugin",
-      "cluster-name",
-      "My Grouparoo Cluster",
-      "My Grouparoo Cluster",
-      "Name of the cluster",
-      "string"
-    );
-    const {
-      error,
-      clusterName,
-      default: defaultValue,
-    } = await specHelper.runAction("settings:clusterName:view", connection);
-    expect(error).toBeUndefined();
-    expect(clusterName).not.toBeUndefined();
-    expect(clusterName).toEqual(setting.value);
-    expect(defaultValue).toEqual(true);
-  });
-
   describe("reader signed in", () => {
     let csrfToken;
     let id;
--- a/core/__tests__/modules/configUser.ts
+++ b/core/__tests__/modules/configUser.ts
@@ -26,6 +26,18 @@ describe("modules/ConfigUser", () => {
     if (fs.existsSync(localFile)) fs.rmSync(localFile);
   });
 
+  test("does nothing unless in cli:config mode", async () => {
+    process.env.GROUPAROO_RUN_MODE = undefined;
+    expect(fs.existsSync(ConfigUser.localUserFilePath())).toEqual(false);
+    await ConfigUser.create({
+      email: "demo@grouparoo.com",
+      company: "My Company",
+    });
+    expect(fs.existsSync(ConfigUser.localUserFilePath())).toEqual(false);
+    const user = await ConfigUser.get();
+    expect(user).toEqual(null);
+  });
+
   // also tests the get() method
   test("writes a file to .local/user.json", async () => {
     expect(fs.existsSync(ConfigUser.localUserFilePath())).toEqual(false);
--- a/core/__tests__/modules/configUser.ts
+++ b/core/__tests__/modules/configUser.ts
@@ -1,0 +1,58 @@
+import fs from "fs";
+import os from "os";
+import { helper } from "@grouparoo/spec-helper";
+
+import { ConfigUser } from "../../src/modules/configUser";
+import { Setting, plugin } from "../../src";
+
+const workerId = process.env.JEST_WORKER_ID;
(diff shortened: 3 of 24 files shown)


.gitignore
```diff
@@ -14,6 +14,7 @@ coverage
 *.sqlite-*
 apps/*/config/*
 apps/*/*.log
+apps/*/.local
```
evantahler, line 17: 👍 confirmed that this is also in the .gitignore we generate for client projects

core/__tests__/actions/settings.ts
```diff
@@ -12,10 +14,33 @@ describe("actions/settings", () => {
...
+
+  test("can publicly read the cluster name", async () => {
+    const setting = await plugin.registerSetting(
+      "testPlugin",
+      "cluster-name",
+      "My Grouparoo Cluster",
+      "My Grouparoo Cluster",
+      "Name of the cluster",
+      "string"
+    );
```
evantahler: The setting should be created by default.  You may not need this.

core/__tests__/modules/configUser.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  test("stores the company as the cluster name", async () => {
+    await plugin.registerSetting(
+      "testPlugin",
+      "cluster-name",
+      "My Grouparoo Cluster",
+      "My Grouparoo Cluster",
+      "Name of the cluster",
+      "string"
+    );
```
evantahler, line 62: Same note as above - settings are all created at boot

core/src/actions/config.ts
```diff
@@ -69,3 +71,24 @@ export class ConfigGenerate extends AuthenticatedAction {
...
+    this.inputs = {
+      email: { required: true },
+      company: { required: true },
+      subscribed: { required: false },
+    };
+    this.outputExample = {};
+  }
+
+  async runWithinTransaction({ params }) {
+    await ConfigUser.create(params);
```
evantahler, line 93: Should this action fail if `GROUPAROO_RUN_MODE !== 'cli:config'`?

core/src/actions/settings.ts
```diff
@@ -29,6 +30,28 @@ export class SettingsList extends AuthenticatedAction {
   }
 }
 
+export class SettingClusterNameView extends OptionallyAuthenticatedAction {
```
evantahler: Doesn't this information also come back from the navigation action?  We do display the cluster name in the UI when you are logged out already.  The navigation action is already pre-loaded and available on every UI page though `props.navigation`

core/src/modules/middleware/authentication.ts
```diff
@@ -93,6 +87,41 @@ async function authenticateTeamMember(
...
+
+// Conditionally choose auth method based on run mode.
+async function authenticateTeamMember(
+  data: { [key: string]: any },
+  optional: boolean
+) {
+  let error: Error;
+
+  if (
+    process.env.GROUPAROO_RUN_MODE === "cli:config" &&
```
evantahler, line 118: 👍 Can you please test that in the regular run modes you cannot authenticate with a .local user.

ui/ui-config/pages/index.tsx
```diff
@@ -33,3 +35,13 @@ export default function IndexPage() {
...
+
+IndexPage.getInitialProps = async (ctx) => {
+  const { execApi } = useApi(ctx);
+  // Hit an auth route so that we redirect to sign in if not already signed in.
+  const apiStatus: Actions.PrivateStatus = await execApi(
+    "get",
+    `/status/private`
+  );
+  return { apiStatus };
+};
```
evantahler, line 47: Nice trick!

evantahler (changes requested):
Nice work... and thank you for writing up such good PR descriptions and stories!

1. I think you may not need the action to read the setting if you can get it from the navigation
2. Please test the new auth method a little more throughly - notes below.

core/src/actions/navigation.ts
```diff
@@ -192,15 +193,42 @@ export class NavigationList extends OptionallyAuthenticatedAction {
...
-        { type: "link", title: "Plugins", href: "/plugins" },
-        { type: "link", title: "Validate", href: "/validate" },
-      ];
+      const user = await ConfigUser.get();
+      if (user) {
+        navigationItems = [
+          {
+            type: "link",
+            title: "Apps",
+            href: "/apps",
```
teallarson, line 202: Per your comment above, my thoughts are [links](https://fontawesome.com/v5.15/icons/link?style=solid), since they're things you can connect, or [squares](https://fontawesome.com/v5.15/icons/th-large?style=solid), since it somewhat mirrors the imagery we have on our site and on the apps page in the ui.  My other thought was a cube since they're the building blocks, but fontawesome's cube looks like a company logo to me.

pedroslopez: +1 for the squares. I've seen this be used for "apps" in other... _apps_... too.

.gitignore
seancdavis: Do you think we should put individual `.gitignore` files in these example projects? Or having one at the root that captures what we add to the template we generate on `init` is fine?

core/src/actions/settings.ts
seancdavis: Sure does. I adjusted the structure of it to get a little more information (to know if it has been set or not). But we're only using it in the nav, so it was a quick fix.

core/src/actions/navigation.ts
seancdavis: @tealjulia @pedroslopez @evantahler Any idea what I have to do to get this working? I can't get anything other than what we're already currently using to work, even though the appropriate files are clearly there. Are we enabling a specific set somehow?

core/__tests__/actions/settings.ts
seancdavis: Put this in the code above it, just to be sure:

```js
const count = await Setting.count();
console.log("...", count);
```

And got back `0`. 

Also tried to add the `resetSettings: true` option to the test server in the helper, but that didn't change the result.

Is there something else I have to do to not need this code?

core/__tests__/modules/configUser.ts
seancdavis: Put this in the code above it, just to be sure:

```js
const count = await Setting.count();
console.log("...", count);
```

And got back `0`. 

Also tried to add the `resetSettings: true` option to the test server in the helper, but that didn't change the result.

Is there something else I have to do to not need this code?

core/__tests__/actions/settings.ts
seancdavis: I ended up deleting this code because it was testing the action that we don't need.

core/src/actions/config.ts
seancdavis: Yep! Added specs to ensure this is happening.

core/src/actions/navigation.ts
teallarson: Did you add it to the [icon library file](https://github.com/grouparoo/grouparoo/blob/7687269d90e4fad52374d63fb2eeb8c068e8c698/ui/ui-components/components/icons.ts)?

core/__tests__/modules/configUser.ts
evantahler: That's super weird.   https://github.com/grouparoo/grouparoo/blob/main/core/src/initializers/settings.ts should be creating all the settings in this initializer.  Let's pair on this!

core/src/actions/navigation.ts
seancdavis: omg you are my hero

PR #1842, opened 2021-05-28
fix: correctly save destinationGroupMemberships for Destination config object

pedroslopez:
A bit unintuitively, in the config object keys are actually the name in the destination and values are group IDs (which is backwards from the object used by `setDestinationGroupMemberships`)


evantahler (approved):
Good catch!

We'll be updating the format of these objects in https://www.pivotaltracker.com/story/show/177734437 to make them more clear

PR #1844, opened 2021-05-28
Run summary count fix

teallarson:
Summary after `roo run` was not generating accurate counts, specifically # of imports from sources on a first run with a new db.

Changing `findById` to `findByPk` seems to have solved the issue (`findById` was deprecated a few versions ago in sequelize)

I've tried running multiple times, both using the same db for multiple runs in a row and destroying/rebuilding and am now consistently getting the same (accurate) counts.


evantahler (approved):
I imagine the changes to "greater than or equals to" have more to do with the fix than the `findBy` change. If the run was started at the same second as the app, the run would have been skipped prior to this PR. 

I also didn't know `findById` was deprecated!

PR #1870, opened 2021-06-01
fix: use convenient rules when generating group config objects

pedroslopez:


evantahler (approved):
Nice! Good find on nulling-out the relative* properties

PR #1872, opened 2021-06-01
Fix issues when deleting multiple objects at the same time

pedroslopez:
This PR addresses a few issues that occur when deleting multiple models at the same time (through code config). In general, deleting each model individually was working fine but when many or all of these are in the deleted state, there were some weird interactions between them that prevented things from happening, such as exporting people one last time to the group.

- **Test with dependencies in `deleted` state**: In multiple places, like export profiles, I'm now running through the tests twice, once with apps/sources/properties in `ready` state and another with them in `deleted` state. In most cases, all of these should still work fine if things are deleted. For example, we need to use the app to get API credentials when exporting people one last time.
- **destinations:destroy kicks off a run even if the group is deleted**: Previouslty there was a condition to prevent this from happening. However, because the destination was deleted before getting to the group run, the final export would not occur.
- **Relax the used scope for destinations/sources**: There were some checks to prevent deletion of dependent models like Groups if they were still being used by a destination. However, because deleted destinations werent being included, the groups were being deleted before the destination was able to 
- **Properties' default scope includes `deleted` props**: This is one I'm unsure about, because it affects the default set of properties returned. However, in all places I could find you still want to be able to access deleted properties and use them. Something that pushed me over the edge to do this is that when exporting profiles we get the properties with `.findAllWithCache`, which needs to include those that have been deleted to properly export.
- **Bootstrapped property is also enqueued to be deleted**: Before, it was deleted immediately when the config file was removed, which caused issues down the line because the property was missing. Now it's deleted when the source:destroy task deletes the source.
- **Destroying schedules now deletes their runs**: Previously, runs were just stopped. While this isn't directly an issue that prevents proper deletion of models, it means that if you bring it back with the same ID the profiles would not be re-imported. This is because the highWatermark is stored on the previous run, which still existed. We're now deleting previous runs so that if you bring it back everything is reimported properly.

### DIFF
--- a/core/__tests__/models/destination/plugins/exportProfile.ts
+++ b/core/__tests__/models/destination/plugins/exportProfile.ts
@@ -6,133 +6,20 @@ import {
   Export,
   Group,
   Profile,
+  Property,
 } from "../../../../src";
 import { api, specHelper } from "actionhero";
 import { Op } from "sequelize";
 
 describe("models/destination", () => {
-  helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
-  beforeAll(async () => await helper.factories.properties());
+  describe.each(["ready", "deleted"])("with app in %p state", (appState) => {
+    helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
+    beforeAll(async () => await helper.factories.properties());
 
-  describe("with custom exportProfile plugin", () => {
-    let app: App;
-    let destination: Destination;
-    let exportArgs = {
-      app: null,
-      appOptions: null,
-      destination: null,
-      destinationOptions: null,
-      syncOperations: null,
-      profile: null,
-      oldProfileProperties: null,
-      newProfileProperties: null,
-      oldGroups: null,
-      newGroups: null,
-      toDelete: null,
-    };
-    let exportArrayProperties = [];
-    let parallelismResponse = Infinity;
-    let exportProfileResponse = {
-      success: true,
-      error: undefined,
-      retryDelay: undefined,
-    };
-    let exportProfileThrow = null;
-
-    beforeAll(async () => {
-      plugin.registerPlugin({
-        name: "test-export-plugin",
-        apps: [
-          {
-            name: "test-template-app",
-            options: [{ key: "test_key", required: true }],
-            methods: {
-              test: async () => {
-                return { success: true };
-              },
-              parallelism: async () => {
-                return parallelismResponse;
-              },
-            },
-          },
-        ],
-        connections: [
-          {
-            name: "export-from-test-app",
-            description: "a test app connection",
-            app: "test-template-app",
-            direction: "export",
-            options: [],
-            syncModes: ["sync", "enrich", "additive"],
-            methods: {
-              destinationMappingOptions: async () => {
-                return {
-                  labels: {
-                    group: {
-                      singular: "list",
-                      plural: "lists",
-                    },
-                    property: {
-                      singular: "var",
-                      plural: "vars",
-                    },
-                  },
-                  properties: {
-                    required: [],
-                    known: [],
-                    allowOptionalFromProperties: true,
-                  },
-                };
-              },
-              exportArrayProperties: async () => exportArrayProperties,
-              exportProfile: async ({
-                app,
-                appOptions,
-                destination,
-                destinationOptions,
-                syncOperations,
-                export: {
-                  profile,
-                  oldProfileProperties,
-                  newProfileProperties,
-                  oldGroups,
-                  newGroups,
-                  toDelete,
-                },
-              }) => {
-                exportArgs = {
-                  app,
-                  appOptions,
-                  destination,
-                  destinationOptions,
-                  syncOperations,
-                  profile,
-                  oldProfileProperties,
-                  newProfileProperties,
-                  oldGroups,
-                  newGroups,
-                  toDelete,
-                };
-                if (exportProfileThrow) {
-                  throw exportProfileThrow;
-                }
-                return exportProfileResponse;
-              },
-            },
-          },
-        ],
-      });
-
-      app = await App.create({
-        name: "test with real methods",
-        type: "test-template-app",
-      });
-      await app.setOptions({ test_key: "abc" });
-      await app.update({ state: "ready" });
-    });
-
-    beforeEach(async () => {
-      exportArgs = {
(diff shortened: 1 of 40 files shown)


core/__tests__/models/destination/plugins/exportProfile.ts
```diff
@@ -6,495 +6,485 @@ import {
...
-          {
-            name: "test-template-app",
-            options: [{ key: "test_key", required: true }],
-            methods: {
-              test: async () => {
-                return { success: true };
-              },
-              parallelism: async () => {
-                return parallelismResponse;
+  describe.each(["ready", "deleted"])("with app in %p state", (appState) => {
```
evantahler, line 15: 👍  Nice use of test.each!

core/__tests__/models/property/localCache.ts
```diff
@@ -10,10 +10,18 @@ describe("models/property", () => {
...
+    const deletedProperty = await helper.factories.property(
+      source,
+      { key: "deletedProp" },
+      { column: "col" }
+    );
+    await deletedProperty.update({ state: "deleted" });
+
     originalPropertyCount = await Property.scope(null).count({
-      where: { state: "ready" },
+      where: { state: ["ready", "deleted"] },
```
evantahler, line 23: Umm... I didn't know there was a shorthand for `Op.In`!

core/src/actions/properties.ts
```diff
@@ -259,7 +259,7 @@ export class PropertyView extends AuthenticatedAction {
 
   async runWithinTransaction({ params }) {
     const property = await Property.findById(params.id);
-    const source = await property.$get("source");
+    const source = await property.$get("source", { scope: null });
```
evantahler, line 262: 👍 makes sense that if you are calling a property explicitly by It's ID, you can see it in any state

core/src/actions/properties.ts
```diff
@@ -333,7 +333,7 @@ export class PropertyProfilePreview extends AuthenticatedAction {
     }
 
     const apiData = await profile.apiData();
-    const source = await property.$get("source");
+    const source = await property.$get("source", { scope: null });
```
evantahler, line 336: Not sure about this one - if the source isn't `ready`, should we not be able to preview any properties against it?

core/src/models/Destination.ts
```diff
@@ -114,6 +115,13 @@ const STATE_TRANSITIONS = [
 @DefaultScope(() => ({
   where: { state: "ready" },
 }))
+@Scopes(() => ({
+  notDraft: {
+    where: {
+      state: { [Op.notIn]: ["draft"] },
+    },
+  },
+}))
```
evantahler, line 124: 👍 we don't have many scopes in the app yet, but this one is doing to be used a few times.

ui/ui-components/pages/properties.tsx
```diff
@@ -115,6 +115,8 @@ export default function Page(props) {
     );
   }
 
+  const readySources = sources.filter((s) => s.state === "ready");
```
evantahler, line 118: This change means you can't see the `draft` properties, which you can today. 

For the UIs in general, I think showing the `deleted` properties is OK - we do that already for Groups and Destinations.

core/src/models/Schedule.ts
```diff
@@ -302,13 +302,13 @@ export class Schedule extends LoggedModel<Schedule> {
...
     const runs = await Run.findAll({
-      where: { creatorId: instance.id, state: "running" },
+      where: { creatorId: instance.id },
     });
 
     for (const i in runs) {
-      await runs[i].update({ state: "stopped" });
+      await runs[i].destroy();
     }
   }
```
evantahler, line 313: 👍 I'm OK with this change.  I think we did this for history reasons (eg: this run made this import which set this profile property)... but... it's OK.

evantahler (commented):
That's one heck of a PR!  Great finds all over the app!  Great PR writeup too. 

Addressing your main **defaultScope** question: In general, I see a lot of `scope: null` being added throughout the app.  We added those scopes in the past because we we had trouble "running" models that weren't yet ready... so most of the time we only wanted to consider the "ready" models.  This PR fixes a lot of those problems!  Do you think we still need the default scope?  I could see a wold in which we drop the default scope, and all models gain a `"notDraft"` scope for use in tasks mostly.

ui/ui-components/pages/properties.tsx
pedroslopez: Note that this is filtering the state on the _sources_, not the properties. Previously we were only getting `ready` sources, but I changed this to also retrieve the deleted ones. Because of this, I added this filtered array to only show sources that are ready in the select box for the form at the bottom of the page to create a new property.

core/src/actions/properties.ts
pedroslopez: If you went to the page of a property that had its source `deleted`, the preview would have an error on it saying that you can't call `importProfileProperty` on `null`. However, the error doesn't break anything else and it's just displayed for that property in the preview. Do you think we should instead just have a better error message for when the source isn't ready?

pedroslopez:
@evantahler Yeah, I think if we're clearing the scope practically every time we use the model then maybe the default scope isn't great. If the issues are mostly around not using `draft` models, maybe the default should just be all those that are not draft. This would include deleted models in the default scope, and then if a certain task really only wants those that are ready, then we can query that state. I didn't want to do it here because I'm not sure what those places are where they expect the scope to only include the models that are ready, and it would probably take some more investigation, but let me know if you'd like me to look into it.

core/__tests__/models/property/localCache.ts
pedroslopez: [Yeah](https://sequelize.org/master/manual/model-querying-basics.html#shorthand-syntax-for--code-op-in--code-)!

ui/ui-components/pages/properties.tsx
evantahler: Ah ha! In that case, nice!

core/src/actions/properties.ts
evantahler: Let's leave this as you have it until we get to https://www.pivotaltracker.com/story/show/178417997

evantahler (approved):
Good call @pedroslopez.  Let's merge this in to fix the deleting bugs, and then we can soon revisit the scopes holistically in another story.

PR #1884, opened 2021-06-02
Clear config require cache when new plugin installed

evantahler:
Now we can install new plugins! 
app-install

### DIFF
--- a/core/src/actions/plugins.ts
+++ b/core/src/actions/plugins.ts
@@ -2,6 +2,10 @@ import { AuthenticatedAction } from "../classes/actions/authenticatedAction";
 import { OptionallyAuthenticatedAction } from "../classes/actions/optionallyAuthenticatedAction";
 import { Plugins } from "../modules/plugins";
 import { api } from "actionhero";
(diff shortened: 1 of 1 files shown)


core/src/actions/plugins.ts
```diff
@@ -81,12 +85,33 @@ export class PluginUninstall extends AuthenticatedAction {
...
   }
 }
+
+function clearPluginConfigCache() {
+  const files = ["plugins.js"];
+
+  files.forEach((file) => {
+    const fullPath = path.join(__dirname, "..", "config", file);
+    if (fs.existsSync(fullPath)) {
+      delete require.cache[require.resolve(fullPath)];
```
evantahler, line 114: `core/(src/dis)/config/plugins` was loaded once at boot and exported the list of plugins at that time.  Node caches improt/require values and assumes they don't change... but in this case, they do!

seancdavis (approved):
Yes! Great find!

evantahler:
I have no idea how to test this however :(

PR #1885, opened 2021-06-02
Add scripts to staging-config to make it easier to work in config mode

seancdavis:
Adds two scripts that are explained in a new README file in the staging-config demo project.

### DIFF
--- a/apps/staging-config/package.json
+++ b/apps/staging-config/package.json
@@ -20,8 +20,8 @@
     "jest": "26.6.3"
   },
   "scripts": {
-    "config": "rm grouparoo_config.sqlite; rm -rf .local; CONFIG_DATABASE_URL=\"sqlite://grouparoo_config.sqlite\" NEXT_DEVELOPMENT_MODE=true ./node_modules/.bin/grouparoo config",
-    "demo": "./node_modules/.bin/grouparoo demo purchases --config"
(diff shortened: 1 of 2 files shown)


apps/staging-config/package.json
```diff
@@ -12,16 +12,17 @@
     "@grouparoo/core": "0.3.3",
     "@grouparoo/demo": "0.3.3",
     "@grouparoo/postgres": "0.3.3",
-    "@grouparoo/sqlite": "0.3.3"
+    "@grouparoo/sqlite": "0.3.3",
+    "grouparoo": "0.3.3"
```
evantahler: I would advise against this, as `@grouparoo/core` already has `grouparoo` as a dependency.  The demo script could be `./node_module/@grouparoo/core/node_modules/.bin/grouparoo demo purchases --config`

seancdavis: @evantahler Making this change then still tries to get into a `cli` directory, which isn't there.

```
Cannot find module '/path/to/monorepo/apps/staging-config/node_modules/@grouparoo/cli/dist/grouparoo.js'
```

evantahler: Fair enough.  There's probably some changes we could make in the demo plugin/CLI loader to work a little better in this case... but that's another story! Carry on.

evantahler (approved):

PR #1887, opened 2021-06-02
More complex filter/aggregationMethod example in Property template

teallarson:
Added an additional default example to template to show how filters/aggregationMethods can work together.


evantahler (approved):
Good example!

PR #1889, opened 2021-06-03
removed `build_static_refs` objects for property rules

teallarson:
Per a discussion in [this PR](https://github.com/grouparoo/www.grouparoo.com/pull/481) we no longer want these being auto-generated from Core as we need to do a bit of manipulation (namely, adding an example and a caption) before it ends up in the www.


evantahler (approved):
(as discussed in slack) - we still need to keep around `buildPropertyOpsDictionary()`

PR #1892, opened 2021-06-04
Demo cmd will already clear just the applicable redis.

bleonard:
so it won't take a hammer to all the dbs.


bin/timed_run
```diff
@@ -21,9 +21,6 @@ cd ../../core
 npm run prepare
 cd $PROJECT_DIR
 
-echo "-----> CLEARING REDIS\r\n"
-redis-cli flushall
-
```
evantahler, line 26: 👍

PR #1895, opened 2021-06-07
Generate pretty ID values on config objects

seancdavis:
I first spent some time exploring building `id` values from name-like attributes on objects by adjusting how `LoggedModel` generates these values. That turned into a complex situation because we have to account for objects being created before they have a name available.

I'm making the broad assumption here that we don't actually care about objects in that state. If an object is in a state in the Config UI app such that it doesn't have a name, then it's probably not ready to write to file anyways.

## Changes

So this PR does a number of things:

1. Adds a `getConfigId()` method to every model that had a `getConfigObject()` method. `getConfigId()` is a synchronous method that uses a helper supplied by the Config Writer to generate some pretty ID value based on its name-like attribute.
2. Adjusts `getConfigObject()` such that associated objects' `getConfigId()` methods are called and populated for ID values.
3. `getConfigObject()` will return `undefined` when it doesn't have the appropriate values.

This way I could bypass how the app works and only manipulate how we're generating the config objects themselves.

## A Note on Speed

I tried to be smart about eager loading and I ran into a few hurdles. So I decided to put that to the side. In fact, I'm actively _not_ eager loading here, such that I can always grab the most up to date associated objects. This way, I don't rely on memory, but every time we generate the config objects, we generate _all of them_, and we make sure that we're explicitly retrieving the latest values and associations in all cases.

## Known Edge Case

One last quick note: If you're using the app and paying attention to ID values, it may seem a little weird. When generating objects, you're going to get the obscure UUID in the URL, though the pretty ID will be written to file. When booting it up the next time, you'll have the pretty ID. 

What that means is that _URLs may change_ between boots, and this is expected. I am assuming that there isn't going to be much URL sharing in this app and that it would be used primarily just to generate the objects. And if there is any sharing, it's probably not between when the object is created and the first time it is updated.

### DIFF
--- a/core/__tests__/modules/configWriter.ts
+++ b/core/__tests__/modules/configWriter.ts
@@ -363,7 +363,7 @@ describe("modules/configWriter", () => {
       const { name, type } = source;
       const app = await source.$get("app");
       const options = await source.getOptions();
-      const mapping = await MappingHelper.getMapping(source, "id");
+      const mapping = await MappingHelper.getConfigMapping(source);
 
       expect(config).toEqual({
         class: "Source",
@@ -384,7 +384,7 @@ describe("modules/configWriter", () => {
       const { name, type } = source;
       const app = await source.$get("app");
       const options = await source.getOptions();
-      const mapping = await MappingHelper.getMapping(source, "id");
+      const mapping = await MappingHelper.getConfigMapping(source);
 
       expect(config.length).toEqual(2);
       expect(config[0]).toEqual({
@@ -471,7 +471,10 @@ describe("modules/configWriter", () => {
     });
 
     test("destinations can provide their config objects", async () => {
-      const destination: Destination = await helper.factories.destination();
+      const destination: Destination = await helper.factories.destination(
+        undefined,
+        { groupId: group.id }
+      );
       const app: App = await destination.$get("app");
 
       const destinationGroupMemberships = {};
@@ -485,7 +488,7 @@ describe("modules/configWriter", () => {
       const { name, type, syncMode } = destination;
 
       const options = await destination.getOptions();
-      const mapping = await MappingHelper.getMapping(destination, "id");
+      const mapping = await MappingHelper.getConfigMapping(destination);
 
       expect(config.id).toBeTruthy();
       expect(config).toEqual({
@@ -499,7 +502,7 @@ describe("modules/configWriter", () => {
         options,
         mapping,
         destinationGroupMemberships: {
-          "My Dest Tag": group.id,
+          "My Dest Tag": group.getConfigId(),
         },
       });
     });
--- a/core/__tests__/modules/configWriter.ts
+++ b/core/__tests__/modules/configWriter.ts
@@ -34,6 +34,20 @@ describe("modules/configWriter", () => {
     process.env.GROUPAROO_RUN_MODE = undefined;
   });
 
+  // ---------------------------------------- | ConfigWriter.generateId()
+
+  describe("generateId()", () => {
+    test("it returns undefined when no name is passed", () => {
+      expect(ConfigWriter.generateId(null)).toBeUndefined();
+    });
+    test("it returns a slugified, lowercase name", () => {
+      let res = ConfigWriter.generateId("Hello World");
+      expect(res).toEqual("hello-world");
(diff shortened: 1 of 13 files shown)


core/src/models/App.ts
```diff
@@ -14,6 +14,7 @@ import {
 } from "sequelize-typescript";
 import { api, redis } from "actionhero";
 import { Op } from "sequelize";
+import slugify from "slugify";
```
evantahler: Doesn't seem to be used?

core/src/modules/configWriter.ts
```diff
@@ -1,12 +1,13 @@
 import { api } from "actionhero";
 import fs from "fs";
 import path from "path";
+import slugify from "slugify";
```
evantahler: Could we use slugify elsewhere in code config... perhaps when providing IDs via the CLI?

evantahler (changes requested):
This is a tricky one, as we talked about.  I too would have first tried an approach which modifies the model's IDs in general, so that it would work for ui-enterprise as well.  This would have also solved the changing-IDs edge case you called out.  But, in order to move things forward, I think this should be merged and we can move on.  I added a [story to the icebox](https://www.pivotaltracker.com/story/show/178451222) to revisit this later. 

Addressing your note on speed - In config-ui, I think it's totally OK to be inefficient. 

Also, I wanted to confirm that Options and Filters are OK when the ID changes.  You've added code to the MappingHelper to update things when the parent ID changes... are those other models OK?  Please add tests for this.

core/src/modules/configWriter.ts
seancdavis: Good idea! I'll add it to the generator.

seancdavis: I looked at what the generator was doing and I ended up going back to a custom method. We had some behavior in the generator that slugify couldn't provide to us, so I ditched it and went with a shared and custom solution and added tests for the conditions we want to solve for.

core/src/classes/configTemplate.ts
```diff
@@ -106,7 +107,7 @@ export abstract class ConfigTemplate {
   }
 
   formatId(s: string) {
-    return s.toLowerCase().replace(/[^a-zA-Z0-9-_\/.]/gi, "_");
+    return ConfigWriter.generateId(s);
```
seancdavis, line 110: @evantahler Big difference here is that we _were_ allowing periods and slashes and not replacing them. But we have solid tests for the generate command, and they are still passing. IOW I was thinking that the slashes and period were unnecessary, and actually may cause and issue in some cases because we don't really want them in the ID values that we're generating.

seancdavis:
@evantahler On this:

> Also, I wanted to confirm that Options and Filters are OK when the ID changes. You've added code to the MappingHelper to update things when the parent ID changes... are those other models OK? Please add tests for this.

I'm not testing these explicitly, but what I've done is adjust the way in which the config writer tests are written such that we can ensure no config object shape ever provides a value that comes from the primary key in the database, but instead uses the various `getConfigId()` methods. By doing this, we ensure that the config shapes are related as they should be.

And this was a great idea! Options were somewhat irrelevant because they don't use primary keys, but this helped me catch a bug with group mappings.

core/src/classes/configTemplate.ts
evantahler: Sold!

core/__tests__/modules/configWriter.ts
```diff
@@ -280,151 +362,237 @@ describe("modules/configWriter", () => {
     let property: Property;
     let group: Group;
 
-    beforeAll(async () => {
+    let sourceOptions = { table: "test-table-02" };
+    let bsPropertyCol: string = faker.database.column();
+    let bsPropertyKey: string = faker.random.alphaNumeric(12);
+    let bsPropertyId: string = faker.datatype.uuid();
```
evantahler: No bad words in the codebase please.

evantahler (approved):
That's some thorough testing! 

> Options were somewhat irrelevant because they don't use primary keys, but this helped me catch a bug with group mappings.

I was struggling with the above note for a while until I realized that the problem with mappings is that their /values/ are IDs. Options, Group Rules, Filters, etc have /keys/ that are IDs, and don't need to be stored in the config files because they are part of the parent object.  I get it now!

PR #1896, opened 2021-06-07
fix: get status report for deleted sources and destinations

pedroslopez:


core/src/modules/statusReporters.ts
```diff
@@ -437,7 +437,7 @@ export namespace FinalSummaryReporters {
         await run.updateTotals();
         const schedule = await Schedule.findByPk(run.creatorId);
 
-        if (schedule) source = await schedule.$get("source");
+        if (schedule) source = await schedule.$get("source", { scope: null });
```
evantahler, line 440: 👍  Makes sense - Some exports may have been sent in the case where something was deleted

evantahler (approved):

PR #1898, opened 2021-06-08
Date group rules to not require `"type": "date"`

teallarson:
BEFORE: had to feed a "type" to the config file for date based rules

NOW: If a group rule is based on a calculation, check the associated property's type to see if it's a date.  If so, it is parsed to unix timestamp.

### DIFF
--- a/core/src/modules/configLoaders/group.ts
+++ b/core/src/modules/configLoaders/group.ts
@@ -57,7 +57,7 @@ export async function loadGroup(
         rules[i].key = property.key;
 
         // if calculating based on a date, parse to unix timestamp
-        if (calculatesWithDate.indexOf(rules[i]["operation"]["op"]) >= 0) {
+        if (calculatesWithDate.includes(rules[i]["operation"]["op"])) {
           if (property.type === "date") {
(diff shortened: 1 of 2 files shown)


core/src/modules/configLoaders/group.ts
```diff
@@ -49,19 +49,19 @@ export async function loadGroup(
...
-      }
 
-      //parses to epoch time if calculated date rule
-      if (
-        calculatesWithDate.indexOf(rules[i]["operation"]["op"]) >= 0 &&
-        rules[i]["type"] === "date"
-      ) {
-        rules[i]["match"] = Date.parse(rules[i]["match"].toString());
+        // if calculating based on a date, parse to unix timestamp
+        if (calculatesWithDate.indexOf(rules[i]["operation"]["op"]) >= 0) {
```
evantahler: You can use [`calculatesWithDate.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) here!  I too am trying to rid myself of of the `indexOf` pattern of yesterday.

evantahler (approved):

PR #1899, opened 2021-06-08
Remove run references in Config UI

seancdavis:
There's probably some refactoring we could do here, especially since some of these components look so similar. Right now I'm just trying to move quickly.

Would you rather I take the time to refactor here and clean up the UI, or is this show/hide based on UI edition enough for MVP for config?


evantahler (approved):
👍 for now!  Speed ahead!

PR #1900, opened 2021-06-08
Groups don't have "updating" state in Config UI

seancdavis:
This approach says that groups can't have a run in config mode. It seemed like that was the preference.

I don't see where we are testing this method elsewhere, and felt like that would be getting into a whole can of worms. If you can show me where we are testing already, or if you feel strongly that I should add new tests for this method, I can spend some time with it.

_Note: There are also some updates (fixes) to the staging-config commands._

### DIFF
--- a/core/__tests__/models/group/group.ts
+++ b/core/__tests__/models/group/group.ts
@@ -60,6 +60,44 @@ describe("models/group", () => {
     expect(log.message).toBe('group "doomed group" destroyed');
   });
 
+  describe("run()", () => {
+    const getRuns = async (group) => {
+      const runs = await Run.findAll({
+        where: { creatorId: group.id, creatorType: "group" },
+      });
+      return runs;
+    };
+
+    afterEach(async () => {
+      process.env.GROUPAROO_RUN_MODE = undefined;
+    });
+    test("creates a run and sets state to initialized", async () => {
+      const group = await Group.create({
+        name: "group that will create a run",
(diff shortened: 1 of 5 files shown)


evantahler (changes requested):
The code change looks good to me!

You are right! We aren't going a good job of testing this.  We test the side effects in `__tests__/models/group/calculatedGroup` (eg: "changing group rules changes the state to initializing and enquires a run...") but do not test the logic that it /doesn't/ make a run in CLI or CONFIG modes.

I'd suggest adding a test for each GROUPAROO_RUN_MODE in `__tests__/models/group/` that `group.run()` doesn't do anything in those modes, and does make a run in the "normal' mode.  I think you can just change `process.env.GROUPAROO_RUN_MODE` and then call the method and check if it returned a run or not.

core/__tests__/models/group/group.ts
```diff
@@ -60,6 +60,44 @@ describe("models/group", () => {
...
+      expect(runs.length).toEqual(1);
+      expect(runs[0].state).toEqual("running");
+    });
+    test("does not create a group when in config mode", async () => {
+      process.env.GROUPAROO_RUN_MODE = "cli:config";
+
+      const group = await Group.create({
+        name: "group that will not create a run",
+        type: "manual",
+      });
```
bleonard, line 92: Probably should check the state here. I think it's "initializing" and not "ready"

PR #1909, opened 2021-06-08
Test `timed_run` weekly with 100,000 profiles

evantahler:
This PR adds a Github Action which will run `./bin/timed_run` every week against a demo database of 100,000 profiles.  Running this regularly will let us learn about performance improvements or reductions.  This action can also be run upon request with a workflow_dispatch

* Will report traces to a special Sentry.io project called `grouparoo-timed-run-action` (10% of all tasks)
* Uses the demo dataset with a scale of 100 (1000 * 100 = 100,000 Profiles)
* Uses `LOG_LEVEL=alert` to be a little less noisy.

We should be OK as long as the total runtime is under 6 hours: https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits

### DIFF
--- a/.github/workflows/timed-run.yml
+++ b/.github/workflows/timed-run.yml
@@ -1,0 +1,77 @@
+name: timed-run
+
+on:
+  schedule:
+    - cron: "0 2 * * 5"
+  workflow_dispatch:
+    inputs:
+      app:
+        description: "app to test?"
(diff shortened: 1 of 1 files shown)


evantahler:
Screen Shot 2021-06-08 at 5 03 08 PM

PR #1913, opened 2021-06-09
Clean up SQLite plugin queries

seancdavis:
What I ended up doing here was to use goofy table names when running the specs in this plugin to ensure that we're using quotes where necessary to escape characters in identifiers.

### DIFF
--- a/plugins/@grouparoo/sqlite/src/lib/export/exportProfile.ts
+++ b/plugins/@grouparoo/sqlite/src/lib/export/exportProfile.ts
@@ -37,11 +37,11 @@ export const exportProfile: ExportProfilePluginMethod = async ({
         );
       }
       // delete
-      const query = `DELETE FROM ${table} WHERE ${primaryKey} = ${newProfileProperties[primaryKey]}`;
+      const query = `DELETE FROM "${table}" WHERE ${primaryKey} = ${newProfileProperties[primaryKey]}`;
       validateQuery(query);
       await connection.asyncQuery(query);
     } else if (newProfileProperties[primaryKey]) {
-      const query = `SELECT * FROM ${table} WHERE ${primaryKey} = ${newProfileProperties[primaryKey]}`;
+      const query = `SELECT * FROM "${table}" WHERE ${primaryKey} = ${newProfileProperties[primaryKey]}`;
       validateQuery(query);
       const existingRecords = await connection.asyncQuery(query);
       if (existingRecords.length === 1) {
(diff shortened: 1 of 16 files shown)


plugins/@grouparoo/sqlite/src/lib/export/exportProfile.ts
```diff
@@ -37,11 +37,11 @@ export const exportProfile: ExportProfilePluginMethod = async ({
         );
       }
       // delete
-      const query = `DELETE FROM ${table} WHERE ${primaryKey} = ${newProfileProperties[primaryKey]}`;
+      const query = `DELETE FROM "${table}" WHERE ${primaryKey} = ${newProfileProperties[primaryKey]}`;
```
evantahler, line 40: While we are on the topic of escaping, I'm wondering if /all/ user provided values should be quoted?  What if `newProfileProperties[primaryKey]` in this case was a UUID?  Would it fail?

> DELETE FROM users WHERE uuid = abc123

Should it be

> DELETE FROM users WHERE uuid = "abc123"

?

plugins/@grouparoo/sqlite/src/lib/export/exportProfile.ts
```diff
@@ -37,11 +37,11 @@ export const exportProfile: ExportProfilePluginMethod = async ({
...
       validateQuery(query);
       await connection.asyncQuery(query);
     } else if (newProfileProperties[primaryKey]) {
-      const query = `SELECT * FROM ${table} WHERE ${primaryKey} = ${newProfileProperties[primaryKey]}`;
+      const query = `SELECT * FROM "${table}" WHERE ${primaryKey} = ${newProfileProperties[primaryKey]}`;
```
evantahler, line 44: Does `primaryKey` need to be quoted here too?  What if the column name was `"user id"` or something like that?

> SELECT * FROM "users" where user id = 1

Should it be 

> SELECT * FROM "users" where "user id" = 1

?

evantahler (commented):
I think this is a good start for sure, and trying wacky values in the test suite is a good path forward.  

I'm a little concerned any time we try to escape things ourselves.  I think with table names that's /probably/ ok, but this PR opens up some questions about other string replacements (see comments).  How far do we need to go?  What's the level of risk that you think we might have here?

There are prepared statements and escaping built into the SQLite driver (https://github.com/mapbox/node-sqlite3#usage) that might be helpful:

```
var stmt = db.prepare("INSERT INTO lorem VALUES (?)")
stmt.run("foo")
```

Maybe `asyncQuery` could be modified to accept a string and then an array of replacements?

seancdavis:
After a conversation, we've decided to merge this and [create a new story](https://www.pivotaltracker.com/story/show/178488236).

PR #1914, opened 2021-06-09
End validation if error is `instanceof Sequelize.DatabaseError`

teallarson:
Checks if error message includes 'transaction aborted' and will stop trying to validate other configs if so.

Also updated the link to the sequelize docs

### DIFF
--- a/core/src/modules/configLoaders/index.ts
+++ b/core/src/modules/configLoaders/index.ts
@@ -269,16 +269,22 @@ export async function processConfigObjects(
       }
     } catch (error) {
       // Normally, we can can keep going after an error and keep checking the other config objects
-      // but, there's some types of errors (like unique key duplicates) which pollute the transaction and we need to stop
+      // but, there's some types of errors (like unique key duplicates and casting incorrect types) which pollute the transaction and we need to stop
 
       const { message, fields } = GrouparooErrorSerializer(error);
 
       const errorMessage = `[ config ] error with ${configObject?.class} \`${
         configObject.key || configObject.name
       }\` (${configObject.id}): ${message}`;
+
       errors.push(errorMessage);
+
       log(error?.stack || error, "debug");
-      if (fields.length === 0) {
+
+      if (errorMessage.includes("current transaction is aborted")) {
+        log(errorMessage, "warning");
+        throw new Error(`Cannot validate additional objects.`);
+      } else if (fields.length === 0) {
         log(errorMessage, "warning");
         continue;
       } else {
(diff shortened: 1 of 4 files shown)


core/src/modules/configLoaders/index.ts
```diff
@@ -269,16 +269,22 @@ export async function processConfigObjects(
...
       }\` (${configObject.id}): ${message}`;
+
       errors.push(errorMessage);
+
       log(error?.stack || error, "debug");
-      if (fields.length === 0) {
+
+      if (errorMessage.includes("current transaction is aborted")) {
+        log(errorMessage, "warning");
+        throw new Error(`Cannot validate additional objects.`);
```
bleonard: Maybe something around Sequelize. Like `Sequelize database error. Cannot validate additional objects.`

evantahler: Even tighter, it may be possible to check if `error instanceof Sequelize.ConnectionError` (+ some other types of Sequelize errors). Here are a few of the error types https://github.com/grouparoo/grouparoo/blob/e414d4bc136336ca13860f7fd1ccacef23a3958f/core/src/config/sequelize.ts#L135

evantahler (changes requested):
I think there's a tighter way to check for a sequelize error than looking at the message string.

teallarson:
Update: 
- Now checks `instanceof` the error instead of message, and checks/stops before running the processing function.  I've only included `Sequelize.DatabaseError` as that was what the story specified -- though I could include others as @evantahler pointed out.  Open to thoughts.
- The error message on log level `info` and `debug` now both more human readable/useful:
Screen Shot 2021-06-10 at 9 55 13 AM

Screen Shot 2021-06-10 at 9 56 22 AM
(`debug` now has a line noting what query the transaction was aborted on)

More thoughts:
- If we remove `DatabaseError` from the `retry` list in the file @evantahler linked, the whole thing fails faster instead of sitting on that failing object for a few seconds.  But are there some `DatabaseError`s that we do want to retry on?  Didn't want to mess with that without knowing.
- If this seems like the right approach, I can clean up the now failing tests.  They're failing because I changed the error object to include the SQL that failed for debug mode.

evantahler (approved):
I like this approach!  I don't think you need to worry about the other types of database errors for now - those connection/timeout errors are unlikely to happen on localhost, and if they do happen, they would have prevented the app from starting up at all.

In general, for the rest of the app, we probably do want to keep `DatabaseError` in the retry list - when the app is running (`grouparoo run/start`) we might hit an error like "cannot apply transaction" if 2 workers try to update the same row at the same time, so this will allow the task to re retried.  Same with `ConnectionError`s - If the network is being weird and we have trouble, we want to let the task/action retry a few times before giving up.

PR #1915, opened 2021-06-10
fix: source with directlyMapped property can be destroyed

pedroslopez:
Previously, deleting a source with a directlyMapped property threw an error saying that it still had a property, and trying to delete the property threw an error saying that it was being used in a mapping. This change allows these to be deleted by deleting the directlyMapped property along with the source, like code config was doing.

### DIFF
--- a/core/src/models/Property.ts
+++ b/core/src/models/Property.ts
@@ -17,7 +17,7 @@ import {
   BeforeCreate,
   AfterSave,
 } from "sequelize-typescript";
-import { Op } from "sequelize";
+import { Op, WhereOptions } from "sequelize";
 import { env, api, redis, config } from "actionhero";
 import { plugin } from "../modules/plugin";
 import { LoggedModel } from "../classes/loggedModel";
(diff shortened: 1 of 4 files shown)


core/src/models/Property.ts
```diff
@@ -17,7 +17,7 @@ import {
   BeforeCreate,
   AfterSave,
 } from "sequelize-typescript";
-import { Op } from "sequelize";
+import { Op, WhereOptions } from "sequelize";
```
evantahler, line 20: Yeah! Sequelize has some really helpful types now!

core/src/models/Source.ts
```diff
@@ -377,26 +377,33 @@ export class Source extends LoggedModel<Source> {
...
+    });
 
     if (properties.length > 0) {
       throw new Error("cannot delete a source that has a property");
     }
   }
 
+  @BeforeDestroy
+  static async ensureDirectlyMappedPropertyNotInUse(instance: Source) {
+    const directlyMappedProperty = await Property.findOne({
```
evantahler, line 398: 👍 using a real database read and not cached values

evantahler (approved):
Nice!

PR #1916, opened 2021-06-10
Move config CLI command to ui-config

seancdavis:
I was able to get this to work in this super weird way. Being that ui-config is a Next project, it already has configuration in its root directory geared toward the Next build. So I used our common src/dist pattern and through a custom `tsconfig.json` file in `src`.

My manual tests in the monorepo proved fruitful, but I imagine we'll want to give this another one after the next release.

I thought about trying to programmatically test this, but it felt like overkill. It'll either work or it won't and would be pretty obvious right away. Beyond ensuring that the file exists in `dist`, it didn't feel worth the effort. But curious to get your take.


ui/ui-config/package.json
```diff
@@ -34,6 +34,7 @@
     "@fortawesome/free-solid-svg-icons": "5.15.3",
     "@fortawesome/react-fontawesome": "0.1.14",
     "@zeit/next-source-maps": "0.0.3",
+    "actionhero": "^26.0.6",
```
evantahler: This is probably should be a devDependency (looking at the other plugins)

evantahler (commented):
I'm a little worried about having 2 typescript build/config files in the same package.   Would it be crazy to write `ui-config/src/bin/config.js` directly in a `dist` folder directly (e.g.: write the CLI command in JS and skip the build step)?

I'm also OK skipping the tests on this for now - it will pretty clear once this is deployed if it works or not.

seancdavis:
@evantahler What do you think of the changes now? It seems to work fine. The weirdest part was figuring out how to track the file with git. [Here's an interesting explanation of how ignoring ignored files works](https://stackoverflow.com/a/20652768/2241124).

evantahler:
Nice! I too never knew how to get the "ignoring of ignored files" to work either!

PR #1918, opened 2021-06-11
Remove group calcs in Config UI

seancdavis:
Removes the following in the Config UI:

* \# of Profiles column
* Button to Count potential group members
* "Total Profiles in this group: #" line
* number of group members in destinations
* List view. remove calculatedAt and Profiles columns
* Edit View:  "never updated" and "Next Member Calculation"

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -20,37 +20,37 @@ importers:
 
   apps/staging-community:
     specifiers:
-      '@grouparoo/bigquery': 0.4.0-alpha.2
-      '@grouparoo/core': 0.4.0-alpha.2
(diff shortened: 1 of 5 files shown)


evantahler (approved):
👍 works for me for now! Confirming that the API will continue to return group counts and calculated at, but they will always be 0 and null, respectively. 

As we add more and more of these `if` checks to the UI, what would be the more "react" way... a conditionally rendering component?

```tsx
<UImode not="config">Group Members: {total}</UImode>
```

seancdavis:
I like that pattern a lot. A conditional component, rather than having to add the conditional everywhere.

PR #1922, opened 2021-06-12
Use regular `specHelper` for UI tests

evantahler:
This PR updates `@grouparoo/specHelper` such that `grouparooTestServer` now works for the UI plugins as well! The old `integrationSpecHelper` is removed.  The old `integrationSpecHelper` was bad because it ran the UI packages in a wacky way - in a sub-process, relying on the staging apps.  This was both slower and made working with the models in the test impossible... something you probably need in test!

Of note:
- A lot less boilerplate!
- Getting Jest, webdriver, and Next to play nicely was a struggle.  For some reason, in the web projects `jest.setup` needs to be typescript files (vs JS in all the other projects) so that we can use `ts-loader` and not whatever Next does.  This way things work without modifying the custom `tsconfig` Next likes.

Also in this PR:
* Now that the UI tests are working better, I noticed that if you visited session/sign-out directly (vs routing to it though react) it caused a problem because you were being signed out as the setup steps were loading.  This PR fixes that.


ui/ui-config/__tests__/integration/happyPath.ts
```diff
@@ -92,16 +102,34 @@ describe("integration", () => {
...
+    await browser.wait(until.elementLocated(by.className("btn-primary")));
+    const button = browser.findElement(by.className("btn-primary"));
+    await button.click();
+
+    await helper.sleep(1000);
+
+    const currentUrl = await browser.getCurrentUrl();
+    expect(currentUrl).toMatch(/\/profiles/);
+    await browser.get(currentUrl);
+  });
```
evantahler, line 133: @tealjulia @seancdavis - here's the test that started all of this! Now, we can update the SetupStep model, reload the page, and see that we are taken to `/profiles` rather than `/setup`

ui/ui-enterprise/__tests__/integration/happyPath.ts
```diff
@@ -90,29 +77,38 @@ describe("integration", () => {
...
+
   test(
     "I can sign out",
     async () => {
-      await browser.get(`${env.url}/session/sign-out`);
+      await browser.get(`${url}/session/sign-out`);
       await browser.wait(until.elementLocated(by.tagName("p")));
-      const url = await browser.getCurrentUrl();
-      expect(url).toMatch(/\//);
-      await browser.get(url);
```
teallarson, line 107: I see that this line is no longer used in the tests... what was its purpose before we switched the specHelper?  Why is it no longer necessary?  I was curious why we would `browser.get(url)` when the url was already derived from `browser.getCurrentUrl()`.

evantahler: I also don't know!  

I agree that it seems unnecessary, since we already confirmed the page was `/`... and the test works without those lines.

PR #1925, opened 2021-06-14
Using env vars for secret values in Config UI

seancdavis:
While secrets worked when _loading_ config files on boot, they were being retrieved and written back to file using their resolved values. This PR introduces the functionality to write the env variable **keys** to file, along with new specs to test that functionality. The behavior is now in place for five of the writable models, though it will likely only apply to apps.

This also brings a note to the app edit page when in Config mode to note how to work with secret values by linking to our documentation.

### DIFF
--- a/core/__tests__/modules/configWriter.ts
+++ b/core/__tests__/modules/configWriter.ts
@@ -426,6 +426,17 @@ describe("modules/configWriter", () => {
       expect(config).toBeUndefined();
     });
 
+    test("apps don't transform env vars", async () => {
+      process.env.GROUPAROO_OPTION__APP__CONFIG_WRITER_ENV_VAR = "my_file_123";
+      const app: App = await helper.factories.app({
+        options: { fileId: "CONFIG_WRITER_ENV_VAR" },
+      });
+      const config = await app.getConfigObject();
+      expect(config.options.fileId).toEqual("CONFIG_WRITER_ENV_VAR");
+      const options = await app.getOptions();
+      expect(options.fileId).toEqual("my_file_123");
+    });
+
     test("sources can provide their config objects", async () => {
       const config = await source.getConfigObject();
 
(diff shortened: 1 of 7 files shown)


core/__tests__/modules/configWriter.ts
```diff
@@ -660,7 +707,20 @@ describe("modules/configWriter", () => {
...
+        "my_table_456";
+      const destination: Destination = await helper.factories.destination(
+        undefined,
+        { groupId: group.id, options: { table: "CONFIG_WRITER_ENV_VAR" } }
+      );
+      const config = await destination.getConfigObject();
+      expect(config.options.table).toEqual("CONFIG_WRITER_ENV_VAR");
+      const options = await destination.getOptions();
+      expect(options.table).toEqual("my_table_456");
+    });
```
evantahler, line 721: These are good tests

core/__tests__/modules/configWriter.ts
```diff
@@ -68,13 +68,13 @@ describe("modules/configWriter", () => {
       await App.destroy({ truncate: true });
     });
 
-    test.skip("provides a url-friendly path from the name in the object", async () => {
```
evantahler, line 71: Are there any more skipped tests?

core/src/models/App.ts
```diff
@@ -238,7 +238,7 @@ export class App extends LoggedModel<App> {
   async getConfigObject() {
     const { type, name } = this;
 
-    const options = await this.getOptions();
+    const options = await this.getOptions(false);
```
evantahler, line 241: Nice! That `sourceFromEnvironment` argument /was/ helpful after all!

ui/ui-components/pages/app/[id]/edit.tsx
```diff
@@ -174,9 +174,22 @@ export default function Page(props) {
...
+                        <a
+                          target="_blank"
+                          href="https://www.grouparoo.com/docs/support/secrets"
+                        >
+                          Learn more
+                        </a>
+                        .
+                      </>
+                    )}
+                  </p>
```
Comment (evantahler) [inline ui/ui-components/pages/app/[id]/edit.tsx:192]: Not something that needs to change now, but maybe we always want to show this note on the page

evantahler (approved):

core/__tests__/modules/configWriter.ts
seancdavis: Nope! Did a global search and seems like we're in good shape.

core/src/models/App.ts
seancdavis: [Thanks, Past Evan!](https://github.com/grouparoo/grouparoo/pull/1533)

PR #1926, opened 2021-06-14
Fix [Object] display bug on Destination Data

evantahler:
Before

Screen Shot 2021-06-14 at 9 27 10 AM

After

Screen Shot 2021-06-14 at 9 27 01 AM

### DIFF
--- a/ui/ui-components/pages/destination/[id]/data.tsx
+++ b/ui/ui-components/pages/destination/[id]/data.tsx
@@ -302,9 +302,8 @@ export default function Page(props) {
                       .map((group) => (
                         <option key={`grp-${group.id}`} value={group.id}>
                           {group.name}
-                          {process.env.GROUPAROO_UI_EDITION !== "config" && (
-                            <>&nbsp;{group.profilesCount} members</>
-                          )}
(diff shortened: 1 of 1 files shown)


ui/ui-components/pages/destination/[id]/data.tsx
```diff
@@ -302,9 +302,8 @@ export default function Page(props) {
                       .map((group) => (
                         <option key={`grp-${group.id}`} value={group.id}>
                           {group.name}
-                          {process.env.GROUPAROO_UI_EDITION !== "config" && (
-                            <>&nbsp;{group.profilesCount} members</>
-                          )}
```
Comment (evantahler) [inline ui/ui-components/pages/destination/[id]/data.tsx:307]: cc @seancdavis 

I don't really know why this didn't work out

Comment (seancdavis) [reply ui/ui-components/pages/destination/[id]/data.tsx:307]: Yikes! I guess I have to be better about testing this in Enterprise when I make a change like. I figured a simple conditional was simple enough.

seancdavis:
@evantahler FWIW the previous version of this feature wrapped the count in parentheses (e.g. `All Emails (2617 members)`).

evantahler:
ah ha! I'll add that back

PR #1928, opened 2021-06-14
Human IDs, Part 2

seancdavis:
This is a continuation of #1895. That story was rejected because the `id` would be updated every time. So whenever a `name` (or `key`) would change, the file would change locations.

Logged Model now makes an inference as to whether its `id` follows the default structure or not. The writable models use this method to determine if they should use the `id` value from the database or slugify their human property.

Through this process I also found that the config writer was using the `name` property on the config object to generate the filename. Now we assume that the `id` was generated properly and use it directly to name the file.

### DIFF
--- a/core/src/classes/loggedModel.ts
+++ b/core/src/classes/loggedModel.ts
@@ -11,6 +11,7 @@ import {
   Length,
   BeforeBulkCreate,
 } from "sequelize-typescript";
+import validator from "validator";
 import * as uuid from "uuid";
 import { Log } from "../models/Log";
 import { config, chatRoom } from "actionhero";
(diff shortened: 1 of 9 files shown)


core/src/classes/loggedModel.ts
```diff
@@ -139,6 +140,11 @@ export abstract class LoggedModel<T> extends Model {
     return message;
   }
 
+  idIsDefault(): boolean {
+    const uuid = this.id.split("_").pop();
+    return this.id.startsWith(`${this.idPrefix()}_`) && validator.isUUID(uuid);
+  }
```
evantahler, line 146: Nice check!

evantahler (approved):
Makes sense! 

To confirm the logic: 
1) If a UUID was in use, the config writer can make up a better ID based on the name
2) Otherwise, keep the ID it has.

The risk of an ID provided by the user matching the `prefix`_`uuid` pattern is low, so this seems OK!

seancdavis:
@evantahler That logic is correct!

PR #1930, opened 2021-06-14
Relax error throwing for `profileProperties:enqueue` task

evantahler:
This PR relaxes error handling for the `profileProperties:enqueue` task.  We will log errors which occur, but not throw. 

This PR also increases the number of times `RetryableTask` will retry when `NODE_ENV=development` to 3 (from 1).


evantahler:
I'm unable to test this because we can't create a non-blocking transaction the test doesn't wait for

PR #1931, opened 2021-06-15
Remove Events

evantahler:
The PR Removes Events from Grouparoo.  As a side effect, we no longer need `@grouparoo/clients` (we only had a web client thus far).

Docs @ https://github.com/grouparoo/www.grouparoo.com/pull/529


evantahler:
Screen Shot 2021-06-14 at 5 27 18 PM

evantahler:
TODO: Remove `profile.annonymusId`

PR #1932, opened 2021-06-15
Write cluster name setting if it has changed

seancdavis:
This expands upon #1841 to write the cluster name setting, but only if its value doesn't match the default value.


evantahler (approved):
Looks good - I like that the plumbing is generic enough that we can add more settings to the config files if we want later

PR #1936, opened 2021-06-15
Set groups right to ready in config

seancdavis:
@evantahler This is a quick change, but wanted your input on it. What I'm trying to accomplish is to set a group's `state` to `ready` in config mode when it has at least one rule. Previously they were stuck in `initializing`.

Is this an appropriate change? Would you have gone about it differently?

### DIFF
--- a/core/__tests__/models/group/calculatedGroup.ts
+++ b/core/__tests__/models/group/calculatedGroup.ts
@@ -31,6 +31,7 @@ describe("models/group", () => {
 
     afterEach(async () => {
       await SharedGroupTests.afterEach();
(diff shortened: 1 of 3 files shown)


evantahler (approved):
Yeah, I think this is OK!

I'm the one that did it, but it's a little weird that `group.setRules()` changes state... that probably should be an `@beforeSave` so we can test it in isolation.

seancdavis:
Yeah, I suppose. On one hand it makes sense so you're not checking the state for every group rule added, but can do it in at one time when adding all the rules.

PR #1938, opened 2021-06-15
Remove UI plugins from the `grouparoo.plugins` section in package.json file

seancdavis:
Makes the following changes:

- UI plugins are loaded automatically if they are installed. They do not have to be in the `grouparoo.plugins` array.
- `grouparoo install` will install ui-config as a dev dependency.
- `grouparoo install` does not add ui-config, ui-community, or ui-enterprise to the `grouparoo.plugins` array.
- Fix an issue with the conflict error on `grouparoo install`. Now it will only fail if trying to install ui-community or ui-enterprise when one or the other has been installed. The error logic ignores ui-config.
- Removes duplicates from `grouparoo.plugins` (on the fly, it doesn't actually make changes to package.json).

This one is weird to test. I did all of the testing manually.

### DIFF
--- a/cli/src/lib/install.ts
+++ b/cli/src/lib/install.ts
@@ -30,15 +30,12 @@ export default async function Update(pkg: string) {
     process.exit(1);
   }
 
-  const installedPackages = Object.keys(
-    pkgJSONContents?.devDependencies || []
-  ).concat(Object.keys(pkgJSONContents?.dependencies || []));
-
+  const dependencyList = Object.keys(pkgJSONContents?.dependencies || []);
   const availableUiPackages = [
     "@grouparoo/ui-community",
     "@grouparoo/ui-enterprise",
   ];
-  const existingUiPackage = installedPackages.find((p) =>
+  const existingUiPackage = dependencyList.find((p) =>
     availableUiPackages.includes(p)
   );
 
--- a/cli/src/lib/install.ts
+++ b/cli/src/lib/install.ts
@@ -24,18 +24,32 @@ export default async function Update(pkg: string) {
   let plugins: string[] = pkgJSONContents?.grouparoo?.plugins;
 
   if (!plugins) {
-    logger.fail("There is no `grouparoo` section in this package.json.");
+    logger.fail(
+      "There is no `grouparoo.plugins` section in this package.json."
+    );
     process.exit(1);
   }
 
-  if (pkg && pkg?.match("@grouparoo/ui-")) {
-    const existingUIPackage = plugins.find((p) => p.match(/@grouparoo\/ui.*/));
-    if (existingUIPackage) {
-      logger.fail(
-        "There is already a ui package in this project. Uninstall the existing ui package before adding another."
-      );
-      process.exit(1);
-    }
+  const installedPackages = Object.keys(
+    pkgJSONContents?.devDependencies || []
+  ).concat(Object.keys(pkgJSONContents?.dependencies || []));
+
+  const availableUiPackages = [
+    "@grouparoo/ui-community",
+    "@grouparoo/ui-enterprise",
+  ];
+  const existingUiPackage = installedPackages.find((p) =>
+    availableUiPackages.includes(p)
+  );
+
+  const isUiPackage =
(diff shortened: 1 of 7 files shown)


cli/src/lib/install.ts
```diff
@@ -24,32 +24,46 @@ export default async function Update(pkg: string) {
...
-  if (pkg && pkg?.match("@grouparoo/ui-")) {
-    const existingUIPackage = plugins.find((p) => p.match(/@grouparoo\/ui.*/));
-    if (existingUIPackage) {
-      logger.fail(
-        "There is already a ui package in this project. Uninstall the existing ui package before adding another."
-      );
-      process.exit(1);
-    }
+  const installedPackages = Object.keys(
+    pkgJSONContents?.devDependencies || []
```
evantahler: What does Object.keys() of an array produce?

evantahler: I don't think this is right... if we are running on `NODE_ENV=production`, and we haven't installed devDependencies, we shouldn't include `pkgJSONContents.devDependencies` right?

evantahler: But... this is just for `grouparoo install`, so maybe it's OK

cli/src/utils/npm.ts
```diff
@@ -10,9 +10,12 @@ export namespace NPM {
...
     pkg?: string,
     exact = true,
-    npm_config_loglevel = "error"
+    npm_config_loglevel = "error",
+    devDependency = false
   ) {
-    const args = ["install"];
+    let args = ["install"];
+
+    if (devDependency) args.push("-D");
```
evantahler: please use the long args, `--save-dev`, it's easier to reason about in the future

core/src/utils/pluginDetails.js
```diff
@@ -63,61 +63,77 @@ function getPluginManifest() {
...
+
+  const availableUiPlugins = [
+    "@grouparoo/ui-enterprise",
+    "@grouparoo/ui-community",
+    "@grouparoo/ui-config",
+  ];
+
+  const installedPackages = Object.keys(
+    parentPkg?.devDependencies || []
+  ).concat(Object.keys(parentPkg?.dependencies || []));
```
evantahler, line 76: Here's where I'm not sure you can be sure `devDependencies` are installed or not.  You might need to test if they exist with a `require.resolve() + fs` check

evantahler (commented):
I think I'm confused about the logic in `core/src/utils/pluginDetails.js`.  I think I'm reading that you are considering any package in `dependencies` or `devDependencies` as available, but I'm worried that in our environment, maybe the `devDependencies` have not been installed.  Does the later try/catch work out for this case?

core/src/utils/pluginDetails.js
seancdavis: Very good point. This was dumb because it should only be looking for the UI packages here. However, all this is doing is collecting them into an array. Lower in the file it is already checking to make sure they are actually installed, but I think could use a tighter check. I'll give this one another shot.

cli/src/lib/install.ts
seancdavis: > What does Object.keys() of an array produce?

`dependencies` and `devDependencies` come through as objects where the key is the plugin name and the value is the version identifier. `Object.keys()` here gives us a list of names.

> I don't think this is right... if we are running on `NODE_ENV=production`, and we haven't installed devDependencies, we shouldn't include `pkgJSONContents.devDependencies` right?

I think what I was trying to protect against was still have you get an error message when trying to installed both production ui packages. But because those are the prod packages, I think you're right — I don't think there's any reason to look in `devDependencies`.

> But... this is just for `grouparoo install`, so maybe it's OK

Yeah, in the end, this is just to check against installing both ui production packages, so it probably doesn't matter, but it could definitely be tighter. I'll clean it up.

seancdavis:
@evantahler Check out the last few commits and see if you think that tightens things up a bit.

evantahler:
Yeah, this looks safer! 
Either way, this is the kind of thing that we can only really test once it's on NPM... so... lets merge it in and get an alpha out!

evantahler (approved):

PR #1943, opened 2021-06-16
Move Schedule Error to after run

teallarson:
I ended up integrating it into the run summary. My reasoning was that it gets lost in the system shutdown tasks, and the summary stands out more. Though I don't have super strong feelings about it @bleonard or @evantahler if we want it back in the "log stream" so to speak.  

Still need to test, but wanted to be sure this is a reasonable approach first:

Screen Shot 2021-06-17 at 3 03 32 PM

A run with no warnings does not display the warnings header:

Screen Shot 2021-06-17 at 3 08 55 PM

### DIFF
--- a/core/__tests__/modules/status.ts
+++ b/core/__tests__/modules/status.ts
@@ -5,7 +5,7 @@ import {
   FinalSummaryReporters,
 } from "../../src/modules/statusReporters";
 import { Status } from "../../src/modules/status";
-import { Destination, Profile, Source } from "../../src";
+import { Destination, Profile, Source, Schedule } from "../../src";
 
 describe("modules/status", () => {
   helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
@@ -362,6 +362,20 @@ describe("modules/status", () => {
           expect(sources[0].profilesImported).toEqual(1);
           expect(sources[0].error).toEqual(null);
         }
+      );
+    });
+
+    test("runs with no source show warning", async () => {
+      await helper.truncateOne(Schedule);
+
+      const warnings = await FinalSummaryReporters.Warnings.getWarnings();
+      const warning = warnings[0];
+      expect(warning.name).toBe("Schedules");
+      expect(warning.message).toBe(
+        `No schedules found.  The run command uses schedules to know what profiles to import.`
+      );
+      expect(warning.link).toBe(
+        `See this link for more info: https://www.grouparoo.com/docs/getting-started/product-concepts#schedule`
       );
     });
 
--- a/core/__tests__/modules/status.ts
+++ b/core/__tests__/modules/status.ts
@@ -366,7 +366,7 @@ describe("modules/status", () => {
     });
 
     test("runs with no source show warning", async () => {
-      await helper.truncateOne(Schedule);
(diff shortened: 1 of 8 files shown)


evantahler:
Sorry I hadn't responded until now! 

I like this approach a lot!  Maybe the object headers of `Message:` and `Link:` don't need to be displayed, but the strategy here is good!

teallarson:
> Sorry I hadn't responded until now!
> 
> I like this approach a lot! Maybe the object headers of `Message:` and `Link:` don't need to be displayed, but the strategy here is good!

Awesome.  Will clean it up and test it then!

core/src/modules/cli.ts
```diff
@@ -196,17 +210,21 @@ export namespace GrouparooCLI {
       GrouparooCLI.logger.log("");
       GrouparooCLI.logger.log(cyanBold(formattedTitle));
       const headings = ["PROFILES", "SOURCES", "DESTINATIONS"];
+      if (finalSummaryLogs[3].length > 0) headings.push("WARNINGS");
```
evantahler, line 213: This might be a little brittle, but we can worry about that when we next add to this summary

core/src/modules/status.ts
```diff
@@ -137,7 +139,12 @@ export namespace FinalSummary {
...
     finalSummaryLogs.push(await FinalSummaryReporters.Sources.getData());
     finalSummaryLogs.push(await FinalSummaryReporters.Destinations.getData());
+    finalSummaryLogs.push(await FinalSummaryReporters.Warnings.getWarnings());
 
     return finalSummaryLogs;
   }
+
+  export async function checkSchedules() {
+    return await Schedule.count();
+  }
```
evantahler: I don't think I saw where this was used

core/src/modules/statusReporters.ts
```diff
@@ -21,6 +21,7 @@ import { TeamMember } from "../models/TeamMember";
 import { Notification } from "../models/Notification";
 import { GroupOps } from "../modules/ops/group";
 import { SourceOps } from "../modules/ops/source";
+import { SchedulesList } from "../actions/schedules";
```
evantahler: Not used?

core/src/tasks/system/status/status.ts
```diff
@@ -1,4 +1,5 @@
 import { api, log, task } from "actionhero";
+import { Schedule } from "../../../models/Schedule";
```
evantahler: Not used?

core/__tests__/modules/status.ts
```diff
@@ -365,6 +365,20 @@ describe("modules/status", () => {
       );
     });
 
+    test("runs with no source show warning", async () => {
+      await helper.truncateOne(Schedule);
```
evantahler: Without adding to the helper, you could do `await Schedule.truncate()`

evantahler (commented):
In general, this looks good! 

There are a few un-used things noted

core/src/modules/cli.ts
teallarson: agree

evantahler (approved):

PR #1958, opened 2021-06-17
Attempt to open web browser when starting config ui

seancdavis:
@evantahler Simple change, but I was wondering if there is a flaw in the way I'm building the URL? I didn't see that ui-config is using the `WEB_URL` environment variable, so I just assumed localhost and used the port env var, which is automatically set in the config script.

### DIFF
--- a/ui/ui-config/dist/bin/config.js
+++ b/ui/ui-config/dist/bin/config.js
@@ -31,7 +31,7 @@ class Apply extends actionhero.CLI {
     await main(); // run the server
     try {
       log("Opening Grouparoo Config in web browser", "info");
-      open(`http://localhost:${process.env.PORT}`);
+      await open(`http://localhost:${process.env.PORT}`);
     } catch (err) {
       log(err.message, "error");
     }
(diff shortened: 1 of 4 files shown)


seancdavis:
Restarting the build. It got hung up while setting up Redis.

ui/ui-config/dist/bin/config.js
```diff
@@ -27,6 +29,12 @@ class Apply extends actionhero.CLI {
     cli.GrouparooCLI.logCLI(this.name, false);
     const { main } = require("@grouparoo/core/dist/grouparoo");
     await main(); // run the server
+    try {
+      log("Opening Grouparoo Config in web browser", "info");
+      open(`http://localhost:${process.env.PORT}`);
```
evantahler: maybe an `await` here?

evantahler (commented):
Implementation looks good! 

`PORT` and `WEB_URL` are different - the port says what I should listen on locally (maybe it's PORT=3000) and `WEB_URL` is the domain the user ends up at (https://foo.grouparoo.app) which is used for headers and auth. They aren't necessarily related, as folk's containers, load balancers, etc will modify things.   

In theory, on localhost, they will match (PORT=3000 and WEB_URL=http://localhost:3000), so I think either would have been correct to use.

PR #1959, opened 2021-06-17
ConfigWriter.generateId() throws an error when it can't build a valid id

seancdavis:
The `generateId()` method will throw an error when given some string for which it can't generate a valid ID value. However, when given an empty value, it will simply return `undefined`. This protects us from throwing an error when trying to build config objects for records that don't have a name-like value available yet.

IOW the following conditions exist:

- The typical workflow remains the same.
- Naming an item something like "$", where the `id` would be empty will throw and error: `Could not generate ID from name.`
- Naming one item "Post" and another "$Post" (where there would be a clash in `id` values) does not throw an error. It will only create the last one it grabs (and that _may_ not be consistent).


evantahler (approved):
It gets the job done!

PR #1960, opened 2021-06-17
Add a first setup step in config mode that is always true

seancdavis:
This was a quick iteration on a workflow bug we found. Suggested solution [here](https://www.pivotaltracker.com/story/show/178546753/comments/224915857).

Based on the way I approached this, it is a little weird to test because we have to _start_ the server in a particular mode to test that we have the right steps. Would you suggest I go about this differently, or can you think of an easy way to test that isn't creating a whole new spec?

### DIFF
--- a/core/__tests__/models/setupStep.ts
+++ b/core/__tests__/models/setupStep.ts
@@ -9,6 +9,10 @@ describe("models/setupStep", () => {
     resetSettings: true,
   });
 
+  afterEach(async () => {
+    process.env.GROUPAROO_RUN_MODE = undefined;
+  });
+
   test("setupSteps will be created when the server boots", async () => {
     const setupSteps = await SetupStep.findAll({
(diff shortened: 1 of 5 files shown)


evantahler (commented):
Does something bad happen if the number of SetupSteps changes between config mode and the 'regular' modes?   Perhaps `SetupStepOps.setupStepDescriptions` is the only array that exists, and that's built up within `core/src/modules/ops/setupSteps.ts` dynamically based on the GROUPAROO_RUN_MODE, so that's the only place that needs to consider the  GROUPAROO_RUN_MODE?

Then, the test could be changing GROUPAROO_RUN_MODE and seeing the changes in the contents of SetupStepOps.setupStepDescriptions

seancdavis:
I like that approach better. I'll give it a shot.

PR #1973, opened 2021-06-17
Do not retry `profile:completeImport` if not ready

evantahler:
* Do not retry `profile:completeImport` if not ready (solves https://www.pivotaltracker.com/story/show/178560767)
* `profile:completeImport` now handles many profiles at once (speed)
* `GroupMembers` now create logs in bulk when created or destroyed (speed)

### DIFF
--- a/core/src/tasks/profile/completeImport.ts
+++ b/core/src/tasks/profile/completeImport.ts
@@ -55,7 +55,7 @@ export class ProfileCompleteImport extends RetryableTask {
 
       if (Object.keys(mergedValues).length > 0) {
         await profile.addOrUpdateProperties(mergedValues);
-        await profile.reload({ include: [ProfileProperty] });
+        delete profile.profileProperties; // will be reloaded
       }
     }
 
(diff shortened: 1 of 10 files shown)


core/src/tasks/profile/completeImport.ts
```diff
@@ -16,57 +16,62 @@ export class ProfileCompleteImport extends RetryableTask {
...
+    const readyProfiles = profiles.filter(
+      (profile) =>
+        profile.state === "ready" &&
+        !profile.profileProperties.find((p) => p.state !== "ready")
     );
+    if (readyProfiles.length === 0) return;
 
-    if (profile.state !== "ready" || pendingProfileProperty) {
-      return CLS.enqueueTaskIn(config.tasks.timeout + 1, this.name, params);
-    }
```
evantahler, line 38: This re-enqueue was likely the problem.  If we keep re-enqueueing n+1 tasks (where n > than the number of workers) we'll never complete this queue, and won't move onto anything else.   The `profiles` queue is higher in priority than `profileProperties`... so we won't import anything.

Now, we'll just wait until the next `checkReady` batch to try again.  It's safer although a little slower... but the case where a profile is made un-ready by a new property or import in the middle of another import is rare.

PR #1975, opened 2021-06-18
Add Eloqua destination plugin integration

pauloouriques:


plugins/@grouparoo/eloqua/README.md
```diff
@@ -0,0 +1,3 @@
+## Attribution
+
+The Eloqua icon used in this plugin is from https://www.brandsoftheworld.com/logo/eloqua-0
```
evantahler: I think the logo they prefer is the "white oval on red" (that doesn't say Eloqua on it).   That's what eloqua.com's favicon is these days.  That said, I can't find a good high-res version. Here's the best one I could find https://azuremarketplace.microsoft.com/en-us/marketplace/apps/aad.eloqua?tab=overview


eloqua_215

plugins/@grouparoo/eloqua/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "dependencies": {
+    "@grouparoo/app-templates": "0.4.0-alpha.4",
+    "axios": "0.21.1"
+  },
+  "devDependencies": {
+    "@grouparoo/core": "0.4.0-alpha.4",
+    "@grouparoo/spec-helper": "0.4.0-alpha.4",
+    "@types/jest": "*",
+    "@types/node": "*",
+    "actionhero": "26.0.6",
```
evantahler: I think the actionhero version we have elsewhere is 26.0.8 - please rebase.  The typescript and jest versions may be bumped as well

plugins/@grouparoo/eloqua/__tests__/utils/shared.ts
```diff
@@ -0,0 +1,8 @@
+import { helper } from "@grouparoo/spec-helper";
+
+export async function indexUsers(isNewNock: boolean) {
+  // wait for search index
+  if (isNewNock) {
+    await helper.sleep(helper.shortTime);
+  }
+}
```
evantahler, line 8: Thanks, Eloqua :/

plugins/@grouparoo/eloqua/public/templates/app/{{{id}}}.js.template
```diff
@@ -0,0 +1,15 @@
+exports.default = async function buildConfig() {
+  return [
+    {
+      class: "app",
+      id: {{{id}}},
+      name: {{{id}}},
+      type: "{{{__pluginName}}}",
+      options: {
+        siteName: "...", // Eloqua Site Name
```
evantahler: Can you give an example of a "site", eg: is it "www.company.com" or "myCompany"?

plugins/@grouparoo/eloqua/src/lib/client/client.ts
```diff
@@ -0,0 +1,86 @@
+import axios, { AxiosRequestConfig } from "axios";
+import Bulk from "./bulk";
+import Contact from "./contact";
+import ContactList from "./contactList";
+
+export default class EloquaClient {
```
evantahler, line 6: This is a nice client API!

plugins/@grouparoo/eloqua/src/lib/client/client.ts
```diff
@@ -0,0 +1,86 @@
...
+  async _init() {
+    if (this.baseUrl) {
+      return this.baseUrl;
+    }
+    this.baseUrl = await this.getBaseUrl();
+  }
+
+  async getBaseUrl() {
+    // disable certificate checks - was getting inconsistent certificate results with this call
+    // const agent = new https.Agent({ rejectUnauthorized: false })
```
evantahler: Not needed anymore

plugins/@grouparoo/eloqua/src/lib/client/client.ts
```diff
@@ -0,0 +1,86 @@
...
+    username: string;
+    password: string;
+  };
+  loginUrl: string;
+  baseUrl: string;
+  bulk: Bulk;
+  contacts: Contact;
+  lists: ContactList;
+
+  constructor(options: any) {
```
evantahler: This might be helpful to type: `options: {username:string, password:string, site:string}`

plugins/@grouparoo/eloqua/src/lib/export/cachedMethods.ts
```diff
@@ -0,0 +1,47 @@
...
+      // No need to do so if we're just removing groups
+      if (!createIfNotExists) return null;
+
+      eloquaList = await client.lists.create({ name: listName });
+      return eloquaList.id;
+    }
+  );
+}
+
+export async function getContact(
```
evantahler, line 37: I don't see any invalidation methods for the cached methods.  If a user is deleted, has their email changed, etc, won't we need to invalidate the local cache?

plugins/@grouparoo/eloqua/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,153 @@
...
+    "address3",
+    "city",
+    "province",
+    "postalCode",
+    "country",
+    "businessPhone",
+    "mobilePhone",
+    "fax",
+    "salesPerson",
+  ];
```
evantahler, line 144: Is this list static for all of Eloqua, or could it change per-account?

plugins/@grouparoo/eloqua/__tests__/fixtures/destination-mapping-options.js
```diff
@@ -0,0 +1,122 @@
+const nock = require('nock');
+
+nock('https://login.eloqua.com:443', {"encodedQueryParams":true})
+  .get('/id')
+  .once().reply(200, {"site":{"id":734972267,"name":"e703652c38bf92ec484832c95344d3be765d58b5"},"user":{"id":222,"username":"t3st","displayName":"t3st","firstName":"Brian","lastName":"Leonard","emailAddress":"t3st@grouparoo.com"},"urls":{"base":"https://secure.p04.eloqua.com","apis":{"soap":{"standard":"https://secure.p04.eloqua.com/API/{version}/Service.svc","dataTransfer":"https://secure.p04.eloqua.com/API/{version}/DataTransferService.svc","email":"https://secure.p04.eloqua.com/API/{version}/EmailService.svc","externalAction":"https://secure.p04.eloqua.com/API/{version}/ExternalActionService.svc"},"rest":{"standard":"https://secure.p04.eloqua.com/API/REST/{version}/","bulk":"https://secure.p04.eloqua.com/API/Bulk/{version}/"}}}}, [
```
evantahler: Nice cleanup on the name and email in the fixtures!

evantahler (changes requested):
This is really great work @pauloouriques!  It looks like you've done most of the hard parts:
* combining all versions of the API into what we need; figuring out the reserved words and secret APIs
* sorting out byte-sizes for bulk requests
* created a custom Axios client for Eloqua
* caching to help reduce requests
* lots of tests

There are a few things to fix up, and I think that @pedroslopez will be a better judge of how the batch helpers are in use

plugins/@grouparoo/eloqua/src/lib/export/destinationMappingOptions.ts
pauloouriques: Tes, these fields are static and I'm calling them primary fields. There are others customs ones that we're getting via getContactFields call.

plugins/@grouparoo/eloqua/README.md
```diff
@@ -0,0 +1,3 @@
+## Attribution
```
evantahler, line 11: There's a little more to add to each README now.  Check out https://github.com/grouparoo/grouparoo/blob/main/plugins/@grouparoo/mysql/README.md for an example.  Basically, we want our NPM pages to look good.

plugins/@grouparoo/eloqua/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "lint": "prettier --check src",
+    "watch": "tsc --watch"
+  },
+  "dependencies": {
+    "@grouparoo/app-templates": "0.4.2-alpha.3",
+    "axios": "0.21.1"
+  },
+  "devDependencies": {
+    "@grouparoo/core": "0.4.2-alpha.3",
+    "@grouparoo/spec-helper": "0.4.2-alpha.3",
```
evantahler: Versions match 👍

evantahler (commented):
Looking good! 

You've addressed most of my comments, but I'm still worried about the cache methods (https://github.com/grouparoo/grouparoo/pull/1975#discussion_r656407917) I still don't see them getting invalidated?

I also added @pedroslopez to see if he has any comments on the async batches.

plugins/@grouparoo/eloqua/src/lib/client/field.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.client = client;
+    this.objectName = objectName;
+    this.bulkObjectName = bulkObjectName;
+  }
+
+  getAll(options?: any) {
+    return this.client._request({
+      method: "GET",
+      url: `/api/bulk/2.0/${this.bulkObjectName}/fields`,
+      // url: `/api/REST/1.0/assets/${this.objectName}/fields`,
```
pedroslopez: Could we remove these comments?

plugins/@grouparoo/eloqua/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,291 @@
...
+import { ErrorCheckExport } from "@grouparoo/app-templates/dist/destination/shared/batch";
+
+let client: EloquaClient;
+let exportedProfileFields = new Set<String>();
+
+const findAndSetDestinationIds = async ({ exports }) => {
+  const batchEmails = exports.map((p) => p.foreignKeyValue);
+  const allResults = await client.contacts.getContactsByEmail(batchEmails);
+  for (const profile of exports) {
+    const filteredContacts = allResults.filter(
```
pedroslopez, line 22: Note: you could also use `allResults.find()` here, but this is fine

plugins/@grouparoo/eloqua/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,291 @@
...
+const buildImportDefinition = async ({ appId }) => {
+  const allFields = await getAllContactFields(client);
+  const fields = {};
+  for (const field of allFields) {
+    if (exportedProfileFields.has(field.key)) {
+      fields[field.key] = field.statement;
+    }
+  }
+  const importName = `grouparoo-import-${appId}`;
+  return await client.bulk.createImport(importName, "emailAddress", fields);
```
pedroslopez, line 41: I think I'm reading that we're always creating an import definition each export using the same name. Are there any issues with using the same name multiple times?

plugins/@grouparoo/eloqua/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,291 @@
...
+      }
+      if (!profile.destinationId) {
+        throw new Errors.InfoError(
+          `destinationId not found to delete: ${profile.foreignKeyValue}`
+        );
+      }
+      await client.contacts.delete(profile.destinationId);
+      profile.processed = true;
+    } catch (error) {
+      profile.error = error;
```
pedroslopez, line 162: Errors are being attached to the specific profile in question 👍🏼

plugins/@grouparoo/eloqua/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,291 @@
...
+  client: currentClient,
+  appId,
+  appOptions,
+  syncOperations,
+  exports,
+}) {
+  if (currentClient) {
+    client = currentClient;
+  }
+  exports = buildBatchExports({ exports });
```
pedroslopez: Are there any issues with using the reserved word `exports` as a variable? I know in some other places of the codebase we use `_exports`.

plugins/@grouparoo/eloqua/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,291 @@
...
+    throw new Error(`emailAddress normalized to no value`);
+  }
+  newValue = newValue.toString();
+  exportedProfile.foreignKeyValue = newValue;
+
+  // record other one if applicable
+  if (oldValue) {
+    oldValue = oldValue.toString();
+    if (newValue !== oldValue && oldValue.length > 0) {
+      exportedProfile.oldForeignKeyValue = oldValue;
```
pedroslopez, line 190: We're saving the oldForeignKeyValue here, but I don't see it being used anywhere?

In the current batch helper we're first searching by the old keys and then the new keys to make sure we get the right contact:

https://github.com/grouparoo/grouparoo/blob/ce4ba281d5c964d776fbd1ee0f54d782ea7fed02/plugins/%40grouparoo/app-templates/src/destination/batch/exportProfiles.ts#L56-L60

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
```diff
@@ -0,0 +1,116 @@
...
+import { getListId, getContact } from "./cachedMethods";
+let client: EloquaClient;
+
+export const processExportedProfiles: ProcessExportedProfilesPluginMethod =
+  async ({ appId, appOptions, remoteKey, exports: _exports }) => {
+    client = await connect(appOptions);
+    const sync = await client.bulk.checkSync(remoteKey);
+
+    if (sync.status === "error") {
+      throw new Error("Something went wrong");
```
pedroslopez: Do we get an error message from Eloqua that we could throw here?

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
```diff
@@ -0,0 +1,116 @@
...
+    ) {
+      const response = await client.lists.update(
+        parseInt(listId),
+        listToUpdate
+      );
+    }
+  }
+}
+
+async function treatListToUpdate({
```
pedroslopez: This function name isn't totally clear for me... maybe something like `getListUpdatePayload`?

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
```diff
@@ -0,0 +1,116 @@
...
+        success: true,
+        processExports: {
+          profileIds: _exports.map((e) => e.profileId),
+          remoteKey,
+          processDelay: 1000,
+        },
+      };
+    }
+
+    if (sync.status === "success" || sync.status === "warning") {
```
pedroslopez: Do we know what could cause a `warning` status? Does `warning` mean that everybody's imported anyway and we should still add them to lists or could there be an issue with some profiles?

In https://docs.oracle.com/en/cloud/saas/marketing/eloqua-develop/Developers/BulkAPI/Reference/Syncs/sync-status-type.htm they refer to `/syncs/<ID>/logs` to get more information - maybe there's something to be done here to perhaps attach errors to specific profiles both in the `warning` and `errored` states

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
```diff
@@ -0,0 +1,116 @@
...
+
+import { getListId, getContact } from "./cachedMethods";
+let client: EloquaClient;
+
+export const processExportedProfiles: ProcessExportedProfilesPluginMethod =
+  async ({ appId, appOptions, remoteKey, exports: _exports }) => {
+    client = await connect(appOptions);
+    const sync = await client.bulk.checkSync(remoteKey);
+
+    if (sync.status === "error") {
```
pedroslopez: I _think_ this is supposed to be `errored`? https://docs.oracle.com/en/cloud/saas/marketing/eloqua-develop/Developers/BulkAPI/Reference/Syncs/sync-status-type.htm

plugins/@grouparoo/eloqua/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,96 @@
...
+    app: null,
+    connection: null,
+    destination: null,
+    destinationId: null,
+    destinationOptions: null,
+  });
+}
+
+describe("eloqua/destinationMappingOptions", () => {
+  test("can fetch user fields", async () => {
```
pedroslopez, line 46: I believe Eloqua supports custom contact fields- if so, we should also test that these can be fetched appropriately.

plugins/@grouparoo/eloqua/__tests__/export/export-profiles.ts
```diff
@@ -0,0 +1,990 @@
...
+      destinationOptions: null,
+      appId,
+      appOptions,
+      remoteKey: processExports.remoteKey,
+      exports,
+    });
+
+    expect(success).toBe(true);
+
+    user1 = await client.contacts.getByEmail(email1); // keep the old contact
```
pedroslopez: Are we keeping the old contact because eloqua doesn't support changing emails? If that's the case and we're in "sync" mode, we should delete the old one.

plugins/@grouparoo/eloqua/__tests__/export/export-profiles.ts
```diff
@@ -0,0 +1,990 @@
...
+    await indexUsers(newNock);
+    expect(success).toBe(true);
+
+    user1 = await client.contacts.getByEmail(email1);
+    expect(user1.emailAddress).toBe(email1);
+    expect(user1.firstName).toBe("John");
+    expect(user1.lastName).toBe("Test");
+
+    user2 = await client.contacts.getByEmail(email2);
+    expect(user2).toBeNull(); // deleted!
```
pedroslopez, line 799: We should check that he's also not in list1 just in case

plugins/@grouparoo/eloqua/__tests__/export/export-profiles.ts
```diff
@@ -0,0 +1,990 @@
...
+          newGroups: [],
+          toDelete: false,
+          profile: null,
+        },
+      ],
+    });
+    await indexUsers(newNock);
+    expect(success).toBe(true);
+
+    // nothing changed
```
pedroslopez, line 949: Does the sync return some status that indicates this? Maybe this is a case where using the logs endpoint could be helpful? The fact that there was an error here should be set on the profile somehow. If this is already happening, please add this to the test.

plugins/@grouparoo/eloqua/__tests__/export/export-profiles.ts
```diff
@@ -0,0 +1,990 @@
...
+          oldGroups: [],
+          newGroups: [],
+          toDelete: false,
+          profile: null,
+        },
+      ],
+    });
+    await indexUsers(newNock);
+
+    expect(success).toBe(true);
```
pedroslopez, line 1067: Same here: i would expect success to be false and an error to be set on the profile whose email was bad.

plugins/@grouparoo/eloqua/__tests__/export/export-profiles.ts
```diff
@@ -0,0 +1,990 @@
...
+    expect(errors.length).toBe(0);
+
+    expect(await client.contacts.getByEmail(email4)).toBeNull();
+  });
+
+  test("can handle batches with lots of prospects", async () => {
```
pedroslopez: `prospects` seems like a new noun in this context- maybe `profiles`?

pedroslopez (changes requested):
Looking good! A few notes:

- I'm a bit worried that we're essentially ignoring the `warning` sync status, as well as not knowing the cause of `errored`.  Maybe there's something to do here with the /sync/<ID>/status endpoint.
- I would like to see `processExports` object that's being returned by `exportProfiles` be checked a bit more - we're checking that it's not undefined, but maybe also check that the profiles to check are correct and that it has a remoteKey. Likewise, I don't think I saw processExportedProfile's return value being tested.
- There are a few interesting things to check out with regards to changing foreign keys. It seems like we're not using the oldForeignKey currently.

On another note, I _think_ we want to support custom objects in addition to contacts (@evantahler or @bleonard please correct me if I'm wrong). Is this possible here?

plugins/@grouparoo/eloqua/src/lib/export/exportProfiles.ts
pauloouriques: I don't think so

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
pauloouriques: the warnings messages have no mention to the profile that causes it:
`
{
          syncUri: '/syncs/807',
          count: 1,
          severity: 'warning',
          statusCode: 'ELQ-00002',
          message: 'Invalid email addresses.',
          createdAt: '2021-07-05T19:51:36.3970000Z'
        }
`

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
pauloouriques: I thought the same thing when I saw the first time
    {
      syncedInstanceUri: '/contacts/imports/1003',
      syncStartedAt: '2021-07-05T20:21:24.3870000Z',
      syncEndedAt: '2021-07-05T20:21:24.9200000Z',
      status: 'error',
      createdAt: '2021-07-05T20:21:22.8730000Z',
      createdBy: 'Brian.Leonard',
      uri: '/syncs/811'
    }

plugins/@grouparoo/eloqua/__tests__/export/export-profiles.ts
pauloouriques: The way we check the lists the users are in is by calling the client.contacts.getLists(user.id)

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
```diff
@@ -0,0 +1,225 @@
...
+    let newValue = contact.newProfileProperties.emailAddress;
+    let oldValue = contact.oldProfileProperties.emailAddress;
+    newValue = newValue.toString().toLowerCase().trim();
+    if (oldValue) {
+      oldValue = oldValue.toString().toLowerCase().trim();
+    }
+    if (oldValue) {
+      oldValue = oldValue.toString();
+      if (newValue !== oldValue && oldValue.length > 0) {
+        exportsToDelete.push(contact);
```
pedroslopez: this array doesn't seem to be used

plugins/@grouparoo/eloqua/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,302 @@
...
+        throw new Errors.InfoError(
+          "Destination sync mode does not delete contacts."
+        );
+      }
+      if (!profile.destinationId) {
+        throw new Errors.InfoError(
+          `destinationId not found to delete: ${profile.foreignKeyValue}`
+        );
+      }
+      await client.contacts.delete(profile.destinationId);
```
pedroslopez: Because we're always setting the `destinationId` based on the new foreign key, what would happen if there's an email change and toDelete at the same time? We should delete by the old email if it exists.

plugins/@grouparoo/eloqua/src/lib/export/exportProfiles.ts
pedroslopez: It still seems like, while `oldForeignKeyValue` is being set, it's not being used anywhere. If we're not using it we should probably just remove this, but it feels like something is missing here.

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
```diff
@@ -0,0 +1,225 @@
...
+        processExports: {
+          profileIds: _exports.map((e) => e.profileId),
+          remoteKey,
+          processDelay: 1000,
+        },
+      };
+    }
+  };
+
+async function postFKChange({
```
pedroslopez, line 93: Here's what I'm reading that this method does:
1. check if the email has been changed
2. delete the contact with the old email

A few things:
1. This confirms that changing email requires deleting the old contact, correct?
2. It's interesting that we're deleting things in this "post-export" step. Could this be done in `exportProfiles`? Since we're looking up people's IDs it may be possible, though if the new export fails we may be left in a state where we've deleted the old one and the new one has not been created.

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
```diff
@@ -0,0 +1,225 @@
...
+    appOptions,
+    remoteKey,
+    exports: _exports,
+    syncOperations,
+  }) => {
+    client = await connect(appOptions);
+    const sync = await client.bulk.checkSync(remoteKey);
+    if (sync) {
+      if (sync.status === "error") {
+        const logsResponse = await client.bulk.getSyncLogs(remoteKey);
```
pedroslopez, line 26: Could we use the `rejects` endpoint here as well? Or does the `error` status mean the same thing was wrong for all profiles?

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
```diff
@@ -0,0 +1,225 @@
...
+            profileIds: _exports.map((e) => e.profileId),
+            remoteKey,
+            processDelay: 1000,
+          },
+        };
+      } else if (sync.status === "success" || sync.status === "warning") {
+        let success = true;
+        const errors = [];
+        if (sync.status === "warning") {
+          const rejectsResponse = await client.bulk.getSyncRejects(remoteKey);
```
pedroslopez, line 53: Nice! We're now able to attribute warnings to specific profiles 👍🏼

pedroslopez (commented):
Nice! I like that we can now set warnings on the correct profile.

Just a few comments, mostly about changing foreign keys. I think to be sure this works we should verify that these cases work as expected and add a few tests as needed. Assuming changing email requires deleting the old one:

- Change email when both exist 
  - SYNC: update new, delete old
  - ADDITIVE: update new, orphan old
- Change email when only new exists
  - Just update data on the new email, nothing to do for old
- Change email when only old exists
  - SYNC: create new, delete old
  - ADDITIVE: create new, orphan old
- Change email and `toDelete: true` when both exist
  - delete the old email if syncmode allows
- Change email and `toDelete: true` when only old exists
   - delete the old email if syncmode allows
- Change email and `toDelete: true` when only new exists
  - delete the new email if syncmode allows

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
pauloouriques: when they send the "error" status it applies to all exports.

plugins/@grouparoo/eloqua/src/lib/export/processExportedProfiles.ts
pauloouriques: The reason why I did it on processExportedProfiles was exactly to avoid the case in which we've deleted the old one and the new one has not been created.

pedroslopez (approved):

evantahler (approved):

evantahler:
FYI I updated the title in the readme (8e8e51414f4aa555e43a304b68daca8b5571b774)

PR #1984, opened 2021-06-23
Remove `ah-next-plugin` + Next.js v11.0.1

evantahler:
This PR "vendors" and modifies `ah-next-plugin` so we can dynamically load next from within a UI plugin (if present).  This means that next no longer is a peerDependency of core (or `ah-next-plugin`), so only one exact version should be installed.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -37,7 +37,7 @@
     "@types/node": "*",
     "@types/validator": "*",
     "actionhero": "26.0.8",
-    "ah-next-plugin": "0.5.4",
+    "ah-next-plugin": "0.5.5",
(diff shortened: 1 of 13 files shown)


core/package.json
```diff
@@ -70,6 +69,7 @@
     "sqlite3": "5.0.2",
     "ts-node": "10.0.0",
     "uuid": "8.3.2",
+    "umzug": "v3.0.0-beta.5",
```
evantahler, line 72: Unrelated, but umzug should be a dependency, as we want to allow users to use the `bin/migrator` command now

PR #1992, opened 2021-06-24
Use absolute times in UI

evantahler:
Data engineers would probably rather see absolute times (`2021-06-22 11:27:28`) rather than relative times (`4 hours ago`)

Screen Shot 2021-06-24 at 4 16 01 PM

Timestamps now display as:
* `YYYY-MM-DD HH:mm:SS` (eg: "database timestamps")
* Times are using 24 hour clocks, so there is no AM or PM
* Times are displayed in the viewer's timezone (but there's a handy toggle to move to UTC display if we want in the future)


andyjih:
I've always preferred absolute time, and find relative time annoying. <3

andyjih:
YYYY-MM-DD 24-hour time UTC

evantahler:
TODO: Also format date (timestamp) profile properties

PR #1997, opened 2021-06-25
Destroy before ending run

teallarson:
Previously, `roo run` could end with rows in a deleted state.

This PR 
(a) adds a check for pending deletions
(b) adds pending deletions to the status reports
(c) processes the deletions (destroys the rows marked as deleted) before allowing the process to shutdown.

Thanks to @pedroslopez for helping me out with how our `destroy` tasks recur.

### DIFF
--- a/core/src/tasks/source/destroy.ts
+++ b/core/src/tasks/source/destroy.ts
@@ -33,6 +33,16 @@ export class SourceDestroy extends CLSTask {
       throw error;
     }
 
+    // check if the property is directly mapped
+    try {
+      await Source.ensureDirectlyMappedPropertyNotInUse(source);
+    } catch (error) {
+      if (error.message.match(/cannot delete property/)) {
+        return; // check back later
+      }
+      throw error;
+    }
+
     // no properties, let's delete it
     await source.destroy();
   }
--- a/core/__tests__/modules/status.ts
+++ b/core/__tests__/modules/status.ts
@@ -111,6 +111,20 @@ describe("modules/status", () => {
 
       const foundMetrics = await Status.get();
       expect(foundMetrics).toEqual({
+        Destination: {
+          deleted: [
+            {
+              metric: {
+                aggregation: "count",
+                collection: "deleted",
+                count: 0,
+                topic: "Destination",
+              },
+              timestamp: expect.any(Number),
+            },
+          ],
+        },
+
         Export: {
           pending: [
             {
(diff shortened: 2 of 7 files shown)


teallarson:
@evantahler I believe the extra check we added is already covered under the test [here](https://github.com/grouparoo/grouparoo/blob/c2e727d9bf2ef8f59c3eada89031722970d7f0c5/core/__tests__/tasks/source/destroy.ts#L72), it just does it without throwing any errors now.  Does that sound right?

evantahler:
> @evantahler I believe the extra check we added is already covered under the test [here](https://github.com/grouparoo/grouparoo/blob/c2e727d9bf2ef8f59c3eada89031722970d7f0c5/core/__tests__/tasks/source/destroy.ts#L72), it just does it without throwing any errors now.  Does that sound right?

I think we should add a similar check that does the reverse - if the property still exists and is in use, the Source won't be deleted.  That test may already exist, on the source model.

teallarson:
> > @evantahler I believe the extra check we added is already covered under the test [here](https://github.com/grouparoo/grouparoo/blob/c2e727d9bf2ef8f59c3eada89031722970d7f0c5/core/__tests__/tasks/source/destroy.ts#L72), it just does it without throwing any errors now.  Does that sound right?
> 
> I think we should add a similar check that does the reverse - if the property still exists and is in use, the Source won't be deleted. That test may already exist, on the source model.

@evantahler that is tested [here](https://github.com/grouparoo/grouparoo/blob/0e2424326f46eeb6fea415809c6cc09202b5adbd/core/__tests__/models/source.ts#L210).

evantahler (approved):
Looks great!

core/src/tasks/source/destroy.ts
```diff
@@ -33,6 +33,16 @@ export class SourceDestroy extends CLSTask {
...
 
+    // check if the property is directly mapped
+    try {
+      await Source.ensureDirectlyMappedPropertyNotInUse(source);
+    } catch (error) {
+      if (error.message.match(/cannot delete property/)) {
+        return; // check back later
+      }
+      throw error;
+    }
```
pedroslopez, line 44: We could probably add a test that the task waits for this before trying to delete, similar to the one that checks that the source isn't being used

pedroslopez (commented):
Nice work! Just one comment about a potential test to add.

PR #1998, opened 2021-06-25
List `bluebird` as a dependency of `@grouparoo/facebook`

evantahler:
I explored a number of ways to invalidate more of the node `require.cache`, but none of them were effective.  This might help... but we really need to publish it to test :/


evantahler:
v0.4.2-alpha.1 published to test this out...

PR #2000, opened 2021-06-25
copy and link update

teallarson:
Could also update the EnterpriseLink component to take an optional parameter to force the link type?  There were a few ways to do this, this way felt least redundant and least likely to unintentionally affect other components.

### DIFF
--- a/ui/ui-components/pages/profile/[id]/edit.tsx
+++ b/ui/ui-components/pages/profile/[id]/edit.tsx
@@ -391,31 +391,30 @@ export default function Page(props) {
           </ListGroup>
 
           <hr />
-
-          <Form onSubmit={(event) => handleAdd(event)} autoComplete="off">
-            <Row>
-              <Col md={9}>
-                <Form.Group controlId="groupId">
-                  <Form.Label>Add Group</Form.Label>
-                  <Form.Control as="select" disabled={loading}>
-                    {allGroups.map((group) => {
-                      const disabled =
-                        group.type !== "manual" ||
-                        groupMembershipIds.includes(group.id);
-                      return (
-                        <option
-                          disabled={disabled}
-                          value={group.id}
-                          key={`group-${group.id}`}
-                        >
-                          {group.name}
-                        </option>
-                      );
-                    })}
-                  </Form.Control>
-                </Form.Group>
-              </Col>
-              {process.env.GROUPAROO_UI_EDITION === "config" ? null : (
+          {process.env.GROUPAROO_UI_EDITION === "config" ? null : (
+            <Form onSubmit={(event) => handleAdd(event)} autoComplete="off">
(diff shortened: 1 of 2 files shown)


ui/ui-components/pages/profile/[id]/edit.tsx
```diff
@@ -376,12 +376,16 @@ export default function Page(props) {
                     &nbsp; &nbsp;
                   </>
                 ) : null}
-                <EnterpriseLink
-                  href="/group/[id]/members"
-                  as={`/group/${group.id}/members`}
-                >
-                  <a>{group.name}</a>
-                </EnterpriseLink>
+                {process.env.GROUPAROO_UI_EDITION === "enterprise" ? (
```
Comment (evantahler) [inline ui/ui-components/pages/profile/[id]/edit.tsx]: I think this should be `process.env.GROUPAROO_UI_EDITION !== "config"`

ui/ui-components/pages/profile/[id]/edit.tsx
```diff
@@ -411,17 +415,19 @@ export default function Page(props) {
...
-                <LoadingButton
-                  variant="outline-primary"
-                  size="sm"
-                  type="submit"
-                  disabled={loading}
-                >
-                  Add
-                </LoadingButton>
-              </Col>
+              {process.env.GROUPAROO_RUN_MODE === "config" ? null : (
```
Comment (evantahler) [inline ui/ui-components/pages/profile/[id]/edit.tsx]: I think this should be `process.env.GROUPAROO_UI_EDITION === "config"`

Comment (evantahler) [reply ui/ui-components/pages/profile/[id]/edit.tsx]: (GROUPAROO_RUN_MODE) is on the server-side

evantahler (changes requested):

ui/ui-components/pages/profile/[id]/edit.tsx
Comment (teallarson) [reply ui/ui-components/pages/profile/[id]/edit.tsx]: oops!  yep.

ui/ui-components/pages/profile/[id]/edit.tsx
```diff
@@ -411,17 +415,19 @@ export default function Page(props) {
...
-                <LoadingButton
-                  variant="outline-primary"
-                  size="sm"
-                  type="submit"
-                  disabled={loading}
-                >
-                  Add
-                </LoadingButton>
-              </Col>
+              {process.env.GROUPAROO_UI_EDITION === "config" ? null : (
```
Comment (evantahler) [inline ui/ui-components/pages/profile/[id]/edit.tsx]: I think above (`<Col md={9}>`) will need to be updated to be `md=12` since there will only be one column to fill the whole area

Comment (teallarson) [reply ui/ui-components/pages/profile/[id]/edit.tsx]: I think we actually want the whole row gone.  The col-9 section is a list of possible groups... which doesn't matter if you can't add them.  Got that updated!

evantahler (approved):

PR #2005, opened 2021-06-28
Update nav on group create

teallarson:
Previously: Because there was no redirect after a user adding a group, there was no update in the progress nav bar (on any ui package)

Now: Triggers a  page reload on submission of group setup (which triggers the nav setup bar to update).

There is definitely a more elegant solution that doesn't reload the entire window, but given that there shouldn't be a huge amount of data back and forth on this page it seems ok.


evantahler (approved):
It looks like we do the same thing `settings/tab` page when there's an update... so OK for now!  We'll do something more elegant in the future.

PR #2006, opened 2021-06-28
Add "Create Sample Profiles" to setup steps in config

teallarson:
**Do not merge until UI Config docs are up.** 

The "Learn More" link in the new setup steps is based on those docs.

Added setting up sample profiles as a step for UI-Config only.

Because there are now multiple differences between UI config setup steps and all other setup steps now, I separated them out rather than complicating the logic.

### DIFF
--- a/core/__tests__/models/setupStep.ts
+++ b/core/__tests__/models/setupStep.ts
@@ -37,12 +37,19 @@ describe("models/setupStep", () => {
     );
   });
 
-  test("setupSteps ops returns config versions when in config mode", async () => {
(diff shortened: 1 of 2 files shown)


evantahler (changes requested):
Implementation looks good! 

Please add a test that checks that the setup steps are changing based on `GROUPAROO_RUN_MODE`.  There are similar tests over here https://github.com/grouparoo/grouparoo/blob/main/core/__tests__/models/setupStep.ts#L40-L46

teallarson:
Converted it back to draft so we don't merge until after docs.

evantahler (approved):

PR #2007, opened 2021-06-28
[Config UI] Don't save mappings for query sources

pedroslopez:
Query sources don't need a mapping, but config ui was writing an empty mapping to file, which caused issues when validating the config:

```json
{
  "class": "Source",
  "id": "query",
  "name": "Query",
  "type": "postgres-query-import",
  "appId": "postgres",
  "mapping": {},
  "options": {}
}
```

```
warning: [ config ] error with Source `Query` (query): missing parent id to find a Property 
warning: [ config ] error with Property `ltv` (ltv): source is not ready and property not unique 
```

This PR does not save a `mapping` object to the config file if it's empty.

### DIFF
--- a/core/__tests__/modules/configWriter.ts
+++ b/core/__tests__/modules/configWriter.ts
@@ -557,6 +557,12 @@ describe("modules/configWriter", () => {
       expect(config).toBeUndefined();
     });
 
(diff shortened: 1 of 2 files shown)


evantahler (approved):
Clear problem with a clear fix.

PR #2008, opened 2021-06-29
fix: save and load profile column group rules in code config

pedroslopez:
This PR addresses three issues:
1. An error occurred when trying to add a group rule when no properties had been created. This was due to trying to use the first property as the initial property set on the rule being created. This changes the behavior to also include profile columns, so if there are no properties, the first profile column will be used instead.
2. Group rules that used profile columns (topLevel) could not be saved by config UI .
3. Group rules that used profile columns (topLevel) could not be loaded through code config.

### DIFF
--- a/core/src/classes/codeConfig.ts
+++ b/core/src/classes/codeConfig.ts
@@ -1,9 +1,11 @@
 import { log } from "actionhero";
 import { PropertyFiltersWithKey } from "../models/Property";
-import { GroupRuleWithKey, TopLevelGroupRules } from "../models/Group";
+import { GroupRuleWithKey } from "../models/Group";
 import { topologicalSort, Graph } from "../modules/topologicalSort";
 import { DestinationSyncMode } from "../models/Destination";
 import { MustacheUtils } from "../modules/mustacheUtils";
+import { TopLevelGroupRules } from "../modules/topLevelGroupRules";
+
 export interface IdsByClass {
   app?: string[];
   source?: string[];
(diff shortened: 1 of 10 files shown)


core/src/classes/codeConfig.ts
```diff
@@ -330,7 +332,14 @@ export async function getParentIds(
       const recordKeys = Object.keys(record);
       for (const j in recordKeys) {
         if (recordKeys[j].match(/.+Id$/)) {
-          prerequisiteIds.push(`property:${record[recordKeys[j]]}`);
+          const value = record[recordKeys[j]];
+
+          // special case: topLevel properties
+          if (TopLevelGroupRules.map((r) => r.key).includes(value)) {
+            continue;
+          }
```
evantahler, line 340: Ah ha!

core/src/models/Group.ts
```diff
@@ -648,7 +644,7 @@ export class Group extends LoggedModel<Group> {
     for (const rule of convenientRules) {
       const property = await Property.findOneWithCache(rule.key, "key");
       rules.push({
-        propertyId: property.getConfigId(),
+        propertyId: rule.topLevel ? rule.key : property.getConfigId(),
```
evantahler, line 647: 👍

evantahler (approved):
Nice work!

I'm surprised by `#2` (could not save top-level group rules in the UI).. I think that works OK now?

Screen Shot 2021-06-30 at 12 15 41 PM

pedroslopez:
> I'm surprised by `#2` (could not save top-level group rules in the UI).. I think that works OK now?

@evantahler Yeah, that was because it was trying to call `property.getConfigId()`, and `property` was null, so it crashed there.

PR #2009, opened 2021-06-29
redirect to app edit after plugin install

teallarson:
After installing a Plugin, an App instance is created users are redirected to `/app/[id]/edit` upon server restart.

### DIFF
--- a/ui/ui-config/pages/app/new.tsx
+++ b/ui/ui-config/pages/app/new.tsx
@@ -130,6 +130,14 @@ export default function Page(props) {
       await timeout(response.checkIn);
       await waitForServer();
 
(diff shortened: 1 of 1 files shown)


ui/ui-config/pages/app/new.tsx
```diff
@@ -130,6 +130,14 @@ export default function Page(props) {
       await timeout(response.checkIn);
       await waitForServer();
 
+      //we only want the plugin type, not the '@grouparoo/'
+      const pluginName = plugin.plugin.name.substring(11);
+      const newApp: Actions.AppCreate = await execApi("post", `/app`, {
```
evantahler, line 135: 👍 adding the type to the API response

evantahler (approved):

PR #2010, opened 2021-06-29
batch properties SUM and AVG return 0 for null

teallarson:
Aggregated properties of type `SUM` and `COUNT` will return `0` if no data, other aggregated properties will continue to return `null`.

Most of the changes are in the plugin test suites, many of which depend on `COUNT()` aggregations to check properties imports.

All aggregation types are tested specifically in the "secondary tables" section of the suites.


plugins/@grouparoo/app-templates/src/source/table/profileProperties.ts
```diff
@@ -55,6 +56,14 @@ export const getProfileProperties: GetProfilePropertiesMethod = ({
...
       }
+      if (
+        aggregationMethod === AggregationMethod.Sum ||
+        aggregationMethod === AggregationMethod.Count
+      ) {
+        //default all entries to 0 to start... will be replaced later if there is a response
+        responsesById[profiles[i].id] = [];
+        responsesById[profiles[i].id].push(0);
+      }
     }
```
evantahler: Helpful comment! 

No change needed, but this could be slightly written as `responsesById[profiles[i].id] = [0]`, skipping the `push`

plugins/@grouparoo/mongo/__tests__/table-import/import-properties.ts
```diff
@@ -276,6 +288,7 @@ describe("mongo/table/profileProperties", () => {
         });
         expect(values[profile.id]).toEqual([1.42]);
         expect(values[otherProfile.id]).toEqual([0.78]);
+        expect(values[thirdProfile.id]).toBeUndefined;
```
evantahler: Should be a function: `toBeUndefined()`.  There's a few of these

plugins/@grouparoo/mongo/__tests__/table-import/import-properties.ts
```diff
@@ -820,8 +869,8 @@ describe("mongo/table/profileProperties", () => {
           },
           [{ op, key: "properties.type", match: "Earthquake" }]
         );
-        expect(values[profile.id]).toBeUndefined();
-        expect(otherProfile[profile.id]).toBeUndefined();
+        expect(values[profile.id]).toEqual([0]);
+        expect(values[profile.id]).toEqual([0]);
```
evantahler: Nice! here's a behavior change.

evantahler (changes requested):
Looks great, except for the missing () in a few tests.  Thank you for the very thourough testing!

plugins/@grouparoo/mongo/__tests__/table-import/import-properties.ts
teallarson: 😳 Oh my!  Copy/paste fail.  Got those fixed.

evantahler (approved):

PR #2013, opened 2021-06-30
set mapped column when bootstrapping google sheets property

pedroslopez:
This caused an issue that prevented a property to be bootstrapped through Enterprise/Config UI


evantahler (approved):

evantahler:
Good catch!

evantahler:
Do other sources that don't use `@grouparoo/app-templates` have the same problem (eg: CSV?)

pedroslopez:
> Do other sources that don't use `@grouparoo/app-templates` have the same problem (eg: CSV?)

@evantahler I checked CSV and it implements this method so it's fine. However, the mailchimp source doesn't, and it does have a required `field` property. This causes the same issue we were seeing with google sheets if you try to bootstrap a property (mailchimp as first source). Adding it to tracker! pivotal-178731826

PR #2017, opened 2021-06-30
Add mapping from Grouparoo fields to Iterable nested objects.

pauloouriques:
- add dot notation to object coversion;
- add flag mergeNestedObjects.


plugins/@grouparoo/iterable/jest.config.js
```diff
@@ -1,8 +1,6 @@
-const { helper } = require("@grouparoo/spec-helper");
-
 module.exports = {
   testEnvironment: "node",
-  testTimeout: helper.longTime,
+  testTimeout: 60 * 2 * 1000,
```
evantahler: A shame!

plugins/@grouparoo/iterable/src/lib/export/exportProfile.ts
```diff
@@ -111,6 +116,7 @@ export const sendProfile: ExportProfilePluginMethod = async ({
       );
     }
 
+    payload["mergeNestedObjects"] = true;
```
evantahler, line 118: This implies that in all cases, we can send incomplete updates to an Iterable object and we will merge the existing data with what we have.  Cool!

What doest it mean when we want to remove a piece of data?  For example, in your test you could set `addres.city` and then remove it... would it be removed from Itterable?

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
```diff
@@ -182,7 +183,15 @@ describe("iterable/exportProfile", () => {
...
       oldProfileProperties: {},
-      newProfileProperties: { email, name },
+      newProfileProperties: {
+        email,
+        name,
+        defaultAddress: {
+          street: street1,
+          number: number1,
+          zipcode: zipcode,
+        },
```
evantahler: I don't think that it's likely Grouparoo would provide an object to the plugin like this

evantahler: The `.` way of defining properties seems more likely

evantahler (changes requested):
Good start! 

I'd like to see more testing on what it means to remove a property from an Itterable Object, and now it relates to `mergeNestedObjects`

plugins/@grouparoo/iterable/__tests__/export/export-profile.ts
pauloouriques: This is just to add the object there for the first time. But I agree we should use dot notation even the first time.

evantahler (approved):
I think this is a great start, and we should merge this in and get it out to our customers!

I'm still interested to know what it means to stop sending an object property.

If we had previously sent `address.city` and then stopped sending `address.city`, I think I read this as Iterable will keep `address.city` on the destination, and so unless we send `""`, there's no way to remove it.

evantahler:
> I think this is a great start, and we should merge this in and get it out to our customers!
> 
> I'm still interested to know what it means to stop sending an object property.
> 
> If we had previously sent `address.city` and then stopped sending `address.city`, I think I read this as Iterable will keep `address.city` on the destination, and so unless we send `""`, there's no way to remove it.

@pauloouriques pointed out that this is already tested here https://github.com/grouparoo/grouparoo/blob/7b18dc8ead9ae28c9c549ffe720fce74d38858ea/plugins/%40grouparoo/intercom/__tests__/integration/export-profile-lead.ts#L164-L193 We are good to go!

PR #2018, opened 2021-06-30
"Import and Export" button will Export Profile to Destinations that it should be removed from

evantahler:
This PR ensures that calling `profile.sync()` will create exports for when Profiles are removed from Destinations... not just when they are added.   There are 3 ways that a Profile might Leave a Destination:

1. The Profile's Properties have changed
2. The Group's defintion has changed
3. The Destination no longer tracks a Group that the Profile is in

1 and 2 happen in-line to the `profile.sync()` call, so we know the Group membership has changed.  3 requires that we look though old Exports for the profile and include those destinations in the check.

### DIFF
--- a/core/src/modules/ops/profile.ts
+++ b/core/src/modules/ops/profile.ts
@@ -4,6 +4,7 @@ import { Property } from "../../models/Property";
 import { Source } from "../../models/Source";
 import { Group } from "../../models/Group";
 import { Destination } from "../../models/Destination";
+import { Export } from "../../models/Export";
 import { GroupMember } from "../../models/GroupMember";
 import { Log } from "../../models/Log";
 import { api } from "actionhero";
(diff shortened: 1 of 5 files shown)


core/src/modules/ops/profile.ts
```diff
@@ -469,16 +470,46 @@ export namespace ProfileOps {
...
+    JOIN (
+      SELECT "destinationId", MAX("createdAt") as "createdAt"
+      FROM "exports"
+      WHERE "profileId" = '${profile.id}'
+      GROUP BY "destinationId"
+    ) AS "latest"
+    ON "latest"."destinationId" = "exports"."destinationId" AND "latest"."createdAt" = "exports"."createdAt"
+    WHERE "profileId" = '${profile.id}'
+    AND "toDelete" = false
+    ;
```
evantahler, line 498: This is inefficient, but it's only for one profile at a time, and only on user-request (not used in tasks)

PR #2034, opened 2021-07-01
fix: set scope on model options

pedroslopez:
Because we now support having different types of models with the same ID, this could create an issue where getting the options of one model included those of another that had the same ID. This adds a scope on the HasMany association to limit results for only the model in question.

We were also deleting options for all models with the same id.


evantahler (approved):
Thank you!

PR #2039, opened 2021-07-01
Upgrade to `jest` v27 and use patched `jest-environment-webdriver` & move to `@testing-library/react`

evantahler:
In addition to the quality-of-life improvements in the new version of Jest, v27 finally will fail a test if the `afterAll` or `afterEach` hook throws.  We ~~have~~ had a number of problems in this space in our test suite.

* We now use a new fork of webdriver https://github.com/grouparoo/jest-webdriver/tree/jest-v27-webdriver
* We now use `@testing-library/react` instead of Enzyme and the hacked v27 enzyme adapter.  It's more ergonomic anyway.


evantahler:
It seems that it's PNPM messing things up.  In a fresh project, or if you `npm install` within `@grouparoo/ui-components`, it's possible to get jest working with an environment of `jsdom`...

Possibly relevant:
* https://github.com/facebook/jest/issues/5369

evantahler:
Error reported with pnpm https://github.com/pnpm/pnpm/issues/3582

tools/merger/data/ci/ui/job.yml.template
```diff
@@ -37,7 +37,7 @@
 {{{custom_steps}}}
       - run:
           name: test
-          command: cd {{{relative_path}}}/{{{test_section}}} && ./node_modules/.bin/jest --ci --maxWorkers 2
+          command: cd {{{relative_path}}}/{{{test_section}}} && ../../node_modules/.pnpm/jest@27.0.6/node_modules/jest/bin/jest.js --ci --maxWorkers 2 --passWithNoTests
```
evantahler: This is the temporary hack to get around https://github.com/pnpm/pnpm/issues/3582

ui/ui-components/__tests__/smoketest.tsx
```diff
@@ -0,0 +1,18 @@
...
+  );
+}
+
+test("has correct welcome text", async () => {
+  render(<Welcome firstName="John" />);
+  expect(screen.getByRole("heading")).toHaveTextContent("Welcome, John");
+  expect(screen.getByTestId("p1")).toHaveTextContent(
+    "We are glad you are here, John."
+  );
+});
```
evantahler, line 18: Here's a smoke test for the new `@testing-library/react` library that replaced enzyme

ui/ui-components/package.json
```diff
@@ -31,18 +31,15 @@
...
-    "@types/jest": "*",
+    "@testing-library/jest-dom": "5.14.1",
+    "@testing-library/react": "12.0.0",
     "@types/react": "*",
     "@types/react-dom": "*",
-    "@wojtekmaj/enzyme-adapter-react-17": "0.6.2",
     "axios": "0.21.1",
     "bootstrap": "4.6.0",
     "date-fns": "2.22.1",
-    "enzyme": "3.11.0",
```
evantahler, line 42: 👋 Bye Enzyme!

ui/ui-components/__tests__/components/tabs.tsx
```diff
@@ -1,86 +1,54 @@
...
-    const link = links.at(0);
-    const domNode = link.getDOMNode();
-
-    expect(link.text()).toBe("Properties");
-    expect(domNode.constructor.name).toBe("HTMLAnchorElement");
-    expect(domNode.getAttribute("href")).toBe("/properties");
+  it("renders the plural word properly", async () => {
+    const link = screen.getByRole("link", { name: "Properties" });
+    expect(link).toHaveAttribute("href", "/properties");
   });
```
pedroslopez, line 23: much cleaner!

ui/ui-components/jest.setup.js
```diff
@@ -1,8 +0,0 @@
-import { configure } from "enzyme";
-import Adapter from "@wojtekmaj/enzyme-adapter-react-17";
-
-configure({ adapter: new Adapter() });
-
-if (!process.env.SERVER_TOKEN) {
-  process.env.SERVER_TOKEN = "test-server-token";
```
pedroslopez, line 7: confirming that this doesn't need to be set

tools/merger/data/ci/ui/job.yml.template
pedroslopez: Interesting- the symlink causes issues so we're using pnpm's installed jest directly

core/__tests__/models/option.ts
```diff
@@ -311,7 +311,7 @@ describe("models/option", () => {
           });
         });
 
-        afterAll(async () => {
+        beforeAll(async () => {
```
pedroslopez: I'm curious about this change- with two `beforeAll` hooks that mess with `connection`, what order do these run in? It seems like before this was cleaning up the destination after running the tests but I'm not sure what the intention is now. I'm reading that the destination is being created and destroyed in this `beforeAll` step, but that's clearly not what's happening since tests are passing 🤔

ui/ui-components/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "scripts": {
     "pretest": "npm run lint",
-    "test": "jest",
+    "test": "../../node_modules/.pnpm/jest@27.0.6/node_modules/jest/bin/jest.js",
```
pedroslopez: Do we need to do this for any of the other packages?

pedroslopez (commented):
Nice! react-testing-library seems easier to work with. Just two questions on some changes in the comments.

tools/merger/data/ci/ui/job.yml.template
evantahler: It's a mystery to me too!

ui/ui-components/__tests__/smoketest.tsx
evantahler: One thing to note, you need to `render()` in each test, or `beforeEach` - it gets cleared at the end of each test.

ui/ui-components/jest.setup.js
evantahler: 👍 Yep, as we aren't booting up the server to test this lib, as it's just react components.

core/__tests__/models/option.ts
evantahler: Yeah... that's a mistake.  I'll update!

evantahler:
Yes! The mysterious pnpm problem was caused by having 2 versions of jest in the same monorepo.  I had previously not upgraded jest in the `apps` Upgrading all packages to the same version of jest fixed things!

PR #2040, opened 2021-07-02
remove warning from source and property pages

teallarson:
- Removed warning from both initial mapping on `source/[id]/mapping` and on the create property pages

"Files changed" tab makes it look more complicated than it is... I just removed the if statement/"are you sure" response.

### DIFF
--- a/ui/ui-components/pages/property/[id]/edit.tsx
+++ b/ui/ui-components/pages/property/[id]/edit.tsx
@@ -59,11 +59,6 @@ export default function Page(props) {
 
   async function onSubmit(event) {
     event.preventDefault();
-    // if (
-    //   window.confirm(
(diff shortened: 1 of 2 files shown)


evantahler (approved):
👍 

Are there any other `window.confirm` we might want to remove?

teallarson:
@evantahler There are quite a few places `window.confirm()` is still in use throughout `ui-components`... in `settings`, on many of the add/edit pages, on the resque page.  Not sure on the wanting to remove them part!

evantahler:
> @evantahler There are quite a few places `window.confirm()` is still in use throughout `ui-components`... in `settings`, on many of the add/edit pages, on the resque page. Not sure on the wanting to remove them part!

Ah, if there are that many, then we should probably talk about them separately and not shoehorn them into this PR :D.  Ship it!

PR #2041, opened 2021-07-02
resilient import completion and exports creation

pedroslopez:
The `profile:completeImport` and `profile:export` tasks were sensitive to redis being cleared and would not ever run if it was cleared in between when they were enqueued and when they were supposed to run. This PR makes two main changes:
1. Merge `profile:checkReady` and `profile:completeImport` tasks. The first one only updated profile states to ready and then enqueued the other task that actually set some things on the import to mark it complete. Now all of this happens in a single task.
2. Add a new recurring task to create exports for profiles whose imports have been completed but have not exported.

I'm not sure on naming for `profiles:enqueueExports` or the new and improved `profile:checkReady`, so open to suggestions on those. There's also a migration to fix imports that are stuck in this pending state.


core/__tests__/tasks/profile/completeImport.ts
```diff
@@ -1,191 +0,0 @@
...
-        where: { profileId: profile.id, propertyId: emailProperty.id },
-      });
-      await profileProperty.update({ state: "pending" });
-
-      await specHelper.runTask("profile:completeImport", {
-        profileIds: [profile.id],
-        toExport: true,
-      });
-
-      const foundTasks = await specHelper.findEnqueuedTasks("profile:export");
```
evantahler, line 74: I'd still suggest keeping these tests - moving them into `core/__tests__/tasks/profile/checkReady.ts` to ensure that profiles that are not ready are not tried to export.

core/src/tasks/profile/completeImport.ts
```diff
@@ -1,18 +1,10 @@
...
+import { Task } from "actionhero";
 
-export class ProfileCompleteImport extends RetryableTask {
+export class ProfileCompleteImport extends Task {
   constructor() {
     super();
     this.name = "profile:completeImport";
-    this.description =
-      "Once the profile properties are imported, update runs, imports, and groups then export";
+    this.description = "deprecated";
```
evantahler, line 7: 👍 nice touch

core/src/migrations/000074-mergeCompleteImportTask.ts
```diff
@@ -0,0 +1,15 @@
+export default {
+  up: async function (migration) {
+    await migration.sequelize.transaction(async () => {
+      // Kick profiles back to "pending" if they're ready but their imports are not
+      await migration.sequelize.query(
+        `UPDATE profiles SET "state"='pending' WHERE "state"='ready' AND "id" IN 
+          (SELECT "profileId" FROM imports WHERE "exportedAt" IS NULL AND "profileUpdatedAt" IS NULL AND "groupsUpdatedAt" IS NULL)`
```
evantahler:
```sql
UPDATE "profiles" SET "state"='pending' WHERE "state"='ready' AND "id" IN 
          (SELECT DISTINCT("profileId") FROM "imports" WHERE "exportedAt" IS NULL AND "profileUpdatedAt" IS NULL AND "groupsUpdatedAt" IS NULL)
```

* add quotes around table names
* added distinct around inner select

evantahler: Confirming that after this PR, we shouldn't be able to get into this state again

core/src/tasks/profile/enqueueExports.ts
```diff
@@ -0,0 +1,49 @@
...
+import { plugin } from "../../modules/plugin";
+import { Import } from "../../models/Import";
+import { CLS } from "../../modules/cls";
+
+export class ProfilesEnqueueExports extends CLSTask {
+  constructor() {
+    super();
+    this.name = "profiles:enqueueExports";
+    this.description =
+      "If a profile is ready and has an import that has not exported, enqueue profile:export for them";
```
evantahler, line 14: Is it possible for a profile to be ready, but for its properties not to be ready?  It shouldn't be, but we should confirm this... otherwise we can add that check here as well.

core/src/tasks/profile/enqueueExports.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.frequency = 1000 * 10;
+    this.queue = "profiles";
```
evantahler, line 16: 👍 
The `profiles` queue is lower in priority than the `exports` and `exports:{destination}` queues in priority.  So, this task should run /after/ we've started to work on the exports, meaning the set of remaining profiles to export should change before this runs again.

core/src/tasks/profile/enqueueExports.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.inputs = {};
+  }
+
+  async runWithinTransaction() {
+    const limit = parseInt(
+      (await plugin.readSetting("core", "runs-profile-batch-size")).value
```
evantahler, line 22: 👍 I think this is the right setting to use.
The `exports-profile-batch-size` setting will be used later on to /send/ the exports, but here we are /creating/ the exports.

plugins/@grouparoo/spec-helper/src/lib/workflows/import.ts
```diff
@@ -35,6 +35,7 @@ export async function ImportWorkflow() {
     }
 
     await specHelper.runTask("profiles:checkReady", {});
+    await specHelper.runTask("profiles:enqueueExports", {});
```
evantahler, line 38: It is good to see that after changing this helper here all the tests remain passing!

evantahler (changes requested):
Really good! 
Some minor testing and SQL notes below

core/src/tasks/profile/enqueueExports.ts
pedroslopez: No, profiles are only marked ready by `profiles:checkReady` when all their properties are ready.

core/__tests__/tasks/profile/completeImport.ts
pedroslopez: I don't think that it's possible to get a `ready` profile with a property that's `pending`, but the `completeImport` task was previously doing this check. I added it back to the `enqueueExports` task and moved the test over.

evantahler (approved):
Awesome!

PR #2044, opened 2021-07-06
Hide setup bar if complete

teallarson:
Open to other suggestions here!

The original story in tracker called for removing the component entirely if completed, but after discussion it sounded like we want the option to minimize instead.

Tried going with always hiding, but then user doesn't get any confirmation that they've completed the steps (OR they get it every time they reopen the app or refresh the page).  Using an accordion component to minimize, it's still a heavy visual element and sticks out a lot.

Went with showing a small success message always that also links to the `/setup` page if they want to refer back to the steps, docs, etc. later.

The updated UI on all three UI packages looks like this if a user has completed the steps:
Screen Shot 2021-07-06 at 11 20 23 AM

At one point in fiddling with things, I switched the style to appear more like a card and not bleed into the rest of the page.  I like it better but have zero strong feelings about it if we want it changed back.

Screen Shot 2021-07-06 at 11 25 07 AM

### DIFF
--- a/core/src/initializers/settings.ts
+++ b/core/src/initializers/settings.ts
@@ -179,14 +179,6 @@ export class Plugins extends CLSInitializer {
 
     const interfaceSettings: SettingsListItem[] = [
       {
-        key: "display-startup-steps",
-        title: "Display Startup Steps",
-        defaultValue: "true",
-        description:
-          "Should Grouparoo display the Setup Steps to all Team Members?  You can always visit /setup to see your setup steps.",
-        type: "boolean",
-      },
-      {
         key: "status-calculation-frequency-seconds",
         title: "Status Calculation Frequency Seconds",
         defaultValue: "10",
--- a/ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
+++ b/ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
@@ -57,7 +57,19 @@ export default function SetupStepsNavProgressBar({
     (100 * completeStepsCount) / totalStepsCount
   );
 
-  if (!shouldDisplay || percentComplete === 100) return null;
+  if (!shouldDisplay) return null;
+
+  if (isOnBoardingComplete) {
+    return (
+      <Row className="pt-1 px-4">
+        <Col>
+          <Link href="/setup">
+            <a>{isOnBoardingComplete ? "Setup Complete 🎉" : "Get Started:"}</a>
+          </Link>
+        </Col>
+      </Row>
+    );
+  }
 
   return (
     <div
--- a/ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
+++ b/ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
@@ -26,7 +26,7 @@ export default function SetupStepsNavProgressBar({
         getSetupSteps
       );
     };
-  }, [!shouldDisplay]);
+  }, []);
 
   async function getSetupSteps(newUrl?: string) {
     if (router.pathname.match(/^\/session\//)) return;
@@ -64,7 +64,7 @@ export default function SetupStepsNavProgressBar({
       <Row className="pt-1 px-4">
         <Col>
           <Link href="/setup">
-            <a>{isOnBoardingComplete ? "Setup Complete 🎉" : "Get Started:"}</a>
+            <a>Get Started:</a>
           </Link>
         </Col>
       </Row>
--- a/ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
+++ b/ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
@@ -1,16 +1,23 @@
 import { useEffect, useState } from "react";
 import { Models } from "../../utils/apiData";
-import { ProgressBar, Row, Col } from "react-bootstrap";
+import { ProgressBar, Row, Col, Button } from "react-bootstrap";
 import Link from "next/link";
 import { useRouter } from "next/router";
+import {
+  FontAwesomeIcon,
+  FontAwesomeIconProps,
+} from "@fortawesome/react-fontawesome";
 import { Actions } from "../../utils/apiData";
 
 export default function SetupStepsNavProgressBar({
   execApi,
   setupStepHandler,
+  successHandler,
 }) {
   const [steps, setSteps] = useState<Models.SetupStepType[]>([]);
   const [shouldDisplay, setShouldDisplay] = useState(false);
+  const [initialOnBoardingState, setInitialOnBoardingState] = useState(null);
+  const [hideCard, setHideCard] = useState(null);
   const router = useRouter();
 
   useEffect(() => {
@@ -40,7 +47,6 @@ export default function SetupStepsNavProgressBar({
 
     if (setupSteps) {
       setShouldDisplay(toDisplay);
-
       setSteps(setupSteps);
     }
   }
@@ -49,6 +55,7 @@ export default function SetupStepsNavProgressBar({
   const isOnBoardingComplete = steps.every(
     (step) => step.complete || step.skipped
   );
+
   const totalStepsCount = steps.length;
   const completeStepsCount = steps.filter(
     (step) => step.complete || step.skipped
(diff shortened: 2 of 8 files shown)


ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
```diff
@@ -58,36 +59,52 @@ export default function SetupStepsNavProgressBar({
...
+    return (
+      <Row className="pt-1 px-4">
+        <Col>
+          <Link href="/setup">
+            <a>Get Started:</a>
+          </Link>
+        </Col>
+      </Row>
+    );
+  }
```
evantahler: Maybe the screenshots don't match this code - I think this says that if you are complete, you'll always see a link with `Get Started:` (with a ":") and no 🎉?

ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
```diff
@@ -58,36 +59,52 @@ export default function SetupStepsNavProgressBar({
...
-          paddingBottom: 10,
           color: "var(--secondary)",
         }}
       >
-        {activeStep?.title}
-      </div>
-      <ProgressBar now={percentComplete} />
+        <Col>
+          <Link href="/setup">
+            <a>{isOnBoardingComplete ? "Setup Complete 🎉" : "Get Started:"}</a>
```
evantahler, line 110: If you are early returning due to `isOnBoardingComplete` above, this check isn't necessary any more.

evantahler (commented):
I think personally I'd prefer the component in the navigation's top level to by hidden once the steps are complete.  It's using a log of visual space... 

> Tried going with always hiding, but then user doesn't get any confirmation that they've completed the steps 

* We could add a link to `/setupSteps` in the "system" section of the nav once they are complete if the user ever wants to go back to the page
* I agree that there should be some fanfare to celebrate the steps being complete... but it doesn't have to be in the nav.  
   * The component could always be in the nav, but render "nothing" if the steps are complete.  If `isOnBoardingComplete` is moved to a react `useState`, it would be possible to monitor it changing with a `useEffect` and trigger something to happen when it does.  
     * This could be where we use the `successHandler` to message that the user is done, or pop up a modal saying "congrats" or something similar.

ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
tealjulia: Ooh.  Yep.  Missed a file in a commit.  This code is definitely wrong.

teallarson: Ohh, yeah.  It looks like I checked an incorrect version in.

teallarson:
>   * The component could always be in the nav, but render "nothing" if the steps are complete.  If `isOnBoardingComplete` is moved to a react `useState`, it would be possible to monitor it changing with a `useEffect` and trigger something to happen when it does.
>     
>     * This could be where we use the `successHandler` to message that the user is done, or pop up a modal saying "congrats" or something similar.

Would I need to use a reducer to watch `isOnBoardingComplete` so it's not showing the `success` message on every page load?  Trying to work it out with `setState()` and `useEffect()` I'm hitting infinite renders loops so I'm wondering I need to be using a reducer and/or three states (`null`, `false`, `true`) instead.

evantahler:
> > * The component could always be in the nav, but render "nothing" if the steps are complete.  If `isOnBoardingComplete` is moved to a react `useState`, it would be possible to monitor it changing with a `useEffect` and trigger something to happen when it does.
> >   
> >   * This could be where we use the `successHandler` to message that the user is done, or pop up a modal saying "congrats" or something similar.
> 
> Would I need to use a reducer to watch `isOnBoardingComplete` so it's not showing the `success` message on every page load? Trying to work it out with `setState()` and `useEffect()` I'm hitting infinite renders loops so I'm wondering I need to be using a reducer and/or three states (`null`, `false`, `true`) instead.

Whoa! Quite possibly!  I was under the impression that as long the `useEffect` that was listening for the change in state didn't itself change any state, things would be ok... but @seancdavis might know more about this?

teallarson:
Two holes with this solution (really one that presents in two ways):  If a user refreshes the page after completing all the steps, the `SetupStepsNavProgressBar` component will hide.  This includes the full page reload when a user renames their cluster (so if that happens to be their final setup step somehow, they won't see the "Setup Complete!" message)

All other cases:
1. Component behaves as normal, tracking `initialOnBoardingState` from its first render and `onBoardingState` from its current render.
2. If/when `initialOnBoardingState` is `incomplete` and `currentOnBoardingState` is `complete`, they'll see the "Setup Complete! 🎉 " message we had before.  ALSO a grey x will appear to hide the component.
Screen Shot 2021-07-07 at 11 41 13 AM
3. The state will keep track of that manual choice to hide if selected.  For subsequent full page reloads/sessions, the `initialOnBoardingState` will be "complete" and the component will be hidden as well.


NOTE: We'd discussed adding the setup steps to the bottom menu so they're accessible if a user wants to refer back.  Does that still seem useful?  Note that Config UI does not have the bottom menu.

ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
```diff
@@ -11,6 +12,9 @@ export default function SetupStepsNavProgressBar({
 }) {
   const [steps, setSteps] = useState<Models.SetupStepType[]>([]);
   const [shouldDisplay, setShouldDisplay] = useState(false);
+  const [initialOnBoardingState, setInitialOnBoardingState] = useState(null);
+  //because shouldDisplay is set on every Setup Step call, track if a user manually hides setup steps separately
+  const [hideCard, setHideCard] = useState(null);
```
evantahler: Typescript thing: If you set an initial `useState` as `null`, Typescript can't guess what the type of the value should be.  You can use `<>` to set a type specifically, in this case:

```tsx
const [initialOnBoardingState, setInitialOnBoardingState] = useState<boolean>(null);
const [hideCard, setHideCard] = useState<boolean>(null);
```

So only `true,false,null` would be allowed

evantahler (commented):
I like this approach with the dismissible `x` in the nav!

* Hopefully we don't do a full page reload too often.  The step to name your cluster is early on, so I'm OK with the risk introduced about not seeing the success message if the cluster name setting was the final step completed.
* I guess we don't need to add an additional link to see the steps again since we are now comfortable with the message auto-hiding... and as you call out, there is no platform nav in config-ui.
  * Does this mean that we need the Setting to hide the setup steps at all any more?  I think that just removing it from `initializers/setupSteps.ts` should remove it from the UI and handle the migration automatically.

ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
teallarson: Nice!  And that will evaluate to `null`, not false?

evantahler: The `<boolean>` doesn't change the /value/ of any JS code - it's just telling Typescrip that the type of `hideCard` is boolean, so you'll get compile-time errors if you did something like `hideCard ='fish'`

teallarson:
We are skipping testing on this component right now, as `useEffect()` is having unexpected effects on the test environment.

core/src/initializers/settings.ts
```diff
@@ -178,14 +178,6 @@ export class Plugins extends CLSInitializer {
...
 
     const interfaceSettings: SettingsListItem[] = [
-      {
-        key: "display-startup-steps",
-        title: "Display Startup Steps",
-        defaultValue: "true",
-        description:
-          "Should Grouparoo display the Setup Steps to all Team Members?  You can always visit /setup to see your setup steps.",
-        type: "boolean",
-      },
```
evantahler, line 188: 👍 remove the setting to show/hide setup setps

ui/ui-components/components/navigation/setupStepsNavProgressBar.tsx
```diff
@@ -56,38 +59,72 @@ export default function SetupStepsNavProgressBar({
...
+    setInitialOnBoardingState(onBoardingState);
+  }
+
+  //if we haven't figured out initial state yet, or the initial state is complete, or someone chooses to... hide the card
+  if (
+    initialOnBoardingState === null ||
+    initialOnBoardingState === true ||
+    hideCard === true
+  )
+    return null;
```
evantahler, line 75: 👍 the UI decides what to show now!

evantahler (changes requested):
Looks good!

It's a shame that async testing is a mess in react... but that's OK for now.

Check the bottom of `@grouparoo/ui-components/pages/setup` - there we tell users about how to hide the setup steps.  That probably should be removed.

evantahler (approved):

PR #2048, opened 2021-07-07
Schedules gain Filters (Polymorphic Filters model)

evantahler:
This PR adds the ability for our Table Source's Schedules to provide Filters.  In this way, you can ignore users of a certain type, or purchases which have been refunded in your Schedules.

Screen Shot 2021-07-08 at 4 34 41 PM

Under the hood:
* Like Options and Mappings, Filters are now a generic polymorphic object.  The previous `ProfileFilter` model and table has been migrated to be `Filter`


core/src/classes/plugin.ts
```diff
@@ -380,9 +382,12 @@ export interface SourceFilterMethod {
...
     sourceMapping: SourceMapping;
-    property: Property;
-    propertyId: string;
-    propertyOptions: SimplePropertyOptions;
+    property?: Property;
+    propertyId?: string;
+    propertyOptions?: SimplePropertyOptions;
+    schedule?: Schedule;
+    scheduleId?: string;
+    scheduleOptions?: SimpleScheduleOptions;
```
evantahler, line 390: `SourceFilterMethod` now is generic and can work for both Properties or Schedules.  To date, this method is static, and basically just returns the list of columns in the table

core/src/models/Filter.ts
```diff
@@ -6,17 +6,18 @@ import {
...
+import { Schedule } from "./Schedule";
 import { APIData } from "../modules/apiData";
 
-@Table({ tableName: "propertyFilters", paranoid: false })
-export class PropertyFilter extends Model {
+@Table({ tableName: "filters", paranoid: false })
+export class Filter extends Model {
   idPrefix() {
-    return "prf";
+    return "flt";
```
evantahler, line 20: New default id prefix

core/src/models/Property.ts
```diff
@@ -197,8 +191,11 @@ export class Property extends LoggedModel<Property> {
   })
   __options: Option[]; // the underscores are needed as "options" is an internal method on sequelize instances
 
-  @HasMany(() => PropertyFilter)
-  propertyFilters: PropertyFilter[];
+  @HasMany(() => Filter, {
+    foreignKey: "ownerId",
+    scope: { ownerType: "property" },
+  })
+  filters: Filter[];
```
evantahler, line 198: polymorphic relationships need a scope (thanks @pedroslopez!)

core/src/modules/modelName.ts
```diff
@@ -0,0 +1,8 @@
+import { Model } from "sequelize-typescript";
+import { LoggedModel } from "../classes/loggedModel";
+
+export function modelName(instance: Model | typeof LoggedModel) {
+  let name = instance.constructor.name;
+  name = name[0].toLowerCase() + name.substr(1);
+  return name;
+}
```
evantahler, line 8: We end up using this a lot within the polymorphic helper (options, mappings, filters)... now with proper Types!

plugins/@grouparoo/mongo/src/lib/table-import/getChangedRows.ts
```diff
@@ -12,21 +12,28 @@ export const getChangedRows: GetChangedRowsMethod = async ({
...
-  if (Object.keys(query).length > 0) {
-    aggPipeline.push({
-      $match: query,
-    });
+
+  for (const condition of matchConditions) {
+    const filterQuery = makeFindQuery(condition);
+    if (Object.keys(filterQuery).length > 0) {
+      aggPipeline.push({ $match: filterQuery });
+    }
```
evantahler, line 34: @pauloouriques This is working just fine - here's the query generated:

```json
[
   {
      "$match":{
         "id":{
            "$gt":4
         }
      }
   },
   {
      "$match":{
         "id":{
            "$lt":7
         }
      }
   },
   {
      "$sort":{
         "stamp":1,
         "id":1
      }
   },
   {
      "$addFields":{
         "__hwm":"$stamp"
      }
   },
   {
      "$limit":100
   },
   {
      "$skip":0
   }
]
```

Is there any perfomance penalty for having 2 separate comparison blocks for the "id" property, rather than something like:

```json 
   {
      "$match":{
         "id":[{ "$gt":4 }, { "$lt":7 }]
      }
   },
```

ui/ui-components/utils/makeLocal.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Sometimes you *really* need to decouple a child object form a parent in a react useState method
+ * TODO: Why is this require to break the object chain to rules?
+ */
+export function makeLocal(obj: any) {
+  return JSON.parse(JSON.stringify(obj));
+}
```
evantahler, line 7: @seancdavis any idea why this is required?  If we are trying to copy an array of an object (eg `property.filters: []` into a local copy with `useState()` so we can manipulate it on the page, there doesn't seem to be a copy (`Array.from(property.filters)`) or spread (`[...property.filters]`) method that really copies the data into a new array.  This was the nuclear option that got things working 💣

core/src/migrations/000075-polymorphicFilters.ts
```diff
@@ -0,0 +1,45 @@
...
+      });
+
+      await migration.addIndex(
+        "filters",
+        ["ownerId", "ownerType", "position"],
+        {
+          fields: ["ownerId", "ownerType", "position"],
+          unique: true,
+        }
+      );
```
evantahler, line 26: You can use the same column multiple filters (userId > 0 and userId < 100), so the only real validation we can do in the database is that there's one entry for each position per parent.

core/src/migrations/000075-polymorphicFilters.ts
```diff
@@ -0,0 +1,45 @@
...
+  up: async function (migration, DataTypes) {
+    await migration.sequelize.transaction(async () => {
+      await migration.renameTable("propertyFilters", "filters");
+
+      await migration.renameColumn("filters", "propertyId", "ownerId");
+
+      await migration.addColumn("filters", "ownerType", {
+        type: DataTypes.STRING(191),
+        allowNull: false,
+        defaultValue: "property",
```
evantahler, line 11: All previously existing filters before this migration belong to a property

PR #2049, opened 2021-07-08
clear profile properties when they no longer exist in the source

pedroslopez:
Currently importing profile properties that have been deleted from their source just marks them as ready and the old value remains. After this change, the value will be cleared instead.

Sources that don't implement an `importProfileProperty` or `importProfileProperties` method (e.g. google sheets, mailchimp) should keep the old value when trying to import it, as they will be updated when their schedule runs (keep old behavior)

### DIFF
--- a/core/__tests__/models/profile/profile.ts
+++ b/core/__tests__/models/profile/profile.ts
@@ -1,4 +1,5 @@
 import { helper } from "@grouparoo/spec-helper";
+import { api } from "actionhero";
 import {
   plugin,
   Profile,
@@ -917,6 +918,38 @@ describe("models/profile", () => {
         email: ["peach@example.com"],
         color: ["pink"],
       });
+    });
+
+    test("importing properties for source that doesn't support direct property imports will keep the old value", async () => {
+      const profile = await Profile.create();
+      await profile.addOrUpdateProperties({
+        userId: [1003],
+        email: ["bowser@example.com"],
+        color: ["green"],
+      });
+      let properties = await profile.properties();
+      expect(Object.keys(properties).sort()).toEqual([
+        "color",
+        "email",
+        "userId",
+      ]);
+
+      const connection = api.plugins.plugins.filter(
+        (p) => p.name === "test-plugin"
+      )[0].connections[0];
+      const oldMethod = connection.methods.profileProperty;
+      delete connection.methods.profileProperty;
+
+      await profile.import();
+
+      properties = await profile.properties();
+      expect(simpleProfileValues(properties)).toEqual({
+        userId: [1003],
+        email: ["bowser@example.com"],
+        color: ["green"],
+      });
+
+      connection.methods.profileProperty = oldMethod;
     });
 
     test("after importing, all missing properties will have created a null profile property", async () => {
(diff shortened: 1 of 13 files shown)


core/__tests__/tasks/profileProperty/enqueue.ts
```diff
@@ -277,6 +278,80 @@ describe("tasks/profileProperties:enqueue", () => {
...
+          expect(importProfilePropertyTasks.length).toBe(0);
+          expect(importProfilePropertiesTasks.length).toBe(0);
+
+          pendingProperties = await ProfileProperty.findAll({
+            where: {
+              state: "pending",
+              profileId: [daisyImport.profileId, peachImport.profileId],
+            },
+          });
+          expect(pendingProperties.length).toBe(0);
```
evantahler: Maybe check that they also still have the values that were set originally?

core/__tests__/tasks/profileProperty/importProfileProperties.ts
```diff
@@ -111,6 +111,35 @@ describe("tasks/profileProperty:importProfileProperties", () => {
       await profile.destroy();
     });
 
+    test("will set value to null for profile properties that no longer exist", async () => {
+      const spy = jest
+        .spyOn(testPluginConnection.methods, "profileProperties")
+        .mockImplementation(() => undefined);
```
evantahler, line 117: This test works because `canImport` will still be true?  If that's the reason, perhaps that should be added to the test somehow.

core/__tests__/models/profile/profile.ts
```diff
@@ -914,6 +920,38 @@ describe("models/profile", () => {
...
+      expect(simpleProfileValues(properties)).toEqual({
+        userId: [1003],
+        email: ["bowser@example.com"],
+        color: ["green"],
+      });
+
+      connection.methods.profileProperty = oldMethod;
+    });
+
     test("after importing, all missing properties will have created a null profile property", async () => {
```
evantahler, line 955: Ah! here's the existing behavior test that ensures null properties are removed

core/src/modules/ops/source.ts
```diff
@@ -262,6 +262,21 @@ export namespace SourceOps {
...
     };
 
+    const { pluginConnection } = await source.getPlugin();
+    if (!pluginConnection) {
+      throw new Error(
+        `cannot find connection for source ${source.type} (${source.id})`
+      );
+    }
+
+    const canImport = pluginConnection.methods.profileProperty;
```
evantahler, line 272: Is there ever a case when a plugin connection has `methods.profileProperties` and not `methods.profileProperty`?

evantahler (approved):
This looks really good!  I've got some minor questions/comments below.

core/__tests__/tasks/profileProperty/importProfileProperties.ts
pedroslopez: No, `canImport` only applies when importing synchronously (e.g. import and export button). In this case, this works because the `importProfileProperties` task now sets `rawValue: null`. This task is never enqueued if the plugin doesn't have an `importProfileProperties` method (as tested by `__tests__/tasks/profileProperty/enqueue: if there is no import method, it will just mark properties as ready`)

core/src/modules/ops/source.ts
pedroslopez: Our plugins implement both, but yeah, it's possible. However, at least at the moment the synchronous profile importing (import and export profile button) exclusively uses the singular `profileProperty` method.

PR #2054, opened 2021-07-13
Import Multiple Profile Properties for Multiple Exact Properties at once

evantahler:
This PR collects the importing of all similar `exact`-aggregation pending Profile Properties for the same Source into a single query, if the plugin supports it.  This means we get queries for the source like:

```sql
SELECT "id",
       "email",
       "language",
       "deactivated",
       "first_name",
       "last_name",
       "id" AS __pk
FROM users
WHERE "id" IN ('1',
               '2')
GROUP BY id
```

And on Grouparoo's side, the writing of Profile Properties is also done in one query, and  looks like:

```sql
INSERT INTO
   "profileProperties" ("id", "profileId", "propertyId", "state", "rawValue", "position", "unique", "valueChangedAt", "stateChangedAt", "confirmedAt", "createdAt", "updatedAt") 
VALUES
   (
      'prp_0e3c1398-57d8-4110-a7b7-08b32edf872e', 'pro_20a9e5d3-2723-48b6-91e9-134ee8808019', 'user_id', 'ready', '1', 0, true, '2021-07-13 00:02:10.652 +00:00', '2021-07-13 00:02:25.009 +00:00', '2021-07-13 00:02:25.009 +00:00', '2021-07-13 00:02:26.404 +00:00', '2021-07-13 00:02:26.404 +00:00'
   )
, 
   (
      'prp_a6391e0c-9589-4ddc-b9b6-81681dd04a4e', 'pro_20a9e5d3-2723-48b6-91e9-134ee8808019', 'email', 'ready', 'ghyndman0@demo.com', 0, true, '2021-07-13 00:02:25.009 +00:00', '2021-07-13 00:02:25.009 +00:00', '2021-07-13 00:02:25.009 +00:00', '2021-07-13 00:02:26.404 +00:00', '2021-07-13 00:02:26.404 +00:00'
   )
ON CONFLICT ("id") DO 
   UPDATE
   SET
      "state" = EXCLUDED."state", "rawValue" = EXCLUDED."rawValue", "unique" = EXCLUDED."unique", "valueChangedAt" = EXCLUDED."valueChangedAt", "stateChangedAt" = EXCLUDED."stateChangedAt", "confirmedAt" = EXCLUDED."confirmedAt", "updatedAt" = EXCLUDED."updatedAt" RETURNING "id", "profileId", "propertyId", "state", "rawValue", "position", "unique", "valueChangedAt", "stateChangedAt", "confirmedAt", "startedAt", "createdAt", "updatedAt";
...

```

## TODO:

- [x] Proof of concept with Core + 1 plugin (Postgres)
- [x] Core tests passing again
- [x] Core tests for multiple column imports
- [x] Core tests to show that only exact properties are grouped
- [x] Each plugin
  - [x] Update to new types/syntax
  - [x] Add multiple column tests


core/src/classes/plugin.ts
```diff
@@ -84,6 +85,29 @@ export interface PluginConnection {
...
+  Equal = "equals",
+  NotEqual = "does not equal",
+  GreaterThan = "greater than",
+  GreaterThanOrEqual = "greater than or equal to",
+  LessThan = "less than",
+  LessThanOrEqual = "less than or equal to",
+  Contain = "contains",
+  NotContain = "does not contain",
+  In = "in",
+}
```
evantahler, line 109: Moved into Core

plugins/@grouparoo/app-templates/src/source/table/meta.ts
```diff
@@ -112,12 +113,13 @@ export const buildConnection: BuildConnectionMethod = ({
...
   });
 
-  return {
+  const plugin: PluginConnection = {
     name,
     direction: "import",
     description,
     app,
     options,
+    groupAggregations: [AggregationMethod.Exact],
```
pedroslopez, line 122: 👍🏼 sources can now define supported groupAggregations in their connection

plugins/@grouparoo/app-templates/src/source/table/profileProperties.ts
```diff
@@ -26,50 +29,55 @@ export const getProfileProperties: GetProfilePropertiesMethod = ({
...
-        if (!primaryKeysHash[k]) primaryKeysHash[k] = [];
-        primaryKeysHash[k].push(profiles[i].id);
-      }
-      if (
-        aggregationMethod === AggregationMethod.Sum ||
-        aggregationMethod === AggregationMethod.Count
-      ) {
-        //default all entries to 0 to start... will be replaced later if there is a response
-        responsesById[profiles[i].id] = [0];
+      for (const property of properties) {
```
pedroslopez, line 58: Is this loop necessary? It seems like most if not everything we're doing applies to every property the same way (using `tableMappingCol`). The only place I'm seeing `property` being used is in line 71 (`responsesById[profiles[i].id] = { [property.id]: [0] };`), but see next comment

plugins/@grouparoo/app-templates/src/source/table/profileProperties.ts
```diff
@@ -26,50 +29,55 @@ export const getProfileProperties: GetProfilePropertiesMethod = ({
...
+          const k = properties[tableMappingCol].values[0].toString();
+          if (!primaryKeysHash[k]) primaryKeysHash[k] = [];
+          primaryKeysHash[k].push(profiles[i].id);
+        }
+        if (
+          aggregationMethod === AggregationMethod.Sum ||
+          aggregationMethod === AggregationMethod.Count
+        ) {
+          //default all entries to 0 to start... will be replaced later if there is a response
+          responsesById[profiles[i].id] = { [property.id]: [0] };
```
pedroslopez: Looping through `properties`, this will set a new object for the profile every time that only contains whatever the latest property was. If the intention is to set all properties to `[0]` for that profile, shouldn't the object contain all properties instead of only the latest one?

plugins/@grouparoo/app-templates/src/source/table/profileProperties.ts
```diff
@@ -95,7 +103,17 @@ export const getProfileProperties: GetProfilePropertiesMethod = ({
 
     for (const pk in responsesByPrimaryKey) {
       primaryKeysHash[pk].forEach((profileId) => {
-        responsesById[profileId] = responsesByPrimaryKey[pk];
+        // responsesById[profileId] = responsesByPrimaryKey[pk];
```
pedroslopez: note: this comment can probably be removed

plugins/@grouparoo/app-templates/src/source/table/profileProperties.ts
```diff
@@ -95,7 +103,17 @@ export const getProfileProperties: GetProfilePropertiesMethod = ({
...
+        responsesById[profileId] = {};
```
pedroslopez: Is this clearing out of properties we've set the value of already intentional? For example, previously we had defaulted some values to `[0]` in

```ts
//default all entries to 0 to start... will be replaced later if there is a response
responsesById[profiles[i].id] = { [property.id]: [0] };
```

pedroslopez (commented):
Looks great! Just a few comments.

plugins/@grouparoo/app-templates/src/source/table/profileProperties.ts
evantahler: Very good catch!

evantahler: Update in 129ebe896cb4e252afee3f0c3de17df9ae58dfc3... but now I need to figure out why none of our existing tests had a problem with this.  It may be the case that since the Properties with these default values are not `"exact"`, so we never considered more than one property anyway...

PR #2056, opened 2021-07-13
Async React Component testing

evantahler:
An example for how to complex react components that fetch data with a `useEffect`

### DIFF
--- a/ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
+++ b/ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
@@ -1,7 +1,11 @@
 import { render, screen } from "@testing-library/react";
 import Component from "../../components/navigation/setupStepsNavProgressBar";
+import mockAxios from "jest-mock-axios";
+import { Actions } from "../../utils/apiData";
+import { useApi } from "../../hooks/useApi";
 
 const useRouter = jest.spyOn(require("next/router"), "useRouter");
+const mockSetupStepHandler = { subscribe: () => {}, unsubscribe: () => {} };
 
 describe("setupStepsNavProgressBar", () => {
   describe("hidden", () => {
@@ -18,7 +22,7 @@ describe("setupStepsNavProgressBar", () => {
         render(
           <Component
             execApi={() => {}}
-            setupStepHandler={{ subscribe: () => {}, unsubscribe: () => {} }}
+            setupStepHandler={mockSetupStepHandler}
           />
         );
         expect(screen.queryByTestId("setupStepsProgressBar")).toBeNull();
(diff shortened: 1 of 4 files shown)


ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+  describe("on a logged in page", () => {
+    const { execApi } = useApi({});
+
+    beforeEach(() => {
+      useRouter.mockImplementation(() => ({
+        pathname: "/dashboard",
+        asPath: "/dashboard",
+        prefetch: async () => {},
+      }));
+    });
```
evantahler, line 42: We need to mock the next router, as it doesn't work well in a jest test environment.  Mocking it in this way (`const useRouter = jest.spyOn(require("next/router"), "useRouter");`) allows us to control and mock what page we are on.

ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+            setupStepHandler={mockSetupStepHandler}
+          />
+        );
+        expect(screen.queryByTestId("setupStepsProgressBar")).toBeNull();
+      });
+    }
+  });
+
+  describe("on a logged in page", () => {
+    const { execApi } = useApi({});
```
evantahler, line 34: We can use our "real" `useApi` hook, as we'll be mocking Axios (network library) below

ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+      useRouter.mockImplementation(() => ({
+        pathname: "/dashboard",
+        asPath: "/dashboard",
+        prefetch: async () => {},
+      }));
+    });
+
+    afterEach(() => {
+      mockAxios.reset();
+    });
```
evantahler, line 46: This clears the Axios mock cache between tests - otherwise you might assert that something happened that didn't

ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <Component execApi={execApi} setupStepHandler={mockSetupStepHandler} />
+      );
+
+      expect(mockAxios).toHaveBeenCalledWith(
+        expect.objectContaining({
+          method: "get",
+          url: "/api/v1/setupSteps",
+        })
+      );
+      mockAxios.mockResponse({ data: stepsResponse });
```
evantahler, line 88: You can assert that axios got a certain request and you can configure the response you want it to give back.  In this case, we are sending a specific set of SetupSteps, `stepsResponse`.  

The component will still be "awaiting" the response at the point, and won't resolve the network request until `mockAxios.mockResponse` is called.

ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+    it("doesn't display without steps", async () => {
+      render(
+        <Component execApi={execApi} setupStepHandler={mockSetupStepHandler} />
+      );
+
+      expect(screen.queryByTestId("setupStepsProgressBar")).toBeNull();
+    });
+
+    it("displays with steps", async () => {
+      let stepsResponse: Actions.SetupStepsList = {
```
evantahler, line 57: Use the Action types to be sure that your mock request data is formatted properly.

ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+        expect.objectContaining({
+          method: "get",
+          url: "/api/v1/setupSteps",
+        })
+      );
+      mockAxios.mockResponse({ data: stepsResponse });
+
+      const element = await screen.findByTestId("setupStepsProgressBar");
+      expect(element).toHaveTextContent("Get Started:");
+      expect(element).toHaveTextContent("do a thing");
```
evantahler, line 92: `@testing-library` has both async and sync finding methods (get* are sync, and find* are async).  Be sure to use the async (find*) methods to automatically handle "act" wrapping and allowing for react state to change.

teallarson (approved):
This is cool!  

Is there a way to trigger the `useEffect()` by updating the state and then checking to see that the component re-rendered correctly without triggering a full remount (ie: the behavior when someone finishes a step and the bar progresses, or finishes all steps and the close button appears)?  That type of behavior seems relevant for components like the dashboard and charts too.

evantahler:
> Is there a way to trigger the `useEffect()` by updating the state and then checking to see that the component re-rendered correctly without triggering a full remount (ie: the behavior when someone finishes a step and the bar progresses, or finishes all steps and the close button appears)? That type of behavior seems relevant for components like the dashboard and charts too.

I don't think you can /directly/ tigger another call to `useEffect` unless it was watching a passed-in prop you could change from the test.  We could modify other external things which would trigger another API call, specifically we could change the page the router was on... but that seems a little tricky to mock

PR #2059, opened 2021-07-14
Heroku can skip caching the .next cache files for smaller slugs

evantahler:


evantahler:
This made our Heroku slugs ~100MB smaller! (280mb ~> 170mb)

PR #2061, opened 2021-07-14
Increase the number of tries to match the time for SQLite conflict issues

bleonard:


evantahler:
This PR roughly has the effect of sleeping less and trying more often... which I guess is fine, since this is only for a local database.

evantahler (approved):

PR #2062, opened 2021-07-14
fix: improve Eloqua destination id retrieval.

pauloouriques:


plugins/@grouparoo/eloqua/src/lib/export/exportProfiles.ts
```diff
@@ -19,11 +19,17 @@ const findAndSetDestinationIds = async ({ exports: _exports }) => {
   const batchEmails = _exports.map((p) => p.foreignKeyValue);
   const allResults = await client.contacts.getContactsByEmail(batchEmails);
   for (const profile of _exports) {
-    const filteredContacts = allResults.filter(
```
bleonard, line 22: What it if was more like this. It would be a little faster (2n instead of n^2) and more clear about what we are skipping.

```js
// map from email to destination id
const foundHash = {};
for (const contact of allResults) {
	const key = (contact.emailAddress || "").toString().toLowerCase().trim();
	if(key.length === 0) {
		continue;
	}
	foundHash[key] = profileToGetIdFrom.id;
}

// lookup destination ids
for (const profile of _exports) {
	const key = (profile.foreignKeyValue || "").toString().toLowerCase().trim();
	if(key.length === 0) {
		continue;
	}
	profile.destinationId = foundHash[key];
}

```

PR #2067, opened 2021-07-15
Sweep profiles that have no directly mapped properties

pedroslopez:


core/src/tasks/profile/sweep.ts
```diff
@@ -0,0 +1,53 @@
...
+    this.description =
+      "Remove profiles that no longer have any directly mapped properties";
+    this.frequency = 1000 * 60 * 30; // every half hour
+    this.queue = "profiles";
+    this.inputs = {};
+  }
+
+  async runWithinTransaction() {
+    const limit = parseInt(
+      (await plugin.readSetting("core", "runs-profile-batch-size")).value
```
evantahler: 👍 good choice for batch size

core/src/tasks/profile/sweep.ts
```diff
@@ -0,0 +1,53 @@
...
+      }
+    );
+
+    // use "destroy" to clean up related models
+    if (profiles.length > 0) {
+      await Profile.destroy({
+        where: {
+          id: profiles.map((p) => p.id),
+        },
+      });
```
evantahler: I *think* that this would trigger the `@AfterBulkDestroy` lifecycle hook, and not the per-instance `@AfterDestroy` hooks.  Please double check this in the test (eg: that the Profile's properties got destroyed).

core/src/tasks/profile/sweep.ts
```diff
@@ -0,0 +1,53 @@
...
+        AND 0 = (
+          SELECT count("exports"."id") FROM "exports" 
+            WHERE "exports"."profileId"="profiles"."id" 
+            AND "exports"."state" IN ('pending', 'processing')
+        ) AND "id" NOT IN (
+          SELECT DISTINCT("profileId") FROM "profileProperties" 
+          JOIN properties ON "properties"."id"="profileProperties"."propertyId" 
+          WHERE "properties"."directlyMapped"=true AND "rawValue" IS NOT NULL
+        ) LIMIT ${limit};
+      `,
```
evantahler: This is complex query, but it's tested well!
Also, it's confirmed that it's OK on SQLite here - https://app.circleci.com/pipelines/github/grouparoo/grouparoo/10107/workflows/48841cdc-003b-426e-9a64-d03f39f01fe0/jobs/215309

evantahler (changes requested):
This is 👍, except for the bulk deleting "After" hooks - please double check those.

evantahler:
We may also want to run this task again at the end of `grouparoo run` - If the Run you just did null-ed out some directly mapped properties, I think folks would expect the Profile to be removed by the time the Run completes.

core/src/tasks/profile/sweep.ts
pedroslopez: good catch! you're right, the hooks weren't running.

core/src/tasks/profile/sweep.ts
```diff
@@ -39,13 +42,12 @@ export class ProfileSweep extends CLSTask {
...
-        where: {
-          id: profiles.map((p) => p.id),
-        },
-      });
+      const profileIds = profiles.map((p) => p.id);
+      await Profile.destroy({ where: { id: profileIds } });
+      await Import.destroy({ where: { profileId: profileIds } });
+      await Export.destroy({ where: { profileId: profileIds } });
+      await ProfileProperty.destroy({ where: { profileId: profileIds } });
     }
```
evantahler: I think I'd be worried about forgetting to add another relationship here in the future.  Since this is in a task, for now, I think `for (const profile of profiles) await profile.destroy()` would be OK... 

In the future, perhaps we make an AfterBatchDestroy which then uses the other hooks, keeping all the destroy logic within the model.

core/src/tasks/system/status/status.ts
```diff
@@ -38,8 +38,10 @@ export class StatusTask extends CLSTask {
       const complete = await this.checkForComplete(samples);
 
       if (runMode === "cli:run" && complete) {
-        await this.logFinalSummary();
+        const sweepProfiles: Task = api.tasks.tasks["profiles:sweep"];
+        await sweepProfiles.run({}, {});
```
pedroslopez: I'm not super convinced this is the right place to put this... suggestions welcome!

evantahler: It's a little weird to have one task call another.  This might be a signal to move your logic into an Op - `const deletedCount = ProfileOps.sweep(limit = 1000)` or similar, so you can call it directly here.  
* You also may need to call this method more than once, if there are a lot of profiles to remove... so maybe the method returns how many profiles it found and you keep looping until it's done.

evantahler: An alternative strategy:
* You could modify `checkForComplete` below, and include a check for Profiles that have a null  directly mapped property, and not let the server stop if there are any to delete
* If there are any Profiles that should be deleted, `await task.enqueue('profiles:sweep')` and let it process out of band

pedroslopez: I agree!
I went with a bit of a combination of both options. 

Something to note is that since we're now waiting for these profiles to be deleted, the task now runs more frequently so that `roo run` completes in a reasonable time. 30s seems okay to me, specially because we're limiting the amount of profiles each time it runs.

evantahler (approved):
Looks good! 

We may one day want to optimize/speed up the time between noticing that there are Profiles to delete and when we delete them, but that's a future story!  The methods/queries introduced here are great.

PR #2068, opened 2021-07-16
clear `directlyMapped` properties when profile has been removed from the source

pedroslopez:
- Allow directlyMapped properties to be cleared
- Send old property values when exporting a deleted profile for one last time
- Move setting import data properties to `associateProfiles` instead of `completeImport`.Setting these properties after everything was imported was setting them to possibly outdated values, and was setting properties that had been cleared by the `importProfileProperty`/`importProfileProperties` tasks, like unique properties from schedules runs.

### DIFF
--- a/core/__tests__/tasks/profileProperty/enqueue.ts
+++ b/core/__tests__/tasks/profileProperty/enqueue.ts
@@ -84,9 +84,7 @@ describe("tasks/profileProperties:enqueue", () => {
           const pendingProfileProperties = await ProfileProperty.findAll({
             where: { state: "pending" },
           });
-          expect(pendingProfileProperties.length).toBe(
-            (propertiesCount - 1) * 2
-          );
+          expect(pendingProfileProperties.length).toBe(propertiesCount * 2);
           for (const p of pendingProfileProperties) {
             expect(p.startedAt).toBe(null);
             await p.update({ startedAt: null });
@@ -103,9 +101,7 @@ describe("tasks/profileProperties:enqueue", () => {
           );
 
           expect(importProfilePropertiesTasks.length).toBe(0);
-          expect(importProfilePropertyTasks.length).toBe(
-            (propertiesCount - 1) * 2
-          );
+          expect(importProfilePropertyTasks.length).toBe(propertiesCount * 2);
 
           for (const p of pendingProfileProperties) {
             await p.reload();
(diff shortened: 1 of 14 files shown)


core/__tests__/tasks/profileProperty/enqueue.ts
```diff
@@ -84,9 +84,7 @@ describe("tasks/profileProperties:enqueue", () => {
           const pendingProfileProperties = await ProfileProperty.findAll({
             where: { state: "pending" },
           });
-          expect(pendingProfileProperties.length).toBe(
-            (propertiesCount - 1) * 2
-          );
+          expect(pendingProfileProperties.length).toBe(propertiesCount * 2);
```
evantahler, line 87: It's a side-effect, but this PR also makes this much easier to reason about!  If there are 10 Properties and 2 Profiles, there should be 20 pending Profile Properties.

core/src/modules/ops/profile.ts
```diff
@@ -222,7 +222,8 @@ export namespace ProfileOps {
     profileProperties: {
       [key: string]: Array<string | number | boolean | Date> | any;
     }[],
-    toLock = true
+    toLock = true,
+    ignoreMissingProperties = false
```
evantahler, line 226: What's the case where we want `ignoreMissingProperties=true`?

plugins/@grouparoo/spec-helper/src/lib/factories/profile.ts
```diff
@@ -9,7 +10,21 @@ const data = async (props = {}) => {
...
-  return Profile.create(await data(props));
+  const profile = await Profile.create(await data(props));
+
+  const { Property } = await import(`@grouparoo/core/${loadPath}`);
+  const allProperties = await Property.findAllWithCache();
+  const directlyMappedProperty = allProperties.find((p) => p.directlyMapped);
+
+  if (directlyMappedProperty) {
+    properties[directlyMappedProperty.key] = faker.datatype.number(99999);
+  }
```
evantahler, line 23: Nice!

core/src/tasks/import/associateProfile.ts
```diff
@@ -37,7 +37,7 @@ export class ImportAssociateProfile extends Task {
             if (_import.profileId) return;
 
             const { profile } = await _import.associateProfile();
-
+            await profile.addOrUpdateProperties(_import.data, undefined, true);
```
evantahler, line 40: 👍 moved from the CompleteImport task.

evantahler (approved):
This is really good work!

A question about applying the `Import.data` properties earlier on in the process: I think that originally we had waited until the Profile was ready to apply the Import's data because there might be conflicting Imports, and we wanted to choose the "most recent" data.  I believe this PR removes that logic, and we'll apply `Import.data` as we see it... which may not be in "most recent" order depending on which tasks get run first.   That said,  especially with our data engineer persona in mind, quickly changing Import data seems unlikely, and I think that's a use-case we don't need to worry about too much any more.

core/src/modules/ops/profile.ts
pedroslopez: This is used in the associateProfiles task. If the import data has a property that no longer exists, we want to just ignore it instead of throwing an error.

PR #2069, opened 2021-07-20
Remove `profile:completeImport` and `schedule:updateSchedules` tasks

evantahler:
The `profile:completeImport` and `schedule:updateSchedules` tasks are no longer in use, but were kept around to prevent errors during upgrade.


pedroslopez (approved):
Do we also want to remote the `schedule:updateSchedules` task? It seems to also be deprecated.

evantahler:
Good find @pedroslopez!

PR #2070, opened 2021-07-21
Periodically confirm that profiles still exist in the source

pedroslopez:
- [x] Confirm after schedule runs (if option is selected) - for sources that want to delete profiles every time the schedule runs (full table)
- [x] Confirm at a set frequency - for sources with large amounts of profiles that work off a high watermark, so this can be done less frequently
  - [x] Setting to configure how often we confirm
- [x] Mark all properties pending (to clear everything else out and correctly remove from groups if needed) 
- [x] Create imports - we're marking everything pending so there could be unrelated changes in other properties. Make imports so we can track these changes.
- [x] Testing

(?) Do we need a run for these imports? What kind of run would it be? Who would be its creator?

### New option on schedules:
Is this name too long?
image

### DIFF
--- a/core/src/actions/schedules.ts
+++ b/core/src/actions/schedules.ts
@@ -76,6 +76,7 @@ export class ScheduleCreate extends AuthenticatedAction {
       name: { required: false },
       sourceId: { required: true },
       recurring: { required: true },
+      confirmProfiles: { required: false },
       state: { required: false },
       options: { required: false },
       recurringFrequency: { required: true, default: 0 },
@@ -89,6 +90,7 @@ export class ScheduleCreate extends AuthenticatedAction {
       sourceId: params.sourceId,
       recurring: params.recurring,
       recurringFrequency: params.recurringFrequency,
+      confirmProfiles: params.confirmProfiles,
     });
 
     if (params.options) await schedule.setOptions(params.options);
--- a/core/src/actions/schedules.ts
+++ b/core/src/actions/schedules.ts
@@ -116,6 +116,7 @@ export class ScheduleEdit extends AuthenticatedAction {
       name: { required: false },
       sourceId: { required: false },
       recurring: { required: false },
+      confirmProfiles: { required: false },
       state: { required: false },
       options: { required: false },
       recurringFrequency: { required: false },
@@ -136,7 +137,11 @@ export class ScheduleEdit extends AuthenticatedAction {
     if (params.options) await schedule.setOptions(params.options);
     if (params.filters) await schedule.setFilters(params.filters);
 
-    await schedule.update({ state: params.state, name: params.name });
+    await schedule.update({
+      state: params.state,
+      name: params.name,
+      confirmProfiles: params.confirmProfiles,
+    });
 
     await ConfigWriter.run();
 
--- a/core/src/initializers/settings.ts
+++ b/core/src/initializers/settings.ts
@@ -148,7 +148,7 @@ export class Plugins extends CLSInitializer {
         title: "Confirm Profile Existence Days",
         defaultValue: 7,
         description:
-          "How many days should we wait before creating an import for a profile that hasn't been seen in a while to confirm they still exist in the source?",
+          "How often should Grouparoo check that a Profile still exists in your Source, deleting the Profile if it has been removed? Setting to `0` will disable this check.",
         type: "number",
       },
       {
--- a/core/src/initializers/settings.ts
+++ b/core/src/initializers/settings.ts
@@ -144,6 +144,14 @@ export class Plugins extends CLSInitializer {
         type: "string",
       },
       {
+        key: "confirm-profiles-days",
+        title: "Confirm Profile Existence Days",
+        defaultValue: 7,
+        description:
+          "How many days should we wait before creating an import for a profile that hasn't been seen in a while to confirm they still exist in the source?",
+        type: "number",
+      },
+      {
         key: "sweeper-delete-old-runs-days",
         title: "Sweeper: Delete Old Runs Days",
         defaultValue: 31,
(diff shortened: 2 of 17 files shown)


core/src/actions/schedules.ts
```diff
@@ -116,6 +116,7 @@ export class ScheduleEdit extends AuthenticatedAction {
       name: { required: false },
       sourceId: { required: false },
       recurring: { required: false },
+      confirmProfiles: { required: false },
```
evantahler, line 121: Probably should be added as an option to `schedule:create` as well.  

Please also test the Actions with these new params.

core/src/initializers/settings.ts
```diff
@@ -143,6 +143,14 @@ export class Plugins extends CLSInitializer {
           "The default country code Grouparoo will use to format phone numbers and display data",
         type: "string",
       },
+      {
+        key: "confirm-profiles-days",
+        title: "Confirm Profile Existence Days",
+        defaultValue: 7,
+        description:
+          "How many days should we wait before creating an import for a profile that hasn't been seen in a while to confirm they still exist in the source?",
```
evantahler: I think this description is a bit too nerdy for the end-users.  How about: 

> How often should Grouparoo check that a Profile still exists in your Source, deleting the Profile if it has been removed?

core/src/tasks/profile/confirm.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Schedule } from "../../models/Schedule";
+import { Run } from "../../models/Run";
+import { ProfileOps } from "../../modules/ops/profile";
+
+export class ProfilesConfirm extends CLSTask {
+  constructor() {
+    super();
+    this.name = "profiles:confirm";
+    this.description = "Confirm that profiles still exist in the source";
+    this.frequency = 1000 * 30;
```
evantahler, line 14: Nice!  So this one task will check both properties owned by a source's schedule (that's a mouthful) that are old AND any properties that are globally old.

evantahler: I'm a little curious about the processing time involved here: 
* This task runs every 30 seconds.  
* If there are a lot of Properties to update, this task will re-enqueue itself
* In that time, 30s may have passed, and this task will be re-enqueued again
* If the server is spending all it's time on  `ProfileOps.confirmExistence`, it's likely that none of the Imports could get processed... and we might have an loop?

While the `import` queue is higher in priority than the `profiles` queue, the properties might not be updated before the next loop as the `profileProperties` priority is after `profiles`

---

Update: I think I figured it out! The import will make the ProfileProperty not `ready`, and therefore not found in the next loop of `confirmExistence`.  Was that the intent @pedroslopez ?

core/src/initializers/settings.ts
evantahler: We should also note that setting this to "0" disables the check.

evantahler (changes requested):
Good work! There's some minor cleanup noted in the comments below.

The question of needing a Run for these imports is really interesting:
* There may be UI and `apiData()` assumptions that all Imports belong to a Run
* Where may be Sweeper assumptions that Imports belong to a Run

Perhaps:
* In the case of a Schedule check, we can attach the Import to the most recent schedule's run (you are already loading it up in `core/src/tasks/profile/confirm.ts`
* The global case is tricker... The task could create a "complete" internal run to apply the Imports too if needed

core/src/tasks/profile/confirm.ts
pedroslopez: You got it! Affected profiles (and their properties) become `pending` and `confirmExistence` explicitly looks for profiles and properties that are `ready`. Once these properties go back to being `ready`, they'll have a new `confirmedAt` timestamp as well so `confirmExistence` again won't pick them up until it's time to do so.

ui/ui-components/pages/import/[id]/edit.tsx
```diff
@@ -42,14 +35,13 @@ export default function Page(props) {
...
-                {_import.creatorId}
-              </a>
-            </Link>
+            {_import.creatorType === "run" ? (
+              <Link href="/run/[id]/edit" as={`/run/${_import.creatorId}/edit`}>
+                <a>{_import.creatorId}</a>
+              </Link>
+            ) : (
+              _import.creatorId
+            )}
```
Comment (pedroslopez) [inline ui/ui-components/pages/import/[id]/edit.tsx:44]: From what I could see, imports never have `creatorType: "group"` and the only `creatorTypes` we're using is `run`, `api` (in the `import:create` action, which I believe itself is unused) and now `task`. I changed this to what was already being used in `ui-components/components/import/list.tsx`.

core/src/modules/ops/profile.ts
```diff
@@ -707,6 +708,69 @@ export namespace ProfileOps {
...
+        },
+      ],
+    });
+
+    const now = new Date();
+    const bulkImports = [];
+
+    const creatorInfo = run
+      ? { creatorType: "run", creatorId: run.id }
+      : { creatorType: "task", creatorId: "profiles:confirm" };
```
evantahler, line 752: 👍

evantahler (changes requested):
There's one more UI component to change - check out `components/import/list` - we link to the run from there too.

pedroslopez:
> There's one more UI component to change - check out `components/import/list` - we link to the run from there too.

@evantahler are you referring to this?

https://github.com/grouparoo/grouparoo/blob/9a5ab2106b7022eab0f6c7cf71cce11cbac19790/ui/ui-components/components/import/list.tsx#L112-L122

If so, it already checks the creatorType, and is the same way I'm doing it for `ui/ui-components/pages/import/[id]/edit.tsx`

evantahler:
> > There's one more UI component to change - check out `components/import/list` - we link to the run from there too.
> 
> @evantahler are you referring to this?
> 
> https://github.com/grouparoo/grouparoo/blob/9a5ab2106b7022eab0f6c7cf71cce11cbac19790/ui/ui-components/components/import/list.tsx#L112-L122
> 
> If so, it already checks the creatorType, and is the same way I'm doing it for `ui/ui-components/pages/import/[id]/edit.tsx`

Oh! yeah, that's a good check :D

evantahler (approved):

PR #2072, opened 2021-07-22
add option to keep old property value if it's no longer found

pedroslopez:
Open to suggestions on a better name for this option!

Once both this and #2070 is merged, we can probably optimize the `profiles:confirm` task to ignore properties that have this set to `true`.


evantahler (approved):
This looks great, and naming things is hard.

Perhaps... `PersistPreviousValueWhenNotFound` ? That's terribly long.  Maybe `PersistPreviousValueWhenMissing`...

Maybe over the weekend inspiration will strike.

PR #2075, opened 2021-07-23
check type on newUrl before matching

teallarson:
Previously, trying to skip a step would cause an error: `newUrl.match is not a funciton`.

`newUrl.match is not a function` error was due to newUrl sometimes not being a string depending where it was called from.  We only want to check matches on strings.


evantahler (commented):
Good find!

I'm going to suggest to continue on https://github.com/grouparoo/grouparoo/pull/2076 instead of this PR.  If we can get the Types for all of the EventDispatchers set, we'll prevent more problems like this in the future.

evantahler:
@tealjulia now that https://github.com/grouparoo/grouparoo/pull/2076 is merged in do you want to include this too, or is it covered?

teallarson:
Closing.  Was fixed with #2076

PR #2076, opened 2021-07-24
Type UI Event Dispatchers

evantahler:
Possibly replaces #2075

This PR adds Typescript Types to the UI Event Dispatchers.  This way, it's possible to know what types of objects the subscriptions respond with.  Now we'll see better errors when we use the subscriptions:

Screen Shot 2021-07-24 at 9 01 29 AM

This works because when we extend the generic EventDispatcher class, we add a `<type>` indicating the types of values it's allowed to use in publishing and subscribing.  

TODO: 
- [x] Ensure all EventDisplatchers are typed in all components - If we don't supply them with the Type they are, we con't know the types of the subscription responses.

### DIFF
--- a/ui/ui-components/components/alerts/error.tsx
+++ b/ui/ui-components/components/alerts/error.tsx
@@ -1,7 +1,12 @@
 import { useState, useEffect } from "react";
 import { Alert } from "react-bootstrap";
+import { ErrorHandler } from "../../utils/errorHandler";
 
-export default function ErrorAlert({ errorHandler }) {
+export default function ErrorAlert({
+  errorHandler,
+}: {
+  errorHandler: ErrorHandler;
+}) {
   const [errorMessage, setErrorMessage] = useState(null);
   const [secondsToShow] = useState(1000 * 4);
   const [show, setShow] = useState(false);
@@ -12,7 +17,7 @@ export default function ErrorAlert({ errorHandler }) {
 
     return () => {
       clearTimeout(timer);
-      errorHandler.unsubscribe("error-alert", subscription.bind(this));
+      errorHandler.unsubscribe("error-alert");
     };
   }, []);
 
--- a/ui/ui-components/components/layouts/main.tsx
+++ b/ui/ui-components/components/layouts/main.tsx
@@ -7,9 +7,21 @@ import ErrorAlert from "../alerts/error";
 import Navigation from "../navigation";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import HydrationError from "../alerts/hydrationError";
+import { SuccessHandler } from "../../utils/successHandler";
+import { ErrorHandler } from "../../utils/errorHandler";
 
 export default function Main(props) {
-  const { children, successHandler, errorHandler, hydrationError } = props;
+  const {
+    children,
+    successHandler,
+    errorHandler,
+    hydrationError,
+  }: {
+    children: any;
+    successHandler: SuccessHandler;
+    errorHandler: ErrorHandler;
+    hydrationError: string;
+  } = props;
   const [navExpanded, setNavExpanded] = useState(true);
   const [alertWidth, setAlertWidth] = useState(500);
   const contentAreaLeftPadding = 250;
--- a/ui/ui-components/pages/property/[id]/edit.tsx
+++ b/ui/ui-components/pages/property/[id]/edit.tsx
@@ -16,6 +16,9 @@ import PropertyTabs from "../../../components/tabs/property";
 import { Models, Actions } from "../../../utils/apiData";
 import { filtersAreEqual } from "../../../utils/filtersAreEqual";
 import { makeLocal } from "../../../utils/makeLocal";
+import { ErrorHandler } from "../../../utils/errorHandler";
+import { SuccessHandler } from "../../../utils/successHandler";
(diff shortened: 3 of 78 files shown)


ui/ui-components/utils/profileHandler.ts
```diff
@@ -1,3 +1,4 @@
+import { Models } from "./apiData";
 import { EventDispatcher } from "./eventDispatcher";
 
-export class ProfileHandler extends EventDispatcher {}
+export class ProfileHandler extends EventDispatcher<Models.ProfileType> {}
```
evantahler, line 4: This is pretty neat - we can just reference the type returned from the API/Model's serializer.

ui/ui-components/components/alerts/error.tsx
```diff
@@ -12,7 +17,7 @@ export default function ErrorAlert({ errorHandler }) {
 
     return () => {
       clearTimeout(timer);
-      errorHandler.unsubscribe("error-alert", subscription.bind(this));
+      errorHandler.unsubscribe("error-alert");
```
teallarson, line 20: adding the typings revealed a few places we unsubscribed incorrectly

ui/ui-components/components/layouts/main.tsx
```diff
@@ -7,9 +7,21 @@ import ErrorAlert from "../alerts/error";
...
+  const {
+    children,
+    successHandler,
+    errorHandler,
+    hydrationError,
+  }: {
+    children: any;
+    successHandler: SuccessHandler;
+    errorHandler: ErrorHandler;
+    hydrationError: string;
```
teallarson, line 23: We may want to refactor some around `hydrationError`s.  Sometimes we use it in reference to the string message itself, and sometimes we use it in reference to the `Error`.  I considered that outside of this PR.

ui/ui-components/pages/property/[id]/edit.tsx
```diff
@@ -27,6 +30,15 @@ export default function Page(props) {
...
     hydrationError,
+  }: {
+    errorHandler: ErrorHandler;
+    successHandler: SuccessHandler;
+    sources: Models.SourceType[];
+    propertiesHandler: PropertiesHandler;
+    types: Actions.AppOptions["types"];
+    filterOptions: Actions.ScheduleFilterOptions["options"];
+    properties: Models.PropertyType[];
+    hydrationError: Error;
```
Comment (teallarson) [inline ui/ui-components/pages/property/[id]/edit.tsx:41]: Here's an example where we're referencing the `Error` type... but we use the same terminology/variable name.

ui/ui-components/pages/property/[id]/edit.tsx
```diff
@@ -71,7 +83,7 @@ export default function Page(props) {
     );
     if (response?.property) {
       setProperty(response.property);
-      propertiesHandler.set(response.property);
+      propertiesHandler.set([response.property]);
```
Comment (teallarson) [inline ui/ui-components/pages/property/[id]/edit.tsx:86]: Thanks Typescript!  We had been calling this with an incorrect type.

ui/ui-config/pages/index.tsx
```diff
@@ -18,12 +19,12 @@ export default function Page(props) {
 
   useEffect(() => {
     if (navigationMode === "config:authenticated") {
-      getSetupSteps(props);
+      checkSetupSteps(props);
```
teallarson, line 22: Changed this for clarity.  It doesn't return the setupSteps so it didn't make sense to me to call it a "getter".  It just checks if they're done and decides what CTA to show.

ui/ui-enterprise/pages/source/[id]/runs.tsx
```diff
@@ -18,7 +32,7 @@ export default function Page(props) {
     try {
       await execApi("post", `/schedule/${source.schedule.id}/run`);
       successHandler.set({ message: "run enqueued" });
-      runsHandler.set({});
+      runsHandler.set(null);
```
Comment (teallarson) [inline ui/ui-enterprise/pages/source/[id]/runs.tsx:35]: Another thank you to Typescript here!

teallarson (approved):

teallarson:
@evantahler Completed that to-do and left some comments along the way... didn't want to merge until you review as well as it touches nearly our entire ui suite.

ui/ui-components/components/layouts/main.tsx
evantahler: Good find, and I agree! Can you write up a story in Tracker please about this?

ui/ui-components/utils/appHandler.ts
```diff
@@ -1,3 +1,4 @@
+import { Models } from "./apiData";
 import { EventDispatcher } from "./eventDispatcher";
 
-export class AppHandler extends EventDispatcher {}
+export class AppHandler extends EventDispatcher<Models.AppType> {}
```
evantahler, line 4: Oops! I missed some!

evantahler (commented):
This is awesome!  I didn't realize that I was setting you up for so many changes... thank you!

This is a great refactor and now we'll be much safer across all of our pages!

evantahler:
It looks like can't approve I PR that I started... but... 👍 !

teallarson:
> This is awesome! I didn't realize that I was setting you up for so many changes... thank you!
> 
> This is a great refactor and now we'll be much safer across all of our pages!

It was an oddly satisfying exercise.  Definitely safer now!

PR #2081, opened 2021-07-27
only show delete button in config ui

teallarson:
Now that we can delete profiles from the source, we only want the "delete" button visible in Config UI.


evantahler (approved):
I guess that's it!

PR #2085, opened 2021-07-27
remove empty options objs in destination templates

teallarson:
If there are no options for a destination, we won't show an empty options object in the config file anymore.

### DIFF
--- a/plugins/@grouparoo/customerio/public/templates/destination/{{{id}}}.js.template
+++ b/plugins/@grouparoo/customerio/public/templates/destination/{{{id}}}.js.template
@@ -9,8 +9,6 @@ exports.default = async function buildConfig() {
       groupId: "...", // The ID of the group whose members you want to export - e.g. `groupId: "high_value_customers"`
       syncMode: {{{syncMode}}}, // How should Grouparoo sync with this destination? Options: "sync", "additive", "enrich"
 
-      options: {},
-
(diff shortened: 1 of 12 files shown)


evantahler (changes requested):
Looks good -  It looks like the CSV App options are also empty and can be removed.

teallarson:
> Looks good - It looks like the CSV App options are also empty and can be removed.

I did see that in the story, but I didn't find any in [the templates](https://github.com/grouparoo/grouparoo/tree/main/plugins/%40grouparoo/csv/public/templates).  Would they be in another place?  I tried searching the repo for `options: {}`.

evantahler:
I don't see it either!   I guess the story is out-of-date. 
* https://github.com/grouparoo/grouparoo/blob/main/plugins/%40grouparoo/csv/public/templates/app/%7B%7B%7Bid%7D%7D%7D.js.template looks good
* CLI looks good:
```
$ roo generate csv:app CSV-APP
...
✅ wrote /Users/evan/workspace/grouparoo/grouparoo/apps/staging-enterprise/config/apps/csv-app.js

$ cat /Users/evan/workspace/grouparoo/grouparoo/apps/staging-enterprise/config/apps/csv-app.js
  return [
    {
      class: "app",
      id: "csv-app",
      name: "csv-app",
      type: "csv",
    },
  ];
};
```

evantahler (approved):

PR #2087, opened 2021-07-28
delete profiles when they are no longer present in the source

pedroslopez:
- Merge integration branch for deleting profiles
- Fix conflict with importing multiple profile properties when a directlyMapped property is also in the group


core/src/tasks/profileProperty/importProfileProperties.ts
```diff
@@ -70,14 +70,16 @@ export class ImportProfileProperties extends RetryableTask {
...
+        dependencies[property.id]
+          .filter((dep) => !params.propertyIds.includes(dep.id))
+          .forEach((dep) => {
+            if (
+              profile.profileProperties.find((p) => p.propertyId === dep.id)
+                ?.state !== "ready"
+            ) {
+              ok = false;
+            }
+          });
```
evantahler, line 82: Yay! The test looks a bit gnarly, but at least the solution wasn't too crazy.

core/__tests__/tasks/profileProperty/importProfileProperties.ts
```diff
@@ -113,6 +114,91 @@ describe("tasks/profileProperty:importProfileProperties", () => {
...
+      expect(profileProperty.startedAt.getTime()).toBeGreaterThan(
+        new Date().getTime() - 1000 * 60 * 5
+      );
+      expect(profileProperty.startedAt.getTime()).toBeLessThan(
+        new Date().getTime()
+      );
+      await profile.destroy();
+    });
+
+    test("will import profile properties that have pending dependencies imported at the same time", async () => {
```
evantahler, line 160: ❤️

core/__tests__/actions/properties.ts
```diff
@@ -336,13 +336,15 @@ describe("actions/properties", () => {
...
         unique: true,
+        keepValueIfNotFound: true,
       };
       const { error, property } = await specHelper.runAction(
         "property:edit",
         connection
       );
       expect(error).toBeUndefined();
       expect(property.unique).toBe(true);
+      expect(property.keepValueIfNotFound).toBe(true);
```
evantahler: Maybe test this on the `create` action too

core/__tests__/tasks/import/associateProfile.ts
```diff
@@ -122,5 +122,39 @@ describe("tasks/import:associateProfile", () => {
       const profile = await Profile.findById(property.profileId);
       expect(profile.state).toEqual("pending");
     });
+
+    test("it will set properties included in the import data", async () => {
```
evantahler, line 126: ... we certainly /should/ have been testing this...

core/__tests__/tasks/profile/checkReady.ts
```diff
@@ -3,7 +3,11 @@ import { api, config, task, specHelper } from "actionhero";
 import { Group, plugin, Profile, ProfileProperty } from "../../../src";
 
 describe("tasks/profile:checkReady", () => {
-  helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
+  helper.grouparooTestServer({
+    truncate: true,
+    enableTestPlugin: true,
+    disableTestPluginImport: true,
```
evantahler, line 9: Ah - looks like you are doing this in more tests so that profiles don't get deleted!

core/__tests__/tasks/profileProperty/enqueue.ts
```diff
@@ -115,9 +115,7 @@ describe("tasks/profileProperties:enqueue", () => {
           const pendingProfileProperties = await ProfileProperty.findAll({
             where: { state: "pending" },
           });
-          expect(pendingProfileProperties.length).toBe(
-            (propertiesCount - 1) * 2
-          );
```
evantahler, line 120: 👍 no more special profile property!

core/__tests__/tasks/system/status/status.ts
```diff
@@ -98,6 +98,27 @@ describe("tasks/status", () => {
       await run.destroy();
     });
 
+    test("will not complete with profiles to delete", async () => {
```
evantahler, line 101: 👍

core/src/actions/properties.ts
```diff
@@ -185,6 +187,7 @@ export class PropertyEdit extends AuthenticatedAction {
       sourceId: { required: false },
       options: { required: false },
       filters: { required: false },
+      keepValueIfNotFound: { required: false },
```
evantahler, line 192: Doesn't look like this new input/param is used in the action?

core/src/classes/codeConfig.ts
```diff
@@ -34,7 +34,9 @@ export interface ConfigurationObject {
   identifying?: boolean;
```
evantahler, line 34: This `ConfigurationObject` is getting out of hand!  This was a lazy choice on my part :( 
It might be better to make the type from the union of a few distinct types.  This way, we can check that each form of using the object is internally consistent. 

```ts
export interface ProfileConfigurationObject = {
// just what profile needs
}

export interface AppConfigurationObject = {
// just what app needs
}

export interface ConfigurationObject = ProfileConfigurationObject & AppConfigurationObject


```

evantahler: This could be another story

evantahler: Added a story to maybe delete the `SyncTable` part of this (https://www.pivotaltracker.com/story/show/179063948)

core/src/initializers/settings.ts
```diff
@@ -143,6 +143,14 @@ export class Plugins extends CLSInitializer {
           "The default country code Grouparoo will use to format phone numbers and display data",
         type: "string",
       },
+      {
+        key: "confirm-profiles-days",
+        title: "Confirm Profile Existence Days",
+        defaultValue: 7,
```
evantahler, line 149: 👍 this is a good default.  We will check once a week for profiles we haven't imported in a while and delete them if they are removed from the source.

cc @andyjih

core/src/migrations/000077-addScheduleConfirmProfiles.ts
```diff
@@ -0,0 +1,15 @@
+export default {
+  up: async function (migration, DataTypes) {
+    await migration.sequelize.transaction(async () => {
+      await migration.addColumn("schedules", "confirmProfiles", {
+        type: DataTypes.BOOLEAN,
```
evantahler, line 5: This migration allows `schedules.confirmProfiles` to be null - is that intended?  I think it might be preferable to not allow it to be null, and in the migration, update all exiting rows to `false`

core/src/migrations/000078-propertyKeepValueIfNotFound.ts
```diff
@@ -0,0 +1,15 @@
+export default {
+  up: async function (migration, DataTypes) {
+    await migration.sequelize.transaction(async () => {
+      await migration.addColumn("properties", "keepValueIfNotFound", {
+        type: DataTypes.BOOLEAN,
```
evantahler, line 5: This migration allows `properties.keepValueIfNotFound` to be null - is that intended?  I think it might be preferable to not allow it to be null, and in the migration, update all exiting rows to `true` to match the existing behavior

core/src/tasks/system/status/status.ts
```diff
@@ -1,4 +1,4 @@
-import { api, log, task } from "actionhero";
+import { api, log, task, Task } from "actionhero";
```
evantahler: unused import

plugins/@grouparoo/spec-helper/src/lib/factories/profile.ts
```diff
@@ -9,7 +10,21 @@ const data = async (props = {}) => {
...
-  return Profile.create(await data(props));
+  const profile = await Profile.create(await data(props));
+
+  const { Property } = await import(`@grouparoo/core/${loadPath}`);
+  const allProperties = await Property.findAllWithCache();
+  const directlyMappedProperty = allProperties.find((p) => p.directlyMapped);
+
+  if (directlyMappedProperty) {
+    properties[directlyMappedProperty.key] = faker.datatype.number(99999);
+  }
```
evantahler: Think there could be collisions in tests because of this.  Does faker return a random number or something smarter?

ui/ui-components/pages/import/[id]/edit.tsx
```diff
@@ -16,13 +16,6 @@ export default function Page(props) {
     _import,
   }: { groups: Models.GroupType[]; _import: Models.ImportType } = props;
 
-  function groupName(groupId) {
-    const group = groups.filter((g) => g.id === groupId)[0];
-    if (group) {
-      return group.name;
-    }
-  }
-
```
Comment (evantahler) [inline ui/ui-components/pages/import/[id]/edit.tsx:25]: Yeah, I guess the only creator with a page to view would be a Run!

evantahler (commented):
This is HUGE - I guess that's what happened with integration branches!  Well done!

I want to give this a second pass, but here are a few comments.   I'm most interested to hear what you think about the comments on the Migrations and typing the ConfigurationObject more tightly.

plugins/@grouparoo/spec-helper/src/lib/factories/profile.ts
pedroslopez: Changed to use `faker.unique` to ensure uniqueness!

core/src/actions/properties.ts
pedroslopez: It is! the action uses `await property.update(params);`, and `keepValueIfNotFound` is in that params obj (along with other keys)

core/__tests__/tasks/profile/checkReady.ts
pedroslopez: Yeah, a lot of tests were relying on the old behavior of keeping the old value instead of clearing it, and setting property values manually.

core/src/classes/codeConfig.ts
pedroslopez: I definitely agree! At the moment it's tough to know which part of that corresponds to each config object type.

I think we should get this in and add another story for this, maybe also using these new interfaces in the config loaders as well. I can work on this right after if prioritized.

core/src/migrations/000077-addScheduleConfirmProfiles.ts
pedroslopez: Good catch!

core/src/migrations/000078-propertyKeepValueIfNotFound.ts
pedroslopez: I updated this migration, but I'm not sure about matching existing behavior. I think setting it to `true` for existing properties is definitely the safest option, that way if they don't want anything to be deleted, nothing will happen. It will however be a bit of extra work if you _do_ want to delete things, which I think may be what most people want to do. Note that if this option is marked `true` for the directlyMapped property, nobody will ever be deleted.

core/src/classes/codeConfig.ts
evantahler: I've got a PR up for this - https://github.com/grouparoo/grouparoo/pull/2118, but I won't merge it until after this!

core/src/initializers/settings.ts
evantahler: @andyjih agrees that once a week is OK for now

core/src/migrations/000077-addScheduleConfirmProfiles.ts
evantahler: I think it's usually better to not assign a defaultValue in the migration.  This way you can still get an error from the database if a null is set, rather than silently assigning defaults - Check out https://github.com/grouparoo/grouparoo/blob/main/core/src/migrations/000060-addTypeToOption.ts

If you do a 2-part migration, first you add the column + a default for all existing rows, and then remove the defaults.

core/src/migrations/000078-propertyKeepValueIfNotFound.ts
evantahler: We can add that to the migration guide! 

Check the note on the other migration too...

PR #2089, opened 2021-07-29
mimic "hello" submenu style on platform

teallarson:
Copies the style of the submenu beneath it.  If there are resque errors, the carets are after the errors... they don't change when the menu was expanded, but neither do the carets on "Hello _______ "

Screen Shot 2021-07-29 at 4 32 23 PM

Screen Shot 2021-07-29 at 4 33 42 PM


evantahler:
This is 👍 with me as is!

I bet we *could* rotate those carets with CSS as the menus expand...

evantahler (approved):

PR #2117, opened 2021-08-02
ApiKey values can be set via codeConfig

evantahler:
Allows users to choose a value for an API Key when using codeConfig, eg: 

```js
// from config/apiKeys/main.js

module.exports = function () {
  return {
    class: "apiKey",
    id: "main_api_key",
    name: "Main API Key",
    apiKey: process.env.MAIN_API_KEY, // <-- reference  MAIN_API_KEY from ENV
    options: {
      permissionAllRead: true,
      permissionAllWrite: true,
    },
  };
};

```


bleonard:
It would be kind of nice if something like `GROUPAROO_OPTION__API_KEY_MAIN_API_KEY="something"` worked, but I don't think that's a priority.

PR #2119, opened 2021-08-03
Sources can map via non-unique mapping

evantahler:
This PR makes is possible to create Source Mappings against non-unique Properties. 

When importing, for efficiency, rather querying the Source multiple times for the same non-unique value, we take the value returned for one value of the mapped property and copy to all other Profiles with the same mapped lookup value.  

There are some new caveats:
* Mappings cannot be made against array properties
* If a Source is mapped though a non-unique Property, it cannot create unique Properties
* If a Source is mapped though a non-unique Property, it cannot create unique array Properties
* If a Source is mapped though a non-unique Property, it cannot have a Schedule

TODO: 
- [x] Tests
- [x] Ensure that batching only works for Exact properties with no filters
- [x] Prevent schedules on sources mapped though a non-unique property
- [x] Check nulls on import chain

### DIFF
--- a/core/__tests__/models/property/property.ts
+++ b/core/__tests__/models/property/property.ts
@@ -749,29 +749,29 @@ describe("models/property", () => {
         await source.setMapping({ id: "userId" });
       });
 
-      test("properties mapped though unique properties can be unique", async () => {
+      test("properties mapped through unique properties can be unique", async () => {
         await source.setMapping({ id: "userId" });
         await property.update({ unique: true });
         expect((await property.reload()).unique).toEqual(true);
       });
 
-      test("properties mapped though unique properties can be arrays", async () => {
+      test("properties mapped through unique properties can be arrays", async () => {
         await source.setMapping({ id: "userId" });
         await property.update({ isArray: true });
         expect((await property.reload()).isArray).toEqual(true);
       });
 
-      test("properties mapped though non-unique properties cannot be unique", async () => {
+      test("properties mapped through non-unique properties cannot be unique", async () => {
         await source.setMapping({ id: "lastName" });
         await expect(property.update({ unique: true })).rejects.toThrow(
-          /A unique Property cannot be mapped though a non-unique Property/
+          /A unique Property cannot be mapped through a non-unique Property/
         );
       });
 
-      test("properties mapped though non-unique properties cannot be arrays", async () => {
+      test("properties mapped through non-unique properties cannot be arrays", async () => {
         await source.setMapping({ id: "lastName" });
         await expect(property.update({ isArray: true })).rejects.toThrow(
-          /An array Property cannot be mapped though a non-unique Property/
+          /An array Property cannot be mapped through a non-unique Property/
         );
       });
     });
--- a/core/__tests__/models/property/property.ts
+++ b/core/__tests__/models/property/property.ts
@@ -280,6 +280,7 @@ describe("models/property", () => {
       appId: app.id,
       type: "source-no-options",
     });
+    await source.setMapping({ id: "userId" });
     await source.update({ state: "ready" });
     const property = await Property.create({
       key: "property-no-options",
@@ -581,7 +582,7 @@ describe("models/property", () => {
     await source.destroy();
   });
 
-  describe("directlyMapping", () => {
+  describe("directlyMapped", () => {
     let userIdProperty: Property;
     let emailProperty: Property;
 
@@ -726,6 +727,53 @@ describe("models/property", () => {
 
     beforeEach(() => {
       queryCounter = 0;
+    });
+
+    describe("mapped though a non-unique property", () => {
+      let property: Property;
+
+      beforeAll(async () => {
+        property = await helper.factories.property(
+          source,
+          { key: "wordInSpanish" },
+          { column: "spanishWord" }
+        );
+      });
+
+      beforeEach(async () => {
+        await property.update({ unique: false, isArray: false });
+      });
+
+      afterAll(async () => {
+        await property.destroy();
+        await source.setMapping({ id: "userId" });
+      });
+
+      test("properties mapped though unique properties can be unique", async () => {
+        await source.setMapping({ id: "userId" });
+        await property.update({ unique: true });
+        expect((await property.reload()).unique).toEqual(true);
+      });
+
+      test("properties mapped though unique properties can be arrays", async () => {
+        await source.setMapping({ id: "userId" });
+        await property.update({ isArray: true });
+        expect((await property.reload()).isArray).toEqual(true);
+      });
+
+      test("properties mapped though non-unique properties cannot be unique", async () => {
+        await source.setMapping({ id: "lastName" });
+        await expect(property.update({ unique: true })).rejects.toThrow(
+          /A unique Property cannot be mapped though a non-unique Property/
+        );
+      });
+
+      test("properties mapped though non-unique properties cannot be arrays", async () => {
+        await source.setMapping({ id: "lastName" });
+        await expect(property.update({ isArray: true })).rejects.toThrow(
+          /An array Property cannot be mapped though a non-unique Property/
+        );
+      });
     });
 
     describe("filters", () => {
--- a/core/src/models/Source.ts
+++ b/core/src/models/Source.ts
@@ -216,10 +216,21 @@ export class Source extends LoggedModel<Source> {
 
   async scheduleAvailable() {
     const { pluginConnection } = await this.getPlugin();
-    if (typeof pluginConnection?.methods?.profiles === "function") {
+    if (typeof pluginConnection?.methods?.profiles !== "function") {
+      return false;
+    }
+
+    const mapping = await this.getMapping();
+    if (Object.values(mapping).length === 0) {
+      return true;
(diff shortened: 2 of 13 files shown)


core/__tests__/models/property/property.ts
```diff
@@ -728,6 +729,53 @@ describe("models/property", () => {
...
+      beforeEach(async () => {
+        await property.update({ unique: false, isArray: false });
+      });
+
+      afterAll(async () => {
+        await property.destroy();
+        await source.setMapping({ id: "userId" });
+      });
+
+      test("properties mapped though unique properties can be unique", async () => {
```
pedroslopez: very nitpicky, but this small typo is copied multiple times :)
```suggestion
      test("properties mapped through unique properties can be unique", async () => {
```

core/src/models/Source.ts
```diff
@@ -216,10 +216,21 @@ export class Source extends LoggedModel<Source> {
...
+
+    const mapping = await this.getMapping();
+    if (Object.values(mapping).length === 0) {
+      return true;
+    } else {
+      const propertyMappingKey = Object.values(mapping)[0];
+      const property = (await Property.findAllWithCache()).find(
+        (p) => p.key === propertyMappingKey
+      );
+      if (!property.unique) return false;
```
pedroslopez, line 231: 👍🏼 can't create a schedule for a source mapped with non-unique property

core/src/modules/ops/source.ts
```diff
@@ -253,6 +258,12 @@ export namespace SourceOps {
         profileIds: profiles.map((p) => p.id),
       });
 
+      await applyNonUniqueMappedResultsToAllProfiles(response, {
+        profiles,
+        properties,
+        sourceMapping,
+      });
+
```
pedroslopez, line 266: I understand that this new method sets the non-uniquely mapped properties values for all other pending profiles, but it's not totally clear to me why this is necessary. Does the `profileProperties` method not already return a response with the values set for each of the profiles in question?

Looking a bit into the `app-template`'s `profileProperties` method implementation, it seems like if the "primaryKey" (the mapping) matches a profile, it will be pushed to an array:
https://github.com/grouparoo/grouparoo/blob/b31360b3418a0a7b9f04b454d9a336d9ca1b6e55/plugins/%40grouparoo/app-templates/src/source/table/profileProperties.ts#L64-L66

...and then the resulting value will be set on each of those profiles:
https://github.com/grouparoo/grouparoo/blob/b31360b3418a0a7b9f04b454d9a336d9ca1b6e55/plugins/%40grouparoo/app-templates/src/source/table/profileProperties.ts#L110-L120

From what I'm reading, it suggests that multiple profiles can have this same mapping value and it should set it on all, no? What am I missing?

pedroslopez (commented):
Looks good! I mainly have one question about the "fanning out"

core/__tests__/models/property/property.ts
evantahler: Typo! fixed up in f06019c64299b7e900fa3ae4e7ddfbe49b875798

core/src/modules/ops/source.ts
evantahler: > I understand that this new method sets the non-uniquely mapped properties values for all other pending profiles, but it's not totally clear to me why this is necessary. Does the `profileProperties` method not already return a response with the values set for each of the profiles in question?
> ...
> From what I'm reading, it suggests that multiple profiles can have this same mapping value and it should set it on all, no? What am I missing?


Yeah, this was the confusing part!

https://github.com/grouparoo/grouparoo/blob/b31360b3418a0a7b9f04b454d9a336d9ca1b6e55/plugins/%40grouparoo/app-templates/src/source/table/profileProperties.ts#L64-L66 sets up the response object to have an entry for each profile.  

https://github.com/grouparoo/grouparoo/blob/b31360b3418a0a7b9f04b454d9a336d9ca1b6e55/plugins/%40grouparoo/app-templates/src/source/table/profileProperties.ts#L110-L120 takes what the plugin returns and applies to the response hash per profile, per property... BUT it doesn't guarantee that every profile sent to the plugin will get a result back in this case.

I've been using a translation [dataset in my tests](https://github.com/grouparoo/grouparoo/blob/f06019c64299b7e900fa3ae4e7ddfbe49b875798/core/__tests__/data/translations.csv).  Imagine that was SQL, and we had 100 Profiles with the property `english_word="Spanish"` and `english_word="English"` and we were trying to load up the property `spanish_word`.  The generated query would be something like:
  
```sql
SELECT "spanish_word", "english_word" as __pk FROM translations WHERE  "english_word" IN ('Spanish','English') GROUP BY english_word,spanish_word
```
... which means we'll only get one row back from the database for each language here (2 rows total), not the 100 we need.  So, the code then "fans out" (eg: copies) the response over and over to all the Profiles that has the same `english_word`.

The alternative would have been to come up with a query like @tealjulia talks about in her [0's blog post](https://www.grouparoo.com/blog/data-cleaning-null-aggregate-values) to union together a result set that duplicates the data in SQL rather than JS.  I chose the in-code way to make this more generic for all sources.

evantahler:
> I understand that this new method sets the non-uniquely mapped properties values for all other pending profiles, but it's not totally clear to me why this is necessary. Does the `profileProperties` method not already return a response with the values set for each of the profiles in question?
> ...
> From what I'm reading, it suggests that multiple profiles can have this same mapping value and it should set it on all, no? What am I missing?


Yeah, this was the confusing part!

This: 
https://github.com/grouparoo/grouparoo/blob/b31360b3418a0a7b9f04b454d9a336d9ca1b6e55/plugins/%40grouparoo/app-templates/src/source/table/profileProperties.ts#L64-L66 sets up the response object to have an entry for each profile.  

And this: 
https://github.com/grouparoo/grouparoo/blob/b31360b3418a0a7b9f04b454d9a336d9ca1b6e55/plugins/%40grouparoo/app-templates/src/source/table/profileProperties.ts#L110-L120 takes what the plugin returns and applies to the response hash per profile, per property... BUT it doesn't guarantee that every profile sent to the plugin will get a result back in this case.

I've been using a translation [dataset in my tests](https://github.com/grouparoo/grouparoo/blob/f06019c64299b7e900fa3ae4e7ddfbe49b875798/core/__tests__/data/translations.csv).  Imagine that was SQL, and we had 100 Profiles with the property `english_word="Spanish"` and `english_word="English"` and we were trying to load up the property `spanish_word`.  The generated query would be something like:
  
```sql
SELECT "spanish_word", "english_word" as __pk FROM translations WHERE  "english_word" IN ('Spanish','English') GROUP BY english_word,spanish_word
```
... which means we'll only get one row back from the database for each language here (2 rows total), not the 100 we need.  So, the code then "fans out" (eg: copies) the response over and over to all the Profiles that has the same `english_word`.

The alternative would have been to come up with a query like @tealjulia talks about in her [0's blog post](https://www.grouparoo.com/blog/data-cleaning-null-aggregate-values) to union together a result set that duplicates the data in SQL rather than JS.  I chose the in-code way to make this more generic for all sources.

pedroslopez:
> > I understand that this new method sets the non-uniquely mapped properties values for all other pending profiles, but it's not totally clear to me why this is necessary. Does the `profileProperties` method not already return a response with the values set for each of the profiles in question?
> > ...
> > From what I'm reading, it suggests that multiple profiles can have this same mapping value and it should set it on all, no? What am I missing?
> 
> Yeah, this was the confusing part!
> 
> This:
> https://github.com/grouparoo/grouparoo/blob/b31360b3418a0a7b9f04b454d9a336d9ca1b6e55/plugins/%40grouparoo/app-templates/src/source/table/profileProperties.ts#L64-L66
> 
> sets up the response object to have an entry for each profile.
> And this:
> https://github.com/grouparoo/grouparoo/blob/b31360b3418a0a7b9f04b454d9a336d9ca1b6e55/plugins/%40grouparoo/app-templates/src/source/table/profileProperties.ts#L110-L120
> 
> takes what the plugin returns and applies to the response hash per profile, per property... BUT it doesn't guarantee that every profile sent to the plugin will get a result back in this case.
> I've been using a translation [dataset in my tests](https://github.com/grouparoo/grouparoo/blob/f06019c64299b7e900fa3ae4e7ddfbe49b875798/core/__tests__/data/translations.csv). Imagine that was SQL, and we had 100 Profiles with the property `english_word="Spanish"` and `english_word="English"` and we were trying to load up the property `spanish_word`. The generated query would be something like:
> 
> ```sql
> SELECT "spanish_word", "english_word" as __pk FROM translations WHERE  "english_word" IN ('Spanish','English') GROUP BY english_word,spanish_word
> ```
> 
> ... which means we'll only get one row back from the database for each language here (2 rows total), not the 100 we need. So, the code then "fans out" (eg: copies) the response over and over to all the Profiles that has the same `english_word`.
> 
> The alternative would have been to come up with a query like @tealjulia talks about in her [0's blog post](https://www.grouparoo.com/blog/data-cleaning-null-aggregate-values) to union together a result set that duplicates the data in SQL rather than JS. I chose the in-code way to make this more generic for all sources.

Ah! Makes sense, thanks for explaining!

pedroslopez (approved):

PR #2124, opened 2021-08-04
Fix Property Option `ownerType`

evantahler:
This PR adds a missing update to the Options' ownerType for Properties.

NOTE: There may be a duplicate key error in this migration if you've partially updated your database between this migration and 000056

### DIFF
--- a/core/src/migrations/000079-fixPropertyOwnerType.ts
+++ b/core/src/migrations/000079-fixPropertyOwnerType.ts
@@ -10,7 +10,7 @@ export default {
   down: async function (migration) {
     await migration.sequelize.transaction(async () => {
       await migration.sequelize.query(
-        `UPDATE "options" SET "ownerType"='profilePropertyRule' WHERE "ownerType"='profilePropertyRule';`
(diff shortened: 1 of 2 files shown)


core/src/migrations/000079-fixPropertyOwnerType.ts
```diff
@@ -0,0 +1,17 @@
...
+      await migration.sequelize.query(
+        `UPDATE "options" SET "ownerType"='property' WHERE "ownerType"='profilePropertyRule'`
+      );
+    });
+  },
+
+  down: async function (migration) {
+    await migration.sequelize.transaction(async () => {
+      await migration.sequelize.query(
+        `UPDATE "options" SET "ownerType"='profilePropertyRule' WHERE "ownerType"='profilePropertyRule';`
```
bleonard: I supose it should be 
```
WHERE "ownerType"='property';`
```

evantahler: Good catch, thanks!

PR #2125, opened 2021-08-04
User can generate API key in template

teallarson:


core/public/templates/apiKeys/{{{id}}}.js.template
```diff
@@ -4,6 +4,7 @@ exports.default = async function buildConfig() {
       class: "apiKey",
       id: {{{id}}},
       name: {{{id}}},
+      apiKey: "API KEY HERE", // you can hard-code the API key you want to use, or source it from `process.env.PRODUCTION_API_KEY`
```
evantahler: Thinking about this from a security lense, I think the thing to do is leave line 7 in the template, but commented out.  This way we provide clear instructions on how to provide your own APIKey, but by default we are relying on grouparoo to generate one randomly.

So:

```
      name: {{{id}}},
      // apiKey: "Custom API Key", // By default, Grouparoo will generate a random APIKey value. However, you can choose to hard-code the API key you want to use, or source it from `process.env.PRODUCTION_API_KEY`.
```

evantahler (changes requested):

evantahler (approved):

PR #2129, opened 2021-08-05
Nodejs v16 Support

evantahler:
Includes switching form `node-sass` to `sass`


krishnaglick:
CI is failing on node-sass. I'm not sure if newer react-bootstrap uses it at all but on node-sass's github page they specify it's deprecated and to use [dart-sass](https://sass-lang.com/dart-sass)
I used it at iHeart without issue, we probably should make that transition.

.nvmrc
```diff
@@ -1 +1 @@
-v15
+v16
```
evantahler, line 1: This will change the default node.js version we use when developing to v16.

evantahler:
One (acceptable) downside is that sass now prints lots of warnings from our bootstrap sass styles

Screen Shot 2021-09-08 at 8 53 13 AM

PR #2135, opened 2021-08-06
Speed up Profile Deletion checks

evantahler:
Using:
```
./bin/timed_run staging-enterprise 100
```

Timed Run Before this PR:
```
🎉 Complete in 0hrs 55min 30sec [3330 seconds]
```

Timed Run After this PR:
```
🎉 Complete in 0hrs 48min 35sec [2915 seconds]
```

Not much faster locally...

### DIFF
--- a/core/__tests__/tasks/profile/sweep.ts
+++ b/core/__tests__/tasks/profile/sweep.ts
@@ -32,32 +32,6 @@ describe("tasks/profiles:sweep", () => {
     expect(found.length).toEqual(1);
   });
 
-  test("deletes profiles without a directlyMapped property", async () => {
-    const profile: Profile = await helper.factories.profile();
-    await profile.addOrUpdateProperties({
-      userId: [1000],
-      isVIP: [true],
-      ltv: [213],
-    });
-
-    await ProfileProperty.update(
-      { state: "ready" },
-      { where: { profileId: profile.id } }
-    );
-    await profile.update({ state: "ready" });
-
-    await ProfileProperty.destroy({
-      where: {
-        profileId: profile.id,
-        propertyId: userIdProperty.id,
-      },
-    });
-
-    const count = await specHelper.runTask("profiles:sweep", {});
-    expect(count).toBe(1);
-    await expect(profile.reload()).rejects.toThrow(/does not exist anymore/);
-  });
-
   test("deletes profiles with a directlyMapped property set to null", async () => {
     const profile: Profile = await helper.factories.profile();
     await profile.addOrUpdateProperties({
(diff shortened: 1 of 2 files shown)


core/src/modules/ops/profile.ts
```diff
@@ -671,24 +671,43 @@ export namespace ProfileOps {
...
+  AND 0 = (
+    SELECT count("exports"."id") FROM "exports"
+      WHERE "exports"."profileId"="profiles"."id"
+      AND "exports"."state" IN ('pending', 'processing')
+  )
+  AND "id" NOT IN (
+    SELECT DISTINCT("profileId") FROM "profileProperties"
+    JOIN properties ON "properties"."id"="profileProperties"."propertyId"
+    WHERE "properties"."directlyMapped"=true AND "rawValue" IS NOT NULL
+  )
```
pedroslopez: This looks the same- did you forget to invert it?

core/src/modules/ops/profile.ts
```diff
@@ -671,24 +671,43 @@ export namespace ProfileOps {
...
     );
 
+    if (directlyMappedProperties.length === 0) {
+      // We have no directly mapped Property and every profile should be removed
+      // It's safe to assume that if there are no Properties, we aren't exporting
+      profiles = await Profile.findAll({
+        attributes: ["id"],
+        where: { state: "ready" },
+        limit,
+      });
```
pedroslopez, line 686: This is definitely faster! 👍🏼

core/src/modules/ops/profile.ts
```diff
@@ -671,24 +671,43 @@ export namespace ProfileOps {
...
+      profiles = await api.sequelize.query(
+        `
+  SELECT "id" FROM "profiles"
+  WHERE "state"='ready'
+  AND 0 = (
+    SELECT count("exports"."id") FROM "exports"
+      WHERE "exports"."profileId"="profiles"."id"
+      AND "exports"."state" IN ('pending', 'processing')
+  )
+  AND "id" NOT IN (
```
pedroslopez: Something interesting about this is that it also worked for when we eventually add more directlyMapped properties for different profile types. If any directlyMapped property has a value, it wouldn't be sweeped, but in the new inverted version if any is null, it would be sweeped.

I think this is fine and not really relevant until we do implement profile/entity types, so not something to worry about now. Just noting that we may need to revisit this later.

pedroslopez (changes requested):
Seems like it's still the same query!

There's currently a test that checks that profiles that dont have a directlyMapped property are sweeped (`"deletes profiles without a directlyMapped property"`), but I think that'll need to be updated too to match the new logic of being triggered only when there are no directlyMapped properties in the system instead of when a specific profile doesn't have a directlyMapped property. The test will probably fail once the logic is inverted.

core/src/modules/ops/profile.ts
evantahler: Good point! I think when we have Entity/Models, we'll have to do this in a loop - once per entity/model.  Not all Directly Mapper properties will belong to each type.  The good news is that we'll hopefully get to a point where each type only has one "root" property - the one and only one directly mapped property.

core/__tests__/tasks/profile/sweep.ts
```diff
@@ -178,4 +152,29 @@ describe("tasks/profiles:sweep", () => {
...
+    await ProfileProperty.update(
+      { state: "ready" },
+      { where: { profileId: profile.id } }
+    );
+    await profile.update({ state: "ready" });
+    let count = await specHelper.runTask("profiles:sweep", {});
+    expect(count).toBe(0);
+
+    // now remove the directly mapped property
+    // @ts-ignore
```
pedroslopez, line 174: curious- why was this needed?

pedroslopez, line 174: curious - why was this needed?

core/src/modules/ops/profile.ts
```diff
@@ -671,24 +671,46 @@ export namespace ProfileOps {
...
+        AND "exports"."state" IN ('pending', 'processing')
+  )
+  AND "id" IN (
+    SELECT DISTINCT("profileId") FROM "profileProperties"
+    JOIN properties ON "properties"."id"="profileProperties"."propertyId"
+    WHERE
+      "properties"."directlyMapped"=true
+      AND "rawValue" IS NULL
+  )
+  LIMIT ${limit};
```
pedroslopez, line 706: Looks good to me!

pedroslopez (approved):
LGTM!

core/__tests__/tasks/profile/sweep.ts
evantahler: It's because I wanted to skip the hooks - Sequelize really doesn't want you doing that.  If I had left the hooks on like normal, the @BeforeDestory checks would fire and I'd get yelled at because the mapping was still in use or something like that

PR #2138, opened 2021-08-10
Prettier format code config

bleonard:
Creates less thrash when editing in apps that use this common tool.


evantahler (approved):
Looks like prettier is already a dependency! 👍

PR #2140, opened 2021-08-11
fix: dont crash on confirmProfiles if the schedule has no runs

pedroslopez:
Interestingly the `required: true` when using sequelize's `include` to fetch other models at the same time doesn't quite work the same if the included model specifies a `limit`. Without the limit, only one combined query is executed and rows that dont have matches for the included model are not returned. With the limit, it's instead split up into two queries and rows that dont have matches for the included model _are_ returned (just with an empty array)


evantahler (approved):
Nice find!

PR #2142, opened 2021-08-12
Adapt migration to handle custom schema.

pauloouriques:
- bump ah-sequelize-plugin version up;
 - adapt migrator to handle queryInterface functions schema injection.


evantahler (approved):
Thank you for also updating the migrator!

PR #2143, opened 2021-08-13
Database name update issue.

pauloouriques:
- get db reference straigt from the MongoDB client passing the appOptions.database every time.

### DIFF
--- a/plugins/@grouparoo/mongo/src/lib/disconnect.ts
+++ b/plugins/@grouparoo/mongo/src/lib/disconnect.ts
@@ -1,5 +1,5 @@
 import { DisconnectPluginAppMethod } from "@grouparoo/core";
 
 export const disconnect: DisconnectPluginAppMethod = async ({ connection }) => {
-  await connection.client.close();
+  await connection.close();
 };
--- a/plugins/@grouparoo/mongo/__tests__/utils/data.ts
+++ b/plugins/@grouparoo/mongo/__tests__/utils/data.ts
@@ -164,7 +164,7 @@ export async function getClient() {
 
 export async function endClient() {
   if (connection) {
-    await connection.client.close();
+    await connection.close();
     connection = null;
   }
 }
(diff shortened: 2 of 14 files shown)


plugins/@grouparoo/mongo/src/lib/disconnect.ts
```diff
@@ -1,5 +1,5 @@
 import { DisconnectPluginAppMethod } from "@grouparoo/core";
 
 export const disconnect: DisconnectPluginAppMethod = async ({ connection }) => {
-  await connection.client.close();
+  await connection.close();
```
evantahler, line 4: What does this change do?  This method should be called already whenever appOptions change... perhaps the problem was that we weren't fully disconnecting the old mongo connections with old appOptions before making a new connection?

evantahler (commented):
This fix looks like it will certainly fix the problem of the database "sticking" to the connection when the user changes it in appOptions.

I'm a little worried about the implementation here - that means that we need to always remember to `connection.db(appOptions.database)` for all future mongo methods.  Is there a way to ensure that we always do set the database before querying?

I'm also a little worried about how we implemented the connection `disconect` method - that should have been called whenever the appOptions change, so `connection.client.close()` should have been called and then a new connection made with the new appOptions.

plugins/@grouparoo/mongo/src/lib/disconnect.ts
pauloouriques: according to my tests when appOptions change both methods are been called normally `connect` and `disconnect`, I also made a query inside the `connect` method and it works well considering the new database. So, I think the problem might be when updating the `connection` reference across the app. So, the solution I found was to follow what is made with postgres plugin.
Also, the connection is made with the mongodb server, so you can access different databases inside this server with the same connection.

evantahler: I guess since the database is never set on the connection, it will be clear that you need to choose a database for each query (and you won't see any data until you do) 👍

evantahler (approved):

PR #2144, opened 2021-08-14
Actionhero v27 + Typed Action tests

evantahler:
Typed `specHelper` from https://github.com/actionhero/actionhero/releases/tag/v27.0.0


core/__tests__/actions/profiles.ts
```diff
@@ -369,31 +377,30 @@ describe("actions/profiles", () => {
...
         expect(groups.length).toBe(0);
       });
     });
 
     describe("search", () => {
-      let mario: Profile;
-      let luigi: Profile;
-      let peach: Profile;
-      let toad: Profile;
+      let mario: AsyncReturnType<ProfileCreate["run"]>["profile"];
```
pedroslopez: aha! we found some wrong types! 👍🏼

core/src/actions/swagger.ts
```diff
@@ -120,12 +138,17 @@ export class Swagger extends Action {
...
-      host: url.replace("http://", "").replace("https://", ""),
-      basePath: `/api/v${API_VERSION}`,
-      tags: tags.map((tag) => {
-        return { name: tag };
-      }),
-      schemes: url.includes("https") ? ["https"] : ["http"],
+      host: process.env.WEB_URL
+        ? process.env.WEB_URL.replace("https://", "").replace("http://", "")
+        : "localhost",
+      basePath: `/api/${API_VERSION}`,
```
pedroslopez, line 145: Does this change what swagger thinks our API endpoints are? (`basePath` changed `/v1/`->`/1/`)

pedroslopez (commented):
This was a big one! Types all around!

Side thought: The use of types like `AsyncReturnType<AppView["run"]>["app"];` is interesting, particularly when being used inside plugins. Maybe if we use these a lot we could type out `AppApiData` or something like that for each of our models and use that. Not a huge deal though, current usage isn't too bad.

core/__tests__/actions/profiles.ts
evantahler: Yeah! That's why I ended up with `AsyncReturnType<...` like you called out above.  But... using AppApiData is much cleaner! I'll update those.

core/src/actions/swagger.ts
evantahler: Simplified and cleaned up in 686fe9b2a8aeb3c2d8a4fb00d19746e6c78dac55

PR #2145, opened 2021-08-16
fix: keepValueIfNotFound is applied when importing synchronously (Import and Export button)

pedroslopez:
- Load keepValueIfNotFound on bootstrapped properties through code config
- Use keepValueIfNotFound when importing through the "Import and Export" button


evantahler (approved):
Good find!

PR #2146, opened 2021-08-16
Calculated properties plugin

### LINKED ISSUE #780: Feature Request: Combined Profile Properties
Destinations may want to combine many Profile Properties into a single Field.  The most common case is for a Property called `name`, consisting of `name = "${firstName} ${lastName}"`.  This is also helpful for `companylocation = ${company} - ${location}`

Another use case would be coalesce a few profile properties - finding the first non-null Profile Property.  Imagine something like `email = coalesce(emailFromUserAccount, emailFromMailChimp, emailFromSignupForm)`

Grouparoo can build new Profile Properties from a special Source pointing to Grouparoo itself.  

N/A

**Additional context**
N/A

teallarson:
DONE:
- generates an app
- generates a source
- generates a property
- replaces mustached variables from a property's customFunction option, checks for illegal strings (require, env)
- Sorts config validations, application, and processing so that calculated properties wait on the properties they reference
- Executes the function from the `customFunction` string in a `Node VM` instance per profile
- saves value returned from `customFunction` as a new `ProfileProperty`
- recalculates on each run (like all other `profileProperties`)
- Testing coverage:
  - cli 
  - validate
  - property-imports for each type (based on query sources)
  - property-imports also tests that the function will throw with invalid input

TO DO AFTER COMPLETE (www):
- (in progress) Docs 
  - basics about the tool in code config docs
  - important to note: 
    - profile property values are parsed into the function as strings
    - we actually use a property's name in the mustache vars, not its id
- What's New 
- Add a calculated property to `roo demo -c`


plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
```diff
@@ -0,0 +1,74 @@
...
+async function getRequiredProperties(
+  i,
+  args,
+  arrayedProperties,
+  requiredPropertyObjects
+) {
+  const propertyOptionId = arrayedProperties[i];
+  const property = await ProfileProperty.findOne({
+    where: { propertyId: propertyOptionId, profileId: args.profile.id },
+  });
```
evantahler: Might want to use `ProfileProperty.findOneWithCache()` to be faster

plugins/@grouparoo/calculated-property/public/templates/property/{{{id}}}.js.template
```diff
@@ -0,0 +1,19 @@
...
+      name: {{{id}}},
+      class: "property",
+      sourceId: {{{sourceId}}}, // The ID of the Source that this Property belongs to - e.g. `sourceId: "calculated_property_source"`
+      type: "string", // The type of the Property.  Options are: {{{__typeOptions}}}
+      unique: false, // Will Profiles have unique records for this Property?
+      identifying: false, // Should we consider this property Identifying in the UI? Only one Property can be identifying.
+      isArray: false, // Is this an Array Property?
+      options: {
+        requiredProperties: [], // This property will receive an array of ALL properties.  What properties do we need to use for the calculation below?  -- e.g. `requiredProperties: [firstName, lastName]`
+        customFunction: `async (requiredProperties) => {}` // The custom function you would like to execute, within a string literal.  Returned value must match property type! -- e.g. `return /`${firstName} ${lastName}`.trim()`
```
evantahler: Once the plugin is further along, we can probably have a more robust example.  I'm curious how we decompose `requiredProperties` into `firstName` and `lastName`

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export interface RequiredPropertiesObject {
+  [propertyId: string]: ProfileProperty;
+}
+
+async function getRequiredProperties(
+  i,
+  args,
+  arrayedProperties,
+  requiredPropertyObjects
```
evantahler: Let's type these!

plugins/@grouparoo/calculated-property/src/lib/test.ts
```diff
@@ -0,0 +1,5 @@
+import { TestPluginMethod } from "@grouparoo/core";
+
+export const test: TestPluginMethod = async () => {
+  return { success: true };
```
evantahler: This is probably always true.  Perhaps it might make sense to try out `vm2` here and run a `() => {return true}` or something like that?

plugins/@grouparoo/csv/src/initializers/plugin.ts
```diff
@@ -23,7 +23,6 @@ import { profileProperties as remoteProfileProperties } from "../lib/remote-impo
 
 import {
   CSVAppTemplate,
-  CSVFileSourceTemplate,
```
evantahler, line 26: Good find!

plugins/@grouparoo/calculated-property/src/lib/propertyOptions.ts
```diff
@@ -0,0 +1,24 @@
...
+export const propertyOptions: PropertyOptionsMethod = async () => [
+  {
+    key: "requiredProperties",
+    required: true,
+    description:
+      "Grouparoo Property IDs for all properties used in customFunction",
+    type: "lix  st",
+    options: async () => {
+      return undefined;
+    },
```
evantahler: This might require some updates in core to sort out

evantahler (commented):
Nice work so far! Adding some comments for later today

plugins/@grouparoo/calculated-property/public/templates/property/{{{id}}}.js.template
teallarson: resolved by using mustache variables

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
teallarson: Resolved by using mustache vars instead.

plugins/@grouparoo/calculated-property/src/lib/propertyOptions.ts
teallarson: Is this referring to the type typo?

evantahler: ... probably! I think it's already fixed up

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
teallarson: removed this function

teallarson:
- topological sort with mustache vars works across plugins now
- completed unit testing
- vm executes the function now instead of just compiling and exporting it

core/src/initializers/codeConfig.ts
```diff
@@ -16,7 +16,7 @@ export class CodeConfig extends CLSInitializer {
     super();
     this.name = "codeConfig";
     this.loadPriority = 10000;
-    this.startPriority = 1;
+    this.startPriority = 100;
```
evantahler: Can you please remind me why we made this change?

core/src/classes/codeConfig.ts
```diff
@@ -361,20 +361,26 @@ export async function getParentIds(
...
-      ...mustachePrerequisiteIds.map((p) => `property:${p}`)
-    );
+  // - property with mustache dependency
+
+  if (configObject["options"]) {
+    for (const [k, v] of Object.entries(configObject["options"])) {
+      if (
+        cleanClass(configObject) === "property" &&
+        typeof v === "string" &&
+        v.includes("{{") &&
```
evantahler, line 370: I think this is a good enough check, but we may want to try/catch around `MustacheUtils.getMustacheVariablesAsPropertyIds()` in the future

core/src/models/Property.ts
```diff
@@ -88,6 +88,38 @@ const STATE_TRANSITIONS = [
...
+  }) => Promise<
+    Array<{
+      key: string;
+      description?: string;
+      default?: boolean;
+      examples?: Array<any>;
+    }>
+  >;
+}
+
```
evantahler: This shouldn't be here... it's now in `core/src/class/plugin` 

https://github.com/grouparoo/grouparoo/blob/main/core/src/classes/plugin.ts#L487-L517

core/src/modules/plugin.ts
```diff
@@ -350,12 +350,15 @@ export namespace plugin {
    */
   export async function replaceTemplateProfileVariables(
     string: string,
-    profile: Profile
+    profile: Profile,
+    strict: Boolean = true
```
evantahler: I think it's lower-case boolean, and you don't need it! The type is determened automatically from the default value you added:

```suggestion
strict = true
```

core/src/modules/plugin.ts
```diff
@@ -350,12 +350,15 @@ export namespace plugin {
...
   ): Promise<string> {
     if (string.indexOf("{{") < 0) return string;
 
     const data = await getProfileData(profile);
-    return MustacheUtils.strictlyRender(string, data);
+    console.log(JSON.stringify(profile));
```
evantahler: to remove

core/src/utils/pluginDetails.js
```diff
@@ -109,7 +110,13 @@ function getPluginManifest() {
       }
     }
 
-    if (!fs.existsSync(pluginPath)) continue;
+    if (!fs.existsSync(pluginPath)) {
+      if (!manifest.missingPlugins.includes(pluginName)) {
+        manifest.missingPlugins.push(pluginName);
```
evantahler, line 115: 👍  Then this is logged in `core/src/config/plugins.ts`

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
```diff
@@ -0,0 +1,58 @@
...
+    console.log(error);
+    return undefined;
+  }
+  let calculatedPropertyValue;
+
+  //this works with vm and node vm... using node vm to allow for node modules in v2 of this feature
+  const vm = new NodeVM({
+    console: "inherit",
+    sandbox: {},
+  });
```
evantahler: Please be sure to also include:
* `argv: []`
* `env: {}`

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  try {
+    calculatedPropertyValue = vm.run(
+      `const toRun = ${populatedFunction}; module.exports = toRun();`
+    );
```
evantahler, line 48: 👍 the method is being defined and run in the VM, not our 'real' code

core/src/initializers/codeConfig.ts
teallarson: There was a sequelize instance error because a previous iteration of the `calculated-properties` plugin needed access to one of the models.  It's no longer needed since the switch to mustache variables.  Thanks!

core/src/models/Property.ts
teallarson: Ooh!  Yep.  Is just removing these lines and committing a way to resolve that without tangling things up more between branches?

core/src/classes/codeConfig.ts
```diff
@@ -360,20 +360,26 @@ export async function getParentIds(
...
-  ) {
-    const mustachePrerequisiteIds =
-      await MustacheUtils.getMustacheVariablesAsPropertyIds(
-        configObject["options"]["query"],
-        otherConfigObjects
-      );
-    prerequisiteIds.push(
-      ...mustachePrerequisiteIds.map((p) => `property:${p}`)
-    );
+  // - property with mustache dependency
```
teallarson, line 363: now will sort ANY property with a mustache dependency, not just query source properties

core/src/modules/plugin.ts
```diff
@@ -350,12 +350,14 @@ export namespace plugin {
...
-    profile: Profile
+    profile: Profile,
+    strict = true
   ): Promise<string> {
     if (string.indexOf("{{") < 0) return string;
 
     const data = await getProfileData(profile);
-    return MustacheUtils.strictlyRender(string, data);
+    if (strict === true) return MustacheUtils.strictlyRender(string, data);
+    return MustacheUtils.render(string, data);
```
teallarson, line 360: This allows non-strict renders for things like returning mustached properties for calculated properties.  That means someone can now do things like calculate `if ("{{last_purchase_category}}" !== ""){...}`

plugins/@grouparoo/calculated-property/public/templates/property/{{{id}}}.js.template
```diff
@@ -0,0 +1,23 @@
...
+      isArray: false, // Is this an Array Property?
+      options: {
+        customFunction: customFunction // Write your customFunction below.  This will tell Grouparoo to execute it.
+      },
+      filters: null,
+    },
+  ];
+}
+
+function customFunction () {
```
teallarson, line 20: The function is written outside of the config object to allow for helpful, human-friendly formatting and is then parsed to a string in the template for the plugin to read.

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
```diff
@@ -0,0 +1,60 @@
...
+  profile: Profile
+): Promise<string> {
+  let populatedFunction;
+  try {
+    populatedFunction = await plugin.replaceTemplateProfileVariables(
+      customFunction,
+      profile,
+      false
+    );
+    //returns a string of the entire function
```
teallarson, line 29: At this point, the function will be a string with all the variables replaced by their value.  "False" here allows for standard rendering so we can return empty strings for undefined variables.

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  let calculatedPropertyValue;
+
+  //this works with vm and node vm... using node vm to allow for node modules in v2 of this feature
+  const vm = new NodeVM({
+    console: "inherit",
+    sandbox: {},
+    argv: [],
+    env: {},
+  });
```
teallarson, line 43: We're allowing console usage, but keeping everything else locked down.  External libraries cannot be imported and neither can env variables.

plugins/@grouparoo/calculated-property/__tests__/import/import-property.ts
```diff
@@ -0,0 +1,116 @@
...
+        return false;
+    }`;
+    const value = await getPropertyValue(fn);
+    expect(value[0]).toBeTruthy;
+  });
+
+  test("it evaluates date strings as expected", async () => {
+    const fn = `() => {
+          const date = new Date("{{lastLoginAt.iso}}");
+          return date.toISOString();
```
teallarson, line 90: important to note in docs that we return date properties as an object with multiple date formats!

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
```diff
@@ -0,0 +1,68 @@
...
+  propertyName: string
+): Promise<string> {
+  let populatedFunction;
+  try {
+    populatedFunction = await plugin.replaceTemplateProfileVariables(
+      customFunction,
+      profile,
+      false
+    );
+    //fail at every level if someone tries to require a library... this should never be allowed to hit vm.run
```
tealjulia, line 21: processing the property in a run and allowing `vm.run` would fail if someone tried to require a library... since that would be an invalid config, we should let a user know before that point that their code is invalid.

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
teallarson: We also want to make sure that users aren't attempting to require any libraries or access env variables... functions doing this would fail when executed, but we want them to fail before even getting to that point.

core/src/utils/pluginDetails.js
```diff
@@ -55,6 +55,7 @@ function getPluginManifest() {
       grouparoo: { plugins: [] },
     },
     plugins: [],
+    missingPlugins: [],
```
teallarson, line 58: for logging purposes

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
```diff
@@ -0,0 +1,72 @@
...
```
teallarson, line 22: This means that running `roo validate` or `roo apply` will fail.  It fails a bit more abruptly than with other config errors, but I think that's warranted.  We don't want anything questionable attempting to be executed.  This is one area not covered in our test suite because I couldn't get it to test without throwing the error straight to the console.  Open to suggestions there or leaving as is!

teallarson:
Known limitation: if I generate a property, then `roo run`, change the function, then `roo run` again, it will appear to be stuck (does stuff, but # of profiles and imports doesn't change) for a decent chunk of time (>1-2 mins) then eventually will run and recalculate.  This is using our 1000 sample profiles, so could be very slow with larger data sets if someone changed their config.

evantahler:
> Known limitation: if I generate a property, then `roo run`, change the function, then `roo run` again, it will appear to be stuck (does stuff, but # of profiles and imports doesn't change) for a decent chunk of time (>1-2 mins) then eventually will run and recalculate. This is using our 1000 sample profiles, so could be very slow with larger data sets if someone changed their config.

That's interesting! Can you please write this up?  I wonder if it works that way for all changed properties (and not just this new plugin).  A changed Property does make an internal run...

plugins/@grouparoo/calculated-property/__tests__/import/import-property.ts
```diff
@@ -0,0 +1,116 @@
...
+    const value = await getPropertyValue(fn);
+    expect(value[0]).toBeTruthy;
+  });
+  test("it evaluates boolean properties as expected", async () => {
+    const fn = `() => {
+        if({{isVIP}} === true) return true;
+        return false;
+    }`;
+    const value = await getPropertyValue(fn);
+    expect(value[0]).toBeTruthy;
```
evantahler: Can you check `.toEqual(true)` here?  Any string or number is also truthy, and we want to make sure it's a boolean here

plugins/@grouparoo/calculated-property/__tests__/import/import-property.ts
```diff
@@ -0,0 +1,116 @@
...
+  });
+  test("it throws if mustached property does not exist", async () => {
+    const fn = `() => {
+      return {{fakeProperty}}
+    }`;
+    await expect(getPropertyValue(fn)).rejects.toThrowError(
+      "Could not calculate property: Error: Calculated property's /`customFunction/` undefined"
+    );
+  });
+});
```
evantahler, line 146: Additional tests:

A throwing function:
```js
const fn = `() => { throw new Error('oh no') }`;
```

An invalid function 
```js
const fn = `() => { not;valid-----js + a}`;
```

No function 

```js
const fs = undefined;
```

An Async function

```js
const fn = `async () => { return new Promise((resolve) => resolve("value")); }`
```

I think you can also test invalid functions and check that they throw here: 

```js
const fn = `() => { require('foo')}`;
```

teallarson:
> > Known limitation: if I generate a property, then `roo run`, change the function, then `roo run` again, it will appear to be stuck (does stuff, but # of profiles and imports doesn't change) for a decent chunk of time (>1-2 mins) then eventually will run and recalculate. This is using our 1000 sample profiles, so could be very slow with larger data sets if someone changed their config.
> 
> That's interesting! Can you please write this up? I wonder if it works that way for all changed properties (and not just this new plugin). A changed Property does make an internal run...

Hmm.  Yeah, it's not a calculated properties only thing!  Adding to tracker.

plugins/@grouparoo/calculated-property/public/templates/property/{{{id}}}.js.template
evantahler: 👍 

We'll probably add a link to the docs here once they are ready too

plugins/@grouparoo/calculated-property/src/lib/propertyOptions.ts
```diff
@@ -0,0 +1,14 @@
+import { Property, PropertyOptionsMethod } from "@grouparoo/core";
+
+export const propertyOptions: PropertyOptionsMethod = async () => [
+  {
+    key: "customFunction",
+    required: true,
+    description:
+      "Javascript function to transform existing Property(ies) to create a new one",
+    type: "text",
```
evantahler: I think `textarea` might render better in the UI

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
evantahler: 👍

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
evantahler: I think this can be tested! Wound't this throw in the way you are already testing in `__tests__/import/import-property.ts ` ?

evantahler: To confirm - this check is to "fail fast" with `grouparoo validate` if you try to require something or read process.env.  Otherwise, the function would be valid, but produce results you don't expect as we block / null out those at runtime.



We can test that `process.env` is empty with a test like 

```js
const fn = `() => { return process.env.DATABASE_URL }`; 
```

and show that it's `undefined` which might be a good idea.

evantahler (changes requested):
This is looking great! I think there are few things more to test and a little cleanup... and then this is ready to be merged 🎉!

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
teallarson: It wasn't throwing the exact same way, but I've changed it to now

plugins/@grouparoo/calculated-property/src/lib/profileProperty.ts
```diff
@@ -20,7 +20,7 @@ async function calculateProfilePropertyValue(
     );
 
     //fail at every level if someone tries to require a library... this should never be allowed to hit vm.run
-    const illegalStrings = [`require(`, `process.env`];
+    const illegalStrings = [`require(`, `process.env`, `async`];
```
evantahler, line 23: lol, that's one way to do it!

We will need to get async functions working in the next version of this (when we want to support fetch, etc) but I guess we don't need it now

evantahler (approved):

PR #2173, opened 2021-08-17
Use unique Profiles IDs when processing `preparePropertyImports`

evantahler:


core/src/modules/ops/profileProperty.ts
```diff
@@ -182,6 +182,10 @@ async function preparePropertyImports(
   if (pendingProfileProperties.length === 0) return;
   if (properties.length === 0) return;
 
+  const uniqueProfileIds = pendingProfileProperties
```
bleonard, line 185: Could you do a `DISTINCT` higher up or does that not work with the `JOIN` situation?

evantahler: no... because we are actually iterating over the profile properties themselves (not profiles) - depending on how many exact profile properties we can grab at once, we could get a few: `user-1 + email`, `user-1 + firstName`, etc

PR #2178, opened 2021-08-18
`profiles:confirm` does not create exports when not necessary

pedroslopez:
Before this PR, confirming profiles would create an `Import` (and later an `Export`) for each profile that was being checked on.

This is a refactoring of the implementation which instead only marks the profile and its `directlyMapped` property as pending without creating an import. Then, if the `directlyMapped` property is `null`, `profile:destroy` is enqueued for the profile that will take care of removing it from groups and making the final export.

It now follows a more similar path to the other models, where the recurring system `destroy` task picks them up and enqueues the related model destroy task.

This new implementation also fixes an issue where the profiles were not properly removed from groups before being deleted. Because the profiles were only being imported and group removal was implicit based on new properties, if the profile was in a calculated group that matched on an inverse rule (e.g. `userId is not 54`), the group was not removed and the final export might not have been a `toDelete`. Now group removal is done explicitly.


core/src/tasks/profile/destroy.ts
```diff
@@ -0,0 +1,42 @@
...
+      },
+    });
+    if (pendingExports > 0) return;
+
+    const oldGroups = await profile.$get("groups");
+    if (oldGroups.length > 0) {
+      // clear groups and export
+      // when the export is done, this task will be enqueued again to destroy it
+      await Profile.destroyGroupMembers(profile);
+      await profile.export(false, oldGroups, true, false);
```
evantahler, line 36: It's good this happens in a transaction - otherwise we could end up in a loop where we keep exporting the profile

evantahler: The last argument to `profile.export()` is kind of interesting here - since this task is only considering one profile at a time, you could do the export `sync=true` in this task maybe?

core/src/modules/ops/profile.ts
```diff
@@ -729,61 +724,37 @@ export namespace ProfileOps {
...
-    await markPendingByIds(profiles.map((p) => p.id));
+    // Only mark profile and directlyMapped property pending
+    await markPendingByIds(profileIds, false);
+    await ProfileProperty.update(
+      { state: "pending", startedAt: null },
+      { where: { id: profileProperties.map((pp) => pp.id) } }
+    );
 
-    return profiles.length;
+    return profileProperties.length;
```
evantahler: The return here used to be `profiles.length` - this is equivalent now only if there is one directlyMapper property.  Perhaps `const profileIds = profileProperties.map((pp) => pp.profileId);` above should be uniqe'd and then counted here?

evantahler (commented):
This is really good!  Just some minor stuff below.

The way the tasks chain up now, it is possible if there's a problem with the final export, the `profile:destroy` task will keep finding the same N profiles to destroy, enqueueing the task, and bailing because they have pending exports.  Perhaps `profile.export()` should be done synchronously in `profile:destroy`?

core/src/tasks/profile/destroy.ts
pedroslopez: Hmm- what would happen if the export takes longer than the `destroy` task frequency and the `profile:destroy` task for this profile is enqueued to run? Would the transaction make the task wait or would it still bail because there's an active export?

Also, a couple differences come to mind between sync/async exporting:
- Export processors currently aren't handled synchronously (there's a story for that in tracker). Using sync exports in this case would still work, but the export processors would be handled async which is a little odd.
- Sync exports throw errors directly, while async auto retries up to whatever limit is set up. 

I'm not sure what the right thing to do would be if this export fails up to a retry limit. Should we delete the profile from Grouparoo anyway? Should we keep retrying? How can we communicate this? Maybe having the `profile:destroy` task fail and become a resque error is clearer as to what happened, but the recurring `destroy` task might keep trying to delete it so something would probably need to be done there as well.

evantahler: I think it is OK to leave this as-is for now, and we can deal with the edge-cases in another story/PR.  I had forgotten about that we can't do sync async exports.

I believe that as currently implemented, if an export fails, it will end up in resque as an error, and then the Profile would get deleted from Grouparoo anyway (as the blocking export would move to the `failed` state).  At least having a Resque failure lets the user know they should double-check the destination.

evantahler (approved):

PR #2179, opened 2021-08-18
Provide types to Property Options via `PluginOptionType`

evantahler:
Provides better types for the Options that plugins can provide when declaring their AppOptions, SourceOptions, PropertyOptions, and DestinationOptions.  These types are used to help display and edit these options in the UIs.

```ts
export type PluginOptionType =
  | "list"
  | "typeahead"
  | "text"
  | "textarea"
  | "password"
  | "pending";
```

So you you'll get an error if you try to add an option type that isn't supported
Screen Shot 2021-08-18 at 9 29 56 AM

### DIFF
--- a/core/src/classes/plugin.ts
+++ b/core/src/classes/plugin.ts
@@ -1,4 +1,4 @@
-import { App, AppOption, SimpleAppOptions } from "../models/App";
+import { App, SimpleAppOptions } from "../models/App";
 import { Source, SimpleSourceOptions, SourceMapping } from "../models/Source";
 import { Errors } from "../modules/errors";
 import {
@@ -9,7 +9,6 @@ import {
 } from "../models/Destination";
 import { Run } from "../models/Run";
 import {
-  PluginConnectionPropertyOption,
   SimplePropertyOptions,
   PropertyFiltersWithKey,
 } from "../models/Property";
@@ -50,6 +49,16 @@ export interface PluginApp {
     appOptions?: AppOptionsMethod;
     parallelism?: AppParallelismMethod;
   };
+}
+
+export interface AppOption {
+  key: string;
+  type?: PluginOptionType;
+  displayName?: string;
+  required: boolean;
+  description?: string;
+  placeholder?: string;
+  defaultValue?: string | number | boolean;
 }
 
 /**
@@ -107,6 +116,14 @@ export enum FilterOperation {
   NotContain = "does not contain",
   In = "in",
 }
+
+export type PluginOptionType =
+  | "list"
+  | "typeahead"
+  | "text"
+  | "textarea"
+  | "password"
+  | "pending";
 
 /**
  * Method to get one or many profiles to be saved/updated.
@@ -298,7 +315,7 @@ export interface ConnectionOption extends AppOption {}
 export interface AppOptionsMethod {
   (): Promise<{
     [optionName: string]: {
-      type: PluginOptionTypes;
+      type: PluginOptionType;
       options?: string[];
       descriptions?: string[];
     };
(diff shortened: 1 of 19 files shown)


ui/ui-components/pages/source/[id]/edit.tsx
```diff
@@ -152,7 +152,7 @@ export default function Page(props) {
   // not every row returned is guaranteed to have the same columns
   const previewColumns = preview
     .map((row) => Object.keys(row))
-    .flat()
+    .reduce((acc, val) => acc.concat(val), [])
```
Comment (evantahler) [inline ui/ui-components/pages/source/[id]/edit.tsx:155]: Unrelated, but `.flat()` isn't supported in our ES version... this is a polyfill from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

core/src/classes/plugin.ts
```diff
@@ -573,13 +622,6 @@ export interface ExportArrayPropertiesMethod {
 
 export type ExportArrayPropertiesMethodResponse = Array<string>;
 
-export type PluginOptionTypes =
```
teallarson, line 576: a ha!  that's where it was!

core/src/models/App.ts
```diff
@@ -24,19 +24,8 @@ import { AppOps } from "../modules/ops/app";
 import { LockableHelper } from "../modules/lockableHelper";
 import { ConfigWriter } from "../modules/configWriter";
 import { APIData } from "../modules/apiData";
-import { PluginOptionTypes } from "../classes/plugin";
 import { AppConfigurationObject } from "../classes/codeConfig";
 
-export interface AppOption {
```
tealjulia, line 30: out of curiosity, what was the reason for moving this to the plugin vs on the app model?

core/src/models/Property.ts
```diff
@@ -88,38 +88,6 @@ const STATE_TRANSITIONS = [
   },
 ];
 
-/**
- * Metadata and methods to return the options a Property for this connection/app.
- * Options is a method which will poll the source for available options to select (ie: names of tables or columns)
- */
-export interface PluginConnectionPropertyOption {
```
tealjulia, line 95: makes sense to move this to the plugin class instead.  nice!

plugins/@grouparoo/csv/src/lib/remote-import/sourceOptions.ts
```diff
@@ -5,7 +5,7 @@ import {
 
 export const sourceOptions: SourceOptionsMethod = async () => {
   const response: SourceOptionsMethodResponse = {
-    url: { type: "string", options: [], descriptions: [] },
+    url: { type: "text", options: [], descriptions: [] },
```
teallarson, line 8: i'm glad we added those types!  lots of mislabeled "text" items.

teallarson (approved):
Looks good!  

I notice the AppOption and PropertyOptions got moved to the classes instead of the model.  What is the reasoning behind keeping the SimpleAppOption and SimplePropertyOptions where they are?

evantahler (commented):
`AppOption` is badly named - there are still `SimpleAppOptions` (which mirror `SimplePropertyOptions` and `SimpleSourceOptions`) which I think belong with the Model, as the Model uses those types.  What's called `AppOption` is probably better named something like `AppOptionsOption`

teallarson:
> `AppOption` is badly named - there are still `SimpleAppOptions` (which mirror `SimplePropertyOptions` and `SimpleSourceOptions`) which I think belong with the Model, as the <odel uses those types. What's called `AppOption` is probably better named something like `AppOptionsOption`

Ah ha!  Yep I think I've ended up lost in this part of the monorepo before.  That is a clearer name!

PR #2184, opened 2021-08-18
Better Action input Formatting

evantahler:
This PR introduces a number of Action input formatters to help with non-string types: Booleans, Number, and nested JSON objects.  This allows the Grouparoo API to be accessed by simple CURL methods:

```bash
curl 
  -X PUT 
  -d 'apiKey=xxxxxxxxxxxxxxx' 
  -d 'options={"table":"users"}' 
  -d 'mapping={"id":"userId"}' 
  -d 'state=ready' 
'http://localhost:3000/api/v1/source/users_table'
```

### DIFF
--- a/core/src/actions/destinations.ts
+++ b/core/src/actions/destinations.ts
@@ -12,6 +12,7 @@ import { OptionHelper } from "../modules/optionHelper";
 import { destinationTypeConversions } from "../modules/destinationTypeConversions";
 import { AsyncReturnType } from "type-fest";
 import { ConfigWriter } from "../modules/configWriter";
+import { APIData } from "../modules/apiData";
 
 export class DestinationsList extends AuthenticatedAction {
   constructor() {
@@ -21,11 +22,12 @@ export class DestinationsList extends AuthenticatedAction {
     this.outputExample = {};
     this.permission = { topic: "destination", mode: "read" };
     this.inputs = {
-      limit: { required: true, default: 100, formatter: parseInt },
-      offset: { required: true, default: 0, formatter: parseInt },
+      limit: { required: true, default: 100, formatter: APIData.ensureNumber },
+      offset: { required: true, default: 0, formatter: APIData.ensureNumber },
       state: { required: false },
       order: {
         required: false,
+        formatter: APIData.ensureObject,
         default: [
           ["name", "desc"],
           ["createdAt", "desc"],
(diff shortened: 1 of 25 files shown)


core/src/actions/destinations.ts
```diff
@@ -215,19 +221,15 @@ export class DestinationConnectionOptions extends AuthenticatedAction {
...
     };
   }
 
   async runWithinTransaction({ params }) {
     const destination = await Destination.findById(params.id);
-
-    const options =
-      typeof params.options === "string"
-        ? JSON.parse(params.options)
-        : params.options;
```
evantahler, line 228: Previously, this type of JSON parsing logic was inconsistently scattered around the codebase.  Now our formatters do the work for us!

core/src/actions/sources.ts
```diff
@@ -163,8 +165,8 @@ export class SourceEdit extends AuthenticatedAction {
       name: { required: false },
       type: { required: false },
       state: { required: false },
-      options: { required: false },
-      mapping: { required: false },
+      options: { required: false, formatter: APIData.ensureObject },
+      mapping: { required: false, formatter: APIData.ensureObject },
```
evantahler, line 169: @tealjulia here's why we couldn't set the Options for a Source via CURL

core/src/modules/apiData.ts
```diff
@@ -1,4 +1,51 @@
...
+    if (param === true || param === false) return param;
+    if (param === 1 || param === "1" || param === "true") return true;
+    if (param === 0 || param === "0" || param === "false") return false;
+    throw new Error(`${param} cannot be converted to a boolean`);
+  }
+
+  export function ensureNumber(param: string | number) {
+    if (typeof param === "number") return param;
+    try {
+      return parseFloat(param);
```
teallarson: I noticed our original validators used `parseInt`.   What happens if someone puts in a decimal for an offset/limit (the two main places this is being used)?  I think it would pass this check but return an error?

teallarson (approved):
Nice!  I did have one question on the use of parseFloat vs parseInt.

core/src/modules/apiData.ts
evantahler: Good question! 

I made the parser look for floats rather than integer because I wanted to reflect what the user had intended at this stage.  If limit/offset fails because the user sent a decimal, that would throw an exception elsewhere - probably by Seqeulzie

Also... it turns out decimal limits/offsets are valid in postgres maybe?
Screen Shot 2021-08-19 at 5 27 27 PM

PR #2185, opened 2021-08-19
`grouparoo apply` also migrates (and rolls back if `apply` fails)

evantahler:
We now conditionally choose to run migrations based on if we are running `grouparoo apply` or not.   To get this working, all the transactions were removed from our migrations and applied by the initializer. 

`grouparoo apply` now runs the migrator within the same transaction as the config apply logic.  This means that if the config apply fails, any pending migrations will also be rolled back.  This allows the `grouparoo apply` command to be used as a blocking step to check if your config can be applied to a new version of Grouparoo as part of the upgrade process. 

This PR also moves any initializer work that may modify data (Teams and Settings) into the `start` lifecycle, not `initialize`.

Lots of changes in `ah-sequelzie-plugin` to make this happen! https://github.com/actionhero/ah-sequelize-plugin/pull/346


evantahler:
@pedroslopez I'm pretty happy with this work.   I'm going to look into what it would mean to port this back to `ah-sequelize-initializer` - ideally, all migrations would be automatically placed in a transaction like the PR does.

Update -> https://github.com/actionhero/ah-sequelize-plugin/pull/346

core/src/config/sequelize.ts
```diff
@@ -110,7 +110,8 @@ export const DEFAULT = {
     return {
       _toExpand: false,
       logging,
-      autoMigrate: true,
+      autoMigrate:
+        process.env.GROUPAROO_RUN_MODE === "cli:apply" ? false : true,
```
pedroslopez, line 114: aha! so we did end up keeping `autoMigrate` for non-apply 👍🏼

core/src/migrations/000001-createApps.ts
```diff
@@ -1,52 +1,53 @@
+import Sequelzie from "sequelize";
```
pedroslopez: Typo!

```suggestion
import Sequelize from "sequelize";
```

core/src/migrations/000001-createApps.ts
```diff
@@ -1,52 +1,53 @@
...
+
 export default {
-  up: async function (migration, DataTypes) {
-    await migration.sequelize.transaction(async () => {
```
pedroslopez, line 3: we're no longer wrapping each migration in a transaction- `ah-sequelize-plugin` now wraps it all in a single transaction.

https://github.com/actionhero/ah-sequelize-plugin/blob/512b20834be56de421330ddf1cfe691820c2b038/src/initializers/sequelize.ts#L25-L32

core/src/migrations/000002-createDestinationGroups.ts
```diff
@@ -1,47 +1,48 @@
-export default {
-  up: async function (migration, DataTypes) {
-    await migration.sequelize.transaction(async () => {
-      await migration.createTable("destinationGroups", {
-        guid: {
-          type: DataTypes.STRING(40),
-          primaryKey: true,
-        },
+import Sequelzie from "sequelize";
```
pedroslopez:
```suggestion
import Sequelize from "sequelize";
```

looks like the typo is in all migrations :/

plugins/@grouparoo/zendesk/package.json
```diff
@@ -42,6 +42,7 @@
     "jest": "27.0.6",
     "nock": "13.1.1",
     "prettier": "2.3.2",
+    "sequelize": "6.6.2",
```
pedroslopez, line 45: Looks like this is required to load types 👍🏼

pedroslopez (approved):
Looks great! Just looks like there are some typos in the migrations. I'm glad the actual changes to core after changing some things around in `ah-sequelize-plugin` weren't so bad.

core/src/migrations/000001-createApps.ts
evantahler: Yeah - It's a breaking change I guess... but I can't really think of any reason you wouldn't want your migrations within a transaction.

evantahler:
Typos fixed up in https://github.com/grouparoo/grouparoo/pull/2185/commits/73b14bf6373ae4a4c9b2d04d260ef04999d65d78

PR #2189, opened 2021-08-19
Mixpanel batch and single export hybrid integration

pauloouriques:
- Add waitUntil function to specHelper.


plugins/@grouparoo/mixpanel/__tests__/export/export-profiles.ts
```diff
@@ -0,0 +1,958 @@
...
+
+// these comments to use nock
+const newNock = false;
+require("./../fixtures/export-profiles");
+// or these to make it true
+// const newNock = true;
+// helper.recordNock(nockFile, updater);
+
+const appOptions = loadAppOptions(newNock);
+const appId = "fgQIQXZKhkKQMmj7TeZjCDWB9+ijp4aAUA3039==";
```
evantahler, line 20: Ha! Is this a hash?

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -145,6 +145,16 @@ export namespace helper {
...
 
+  export async function waitUntil(newNock, guard) {
+    let attempts = 0;
+    while (!(await guard()) && attempts < 10) {
+      attempts++;
+      if (newNock) {
+        await this.sleep(shortTime * 3);
+      }
+    }
+  }
```
evantahler, line 156: Nice! This is really helpful the way you implemented it for nock requests.  I was imagining something more generic that had an expectation (`expect(bla).toBe(true)`) passed in that would keep retrying until it passed, but I think this is better suited for the nock use-case.

plugins/@grouparoo/mixpanel/src/lib/export/exportProfiles.ts
```diff
@@ -0,0 +1,252 @@
...
+  BatchMethodFindAndSetDestinationIds,
+  BatchMethodGetClient,
+  BatchMethodNormalizeForeignKeyValue,
+  BatchMethodNormalizeGroupName,
+  BatchMethodRemoveFromGroups,
+  BatchMethodUpdateByDestinationIds,
+  BatchSyncMode,
+  buildBatchExports,
+  exportProfilesInBatch,
+} from "@grouparoo/app-templates/dist/destination/batch";
```
evantahler, line 16: 👍 I'm glad the batchHelpers are helpful!

evantahler (approved):
Looking Good!  I'll merge this in so we can start acceptance!

PR #2190, opened 2021-08-19
Non-unique mapping returns `undefined` when a match doesn't exist

pedroslopez:
For unique mappings, no value (`undefined`) is returned for properties of non-existent profiles. `keepValueIfNotFound` relies on this behavior to know when a value was set or not for a property and either set the value to `null` or keep the old value. 

This PR changes non-unique mappings to exhibit the same behavior so `keepValueIfNotFound` works the same way. Because we're still setting the value to `null` if needed, there shouldn't be an issue with returning `undefined` at this step.

### DIFF
--- a/core/__tests__/models/source/source.ts
+++ b/core/__tests__/models/source/source.ts
@@ -1035,7 +1035,7 @@ describe("models/source", () => {
         });
       });
 
-      test("it returns null when the source does not have a record for the property", async () => {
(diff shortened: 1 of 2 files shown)


evantahler (approved):
👍 from me!  I don't think there was a special reason about why this was returning `[null]` rather than undefined.

PR #2192, opened 2021-08-19
Prevent data in Secondary Sources from Creating Profiles that do not exist in the Primary Sources

evantahler:
If a Schedule creates an Import for a profile we don't have, and the Schedule's Source doesn't own the directly mapped Property, we no longer create a new profile. This is accomplished by checking the source which created the Run, and checking if that source owns the unique Property which would be used to make the new Profile.  If the Source does not own that unique Property, we add an error to the Import.

This error is not surfaced in Resque, as it may happen multiple times per Run. E.g. if there are 10 purchases for a non-existant profile, we don't want 10 resque errors.  We do continue to store the Import until it would be swept away in the normal way. 

Screen Shot 2021-08-19 at 3 52 36 PM


core/src/models/Run.ts
```diff
@@ -64,8 +72,8 @@ export class Run extends Model {
   @Column
   creatorId: string;
 
-  @Column
-  creatorType: string;
+  @Column(DataType.ENUM(...RUN_CREATORS))
+  creatorType: typeof RUN_CREATORS[number];
```
pedroslopez, line 76: more typings! :D

core/src/modules/ops/profile.ts
```diff
@@ -593,10 +593,15 @@ export namespace ProfileOps {
   /**
    * The method you'll be using to create profiles with arbitrary data.
    * Hash looks like {email: "person@example.com", id: 123}
+   *
+   * This method today always returns a profile by finding it or making a a new one... is that right?.
```
pedroslopez: I believe that's right! Though now it can also throw an error if the schedule's source doesn't own the unique property being set

core/src/modules/ops/import.ts
```diff
@@ -66,11 +68,22 @@ export namespace ImportOps {
...
-    const {
-      profile,
-      isNew,
-      // will throw if there are no unique profile properties
-    } = await ProfileOps.findOrCreateByUniqueProfileProperties(_import.data);
+    let run: Run;
+    let schedule: Schedule;
+    let source: Source;
+
+    run = await Run.findOne({ where: { id: _import.creatorId } });
```
pedroslopez: We're probably ok since the other types aren't really used, but maybe also check that the import's creatorType refers to a run?

core/src/models/Import.ts
```diff
@@ -33,6 +33,8 @@ export interface ImportProfileProperties {
   [key: string]: any;
 }
 
+const IMPORT_CREATORS = ["schedule", "run", "api"] as const;
```
pedroslopez: Can schedules create imports directly? I believe they're always created through runs 🤔

Also, I know we have an action that creates imports with `creatorType` of `api`- mainly out of curiosity, what was the use case for that? I don't think I see it being called anywhere.

core/src/modules/ops/import.ts
```diff
@@ -66,11 +68,22 @@ export namespace ImportOps {
...
+    if (run && run.creatorType === "schedule")
+      schedule = await Schedule.findOne({ where: { id: run.creatorId } });
+    if (schedule)
+      source = await Source.findOne({ where: { id: schedule.sourceId } });
+
+    // will throw if there are no unique profile properties
+    const { profile, isNew } =
+      await ProfileOps.findOrCreateByUniqueProfileProperties(
+        _import.data,
+        source
```
pedroslopez, line 87: `source` will only be set when imports are created via a schedule's run, which will trigger verification inside `findOrCreateByUniqueProfileProperties`. Other non-schedule-created imports will still be able to create profiles at will, is that right?

Schedules should be the only type of run that creates profiles, so I think this is 👍🏼

pedroslopez (commented):
Nice! I think this is definitely the right thing to enforce and prevents all sorts of weird shenanigans. Attaching the error to the import and not clearing it right away is also nice, so users can see they have some sort of data integrity issue going on.

Just a couple non blocking comments, but otherwise looks good 👍🏼

core/src/modules/ops/import.ts
evantahler: It's the other way around actually - if you don't provide a Source (eg: the run is crated by an internal run) you cannot create a new Profile if you can't find one.  So this means that only Schedules can create Profiles now.

Some of the updates to the test suite reflect this.

core/src/models/Import.ts
evantahler: That's a very good point - Keeping that API around makes even less sense for our Data Engineer persona.  Removing it in https://github.com/grouparoo/grouparoo/pull/2209 (a PR into this PR).  The only `IMPORT_CREATORS ` will be `run`

core/src/modules/ops/import.ts
pedroslopez: Oh- that would make more sense, but it doesn't seem like that's what's happening.

I'm seeing the `prevents import when unable to create profile from secondary source` test for `associateProfile`, which tests imports created by a schedule from a different source don't create profiles. 

However, I'm not seeing any test for imports not created from a schedule (e.g. internal runs or group runs). From what I'm reading in the code, it seems like if no schedule is passed, `canCreateNewProfile` is set to `true`, so the profile will be created:

```js
const canCreateNewProfile = source
  ? (await Property.findAllWithCache())
      .filter((p) => p.unique === true && p.sourceId === source.id)
      .map((p) => p.key)
      .filter((key) => !!hash[key]).length > 0
  : true; // <- if source is falsy, canCreateNewProfile = true
```

maybe that final `true` should be `false`? In any case, I think we should add a test for the intended behavior when a source isn't passed.

evantahler: Whoops! I was wrong in my reply, and your first 👍 note was correct.  That's what the code and test say - that Profile creation by methods that don't have a source (eg: internal run) is still allowed.  I'm going to dig into why that is...

evantahler: Ok! https://github.com/grouparoo/grouparoo/pull/2192/commits/765d9708721745e83164408f7295a8d16eb169de should make things make a lot more sense!  

The spec helper `hepler.getProfile()` still needs to make profiles all willy-nilly, but now the rest of the app is locked down - you need a Source to check if you can make a new Profile.

PR #2206, opened 2021-08-23
Stricter types for Authentication

evantahler:


core/src/actions/session.ts
```diff
@@ -13,7 +13,6 @@ export class SessionCreate extends CLSAction {
       email: { required: true },
       password: { required: true },
     };
-    this.permission = { topic: "session", mode: "write" };
```
evantahler, line 16: This type shouldn't have existed - this action is not authenticated.

bleonard: It's possible this was here because it had a `write` as a side effect and the `mode: write` was needed for SQLite reasons.

bleonard: At the very least, you'll need to override `isWriteTransaction` to be true then.

evantahler: Ah ha! Fixed up in https://github.com/grouparoo/grouparoo/pull/2207

PR #2209, opened 2021-08-24
Remove `import:create` Action

evantahler:
This PR removes an un-used Aciton to create Imports (`POST /api/v1/import`).  This solidifies that only Runs (eg: Schedules + Sources) can create Imports... and therefore Profiles.


pedroslopez (approved):
Bye! 👋🏼

PR #2213, opened 2021-08-25
load config files from a `.tar.gz` archive

pedroslopez:
Adds the ability to load and automatically extract config files from a `.tar.gz` file.

- New env var `GROUPAROO_CONFIG_ARCHIVE` can be set to either a local or remote path. If the value starts with `https://` or `http://`, it will be considered remote and will be automatically downloaded on startup.
- When the above var is set, `GROUPAROO_CONFIG_DIR` controls which directory within the archive contains the config files. By default these are assumed to be in the `/config` folder.
- `GROUPAROO_CONFIG_DIR` can now be set to `false` to prevent processing config. This can be used to avoid deleting locked objects when the config directory has purposefully not been included along with the running instance.


core/__tests__/modules/codeConfig/codeConfig.ts
```diff
@@ -1559,4 +1588,90 @@ describe("modules/codeConfig", () => {
...
+      delete process.env.GROUPAROO_CONFIG_DIR;
+      delete process.env.GROUPAROO_CONFIG_ARCHIVE;
+    });
+
+    afterAll(() => {
+      process.env.GROUPAROO_CONFIG_DIR = previousConfigDir;
+      process.env.GROUPAROO_CONFIG_ARCHIVE = previousConfigArchive;
+    });
+
+    test("can get the config directory from local path", async () => {
```
evantahler: maybe called "absolute path"?

evantahler: Can a test for a relative path be added too?

core/src/bin/generate.ts
```diff
@@ -116,7 +116,13 @@ Commands:
...
     if (template) params.template = template;
     if (id) params.id = id;
-    params.path = getConfigDir();
+    params.path = await getConfigDir();
+
+    if (!params.path) {
+      return GrouparooCLI.logger.fatal(
+        `The config directory has been disabled, so config files cannot be generated`
+      );
+    }
```
evantahler: 👍

core/src/initializers/codeConfig.ts
```diff
@@ -26,7 +27,12 @@ export class CodeConfig extends CLSInitializer {
   }
 
   async startWithinTransaction() {
-    const configDir = getConfigDir();
+    const configDir = await getConfigDir();
+    if (process.env.GROUPAROO_RUN_MODE === "cli:config" && !configDir) {
+      return GrouparooCLI.logger.fatal(
```
evantahler: Within initializers, I think it might be better to just `throw` a new error - it's up to the CLI to then catch that error and format it / exit as needed

core/src/modules/configWriter.ts
```diff
@@ -204,7 +204,13 @@ export namespace ConfigWriter {
   }
 
   async function writeFile({ filePath, object }: WritableConfigObject) {
-    const configFilePath = path.join(getConfigDir(), filePath);
+    const configDir = await getConfigDir();
+    if (!configDir)
+      throw new Error(
+        "Cannot write config files because the config directory has been disabled. Make sure GROUPAROO_CONFIG_DIR is not set to `false`."
```
evantahler: This message is re-used quite a bit - want to save it somewhere?

core/src/utils/pluginDetails.js
```diff
@@ -38,9 +41,40 @@ function getParentPath() {
...
+  let configDir =
     process.env.GROUPAROO_CONFIG_DIR || path.join(getParentPath(), "config");
+
+  if (process.env.GROUPAROO_CONFIG_ARCHIVE) {
+    if (!extractedConfigDir) {
+      const workingDir = fs.mkdtempSync(path.join(os.tmpdir(), "grouparoo-"));
+
+      let archivePath = process.env.GROUPAROO_CONFIG_ARCHIVE;
+      if (archivePath.startsWith("http://" || "https://")) {
+        const res = await fetch(archivePath);
```
evantahler, line 85: Noting that we are now making external network requests based on an ENV variable - could be dangerous.  Maybe in the future we need to have an allow-list of domains... 

There are tons of things that can go wrong when downloading files... but I think since this is an async function, whatever goes wrong will be `thrown` and passed up properly.

core/src/utils/pluginDetails.js
```diff
@@ -38,9 +41,40 @@ function getParentPath() {
   return path.join(__dirname, "..", "..", "..", "..", "..");
 }
 
-function getConfigDir() {
-  const configDir =
+let extractedConfigDir;
+async function getConfigDir() {
```
evantahler: Utils/pluginDetails is getting pretty complex.  It is time to move this into a module and be made TS?  This was originally made as a JS file so it could be consumed by the UI or plugins, but we've since learned that importing `import x from "@grouparoo/core/dist/modules/plugins"` is possible

core/src/utils/pluginDetails.js
```diff
@@ -38,9 +41,40 @@ function getParentPath() {
...
+        await new Promise((resolve, reject) => {
+          res.body.pipe(fileStream);
+          res.body.on("error", reject);
+          fileStream.on("finish", resolve);
+        });
+      }
+
+      extractedConfigDir = path.join(workingDir, "extracted");
+      fs.mkdirSync(extractedConfigDir);
+      await tar.extract({ cwd: extractedConfigDir, file: archivePath });
```
evantahler, line 93: Are there any checks we should do on the files we've dowloaded?  Some ideas:
* they are only JS or JSON files
* They... don't have `eval()` in them?
* They... don't have `require()` in them?

core/__tests__/modules/codeConfig/codeConfig.ts
```diff
@@ -1559,4 +1588,90 @@ describe("modules/codeConfig", () => {
...
+          },
+          ["."]
+        );
+
+        expect(fs.existsSync(archivePath)).toBe(true);
+      });
+
+      test("config can be loaded from an archive", async () => {
+        process.env.GROUPAROO_CONFIG_ARCHIVE = archivePath;
+        process.env.GROUPAROO_CONFIG_DIR = "/";
```
evantahler: It took me a minute to realize that if you have `GROUPAROO_CONFIG_ARCHIVE`, `GROUPAROO_CONFIG_DIR` is within the unzipped archive.

evantahler (changes requested):
Nice work! This is what we need to download configs from S3!

I think it's time to move `utils/pluginDetails.js` into a "real" module so it can be typed.  I'm also curious what other forms of safety we can add around the downloaded config archive.  

Also, looks like `getConfigDir()` is used in the demo plugin too and needs to be made `async` too.

core/src/utils/pluginDetails.js
pedroslopez: hmm- is this any different from processing config files you already have? Presumably if you're setting this environment variable you have some level of authority to set these things. A concern could be something modifying the `process.env` and potentially pulling it in after startup, but config files are intentionally only downloaded the first time you invoke it.

core/src/initializers/codeConfig.ts
pedroslopez: I'm not sure how I would catch the error from the initializer...

The `config` command runs grouparoo like this:
```js
async run({ params }) {
    cli.GrouparooCLI.logCLI(this.name, false);
    const { main } = require("@grouparoo/core/dist/grouparoo");
    await main(); // run the server
```

and throwing from the initializer looks like this:
```
info: Using SQLite database: :memory: 
warning: Plugin @grouparoo/files-s3 requires environment variable S3_ACCESS_KEY to be set 
warning: Plugin @grouparoo/files-s3 requires environment variable S3_SECRET_ACCESS_KEY to be set 
warning: Plugin @grouparoo/files-s3 requires environment variable S3_REGION to be set 
warning: Plugin @grouparoo/files-s3 requires environment variable S3_BUCKET to be set 
warning: Plugin @grouparoo/sentry requires environment variable SENTRY_DSN to be set 
warning: Plugin @grouparoo/sentry requires environment variable SENTRY_TRACE_SAMPLE_RATE to be set 

🦘 Grouparoo: config

notice: Starting Grouparoo v0.5.3-alpha.5 on node.js v15.14.0 
notice: pid: 19387 
notice: environment: development 
info: *** Starting @grouparoo/core *** 
error: Error from Initializer name=codeConfig file=/Users/pedro/grouparoo/grouparoo/core/dist/initializers/codeConfig.js type=start stack=The config directory has been disabled. Make sure that the `GROUPAROO_CONFIG_DIR` environment variable is not set to `false`. 
notice: stopping process... 
notice: Stopping server: web 
notice: Stopping server: websocket 
notice: *** @grouparoo/core Stopped *** 
```

I wanted to be a bit clearer about what was going on by logging the fatal error as the last line since there are some other log statements in there and could be confusing, but I can't wrap `main()` with a try/catch- it looks like the error is caught before it gets to the command.

What do you think? Is throwing like this enough?

core/src/modules/pluginDetails.ts
```diff
@@ -38,18 +55,75 @@ function getParentPath() {
...
+let extractedConfigDir: string;
+export async function getConfigDir(throwIfDisabled: true): Promise<string>;
+export async function getConfigDir(
+  throwIfDisabled: false
+): Promise<string | false>;
+export async function getConfigDir(
+  throwIfDisabled: boolean
+): Promise<string | false>;
+export async function getConfigDir(): Promise<string | false>;
+export async function getConfigDir(throwIfDisabled?: boolean) {
```
pedroslopez, line 67: I think this is the first time we've used TS overlaods in this project! Looks a bit weird, but I wanted to use them here because the return type when `throwIfDisabled=true` should always be a valid string, so in other places where we want to throw or use the value we don't have to do unnecessary checks.

core/src/modules/pluginDetails.ts
```diff
@@ -38,18 +55,75 @@ function getParentPath() {
...
+  if (process.env.GROUPAROO_CONFIG_ARCHIVE) {
+    if (!extractedConfigDir) {
+      const workingDir = fs.mkdtempSync(path.join(os.tmpdir(), "grouparoo-"));
+
+      let archivePath = process.env.GROUPAROO_CONFIG_ARCHIVE;
+      if (archivePath.startsWith("http://" || "https://")) {
+        const res = await fetch(archivePath);
+        archivePath = path.join(workingDir, "grouparoo.tar.gz");
+        const buffer = await res.arrayBuffer();
+        fs.writeFileSync(archivePath, Buffer.from(buffer));
```
pedroslopez, line 88: I had to change the implementation when moving to TypeScript because technically the read stream returned in `res.body` by `isomorphic-fetch` is the Web API's ReadableStream and not Node's ReadStream (because isomorphic-fetch is designed to work in browser contexts as well). In reality, when running in NodeJS it does return node's ReadStream, which is why my previous implementation worked, but TypeScript was not happy with it because it's not supposed to work.

This new implementation has the downside of having to load the archive into memory when creating the arraybuffer. I can't seem to find a compatible way to just pipe the readable stream to a node writeable file stream 🤷🏼‍♂️ . Because the config archives should be small I don't think it's an issue, but something to keep in mind.

core/src/initializers/codeConfig.ts
evantahler: Yeah - I think throwing like that is OK - this isn't something we expect our users to do.  I'm worried that we won't shut down properly with `GrouparooCLI.logger.fatal` (running the stop initializers)

core/src/utils/pluginDetails.js
evantahler: That's a good point - I guess the right place to these checks would be on the file upload (cloud app) before telling grouparoo core to run them

core/src/modules/pluginDetails.ts
evantahler: Nice!  This clears up the repeated use of the error message too!

core/src/modules/pluginDetails.ts
evantahler: I think this is OK for now too.  This is another safety check we can add to the cloud app - maybe rejecting configurations over 5mb

evantahler (approved):
🥳

PR #2221, opened 2021-08-27
Fix Actionhero websocket client

evantahler:
Actionhero [v27](https://github.com/actionhero/actionhero/releases/tag/v27.0.0) removed support for minimizing the generated client JS.  We were still pointing to the old file name.  Oops! 

In some future story we will bundle this client JS into the next build, which is the "right" way to handle this lib.

### DIFF
--- a/core/src/config/api.ts
+++ b/core/src/config/api.ts
@@ -89,7 +89,6 @@ export const DEFAULT = {
             ]
           : [],
         plugin: [path.join(process.cwd(), "..", "node_modules")],
-        locale: [path.join(process.cwd(), "locales")],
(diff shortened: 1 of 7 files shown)


core/src/config/api.ts
```diff
@@ -89,7 +89,6 @@ export const DEFAULT = {
             ]
           : [],
         plugin: [path.join(process.cwd(), "..", "node_modules")],
-        locale: [path.join(process.cwd(), "locales")],
```
evantahler, line 92: Avtionhero v27 also removed the localization system - we weren't using it, and so we can remove these paths.

PR #2226, opened 2021-08-30
Whenever confirming data, reset startedAt

bleonard:
Just a draft to show where I think we need to set `startedAt` back to `null` to show that it's not currently running.
This will allow it get picked up earlier by `profileProperties:enqueue` because this uses a 5 minute `defaultProfilePropertyProcessingDelay` in `processPendingProfileProperties`.

There might be some more spots or something I'm not considering.

### DIFF
--- a/core/src/modules/ops/profile.ts
+++ b/core/src/modules/ops/profile.ts
@@ -306,6 +306,7 @@ export namespace ProfileOps {
               position,
               rawValue,
               state: "ready",
+              startedAt: null,
               stateChangedAt: now,
               confirmedAt: now,
               valueChangedAt:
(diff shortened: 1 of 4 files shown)


evantahler:
I think perhaps the only changes to reset the `startedAt` would need to be to the internalRun task https://github.com/grouparoo/grouparoo/blob/9d716ae2eb85c76ad802b53c86081fd87e60f4ef/core/src/tasks/run/internalRun.ts#L46-L60.

If the slowdown in https://www.pivotaltracker.com/n/projects/2401683/stories/179363811 was due to an updated task, that would be due to a new internalRun

bleonard:
Yes, setting it when you make it pending might work even better. Let's go with that.

PR #2229, opened 2021-08-31
Improvements to Eloqua error in case of bad credentials

pauloouriques:


plugins/@grouparoo/eloqua/src/lib/client/client.ts
```diff
@@ -57,9 +57,12 @@ export default class EloquaClient {
     const response = await axios(params);
     const { data = {} } = response;
     const { urls = {} } = data;
-
     if (urls["base"]) {
       return urls["base"];
+    } else if (data === "Not authenticated.") {
```
evantahler, line 62: Just confirming that the type of this is correct - On like 58 you init `data` to be an object, and here you are checking that it's a string.  I certainly believe that Eloqua would change types on us... just double checking!

evantahler (approved):

plugins/@grouparoo/eloqua/src/lib/client/client.ts
pauloouriques: yep, they change the type. I created data = {} to not crash in the next line if it's a string or it's not there.

PR #2258, opened 2021-09-01
Log queries for datasources [179274982]

krishnaglick:


ui/ui-components/pages/index.tsx
```diff
@@ -1,6 +1,6 @@
-import { useRouter } from "next/router";
 import Head from "next/head";
-import { Row, Col, Image, Button } from "react-bootstrap";
+import { useRouter } from "next/router";
+import { Button, Col, Image, Row } from "react-bootstrap";
```
krishnaglick, line 3: Prettier import sorting

plugins/@grouparoo/postgres/src/lib/connect.ts
```diff
@@ -43,6 +44,18 @@ export const connect: ConnectPluginAppMethod = async ({ appOptions }) => {
...
+  // Unfortunately there's no event listener for queries so we have to intercept the call.
+  // https://github.com/brianc/node-postgres/issues/2580
+  const shim: typeof pool.query = pool.query.bind(pool);
+  (pool as any).query = (
+    query: Parameters<typeof pool.query>[0],
+    data?: Parameters<typeof pool.query>[1],
+    callback?: Parameters<typeof pool.query>[2]
+  ) => {
+    log(`[ postgres ] ${query}`, "debug", data);
+    return shim(query, data, callback);
```
bleonard, line 56: These might be a little more resilient with sending `(*arguments)` or something like that and only opting into the ones needed for logging. But maybe just something to look out for if there is churn.

bleonard (approved):

plugins/@grouparoo/postgres/src/lib/connect.ts
krishnaglick: I'm intentionally doing this so if the method signature changes we'll need to update this as well.
While unlikely, the first argument could be changed to something else in an update.

plugins/@grouparoo/bigquery/__tests__/integration/log-checking.ts
```diff
@@ -0,0 +1,78 @@
...
+    propertyId: null,
+    propertyFilters: null,
+  });
+}
+
+describe("bigquery/integration/log-checking", () => {
+  helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
+  const actionhero = require("actionhero");
+  const logMock = jest.fn();
+  Object.defineProperty(actionhero, "log", { value: logMock, writable: false });
```
evantahler, line 53: Nice! I'm glad you found a way to mock the logger!

plugins/@grouparoo/snowflake/__tests__/utils/nockHelper.ts
```diff
@@ -85,11 +85,11 @@ export const updater = {
...
       let newBytes = zipped.toString("hex");
 
-      // somehow, these bytes are slightly different (1 hex digit) on mac/ci
+      // somehow, these bytes are slightly different on mac/ci
       // mac: 1f8b0800000000000013558...
       //  ci: 1f8b0800000000000003558...
-      const regexBytes =
-        "/" + newBytes.slice(0, 18) + "." + newBytes.slice(19) + "/";
+
+      const regexBytes = "/" + newBytes.slice(0, 18) + ".+" + "/";
```
evantahler, line 92: whoa! good fix

evantahler (approved):
Great work, and thanks for diving so deep on this!

plugins/@grouparoo/snowflake/__tests__/utils/nockHelper.ts
krishnaglick: Credit where credit's due, this was Brian :D

PR #2259, opened 2021-09-01
Demo now generates a calculated full name property on profiles [179343542]

krishnaglick:


teallarson (changes requested):
Looks really close!  I noticed two things we missed while pairing:
- making the names of the app/source/etc "Demo Calculated Property ____" instead of "My Calculated Property ____" (and the ids too, but with snake case) would help it flow better with what's already generated
- it looks like the `demo/config/shared/all/identity/source/` directory should be `/sources` instead (otherwise it won't end up in the right directory when the files generate)

krishnaglick:
Pushed up those changes!

plugins/@grouparoo/demo/config/shared/all/identity/apps/my_calculated-property_app.json
```diff
@@ -0,0 +1,7 @@
+{
+  "class": "App",
+  "id": "demo_calculated-property_app",
```
krishnaglick: It feels a little strange that this is a mix of `_` and `-` but I'm assuming we do this elsewhere.

teallarson: Good point!  Let's do all snake casing, I'll update the docs to match.  This is our first multi-word plugin!

teallarson (approved):
The failing tests look like a timeout (which for some reason has been happening a lot this week!).  I'd try rerunning them and then it looks good!

krishnaglick:
It's green! 🎉

PR #2261, opened 2021-09-01
Change max node version

vklimontovich:
16.0.0 is not the latest version of node, 16.8.0 is

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "license": "SEE LICENSE IN LICENSE.txt",
   "private": true,
   "engines": {
-    "node": ">=12.0.0 <16.0.0"
+    "node": ">=12.0.0 <16.8.0"
   },
   "description": "The main Grouparoo Application. Learn more at www.grouparoo.com.",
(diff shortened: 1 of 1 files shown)


bleonard:
Thanks @vklimontovich - this PR  https://github.com/grouparoo/grouparoo/pull/2129 has a blocker that is preventing v16 from working fully in our setup.

bleonard:
The one this PR changes is actually only for that development. It's possible that the one that `grouparoo init` does could be upgraded because it doesn't have those compile times issues. Maybe @evantahler ?

I think that would be this file: https://github.com/grouparoo/grouparoo/blob/main/cli/templates/package.json

bleonard:
When it does work, I imagine we'll do `< 17`

evantahler:
This sparked some research and I _think_ that all the built Grouparoo packages will run on node.js v16... we just can't build them with that version yet.  

https://github.com/grouparoo/grouparoo/pull/2270 replaces this PR

PR #2262, opened 2021-09-01
`importProfileProperties` should `filter` instead of `find` matching properties

teallarson:
Previously, when getting a Property Value for a table source, we would `find` what Property was looking for the data from a particular table column.  This caused a bug if a user tried to map more than one exact Property to a Table Source. Only one Property could return a value (the other would remain null).

By using `filter` instead, we can populate all properties that match a given column.


plugins/@grouparoo/app-templates/src/source/table/profileProperties.ts
```diff
@@ -109,13 +108,15 @@ export const getProfileProperties: GetProfilePropertiesMethod = ({
     for (const pk in responsesByPrimaryKey) {
       primaryKeysHash[pk].forEach((profileId) => {
         for (const column of Object.keys(responsesByPrimaryKey[pk])) {
-          const property = properties.find(
+          const propertiesFound = properties.filter(
             (p) => propertyOptions[p.id][columnNameKey] === column
           );
-          if (!property) continue;
+          if (!propertiesFound) continue;
```
bleonard: probably don't need this now because I think it's just an empty array.

plugins/@grouparoo/app-templates/src/source/table/profileProperties.ts
```diff
@@ -109,13 +108,14 @@ export const getProfileProperties: GetProfilePropertiesMethod = ({
...
             (p) => propertyOptions[p.id][columnNameKey] === column
           );
-          if (!property) continue;
 
-          responsesById[profileId][property.id] =
-            responsesByPrimaryKey[pk][column];
+          propertiesFound.forEach((property) => {
+            responsesById[profileId][property.id] =
+              responsesByPrimaryKey[pk][column];
+          });
```
evantahler: 👍 Thanks for tracking this down!

teallarson:
This fix does not behave as expected with bigquery (it does not populate both properties).

bleonard:
> This fix does not behave as expected with bigquery (it does not populate both properties).

The tests work!

teallarson:
> > This fix does not behave as expected with bigquery (it does not populate both properties).
> 
> The tests work!

That's because I didn't push the non-working Bigquery ones haha

bleonard:
> > > This fix does not behave as expected with bigquery (it does not populate both properties).
> > 
> > 
> > The tests work!
> 
> That's because I didn't push the non-working Bigquery ones haha

Ha. I actually meant testing in general leads to use finding these nuances. They "work" in that they save us from crazy stuff.

Btw, there might be something funny here about how we are mapping to upper and lower case things. We could miss that `filter` or something if the plugin gets funny with the columns names.

teallarson:
> > > > This fix does not behave as expected with bigquery (it does not populate both properties).
> > > 
> > > 
> > > The tests work!
> > 
> > 
> > That's because I didn't push the non-working Bigquery ones haha
> 
> Ha. I actually meant testing in general leads to use finding these nuances. They "work" in that they save us from crazy stuff.
> 
> Btw, there might be something funny here about how we are mapping to upper and lower case things. We could miss that `filter` or something if the plugin gets funny with the columns names.

Are you referencing the Snowflake test using capital letters for the column name?  I assumed that was a Snowflake convention as they are all caps in every test.

bleonard:
not sure, I was thinking of cases where the property they configure might be lower case and it returns upper case. Maybe it was Snowflake that's weird like that. But it occurred to me when thinking of why it might miss data. Maybe a separate investigation at some point.

teallarson:
Interesting.  Got it hooked up to a BigQuery source and this error occurred.  Did not occur for any other table source type so far.

```
error: Error: Error with Postgres SQL Statement: Query - `SELECT `first_name`, `first_name`, `id` as __pk FROM `profiles` WHERE  `id` IN UNNEST (?) GROUP BY `id`, `first_name`, `first_name``, Error - Error: Name first_name in GROUP BY clause is ambiguous; it may refer to multiple columns in the SELECT-list at [1:106]
    at getPropertyValues (/Users/teal/workspace/grouparoo/grouparoo/plugins/@grouparoo/bigquery/dist/lib/table-import/getPropertyValues.js:100:15)
```

More interesting is that it somehow resolves that and does return what we're looking for in the run (in this case two exact properties mapped from `first_name` in bq)

bleonard:
Oh! it's in the SELECT query twice!
Core (or the plugins?) should make those unique for all the plugins before the queries.
Probably core.

plugins/@grouparoo/bigquery/src/lib/table-import/getPropertyValues.ts
```diff
@@ -23,7 +23,7 @@ export const getPropertyValues: GetPropertyValuesMethod = async ({
   let responses: { [key: string]: { [column: string]: DataResponse[] } } = {};
   let aggSelect = columnNames.map((col) => `\`${col}\``).join(", ");
   let orderBy = "";
-  let groupByColumns = [tablePrimaryKeyCol];
+  let groupByColumns = [...new Set(tablePrimaryKeyCol)];
```
bleonard: 💥

teallarson: Bigquery is the only one who had these shenanigans... so I only changed Bigquery.

plugins/@grouparoo/mysql/__tests__/table-import/import-properties.ts
```diff
@@ -173,6 +173,31 @@ describe("mysql/table/recordProperties", () => {
...
+          const [values, properties] = await getPropertyValues({
+            columns,
+            sourceMapping,
+            aggregationMethod,
+          });
+          expect(values[record.id][properties[0].id]).toEqual(["Erie"]);
+          expect(values[otherRecord.id][properties[0].id]).toEqual(["Cacilie"]);
+          expect(values[record.id][properties[1].id]).toEqual(["Erie"]);
+          expect(values[otherRecord.id][properties[1].id]).toEqual(["Cacilie"]);
+        });
```
evantahler: Looks like the same test in here 2x

evantahler (commented):
This looks great! 

Thanks for also fixing up a few errant `profile` -> `record` renames too.

I think there's a test needed for Mongo as well?

teallarson:
> This looks great!
> 
> Thanks for also fixing up a few errant `profile` -> `record` renames too.
> 
> I think there's a test needed for Mongo as well?

Thanks.  For Mongo, the tests we have [only grab a single column at a time](https://github.com/grouparoo/grouparoo/blob/f13ef744691fc2ebf54c5b161a920fed970c0a82/plugins/%40grouparoo/mongo/__tests__/table-import/import-properties.ts#L23), but I can spend some time getting that running with multiple columns if we want.

evantahler (approved):
🥳

PR #2268, opened 2021-09-02
Do not exclude contacts from campaigns by default

### LINKED ISSUE #2266: Mailjet users excluded from campaign sends
**Describe the bug**

When users are synced to the Mailjet destination, they are all added to the Campaign Exclusion List. Thus, if they're added to contact lists that are created when Grouparoo syncs groups across, we're unable to send emails to any of them through campaigns or automations.

**To Reproduce**
1. Configure a source.
2. Configure Mailjet as a destination, with data to reflect the following. In particular, note that a Grouparoo group is synced to a Mailjet list.

3. Wait for the sync task to complete.
4. Navigate to https://app.mailjet.com/contacts and open the synced list(s)
5. See error - red dots appear beside every listed contact.

**Expected behavior**

All the listed users should not be in the Campaign Exclusion List.

**Environment (please complete the following information):**

Hosted instance of Grouparoo running at v0.6.0.

boardfish:
Closes #2266

Haven't tested this against Mailjet itself - assuming it'll do the trick. It's just a search and replace. If it's possible for one of the folks on the team to do so, I'd greatly appreciate it.


plugins/@grouparoo/mailjet/src/lib/client.ts
```diff
@@ -125,7 +125,7 @@ class MailjetClient {
     const request = await this.client
       .post("contact", { version: "v3" })
       .request({
-        IsExcludedFromCampaigns: "true",
+        IsExcludedFromCampaigns: "false",
```
evantahler, line 128: This might be a good place for a Destination Option so you can choose how to handle `IsExcludedFromCampaigns` per destination. 

cc @pauloouriques

bleonard: For now, we can probably assume people want to have campaigns and do the option later if we find someone that doesn't.

pauloouriques: Agreed. I'll keep the Destination Option alternative in the loop.

pauloouriques (approved):

bleonard:
Thanks @boardfish, I'll test it out now!

bleonard:
Looks like they aren't on the exclusion list!
Screen Shot 2021-09-02 at 1 18 41 PM

bleonard:
Released this in `0.6.1` release. Thanks @boardfish 
Your Grouparoo Cloud is updated, too. Open source at work!

PR #2272, opened 2021-09-03
MySQL source in demo

bleonard:
Defaults to localhost, user `root`, no password, and database `grouparoo_demo`
If other, use env variable: `DEMO_MYSQL_URL=mysql://root:pw@127.0.0.1:3306/my_db_name`

example commands:
`roo demo mysql`
`roo demo mysql accounts mailchimap`


plugins/@grouparoo/demo/src/connections/mysql.ts
```diff
@@ -0,0 +1,201 @@
...
+  payments: {
+    id: "INT NOT NULL PRIMARY KEY",
+    account_id: "INT NOT NULL",
+    amount: "DECIMAL",
+    state: "VARCHAR(191)",
+    created_at: "TIMESTAMP NOT NULL",
+  },
+};
+
+function findConfig(): any {
```
krishnaglick, line 57: Are these `any`'s because they're generic?

plugins/@grouparoo/demo/src/connections/mysql.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    await this.query(1, `DROP TABLE IF EXISTS ${sqlTable}`);
+
+    const createQuery = `CREATE TABLE ${sqlTable} (${columnTypes.join(", ")})`;
+    await this.query(1, createQuery);
+
+    if (typeColumn !== "id") {
+      const indexQuery = `CREATE INDEX \`${tableName}_${typeColumn}\` ON ${sqlTable} (${typeColumn});`;
+      await this.query(1, indexQuery);
+    }
```
krishnaglick, line 159: It looks like we build a lot of SQL ourselves. Out of curiosity, why not just use an ORM?

plugins/@grouparoo/demo/src/connections/mysql.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+  if (value === "false") {
+    return false;
+  }
+  if (!value) {
+    return null;
+  }
+  if (value.length === 0) {
+    return null;
+  }
```
krishnaglick, line 199:
```suggestion
  if (!value || value.length === 0) {
    return null;
  }
```

krishnaglick (commented):
Does this need any dockerfile updates or anything?

plugins/@grouparoo/demo/src/connections/mysql.ts
bleonard: I'm sure there's a better type, even just key/value, but I didn't prioritize figuring it out.

plugins/@grouparoo/demo/src/connections/mysql.ts
bleonard: I was doing things outside of sequelize for the Postgres one so it didn't get mixed up with the Grouparoo sequelize situation in any way. Then I copied that one here. Could go either way.

PR #2211, opened 2021-08-24
`Profiles` renamed to `Records`

evantahler:
This PR Prepares the Grouparoo codebase to support multiple Model types by renaming the existing `Profile` to `Record`.  this includes related Migrations and UI changes.

## TODO

- [x] Big Migration
- [x] Packages compile
- [x] Tests Pass
~~- [ ] Adds `Models` table~~ (moved to next story)
- [x] Plugin Methods all change
- [x] UI Updates

## Migrate new packages since the start of this branch
- [x] Calculated Properties
- [x] Mailjet

## Breaking Changes
None!  There were 2 DEPRECATED notes on config loaders so both `schedule#confirmProfiles` and `profiles` would still load.

## Backup Branches
* `179245806-profiles-to-records-backup-aug-30`

## NOTES
* `Record` is a reserved name in Typescript, so our Model is now `GrouparooRecord`

PR #2276, opened 2021-09-07
Null dates don't show 1970 anymore

krishnaglick:


ui/ui-components/components/property/profilePreview.tsx
```diff
@@ -1,10 +1,17 @@
-import { useState, useEffect } from "react";
-import { useApi } from "../../hooks/useApi";
+import { useRouter } from "next/router";
+import { useEffect, useState } from "react";
 import { Card, ListGroup } from "react-bootstrap";
+import { useApi } from "../../hooks/useApi";
+import { Actions } from "../../utils/apiData";
```
krishnaglick: Forgot to turn off import sorting for `.tsx` files.

ui/ui-components/components/property/profilePreview.tsx
```diff
@@ -1,10 +1,17 @@
...
 import Loader from "../loader";
 import ProfileImageFromEmail from "../visualizations/profileImageFromEmail";
-import { useRouter } from "next/router";
-import { Actions } from "../../utils/apiData";
+
+const calculatePreviewDate = (date: any) => {
+  if (!date) {
+    return new Date().toLocaleString();
```
krishnaglick: Unsure what we want the default behavior to be. Assuming current time for now.

ui/ui-components/components/property/profilePreview.tsx
```diff
@@ -178,7 +185,7 @@ export default function ProfilePreview(props) {
             <ListGroup.Item key={`profile-preview-row-${k}`} variant="light">
               <strong>{k}</strong>:{" "}
               {otherProperties[k]?.type === "date"
-                ? new Date(otherProperties[k]?.values[0]).toLocaleString()
+                ? calculatePreviewDate(otherProperties[k]?.values[0])
```
krishnaglick: I'm not sure if this incoming value can be typed.

evantahler:
Please add a PR label (`bug` in this case) so this shows up in the changelog!

evantahler:
I removed the story ID from the title of the PR - that's only needed in the branch.  It's not meaningful to folks outside the org

ui/ui-components/components/property/profilePreview.tsx
evantahler: It's OK for now, but in the future I think this will cause some problems

ui/ui-components/components/property/profilePreview.tsx
evantahler: This component is trying to render the Properties of a Profile (e.g.: "Most Recent Purchase Timestamp").  If they don't have a purchase we want to show null, "", or "N/A" and not imply that they've made a purchase

ui/ui-components/components/property/profilePreview.tsx
evantahler: `otherProperties[k]?.values[0]` should always be a number here if you want to check for that

evantahler (changes requested):

ui/ui-components/components/property/profilePreview.tsx
krishnaglick: I turned it off after this. It did cause issues elsewhere already

ui/ui-components/components/property/profilePreview.tsx
krishnaglick: https://www.pivotaltracker.com/story/show/179394479
I left a comment asking what to do. I'm good with any of those.

bleonard: go ahead and return null here or whatever the other ones do. I imagine it's null.
It might be the most correct to check for `date === null || date === undefined` here as I suppose a date of `0` is supported.

ui/ui-components/components/property/profilePreview.tsx
```diff
@@ -177,8 +199,9 @@ export default function ProfilePreview(props) {
           {Object.keys(otherProperties).map((k) => (
             <ListGroup.Item key={`profile-preview-row-${k}`} variant="light">
               <strong>{k}</strong>:{" "}
-              {otherProperties[k]?.type === "date"
-                ? new Date(otherProperties[k]?.values[0]).toLocaleString()
+              {otherProperties[k]?.type === "date" &&
+              typeof otherProperties[k]?.values[0] === "number"
```
bleonard: it's a little funny to be falling through this `if` when it's a date but somehow not a number. maybe "if date, then generatePreviewDate" and handle null and otherwise in `generatePreviewDate` 
Not a huge deal per se. Just an idea.

krishnaglick:
Threw out my refactor work as it was breaking things and spiraling.

evantahler (commented):
The implementation looks good! 

Regarding the test, can you share more about why you added `react-test-renderer` and didn't use `@testing-library` which is already in the project ([for example](https://github.com/grouparoo/grouparoo/blob/main/ui/ui-components/__tests__/components/setupSteps/setupStepCard.tsx)).  What's testing-library missing?

krishnaglick:
> The implementation looks good!
> 
> Regarding the test, can you share more about why you added `react-test-renderer` and didn't use `@testing-library` which is already in the project ([for example](https://github.com/grouparoo/grouparoo/blob/main/ui/ui-components/__tests__/components/setupSteps/setupStepCard.tsx)). What's testing-library missing?

This was my bad. I tried to import `act` and it grabbed it from something strange so I thought we needed something different. Fixed.

core/jest.setup.js
```diff
@@ -2,3 +2,4 @@ require("@grouparoo/spec-helper/dist/lib/environment");
 
 // tell `@grouparoo/spec-helper` to load everything from `src` in core, so we can test without re-compiling
 process.env.GROUPAROO_SPEC_HELPER_LOAD_PATH = "src";
+process.env.TZ = "UTC";
```
krishnaglick: This theoretically should make CI consistent with my local but I'm not sure it's going to do it.

pnpm-lock.yaml
```diff
@@ -14012,7 +14012,7 @@ packages:
       '@types/jest': 27.0.1
       bs-logger: 0.2.6
       fast-json-stable-stringify: 2.1.0
-      jest: 27.1.0_ts-node@10.2.1
+      jest: 27.1.0
```
krishnaglick, line 14015: I can revert this if we want, don't think it matters.

ui/ui-components/__tests__/components/__snapshots__/profilePreview.tsx.snap
```diff
@@ -0,0 +1,135 @@
...
+      </div>
+      <div
+        class="list-group-item list-group-item-light"
+      >
+        <strong>
+          date
+        </strong>
+        :
+         
+        9/8/2021, 2:58:38 PM
```
krishnaglick: Correctly show a timestamp!

ui/ui-components/__tests__/components/__snapshots__/profilePreview.tsx.snap
```diff
@@ -0,0 +1,135 @@
...
+         
+        ...
+      </div>
+      <div
+        class="list-group-item list-group-item-light"
+      >
+        <strong>
+          date
+        </strong>
+        :
```
krishnaglick, line 129: No timestamp.

core/jest.setup.js
evantahler: 👍 might as well lock in a timezone for testing

ui/ui-components/__tests__/components/profilePreview.tsx
```diff
@@ -0,0 +1,112 @@
...
+  nextRouter.useRouter = jest.fn(() => ({
+    route: "/",
+    query: {
+      profileId: "asdf",
+    },
+    replace: jest.fn(),
+    push: jest.fn(),
+  }));
+  return nextRouter;
+});
```
evantahler, line 17: 👍 I bet at some point we extract this router mock into a utility

ui/ui-components/__tests__/components/profilePreview.tsx
```diff
@@ -0,0 +1,112 @@
...
+            id: "test-id",
+            key: "test-key",
+            type: "test-type",
+            options: {},
+          }}
+        />
+      );
+      await new Promise((res) => setTimeout(res, 2000));
+    });
+    expect(profile.container).toMatchSnapshot();
```
evantahler, line 75: Newbie question: Why is snapshot testing preferred vs an explicit expectation?

ui/ui-components/__tests__/components/profilePreview.tsx
```diff
@@ -0,0 +1,112 @@
...
+          property={{
+            id: "test-id",
+            key: "test-key",
+            type: "test-type",
+            options: {},
+          }}
+        />
+      );
+      await new Promise((res) => setTimeout(res, 2000));
+    });
```
evantahler: So that's how you do it! 

We were struggling in the past to know how to test things thad made API requests in a `useEffect` - I guess the `sleep()`  is unavoidable here.

Presumably we'll be doing this kind of thing a lot in tests... can this be extracted into a utility of some sort? `renderAndWait()`?

evantahler (approved):
👍 from me!  

There are some questions below, but none of them prevent merging this in

core/jest.setup.js
krishnaglick: This is _not_ having the intended effect and I am upset.

ui/ui-components/__tests__/components/profilePreview.tsx
krishnaglick: Querying into the rendered element or doing text matching is fine, but I have to reasons:
1. When making changes to the component you'll see structural changes reflected in the snapshot. _Ideally_ you'll be able to determine that you wanted those changes.
2. It's a super easy way to test and keep test data/direct comparisons out of the test code itself. You do have to eyeball it the first run, but then it's pretty consistent.

ui/ui-components/__tests__/components/profilePreview.tsx
krishnaglick: For sure!

ui/ui-components/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "scripts": {
     "pretest": "npm run lint",
-    "test": "jest",
+    "test": "TZ=UTC jest",
```
krishnaglick, line 25: Setting this in the jest config did not seem to have the desired effect.

evantahler:
(normally @krishnaglick would merge your own PR, but I wanted get this landed before https://github.com/grouparoo/grouparoo/pull/2211)

PR #2277, opened 2021-09-07
Use date in `specHelper` directlyMapped property to ensure uniqueness

evantahler:


evantahler:
Tested 20 times in a row locally... on collisions!

plugins/@grouparoo/spec-helper/src/lib/factories/profile.ts
```diff
@@ -19,8 +19,8 @@ export default async (props = {}, properties = {}) => {
   const directlyMappedProperty = allProperties.find((p) => p.directlyMapped);
 
   if (directlyMappedProperty) {
-    properties[directlyMappedProperty.key] = faker.unique(
-      faker.datatype.number
+    properties[directlyMappedProperty.key] = parseInt(
+      `${new Date().getTime()}${faker.unique(faker.datatype.number)}`
```
bleonard, line 23: My understanding is that this will still fail intermittently. The crash is while generating the unique number itself within `faker.unique` - it must keep track which ones it has made before.

Maybe something like this:
```
properties[directlyMappedProperty.key] = parseInt(
      `${new Date().getTime()}${Math.round(Math.random() * 999999)}`)
```

or maybe you need to be clear to give faker a higher max
```
properties[directlyMappedProperty.key] =  faker.unique(faker.datatype.number, 999999999)
```

Faker looks to use 99999 if not given.
```
    if (typeof options === "number") {
      options = {
        max: options
      };
    }

    options = options || {};

    if (typeof options.min === "undefined") {
      options.min = 0;
    }

    if (typeof options.max === "undefined") {
      options.max = 99999;
    }
```

evantahler: I thought so too... but I just couldn't get it to fail?  Either way https://github.com/grouparoo/grouparoo/pull/2279 is better using faker property

PR #2279, opened 2021-09-08
`fake.unique` can use big numbers

evantahler:
A better solution to https://github.com/grouparoo/grouparoo/pull/2277


plugins/@grouparoo/spec-helper/src/lib/factories/profile.ts
```diff
@@ -19,8 +19,9 @@ export default async (props = {}, properties = {}) => {
   const directlyMappedProperty = allProperties.find((p) => p.directlyMapped);
 
   if (directlyMappedProperty) {
-    properties[directlyMappedProperty.key] = parseInt(
-      `${new Date().getTime()}${faker.unique(faker.datatype.number)}`
+    properties[directlyMappedProperty.key] = faker.unique(
+      faker.datatype.number,
+      [{ min: 1, max: 999999999999 }]
```
bleonard, line 24: That's a big number! 👍

PR #2281, opened 2021-09-08
Abort publish if no dist folder

krishnaglick:


bin/publish
```diff
@@ -124,6 +124,14 @@ rm -rf ui/*/.next
 pnpm install --prefer-frozen-lockfile
 git checkout pnpm-lock.yaml # the other github action will take care of this, but we need a clean working directory to publish
 
+# Check to make sure it was built correctly
+rm -rf core/dist
+if [ ! -d "core/dist" ]; then
```
bleonard: think this could fail on other plugins too?

krishnaglick: Added a few more checks

bin/publish
```diff
@@ -124,6 +124,22 @@ rm -rf ui/*/.next
 pnpm install --prefer-frozen-lockfile
 git checkout pnpm-lock.yaml # the other github action will take care of this, but we need a clean working directory to publish
 
+# Check to make sure it was built correctly
+rm -rf core/dist
+if [ ! -d "core/dist" ] ||\
+   [ ! -d "cli/dist" ] ||\
+   [ ! -d "ui/ui-community/.next" ] ||\
+   [ ! -d "ui/ui-config/.next" ] ||\
+   [ ! -d "ui/ui-enterprise/.next" ] ||\
```
evantahler, line 136: I like testing the UI packages, core, and cli explicitly

bin/publish
```diff
@@ -124,6 +124,22 @@ rm -rf ui/*/.next
...
+   [ ! -d "plugins/postgres/dist" ] ||\
+   [ ! -d "plugins/mysql/dist" ] ||\
+   [ ! -d "plugins/mongo/dist" ] ||\
+   [ ! -d "plugins/bigquery/dist" ]; then
```
evantahler: Rathe then check some of these plugins, you could check them all.  There are lots of ways, but here's one suggestion:

```bash
# Count up the number of plugins with `src` folders:
PLUGINS=`find 'plugins' -name src -and -type d | wc -l`
# Count the number of dist folders in the plugins directory
BUILT_PLUGINS=`find 'plugins' -name dist -and -type d | wc -l`
# Check if they are equal
if [[ "$PLUGINS" != "$BUILT_PLUGINS" ]]; then echo 'OH NO'; fi
```

evantahler (changes requested):

bin/publish
```diff
@@ -124,6 +124,22 @@ rm -rf ui/*/.next
 pnpm install --prefer-frozen-lockfile
 git checkout pnpm-lock.yaml # the other github action will take care of this, but we need a clean working directory to publish
 
+# Check to make sure it was built correctly
+rm -rf core/dist
```
pedroslopez:
```suggestion
```

was this left in from a test? looks like this explicitly removes the `dist` dir from core

pedroslopez (changes requested):

bin/publish
krishnaglick: Nice catch, it was!

bin/publish
```diff
@@ -124,6 +124,22 @@ rm -rf ui/*/.next
 pnpm install --prefer-frozen-lockfile
 git checkout pnpm-lock.yaml # the other github action will take care of this, but we need a clean working directory to publish
 
+# Check to make sure it was built correctly
+
+PLUGINS=`find 'plugins' -name src -and -type d | wc -l`
+BUILT_PLUGINS=`find 'plugins' -name dist -and -type d | wc -l`
```
krishnaglick, line 130: Thanks for the sick bash-foo Evan

evantahler (approved):
🐧

bin/publish
```diff
@@ -124,6 +124,22 @@ rm -rf ui/*/.next
...
+
+if [ ! -d "core/dist" ] ||\
+   [ ! -d "cli/dist" ] ||\
+   [ ! -d "ui/ui-community/.next" ] ||\
+   [ ! -d "ui/ui-config/.next" ] ||\
+   [ ! -d "ui/ui-enterprise/.next" ] ||\
+   [ "$PLUGINS" != "$BUILT_PLUGINS" ]; then
+  echo "Missing dist folder(s), exiting"
```
bleonard, line 138: might be nice in some future iteration to see which are missing.

PR #2286, opened 2021-09-09
Added eslint to /ui/

krishnaglick:
This work spawned out of some attempted work for https://github.com/grouparoo/grouparoo/pull/2276 that ended up being out of scope.
The goal of this PR is to simply get the linter in our system, and take action on it at a later point.

Screen Shot 2021-09-09 at 11 47 16 AM
Screen Shot 2021-09-09 at 11 47 41 AM

The big win of adding this is the react hook linting, which lets us know what does and does not need to go into a react hook's dependencies, but we get far more.
It does prop checking to ensure everything being references by a components props was actually defined, and thus expected.
It lets us know if we have unused imports in a file, can catch minor logic issues:
Screen Shot 2021-09-09 at 11 51 31 AM
or let us know if we have a variable lying around that we don't need any more!
Screen Shot 2021-09-09 at 11 52 27 AM

There's a number of very "Javascript" habits and this should catch us on them. We can adjust the sensitivity as we like in the future but most of these defaults are pretty reasonable I felt.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -9,10 +9,10 @@
   "license": "MPL-2.0",
   "private": false,
   "devDependencies": {
-    "@typescript-eslint/eslint-plugin": "^4.31.0",
-    "@typescript-eslint/parser": "^4.31.0",
-    "eslint": "^7.32.0",
-    "eslint-plugin-react": "^7.25.1",
-    "eslint-plugin-react-hooks": "^4.2.0"
+    "@typescript-eslint/eslint-plugin": "4.31.0",
+    "@typescript-eslint/parser": "4.31.0",
+    "eslint": "7.32.0",
+    "eslint-plugin-react": "7.25.1",
+    "eslint-plugin-react-hooks": "4.2.0"
   }
 }
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,0 +1,18 @@
+{
+  "name": "ui",
+  "description": "",
+  "scripts": {
+    "test": "eslint ."
+  },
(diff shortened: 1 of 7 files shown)


ui/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "ui",
+  "description": "",
+  "scripts": {
+    "test": "eslint ."
+  },
+  "author": "Grouparoo Inc <hello@grouparoo.com>",
+  "version": "0.7.0-alpha.0",
+  "license": "MPL-2.0",
+  "private": false,
```
evantahler, line 10: 👍 This "container" project shouldn't be published

ui/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^4.31.0",
+    "@typescript-eslint/parser": "^4.31.0",
+    "eslint": "^7.32.0",
+    "eslint-plugin-react": "^7.25.1",
+    "eslint-plugin-react-hooks": "^4.2.0"
+  }
```
evantahler: Please use exact dependency versions

evantahler (changes requested):
This is cool! 

A few questions:
* Without the `ignoreDuringBuilds` flag, does that mean that CI will fail with a problem?  Perhaps we always leave `ignoreDuringBuilds=true`, but move an explicit check into our `linter` part of the test suite
* Why only for the UI packages... won't this be helpful everywhere?

evantahler:
Replaced by https://github.com/grouparoo/grouparoo/pull/2288

PR #2289, opened 2021-09-10
When Grouparoo versions change, clear redis and resque

evantahler:
This will prevent resque errors that occur when upgrading Grouparoo (eg: task names or input changes).  Grouparoo is resilient to redis failures and after a short delay, all processing will resume again.

### DIFF
--- a/core/src/modules/reset.ts
+++ b/core/src/modules/reset.ts
@@ -146,10 +146,23 @@ export namespace Reset {
   }
 
   async function deleteKeys(pattern: string) {
-    const keys = await api.resque.queue.connection.redis.keys(pattern);
-    return Promise.all(
-      keys.map((k) => api.resque.queue.connection.redis.del(k))
-    );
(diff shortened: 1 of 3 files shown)


core/src/modules/reset.ts
```diff
@@ -146,10 +146,23 @@ export namespace Reset {
...
+      scanStream.on("data", async (keys: string[]) => {
+        scanStream.pause();
+        await Promise.all(keys.map((k) => client.del(k)));
+        count += keys.length;
+        scanStream.resume();
+      });
+    });
+
+    return result;
   }
```
evantahler, line 166: This method is moved form using the `KEYS` command in redis which is slow, blocking, and may return way too much data for node to handle to the `SCAN` method + a streaming handler.

evantahler: This method is tested previously in `__tests__/modules/reset.ts`

PR #2299, opened 2021-09-15
remove csv file import test

pedroslopez:
This was causing CI to fail. Don't quite understand why this was not failing in #2293 - maybe something in the merge with records?

edit

Looks like the PR removed `plugins/@grouparoo/csv/__tests__/file-import/profileProperty.ts` and #2294 renamed it to `plugins/@grouparoo/csv/__tests__/file-import/recordProperty.ts`


evantahler:
Thank you!

PR #2300, opened 2021-09-15
fix: dont throw on `profile:destroy` if it has already been deleted (backports #2298)

pedroslopez:
backports #2298 to v0.6.2

To make tests pass I had to swap some things in the csv tests (it was pulling a file that has been renamed from main) and cherry-pick a fix @evantahler had made that uses node 16.8 for CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
   build:
     resource_class: "medium+"
     docker:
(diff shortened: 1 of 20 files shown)


evantahler (approved):
This is great work!

Thank you for sorting though the multiverse-level of problems with CI.  This looks right!

PR #2305, opened 2021-09-16
Reducing Imports under load

evantahler:
This PR includes 2 enchantments to reduce the load created by Grouparoo.

1. Do not recalculate GroupMembers if the Group does not contain any relative rules (e.g: people in this group have signupDate less than 7 days ago).  Static Group rules do not need to be checked periodically as we will always re-check a Grouparoo Record's membership any time a Property changes.
2. Sort pending RecordProperties about to be imported by updatedAt (oldest first). 
 
(2) prevents a situation that's possible to get into wherin a recurring schedule runs so quickly that we get in a cycle where:
* Run A created a bunch of Imports, which made a bunch of profiles pending.  
* Run A completed, but the Profiles were still pending
* Run B starts, and makes the profiles all pending again
* The bottom chunk of those imports + profiles never get to be complete.

### DIFF
--- a/cli/__tests__/test.sh
+++ b/cli/__tests__/test.sh
@@ -169,7 +169,7 @@ PID=$!
 echo ""
 echo ""
 
-sleep 10
+sleep 30
 
 echo ""
 echo "--- test: status ---"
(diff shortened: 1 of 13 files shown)


evantahler:
Let's remove the Circut breaker but instead lets make sure that we are ordering the RecordProperties we are choosing to import by updatedAt so we do the old ones first.  See here: https://github.com/grouparoo/grouparoo/blob/ed4ab9cc3cbece374d9ec794d6dcd353cdc1f836/core/src/modules/ops/recordProperty.ts#L129-L140

PR #2306, opened 2021-09-17
only show "add multiple properties" if possible to do

teallarson:
Before:  
All Sources would show an "Add Multiple Properties" button.  The issue with this is that Query sources and Calculated Property sources don't support this feature.  Clicking the button on an unsupported Source produced an API error stating that it couldn't return a source preview. 

Screen Shot 2021-09-17 at 10 50 51 AM

After:
Now we only show that button if `previewAvailable = true`.  So Query and Calculated Property sources show:

Screen Shot 2021-09-17 at 10 52 17 AM

And all other source types remain the same.


evantahler (approved):
Thanks!

PR #2308, opened 2021-09-18
`grouparoo demo` can provide junk data

evantahler:
`grouparoo demo -c --junkScale=20` would make 20% of your source data have garbage data! 

* We don't mess with primary keys
* We don't mess with `*_id` relationship keys

What kind of junk do we have so far?
* Numbers might go negative or become null
* Strings might get random spaces added or become null 
* Sometimes email addresses loose the `@`

Help Text:

```
$ grouparoo demo -h

--------

Usage: grouparoo demo [options] [type] [type]

Load data into a source database, create properties, destinations, and other config.
A demo@grouparoo.com Team Member is created. Password: "password"

Valid Types:
b2c            (default) loads users and purchases into source
b2b            loads users, accounts, and payment data into source
purchases      same as b2c
accounts       same as b2b
reset          only clear Grouparoo database and don't load config
setup          only create the login Team Member
postgres       (default) load source data into local Postgres database
mongo          load specified source data into local MongoDB database
mysql          load specified source data into local MySQL database
snowflake      assumes a Snowflake instance with data already present
bigquery       assumes a BigQuery instance with data already present
mailchimp      create Mailchimp destination for data
salesforce     create Salesforce destination for data

Options:
  --scale [scale]              make the number more than 1 to multiple amount of data. (default: "1")
  --junkPercent [junkPercent]  what percent of the data should have problems? (default: "0")
  -c, --config                 add flag to write to config directory and not populate configuraiton into Grouparoo database
  -s, --seed                   add flag to only write (or output) demo source data and not touch Grouparoo database
  -f, --force                  add flag to ensure deletion of config directory
  -h, --help                   display help for command
```

### DIFF
--- a/plugins/@grouparoo/demo/src/util/data.ts
+++ b/plugins/@grouparoo/demo/src/util/data.ts
@@ -2,9 +2,9 @@ export const TYPES = {
   users: {
     id: "INT NOT NULL PRIMARY KEY",
     account_id: "INT",
-    first_name: "VARCHAR(191) NOT NULL",
-    last_name: "VARCHAR(191) NOT NULL",
-    email: "VARCHAR(191) NOT NULL",
+    first_name: "VARCHAR(191)",
+    last_name: "VARCHAR(191)",
+    email: "VARCHAR(191)",
     gender: "VARCHAR(191)",
     ip_address: "VARCHAR(191)",
     avatar: "VARCHAR(191)",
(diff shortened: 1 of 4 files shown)


plugins/@grouparoo/demo/src/util/data.ts
```diff
@@ -2,9 +2,9 @@ export const TYPES = {
   users: {
     id: "INT NOT NULL PRIMARY KEY",
     account_id: "INT",
-    first_name: "VARCHAR(191) NOT NULL",
-    last_name: "VARCHAR(191) NOT NULL",
-    email: "VARCHAR(191) NOT NULL",
```
evantahler, line 7: Can't have garbage data with strong constraints!

evantahler:
now you can make a group of people without email addresses, for example, and see the junk!

Screen Shot 2021-09-17 at 6 04 13 PM

PR #2311, opened 2021-09-20
Retry RecordProperty Import errors outside of batch

evantahler:
If there's a problem importing a batch of RecordProperties (`recordProperty:importRecordProperties`) we will fall back to trying each Record + RecordProperty one-at-a time (`recordProperty:importRecordProperty`) so the culprit won't spoil the batch.

### DIFF
--- a/core/src/tasks/recordProperty/importRecordProperties.ts
+++ b/core/src/tasks/recordProperty/importRecordProperties.ts
@@ -110,11 +110,45 @@ export class ImportRecordProperties extends RetryableTask {
 
     if (recordsToImport.length === 0) return;
 
-    let propertyValuesBatch: RecordPropertiesPluginMethodResponse = {};
     try {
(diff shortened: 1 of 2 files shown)


core/src/tasks/recordProperty/importRecordProperties.ts
```diff
@@ -110,12 +110,46 @@ export class ImportRecordProperties extends RetryableTask {
...
+          where: {
+            propertyId: { [Op.in]: properties.map((p) => p.id) },
+            recordId: {
+              [Op.in]: recordsToImport.map((p) => p.id),
+            },
+            state: "pending",
+          },
+        }
+      );
     } catch (error) {
```
evantahler, line 153: Prior to this PR only the import from the source (`source.importRecordProperties`) was in the try/catch.  Now we'll wrap the import and the saving of that data in the try/catch

PR #2314, opened 2021-09-20
UI and Logging updates for Recods

evantahler:
* Log `Record` not `Grouparoo Record` with grouparoo run output
* Remove "GrouparooRecord" from SetupSteps
* Show logs for records returns `grouparooRecords` when filtering for `records`
* Update old log rows from profile to grouparooRecord (migration)
* We now include the `recordId` (when possible) if we cannot save an invalid property
```
2021-09-20T22:48:39.955Z - error: email "asterman5h-demo.com" is not valid for record rec_742c016d-727c-417e-aaaa-8f6e2dab8b2f
2021-09-20T22:48:40.123Z - error: email "tkleimt2c-demo.com" is not valid for record rec_d621ae68-e879-488c-8698-e9bb3c5c7986
2021-09-20T22:48:40.227Z - error: email "colliar92-demo.com" is not valid for record rec_a10f31ef-3528-45d2-a371-7f323c323052
```

### DIFF
--- a/core/src/migrations/000081-profilesToRecords.ts
+++ b/core/src/migrations/000081-profilesToRecords.ts
@@ -97,6 +97,10 @@ export default {
 
     await queryInterface.renameTable("profileProperties", "recordProperties");
 
+    await queryInterface.sequelize.query(
+      `UPDATE "logs" SET "topic"='grouparooRecord' WHERE "topic"='profile'`
+    );
+
     if (config.sequelize?.dialect === "sqlite") {
       await queryInterface.addIndex(
         "recordProperties",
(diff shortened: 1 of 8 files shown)


core/src/migrations/000081-profilesToRecords.ts
```diff
@@ -97,6 +97,10 @@ export default {
 
     await queryInterface.renameTable("profileProperties", "recordProperties");
 
+    await queryInterface.sequelize.query(
+      `UPDATE "logs" SET "topic"='grouparooRecord' WHERE "topic"='profile'`
```
bleonard, line 101: I don't think this was the full list.
`UPDATE "logs" SET "topic"='recordProperty' WHERE "topic"='profileProperty'` at least.

PR #2319, opened 2021-09-21
`grouaproo run` should end if there are Imports that had an error

bleonard:
This looks to be what was preventing `grouparoo run` from completing in the case that some imports failed.
For example, because of:
```
Error on step import:associateRecord: could not create a new record because no record property in {"User ID":[110821982]} is unique and owned by the source
```


evantahler:
👍 makes sense!

Please add a test for this one.

PR #2320, opened 2021-09-21
Moving agent value to X-Grouparoo-Client header

krishnaglick:


ui/ui-components/client/client.ts
```diff
@@ -124,8 +125,6 @@ export class Client {
       data: null,
       url: `${this.webUrl}/api/${this.apiVersion}${path}`, // path comes with a leading "/"
       withCredentials: true,
-      //@ts-ignore
```
evantahler, line 127: I guess the `//@ts-ignore` should have been a hint this wasn't quite right :D

evantahler (approved):

ui/ui-components/client/client.ts
krishnaglick: What does _typescript_ know 😂

PR #2323, opened 2021-09-21
Grouparoo Cloud CLI commands

pedroslopez:
New commands:
- `grouparoo pack`: generates a `.tar.gz` file with the project's package.json and config directory contents. It uses `getConfigDir()` to get the config directory, so no matter how the project's configured it should find it. Files that are not `.json` or `.js` are filtered out.
- `grouparoo push`: uploads a `.tar.gz` file to a Grouparoo Cloud project, creating a Configuration. Token can be loaded from an argument or env var (`GROUPAROO_CLOUD_API_TOKEN`). By default will only run `validate`. Pass `--apply` and will also apply the config.
- `grouparoo ci`: equivalent to running `grouparoo pack` and `grouparoo push`. The archive is saved to a temp dir and then deleted.
- `grouparoo deploy`: equivalent to running `grouparoo pack` and `grouparoo push --apply`. The archive is saved to a temp dir and then deleted.

**Note:** These commands must be run from within an installed Grouparoo project.

To use a different cloud api endpoint (staging, local dev), the `GROUPAROO_CLOUD_API_URL` env var can be set. By default it will point to `https://cloud.grouparoo.com`

### DIFF
--- a/apps/staging-enterprise/.env.example
+++ b/apps/staging-enterprise/.env.example
@@ -60,3 +60,13 @@ S3_BUCKET=""
 
 # change where the postgres demo database is, default to main database
 # DEMO_DATABASE_URL="postgresql://localhost:5432/grouparoo_development"
+
+###########
+## CLOUD ##
+###########
+
+# set a token for all Cloud API calls
+# GROUPAROO_CLOUD_API_TOKEN=""
+
+# change the API url to point to a different cloud instance
+# GROUPAROO_CLOUD_API_URL="http://localhost:8080"
--- a/core/__tests__/bin/push.ts
+++ b/core/__tests__/bin/push.ts
@@ -1,0 +1,218 @@
+import path from "path";
+import nock from "nock";
+import { helper } from "@grouparoo/spec-helper";
+import { Push } from "../../src/bin/push";
+
+describe("bin/push", () => {
+  helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
+
+  let archivePath = path.join(__dirname, "../../package.json");
+  let messages = [];
+  let spies = [];
+
+  beforeEach(async () => {
+    messages = [];
+    spies.push(
+      jest
+        .spyOn(console, "log")
+        .mockImplementation((message) => messages.push(message))
+    );
+    spies.push(
+      jest
+        .spyOn(console, "error")
+        .mockImplementation((message) => messages.push(message))
+    );
+  });
+
+  afterEach(async () => {
+    spies.map((s) => s.mockRestore());
+  });
+
+  test("an auth token is required", async () => {
+    const command = new Push();
+    const toStop = await command.run({
+      params: {
+        archivePath,
+        projectId: "my-project",
+        apply: false,
+      },
+    });
+
+    expect(toStop).toBe(true);
+
+    const output = messages.join(" ");
+    expect(output).toContain("You must provide a Grouparoo Cloud token");
+  });
+
+  test.each([true, false])(
+    "uploads config and waits for `finished` state (toApply=%p)",
+    async (toApply) => {
+      nock("https://cloud.grouparoo.com")
+        .post("/api/v1/configuration")
+        .query(true)
+        .reply(200, {
+          configuration: { id: "config-id" },
+        });
+
+      nock("https://cloud.grouparoo.com")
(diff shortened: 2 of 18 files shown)


core/src/modules/cloud.ts
```diff
@@ -0,0 +1,149 @@
+import os from "os";
+import path from "path";
+import tar from "tar";
+import fs from "fs";
+import FormData from "form-data";
+import { mkdtemp, copy, remove } from "fs-extra";
+import axios, { AxiosError, AxiosInstance } from "axios";
```
pedroslopez: using axios because isomorphic-fetch doesn't seem to play well with formData in node contexts (used when uploading files)

apps/staging-enterprise/.env.example
```diff
@@ -60,3 +60,13 @@ S3_BUCKET=""
 
 # change where the postgres demo database is, default to main database
 # DEMO_DATABASE_URL="postgresql://localhost:5432/grouparoo_development"
+
+###########
+## CLOUD ##
+###########
```
evantahler, line 66: Please add to apps/staging-community as well

core/__tests__/bin/push.ts
```diff
@@ -0,0 +1,218 @@
...
+        archivePath,
+        projectId: "my-project",
+        token: "some-token",
+        apply: false,
+      },
+    });
+    expect(toStop).toBe(true);
+
+    const output = messages.join(" ");
+    expect(output).toContain("❌ something weird happened here");
```
evantahler: Maybe logs from the server should get a prefix of some sort

core/src/modules/cloud.ts
evantahler: That's a shame!  
... would it be terrible to remove isomprhic-fetch then?  Seems silly to have 2 request libraries...

core/src/modules/cloud.ts
```diff
@@ -0,0 +1,149 @@
...
+    this.code = code;
+  }
+}
+
+export async function packageConfig(
+  projectPath: string,
+  configDirPath: string,
+  tarballPath: string
+) {
+  const tempDir = await mkdtemp(path.join(os.tmpdir(), "grouparoo-pack-"));
```
evantahler, line 23: trailing `-`?  Also, should we use a new tmpdir for each pack attempt?

core/src/modules/cloud.ts
```diff
@@ -0,0 +1,149 @@
...
+  await copy(
+    path.join(projectPath, "package.json"),
+    path.join(tempDir, "package.json")
+  );
+
+  await copy(configDirPath, path.join(tempDir, "config"), {
+    filter: (src) => {
+      const ext = path.extname(src);
+      return !ext || [".json", ".js"].includes(ext);
```
evantahler, line 32: 👍 safety!

core/src/modules/cloud.ts
```diff
@@ -0,0 +1,149 @@
...
+  configurationId: string;
+  logs: string;
+  completedAt: string;
+  createdAt: string;
+  updatedAt: string;
+}
+
+export class CloudClient {
+  request: AxiosInstance;
+  projectId: string;
```
evantahler: Well designed interface!

evantahler (approved):
This looks really solid - good mocking and typing the cloud API responses and keeping things clear!

Some comments below, but nothing blocking!

core/src/modules/cloud.ts
pedroslopez: We are! `mkdtemp` actually makes a new directory. `grouparoo-pack-` is a prefix, to which `mkdtemp` appends a random series of characters. Here's an example one it generated: 
`/private/var/folders/1m/m2fpq2hx6t90xbh_j52j8cq00000gn/T/grouparoo-pack-OmQnwR`

core/src/modules/cloud.ts
pedroslopez: didn't want to remove isomorphic-fetch since it looks like some of these functions that use it may be called from the ui packages?

went back in to try to figure out why isomorphic-fetch wasn't working for me, and it looks like it was an issue with the typings. adding a `@ts-ignore` for it worked so I re-implemented the CloudClient with isomorphic-fetch and removed axios. (bc12180)

PR #2326, opened 2021-09-23
Type specHelper factoires

evantahler:
Attempt to provide a type to the specHelper factories.

We dynamically load the models the factories use from either src or dist, depending... so they might not always line up properly... but some typing is better than none!

### DIFF
--- a/plugins/@grouparoo/spec-helper/src/lib/factories/apiKey.ts
+++ b/plugins/@grouparoo/spec-helper/src/lib/factories/apiKey.ts
@@ -1,5 +1,6 @@
 import { loadPath } from "../loadPath";
 import faker from "faker";
+import { ApiKey } from "@grouparoo/core";
 
 const data = async (props = {}) => {
   const defaultProps = {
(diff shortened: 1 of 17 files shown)


plugins/@grouparoo/spec-helper/src/lib/factories/apiKey.ts
```diff
@@ -1,5 +1,6 @@
 import { loadPath } from "../loadPath";
 import faker from "faker";
+import { ApiKey } from "@grouparoo/core";
```
evantahler, line 3: Note we import the Type form "@grouparoo/core", which needs to be a static import

plugins/@grouparoo/spec-helper/src/lib/factories/apiKey.ts
```diff
@@ -14,5 +15,5 @@ const data = async (props = {}) => {
 
 export default async (props = {}) => {
   const { ApiKey } = await import(`@grouparoo/core/${loadPath}`);
-  return ApiKey.create(await data(props));
+  return ApiKey.create(await data(props)) as ApiKey;
```
evantahler, line 18: The instance of the object the factory creates is dynamically sourced from the proper location (src or dist), but then the type is applied to the response

PR #2330, opened 2021-09-23
ImportAndExport does not use transaction isolation in order to save partial progress

evantahler:
This PR enhances the `import and update` button on the `/record/{id}/edit` page in 2 ways:

* Stores the results of the import recordProperty-by-recordProperty (rather than in bulk) to allow any failures to be isolated and handled.  This allows a failing unique recordProperty to not cancel the whole import
* The action is not in a transaction so we can save as much of the import as we can if something fails.  Perhaps the import is OK but the export fails - now we will save the import data regardlesss

### DIFF
--- a/core/src/modules/ops/record.ts
+++ b/core/src/modules/ops/record.ts
@@ -393,7 +393,7 @@ export namespace RecordOps {
                   {
                     state: "ready",
                     rawValue: null,
-                    invalidValue: `${attemptedRecordProperty.rawValue} (duplicate)`,
+                    invalidValue: `${attemptedRecordProperty.rawValue}`,
                     stateChangedAt: new Date(),
                     confirmedAt: new Date(),
                   },
--- a/core/src/modules/ops/record.ts
+++ b/core/src/modules/ops/record.ts
@@ -13,6 +13,8 @@ import Sequelize, {
   OrderItem,
   WhereAttributeHash,
   QueryTypes,
+  UniqueConstraintError,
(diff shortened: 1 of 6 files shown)


core/src/tasks/recordProperty/importRecordProperty.ts
```diff
@@ -77,35 +74,7 @@ export class ImportRecordProperty extends RetryableTask {
         ? propertyValues
         : [propertyValues];
 
-      try {
-        await record.addOrUpdateProperties(hash);
-      } catch (error) {
```
evantahler, line 82: This retry logic is moved into the `record.addOrUpdateProperties` method itself

core/src/modules/ops/record.ts
```diff
@@ -502,8 +551,11 @@ export namespace RecordOps {
         const { canImport, properties } = await source.import(record);
 
         // We need to save each property as it is loaded so it can be used as a mapping for the next source
+        // We also don't want to save more than one recordProperty at a time so we can isolate any problem values
         if (canImport && toSave) {
-          await addOrUpdateProperties([record], [properties], false);
+          for (const [k, property] of Object.entries(properties)) {
+            await addOrUpdateProperties([record], [{ [k]: property }], false);
+          }
```
evantahler, line 558: This split here is slower, but allows us to retry just one failing write of a RecordProperty.  This method isn't used in the schedule/run workflow so it is OK to slow it down here

PR #2331, opened 2021-09-23
fix: `grouparoo config` runs migrations

pedroslopez:
As of #2309 only worker processes run migrations, but `grouparoo config` runs with 0 workers and _should_ run migrations. This was preventing config ui from starting up properly. 
The integration test did not catch this because `test` environments also auto run migrations.

Also cleans up ui-config a bit - looks like we copied some code from the Apply command that we forgot to update.


evantahler:
<3

PR #2332, opened 2021-09-23
Optimize `bulkDeletes` query cleaning up old RecordProperties after update

evantahler:
* Remove ORs
* Rely on previously loaded recordProperties to determine if delete step is required

### DIFF
--- a/core/__tests__/models/record/record.ts
+++ b/core/__tests__/models/record/record.ts
@@ -728,6 +728,38 @@ describe("models/record", () => {
           ).toBeGreaterThan(firstUpdate.getTime());
         });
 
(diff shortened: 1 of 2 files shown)


core/__tests__/models/record/record.ts
```diff
@@ -728,6 +728,38 @@ describe("models/record", () => {
           ).toBeGreaterThan(firstUpdate.getTime());
         });
 
+        test("array property length can grow and shrink", async () => {
```
evantahler, line 731: This test would have passed with the old query, but this explicitly confirms that previous array-Property recordProperties are getting removed

PR #2334, opened 2021-09-24
Add Hubspot marketplace headers.

pauloouriques:


plugins/@grouparoo/hubspot/src/lib/client.ts
```diff
@@ -1,95 +1,115 @@
...
 
   async createOrUpdateContact(payload: any): Promise<any> {
-    return await this.client.contacts.createOrUpdateContact(payload);
+    const properties = Object.keys(payload).map((key) => ({
+      property: key,
+      value: payload[key],
+    }));
+    return await this._request({
+      method: "POST",
+      url: `/contacts/v1/contact/createOrUpdate/email/${payload.email}/`,
```
bleonard, line 89: I think we might need some url encoding on these url parts.

bleonard: Can you check what the library was doing?

bleonard (changes requested):

bleonard:
Oops. I messed up. These headers were supposed to be for Zendesk. 😿

PR #2338, opened 2021-09-27
Reload the record if import-export failed

evantahler:
This PR follows https://github.com/grouparoo/grouparoo/pull/2330 - now if the `importAndExport` API returns an error, we still reload the record's information in the UI, as there may have been a partial import.

### DIFF
--- a/core/src/modules/ops/record.ts
+++ b/core/src/modules/ops/record.ts
@@ -574,8 +574,6 @@ export namespace RecordOps {
       }
 
       return record;
-    } catch (error) {
-      throw error;
     } finally {
       if (toLock) await releaseLock();
     }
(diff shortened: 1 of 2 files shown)


core/src/modules/ops/record.ts
```diff
@@ -574,8 +574,6 @@ export namespace RecordOps {
       }
 
       return record;
-    } catch (error) {
-      throw error;
```
evantahler, line 578: This is doing nothing

krishnaglick: Shouldn't you remove the `try {}` as well?

evantahler: There's a `finally` block below that we still want to execute. 

We've got

```ts
try {
 // whatever happens
} finally {
  if (toLock) await releaseLock();
}
```

krishnaglick (approved):

PR #2340, opened 2021-09-27
`@grouparoo/cloudwatch` plugin

evantahler:
The `@grouparoo/cloudwatch` plugin will send Grouparoo application logs to AWS cloudwatch!

Screen Shot 2021-09-27 at 12 11 08 PM

Outstanding Questions:
* If we use the default `info` level, will that be too much data being stored? If we send log messages with the `notice` level or higher, will we miss important messages (eg: a task is running slow)?
* Is it OK to use the environment variables as chosen (eg: assume there is one collection of S3 credentials for the application to use)

### DIFF
--- a/plugins/@grouparoo/cloudwatch/package.json
+++ b/plugins/@grouparoo/cloudwatch/package.json
@@ -35,9 +35,12 @@
   "devDependencies": {
     "@grouparoo/core": "0.7.0-alpha.7",
     "@grouparoo/spec-helper": "0.7.0-alpha.7",
+    "@types/jest": "*",
     "@types/node": "*",
     "actionhero": "27.1.4",
+    "jest": "27.1.0",
     "prettier": "2.3.2",
+    "ts-jest": "27.0.5",
     "type-fest": "2.1.0",
     "typescript": "4.4.2"
   },
--- a/plugins/@grouparoo/cloudwatch/package.json
+++ b/plugins/@grouparoo/cloudwatch/package.json
@@ -1,0 +1,57 @@
+{
+  "author": "Grouparoo Inc <hello@grouparoo.com>",
+  "name": "@grouparoo/cloudwatch",
+  "description": "The Grouparoo Cloudwatch Plugin",
+  "version": "0.7.0-alpha.7",
+  "license": "MPL-2.0",
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "engines": {
+    "node": ">=12.0.0"
+  },
+  "homepage": "https://www.grouparoo.com",
+  "bugs": {
+    "url": "https://github.com/grouparoo/grouparoo/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/grouparoo/grouparoo.git",
+    "directory": "plugins/@grouparoo/cloudwatch"
+  },
+  "scripts": {
+    "prepare": "rm -rf dist && tsc --declaration",
+    "test": "echo 'no tests'",
+    "pretest": "npm run lint && npm run prepare",
+    "lint": "prettier --check src __tests__",
+    "watch": "tsc --watch"
+  },
+  "dependencies": {
+    "aws-sdk": "2.995.0",
+    "winston": "3.3.3",
+    "winston-cloudwatch": "3.0.2"
+  },
+  "devDependencies": {
+    "@grouparoo/core": "0.7.0-alpha.7",
+    "@grouparoo/spec-helper": "0.7.0-alpha.7",
+    "@types/jest": "*",
+    "@types/node": "*",
+    "actionhero": "27.1.4",
+    "jest": "27.1.0",
+    "prettier": "2.3.2",
+    "ts-jest": "27.0.5",
+    "type-fest": "2.1.0",
+    "typescript": "4.4.2"
+  },
+  "grouparoo": {
+    "env": {
+      "api": [
+        "S3_ACCESS_KEY",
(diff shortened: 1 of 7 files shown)


plugins/@grouparoo/cloudwatch/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "url": "https://github.com/grouparoo/grouparoo/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/grouparoo/grouparoo.git",
+    "directory": "plugins/@grouparoo/cloudwatch"
+  },
+  "scripts": {
+    "prepare": "rm -rf dist && tsc --declaration",
+    "test": "echo 'no tests'",
```
evantahler, line 25: I'm not really sure how to test this.  I also don't know if it is required, since we are just relying on the library.

plugins/@grouparoo/cloudwatch/src/initializers/plugin.ts
```diff
@@ -0,0 +1,53 @@
...
+    const cloudwatchLogger = new WinstonCloudWatch({
+      name,
+      level: "notice" as ActionheroLogLevel,
+      logGroupName: name,
+      logStreamName: () => {
+        let date = new Date().toISOString().split("T")[0];
+        return `${name}-${date}`;
+      },
+      jsonMessage: true,
+      retentionInDays: 30, // store the logs for 30 days
```
evantahler: Store logs for 30 days

plugins/@grouparoo/cloudwatch/src/initializers/plugin.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (!process.env.S3_SECRET_ACCESS_KEY) return;
+    if (!process.env.S3_REGION) return;
+
+    // try to extract the name of the instance from the URL we are using
+    // we could alternatively use the ClusterName setting
+    let name = `grouparoo-${env}`;
+    if (process.env.WEB_URL) {
+      const parsed = new URL(process.env.WEB_URL);
+      name = parsed.host.split(":")[0].replace(/\./g, "-");
+    }
```
evantahler: @pedroslopez is there a better way to get the name of a running cloud instance?

pedroslopez: we could always inject a new env var `GROUPAROO_CLOUD_INSTANCE_NAME` or something similar

plugins/@grouparoo/cloudwatch/src/initializers/plugin.ts
```diff
@@ -0,0 +1,53 @@
...
+  constructor() {
+    super();
+    this.name = packageJSON.name;
+    this.loadPriority = 1;
+  }
+
+  async initialize() {
+    if (!process.env.S3_ACCESS_KEY) return;
+    if (!process.env.S3_SECRET_ACCESS_KEY) return;
+    if (!process.env.S3_REGION) return;
```
pedroslopez, line 18: These are the same env vars used in other parts of the app, right? Something interesting here is that we would be pointing these to our own S3 bucket, so other things like loading csvs from s3 won't work, is that right? I think these should be distinct so that users can set up s3 with their details if necessary.

We would also be passing our s3 credentials to the app now, which depending on how we set it up might mean this access key having access to push other people's logs as well if they somehow get to the env variables. They can't be accessed through calculated properties and I believe that's the only code hosted users can inject at the moment, so it's fine. Just something to keep in mind. It might be good to give each of these apps a different key that restricts access to their stuff. More to talk about here when we set it up.

pedroslopez (changes requested):
Nice! I think we want to use different auth credentials though.

teallarson:
Notice level seems right.  We may want to keep an eye on pricing as CloudWatch charges to ingest the logs we're sending it. 
 Though it does have the benefit of being more human-friendly.  https://stackoverflow.com/questions/55472308/cloudwatch-log-store-costing-vs-s3-costing

plugins/@grouparoo/cloudwatch/src/initializers/plugin.ts
evantahler: Well, we did recently remove the files systems, so there is no other place in the App that uses `AWS*` environment variables.  In the cloud app, since you are pre-signing the config download, I *think* this is the only S3 Auth in use?

For customers deploying on their own, I *think* they would want to make only one s3 account "for the app", and then give that user only the permissions it needs.

evantahler:
> Though it does have the benefit of being more human-friendly. https://stackoverflow.com/questions/55472308/cloudwatch-log-store-costing-vs-s3-costing

I think that's the goal! We want to be able to read these logs in real time to diagnose problems with the app. 

We could store the logs for less than 30 days (7 days is an option, as is a few hours) to help control costs as well.

evantahler:
Went down to 7 days of log storage

plugins/@grouparoo/cloudwatch/src/initializers/plugin.ts
pedroslopez: ah! you're right- these env variables were required when _uploading_ files to s3 which we no longer support. I think this is ok then.

pedroslopez (approved):
👍🏼  we're no longer using the s3 env vars in other plugins since we're not uploading things directly.

evantahler:
This plugin now also responds to `GROUPAROO_LOG_LEVEL` so we can toggle on more or less logging as needed for future debugging.

PR #2341, opened 2021-09-27
View records with invalid values

krishnaglick:


core/src/modules/ops/record.ts
```diff
@@ -144,16 +146,40 @@ export namespace RecordOps {
...
+      }
+    }
 
     // Are we searching for a specific RecordProperty?
     if (searchKey && searchValue) {
       countRequiresIncludes = true;
-      include.push(RecordProperty);
+      if (state !== "invalid") {
+        include.push(RecordProperty);
+      }
```
krishnaglick: Don't love this, open to ideas.

core/src/modules/ops/record.ts
```diff
@@ -260,7 +286,8 @@ export namespace RecordOps {
       recordId: string;
       position: number;
       rawValue: string;
-      invalidValue?: string;
+      invalidValue?: string | null;
+      invalidReason?: string | null;
```
krishnaglick: We assign these things to null a lot. Doing this will help once we turn on strict mode.

core/src/actions/properties.ts
```diff
@@ -48,17 +54,19 @@ export class PropertiesList extends AuthenticatedAction {
...
-            where: { state: ["draft", "ready"] },
-          },
-        ],
-        limit: params.limit,
-        offset: params.offset,
-        order: params.order,
+        include,
+        limit: Number(params.limit),
+        offset: Number(params.offset),
+        order: params.order as unknown as Order,
```
krishnaglick: I originally added in the `invalid` checking functionality here, but ended up not needing it. These changes don't really hurt anything I feel, but can revert.

core/__tests__/actions/records.ts
```diff
@@ -229,6 +230,50 @@ describe("actions/records", () => {
...
+        csrfToken,
+        state: "invalid",
+      };
+      const { records: invalidRecords, total: invalidTotal } =
+        await specHelper.runAction<RecordsList>("records:list", connection);
+      expect(invalidRecords.length).toBe(1);
+      expect(invalidTotal).toBe(1);
+    });
+
+    test("a writer can find a record with an invalid state, even if the queried record is not invalid", async () => {
```
krishnaglick: There might be a better way to phrase this but effectively if any recordProperty is invalid for a user and you're looking for a recordProperty on that user that _is not invalid AND you want invalid records_ this is confirming it will show up.

core/src/modules/ops/record.ts
```diff
@@ -207,6 +211,20 @@ export namespace RecordOps {
...
 
+    // are we looking for invalid records
+    if (state === "invalid") {
+      ands.push({
+        id: {
+          [Op.in]: Sequelize.literal(`(
+            SELECT DISTINCT "GrouparooRecord"."id"
+            FROM "records" AS "GrouparooRecord"
+            LEFT OUTER JOIN "recordProperties" AS "recordProperties" ON "GrouparooRecord"."id" = "recordProperties"."recordId"
+            WHERE "recordProperties"."invalidReason" IS NOT NULL
```
krishnaglick: Room for improvement likely

core/src/modules/ops/record.ts
```diff
@@ -144,11 +146,13 @@ export namespace RecordOps {
       caseSensitive = true;
 
     const ands: (Sequelize.Utils.Where | WhereAttributeHash)[] = [];
-    const include: Array<any> = [];
+    const include: Sequelize.Includeable[] = [];
     let countRequiresIncludes = false;
 
     // Are we searching for GrouparooRecords in a specific state?
-    if (state) ands.push({ state });
+    if (state && state !== "invalid") {
```
krishnaglick, line 153: I think we should consider making state an enum.

core/__tests__/actions/records.ts
```diff
@@ -126,7 +127,7 @@ describe("actions/records", () => {
       const record = await GrouparooRecord.findOne({
         where: { id: recordData.id },
       });
-      await record.destroy();
+      await record?.destroy();
```
evantahler: Why a `?` - there are a few of these added

core/__tests__/actions/records.ts
evantahler: I think the test case is clear!

core/src/modules/ops/record.ts
evantahler: They already are!  You can say that state comes from the model:

Screen Shot 2021-09-29 at 10 31 22 AM

core/src/modules/ops/record.ts
evantahler: I don't think I understand this comment - we've been using `undefined` previously - why change them to null?

core/src/modules/ops/recordProperty.ts
```diff
@@ -66,10 +67,11 @@ export namespace RecordPropertyOps {
       if (recordProperty && error instanceof Error) {
         error.message += ` for record ${recordProperty.recordId}`;
       }
+      invalidReason = `Invalid ${type} Value`;
```
evantahler: Probably a lowercase `v` in `Value`

core/src/models/Property.ts
```diff
@@ -286,6 +286,9 @@ export class Property extends LoggedModel<Property> {
       isArray: this.isArray,
       createdAt: APIData.formatDate(this.createdAt),
       updatedAt: APIData.formatDate(this.updatedAt),
+      recordProperties: this.RecordProperties
+        ? await Promise.all(this.RecordProperties.map((rp) => rp.apiData()))
+        : undefined,
```
evantahler, line 291: What's this doing?  the collection of record properties is likely huge for any given Property (eg: there are a million instances of "email address").

core/src/actions/properties.ts
evantahler: Typing params is always helpful, but it looks like you made the inputs all a multi type then coerced them back - wouldn't it have skipped a step if you typed limit to be a number at the start of the `runWithinTransaction` method?

core/src/modules/ops/record.ts
evantahler: We've seen in the past that sub-queries really harm performance given our data-size.  Let's brainstorm if there's another way to get at this.  If not, maybe the UI changes when you choose the "invalid state" so you can only choose to pick profiles with an invalid property key, not a value?

core/__tests__/actions/records.ts
krishnaglick: Ah, yeah. That strictness thing. Lemme revert.

core/src/modules/ops/record.ts
krishnaglick: Hmm, yeah. Reverting this.

core/src/actions/properties.ts
krishnaglick: I think this was left over from before I understood the formatter work. Removing

core/__tests__/actions/records.ts
```diff
@@ -741,6 +792,19 @@ describe("actions/records", () => {
...
+        connection.params = {
+          csrfToken,
+          searchKey: "ltv",
+          searchValue: "null",
+        };
+        const { error, records, total } =
+          await specHelper.runAction<RecordsList>("records:list", connection);
+        expect(error).toBeUndefined();
+        expect(records.length).toBe(4);
+        expect(total).toBe(4);
```
krishnaglick: I feel like this got shuffled in as part of a rebase or something.

core/src/modules/ops/record.ts
```diff
@@ -969,11 +984,46 @@ export namespace RecordOps {
...
+      `
+      SELECT DISTINCT
+        "GrouparooRecord"."id"
+      FROM
+        "records" AS "GrouparooRecord"
+        INNER JOIN "recordProperties" AS "recordProperties" ON "GrouparooRecord"."id" = "recordProperties"."recordId"
+      WHERE
+        "GrouparooRecord"."invalid" = FALSE
+        AND "recordProperties"."invalidReason" IS NOT NULL
+    `,
```
krishnaglick: Please 👀

apps/staging-enterprise/package.json
```diff
@@ -50,7 +50,8 @@
   },
   "scripts": {
     "start": "cd node_modules/@grouparoo/core && ./bin/start",
-    "dev": "cd node_modules/@grouparoo/core && ./bin/dev"
+    "dev": "cd node_modules/@grouparoo/core && ./bin/dev",
+    "debug": "cd node_modules/@grouparoo/core && ./bin/debug"
```
evantahler, line 54: Please also add this to the other 2 apps

krishnaglick:
Working on a reverse portion of the task where records without any invalid properties are set to valid.

core/src/actions/properties.ts
```diff
@@ -37,25 +42,43 @@ export class PropertiesList extends AuthenticatedAction {
...
   }
 
-  async runWithinTransaction({ params }) {
-    const includeExamples =
-      params.includeExamples === "true" || params.includeExamples === true;
+  async runWithinTransaction({
+    params,
+  }: {
+    params: {
+      state: GrouparooRecord["state"];
```
evantahler: this is actually more like `GrouparooRecord["state"] & "invalid"`

core/src/migrations/000085-addRecordsPropertyInvalid.ts
```diff
@@ -0,0 +1,18 @@
...
+export default {
+  up: async (
+    queryInterface: Sequelize.QueryInterface,
+    DataTypes: typeof Sequelize
+  ) => {
+    await queryInterface.addColumn("records", "invalid", {
+      type: DataTypes.BOOLEAN,
+      allowNull: false,
+      defaultValue: false,
+    });
```
evantahler: This migration would 'suppress' any future failures to set `invalid` by converting them into "false" at the DB level.  Usually, we do a 2-phase migration that 1) adds the column and updates the old values and 2) changes the column to be non-null without a default so we get warnings if the code forgets:

```ts
await queryInterface.addColumn("records", "invalid", {
      type: DataTypes.BOOLEAN,
      allowNull: false,
      defaultValue: false,
    });

await queryInterface.updateColumn("records", "invalid", {
      type: DataTypes.BOOLEAN,
      allowNull: false,
    });
```

core/src/models/GrouparooRecord.ts
```diff
@@ -55,6 +55,11 @@ export class GrouparooRecord extends LoggedModel<GrouparooRecord> {
   @Column(DataType.ENUM(...STATES))
   state: typeof STATES[number];
 
+  @AllowNull(false)
+  @Default(false)
```
evantahler, line 59: 👍 this is the right place for the default value

evantahler: This should be added to the `GrouparooRecord.apiData()` serializer

core/src/modules/ops/record.ts
```diff
@@ -207,6 +210,14 @@ export namespace RecordOps {
       ands.push({ modelId });
     }
 
+    // are we looking for invalid records
+    if (state === "invalid") {
+      countRequiresIncludes = true;
```
evantahler, line 215: is it required?  Wouldn't this be like the state filter?

core/src/modules/ops/record.ts
evantahler: I don't think you need to return anything here - this whole method should only be operating on the collection of `records` found by the first query.  That's the group of records found that are ready to go ready, limited by `limit`.

So this query's job is to update `record.invalid` for only those already found record IDs.

```ts
// psuedocode

await GrouparooRecord.update(
  {invalid:true},
  {where:{id:{in:  `select distinct recordId from recordProperties where recordId in PRV_RECORD_IDS and invalidReason IS NOT NULL` }}}
})

// and then 

await GrouparooRecord.update(
  {invalid:false},
  {where:{id:{in:  `select distinct recordId from recordProperties where recordId in PRV_RECORD_IDS and invalidReason IS NULL` }}}
})
```

evantahler (changes requested):

core/src/modules/ops/record.ts
krishnaglick: Record state isn't being set to invalid, a top level flag is.

core/src/migrations/000085-addRecordsPropertyInvalid.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default {
+  up: async (
+    queryInterface: Sequelize.QueryInterface,
+    DataTypes: typeof Sequelize
+  ) => {
+    await queryInterface.addColumn("records", "invalid", {
+      type: DataTypes.BOOLEAN,
+      allowNull: false,
+    });
```
evantahler: If you add a new column in a migration that is non-null, you'll need to immediately then set the existing rows to /something/.

core/src/modules/ops/record.ts
```diff
@@ -945,38 +960,112 @@ export namespace RecordOps {
...
-    ;
+      SELECT "id" from "records" where "state" = 'pending'
+      EXCEPT
+      SELECT DISTINCT("recordId") FROM "recordProperties" WHERE "state" = 'pending'
+      LIMIT ${limit};
     `,
       {
         type: QueryTypes.SELECT,
         model: GrouparooRecord,
+        include: RecordProperty,
```
evantahler: whoa! I wouldn't have expected this to work since you aren't joining the record properties here.  How does this modify the SQL generated?

core/src/modules/ops/record.ts
```diff
@@ -945,38 +960,112 @@ export namespace RecordOps {
...
+
+    // const [, updateResponse] = await GrouparooRecord.update(
+    //   { state: "ready" },
+    //   {
+    //     where: {
+    //       id: { [Op.in]: records.map((p) => p.id) },
+    //       state: "pending",
+    //     },
+    //   }
+    // );
```
evantahler: Remove please

core/src/modules/ops/record.ts
```diff
@@ -945,38 +960,112 @@ export namespace RecordOps {
...
       {
         where: {
-          id: { [Op.in]: records.map((p) => p.id) },
-          state: "pending",
+          id: {
+            [Op.in]: invalidRecords.map((record) => record.id),
+          },
         },
       }
     );
```
evantahler: I bet there's a single statement you can use to update the records without selecting them

core/src/modules/ops/record.ts
```diff
@@ -969,11 +987,21 @@ export namespace RecordOps {
...
+      SET
+        "invalid" = TRUE
+      WHERE
+        "records"."id" IN(
+          SELECT
+            "recordProperties"."recordId" FROM "recordProperties"
+          WHERE
+            "recordProperties"."recordId" IN(${records.map((p) => `'${p.id}'`)})
+            AND "recordProperties"."invalidReason" IS NOT NULL)
+        AND "records"."invalid" = FALSE;
```
krishnaglick: I thought I had this working without a sub-query but I was wrong. Not including a sub-query just resulted in all matching records getting set to invalid if _any_ were invalid.

evantahler: Got it - let's go with this strategy then!  Since there's a strong limit in play (`"recordProperties"."recordId" IN(${records.map((p) => `'${p.id}'`)})`) the size of that sub-select shouldn't be too big

evantahler: This last AND `AND "records"."invalid" = FALSE;` is probably not needed as you just updated all the rercods in the set to be `invalid:false` in the previous query

evantahler (approved):
🥳 

Just some minor cleanup on that SQL, a rebase, and then I think this is good to go!

PR #2343, opened 2021-09-28
Salesforce option and config issue

pauloouriques:
- add config key migration;
 - rename shared config in demo plugin.

### DIFF
--- a/core/src/migrations/000084-optionsProfileToRecords.ts
+++ b/core/src/migrations/000084-optionsProfileToRecords.ts
@@ -2,44 +2,58 @@ import Sequelize from "sequelize";
 
 export default {
   up: async (queryInterface: Sequelize.QueryInterface) => {
-    await queryInterface.sequelize.query(
-      `UPDATE "config" SET "key"='membershipRecordField' WHERE "key"='membershipProfileField'`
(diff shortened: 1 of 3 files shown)


bleonard (approved):
Yeah. That might be it!

core/src/migrations/000084-optionsProfileToRecords.ts
```diff
@@ -0,0 +1,45 @@
+import Sequelize from "sequelize";
+
+export default {
+  up: async (queryInterface: Sequelize.QueryInterface) => {
+    await queryInterface.sequelize.query(
+      `UPDATE "config" SET "key"='membershipRecordField' WHERE "key"='membershipProfileField'`
```
evantahler: ... we don't have a `config` table?

evantahler: Perhaps this should be about updating the `options` table?   Also, you'll likely need to do a SELECT query first to find the destinations that are of the type `salesforce-object-export` (I *think* that's the same), then update the options that are owned by those destinations

evantahler (changes requested):

PR #2344, opened 2021-09-28
Update cloudwatch ENV variables to be generic for AWS

evantahler:
Use better names for the AWS Environment Variables 

- `AWS_ACCESS_KEY_ID`
- `AWS_SECRET_ACCESS_KEY`
- `AWS_REGION`

### DIFF
--- a/plugins/@grouparoo/cloudwatch/src/initializers/plugin.ts
+++ b/plugins/@grouparoo/cloudwatch/src/initializers/plugin.ts
@@ -13,9 +13,9 @@ export default class CloudwatchInitializer extends Initializer {
   }
 
   async initialize() {
-    if (!process.env.S3_ACCESS_KEY) return;
-    if (!process.env.S3_SECRET_ACCESS_KEY) return;
-    if (!process.env.S3_REGION) return;
+    if (!process.env.AWS_ACCESS_KEY) return;
+    if (!process.env.AWS_SECRET_ACCESS_KEY) return;
+    if (!process.env.AWS_REGION) return;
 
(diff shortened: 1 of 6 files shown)


plugins/@grouparoo/cloudwatch/src/initializers/plugin.ts
```diff
@@ -35,9 +35,9 @@ export default class CloudwatchInitializer extends Initializer {
       },
       jsonMessage: true,
       retentionInDays: 7, // store the logs for 7 days
-      awsAccessKeyId: process.env.S3_ACCESS_KEY,
-      awsSecretKey: process.env.S3_SECRET_ACCESS_KEY,
-      awsRegion: process.env.S3_REGION,
+      awsAccessKeyId: process.env.AWS_ACCESS_KEY,
```
pedroslopez: Can we use `AWS_ACCESS_KEY_ID` instead of `AWS_ACCESS_KEY`? When setting it up I was a bit confused what exactly it was. I've seen `AWS_ACCESS_KEY_ID` used more often by other tools and is even auto loaded by the sdk https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/loading-node-credentials-environment.html

pedroslopez (changes requested):

plugins/@grouparoo/cloudwatch/src/initializers/plugin.ts
evantahler: You got it!

pedroslopez (approved):
Sweet!

PR #2346, opened 2021-09-29
Rename configs and create options migrations.

pauloouriques:
- add config key migration;
 - rename shared config in demo plugin.


core/src/migrations/000084-optionsProfileToRecords.ts
```diff
@@ -0,0 +1,59 @@
+import Sequelize from "sequelize";
```
evantahler: This migration looks good now, but it probably belongs within the salesforce plugin - it only applies to options from salesforce destinations.  Please move this into the plugin and it will be good-to-go!

evantahler (changes requested):

evantahler (changes requested):
There already is a salesforce/migration-000001 - this would be 000002

Also, please note the naming convention of `######-plugin-${pluginName}-${description}.ts`. We do this to "namespace" the migrations so they don't conflict with core or another plugin.  The name of this migration might be something like `000002-plugin-salesforce-updateOptionNamesForRecords.ts`

evantahler (approved):

PR #2350, opened 2021-09-29
`RecordProperties` do not create Logs

evantahler:
This PR removes Logging from RecordProperties.  This was creating a large volume of logs on every import.  These logs are not required as we store various timestamps on each RecordProperty to share the value's freshness.  This also removed a place in which PII was stored, as we were previously storing old and new property values in the log messages.  

In our demo dataset importing 1000 records:
* Before this PR: 26,247 Log Rows
* After this PR: 4,282 Log rows 

... that's ~85% less logs!

### Development

### Impact

 * This PR should /improve/ database performance! 

### Code review

### DIFF
--- a/core/src/models/Export.ts
+++ b/core/src/models/Export.ts
@@ -293,10 +293,8 @@ export class Export extends Model {
   }
 
   @BeforeCreate
-  static generateId(instance) {
-    if (!instance.id) {
-      instance.id = `${instance.idPrefix()}_${uuid.v4()}`;
-    }
+  static generateId(instance: Export) {
+    if (!instance.id) instance.id = `${instance.idPrefix()}_${uuid.v4()}`;
   }
 
   @BeforeSave
--- a/core/src/models/Export.ts
+++ b/core/src/models/Export.ts
@@ -1,19 +1,14 @@
 import {
-  Model,
   Table,
   Column,
-  CreatedAt,
-  UpdatedAt,
   AllowNull,
   Is,
   BelongsTo,
-  BeforeCreate,
   BeforeSave,
   ForeignKey,
   DataType,
   Default,
 } from "sequelize-typescript";
-import * as uuid from "uuid";
 import { Destination } from "./Destination";
 import { GrouparooRecord } from "./GrouparooRecord";
 import { plugin } from "../modules/plugin";
@@ -26,6 +21,7 @@ import { api, config } from "actionhero";
 import { ExportProcessor } from "./ExportProcessor";
 import { Errors } from "../modules/errors";
 import { PropertyTypes } from "./Property";
+import { CommonModel } from "../classes/commonModel";
 
 /**
  * The GrouparooRecord Properties in their normal data types (string, boolean, date, etc)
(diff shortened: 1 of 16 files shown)


core/src/models/GrouparooModel.ts
```diff
@@ -90,13 +90,6 @@ export class GrouparooModel extends LoggedModel<GrouparooModel> {
     return instance;
   }
 
-  @BeforeCreate
-  static generateId(instance: GrouparooModel) {
-    if (!instance.id) {
-      instance.id = `${instance.idPrefix()}_${uuid.v4()}`;
-    }
-  }
```
evantahler, line 98: This was redundant with the method within `LoggedModel` that GrouparooModel extends

core/src/models/Import.ts
```diff
@@ -230,9 +230,7 @@ export class Import extends Model {
 
   @BeforeCreate
   static generateId(instance: Import) {
-    if (!instance.id) {
-      instance.id = `${instance.idPrefix()}_${uuid.v4()}`;
-    }
+    if (!instance.id) instance.id = `${instance.idPrefix()}_${uuid.v4()}`;
```
evantahler: Cleaning up other instances that don't extend LoggedModel too

core/src/models/RecordProperty.ts
```diff
@@ -25,7 +30,10 @@ const STATE_TRANSITIONS = [
 ];
 
 @Table({ tableName: "recordProperties", paranoid: false })
-export class RecordProperty extends LoggedModel<RecordProperty> {
+export class RecordProperty extends Model {
+  @Column({ primaryKey: true })
+  id: string;
```
evantahler: Brining back id, createdAt, and updatedAt that we previously inherited from LoggedModel

core/src/models/Export.ts
```diff
@@ -293,10 +293,8 @@ export class Export extends Model {
   }
 
   @BeforeCreate
-  static generateId(instance) {
-    if (!instance.id) {
-      instance.id = `${instance.idPrefix()}_${uuid.v4()}`;
-    }
+  static generateId(instance: Export) {
+    if (!instance.id) instance.id = `${instance.idPrefix()}_${uuid.v4()}`;
```
bleonard: Why do these keep happening. Some sort of linter?

core/src/models/RecordProperty.ts
```diff
@@ -25,7 +30,10 @@ const STATE_TRANSITIONS = [
 ];
 
 @Table({ tableName: "recordProperties", paranoid: false })
-export class RecordProperty extends LoggedModel<RecordProperty> {
```
bleonard, line 28: Not 100% sure we'll be happy with the decision if `RecordProperty` should share any behaviors now or in the future with all the other ones. Maybe `LoggedModel` should really be called `CommonModel` and logging is turned off. Something to keep an eye on.

core/src/models/Export.ts
bleonard: fwiw and i don't feel that strongly about it, but I think the braces prevent mistakes even though one line looks nicer.

core/src/models/RecordProperty.ts
bleonard: For example, id generation, timestamps, etc

bleonard (approved):

core/src/models/Export.ts
evantahler: Nah, it's me manually doing it with a find/replace when I find some duplicated code to clean up.  Prettier seems to be OK with it, and I find them much easier to read.

core/src/models/RecordProperty.ts
evantahler: Yeah - we've got ~6 models now that don't extend Logged Model.  I like that though - I'll make a chain `CommonModel -> LoggedModel -> ActualModel`

evantahler:
Now all models extend `CommonModel<T>`, which includes the id and timestamp columns, as well as some utilities around formatting IDs.

PR #2351, opened 2021-09-29
Demo gets models

bleonard:
`roo demo` now adds two models adds two models: `users` and `admins`. This is the same as `roo demo b2c`
There is no more `roo demo purchases`
You can do `roo demo users` or `roo demo admins` or `roo demo accounts` to get specific models.

Also removes the `-f` flag. I always wanted to delete the config directory even when not using config. This prevents errors when it still looks at the config after running the command.

Generally restrutectured the files to account for model specific ids and such


evantahler (approved):
Lots changed, but this looks great!

I also always delete the config folder, so 👍 on that too.

PR #2353, opened 2021-09-30
Destinations can track all Records of a Model

evantahler:
This PR adds the ability for a Destination to track and export all Records in a Model, instead of the members of a Group.

Screen Shot 2021-09-30 at 10 42 33 AM

Code-wise, the previous destination tracking methods have all been replaced with a single utility method: `destination.updateTracking(type:string, id: string)`

```ts
// track a group
await destination.updateTracking('group, 'abc123')

// track a model
await destination.updateTracking('model')

// track nothing
await destination.updateTracking('none')
```

### Development

### Impact

* Migration and backfill

### Code review

### DIFF
--- a/core/__tests__/actions/destinations.ts
+++ b/core/__tests__/actions/destinations.ts
@@ -658,10 +658,13 @@ describe("actions/destinations", () => {
         id,
         groupId: "_none",
       };
-      const { destination } = await specHelper.runAction<DestinationEdit>(
-        "destination:edit",
-        connection
-      );
+      const { destination, error } =
+        await specHelper.runAction<DestinationEdit>(
+          "destination:edit",
+          connection
+        );
+
+      expect(error).toBeUndefined();
       expect(destination.group).toBe(null);
     });
 
--- a/core/__tests__/actions/destinations.ts
+++ b/core/__tests__/actions/destinations.ts
@@ -331,7 +331,7 @@ describe("actions/destinations", () => {
         connection.params = {
           csrfToken,
           id,
-          trackedGroupId: group.id,
+          groupId: group.id,
         };
 
         const { destination, run, error } =
@@ -505,7 +505,7 @@ describe("actions/destinations", () => {
         connection.params = {
           csrfToken,
           id,
-          trackedGroupId: null,
+          groupId: null,
         };
         const {
           destination: updatedDestination,
@@ -526,7 +526,7 @@ describe("actions/destinations", () => {
         connection.params = {
           csrfToken,
           id,
-          trackedGroupId: group.id,
+          groupId: group.id,
         };
         const { destination: _destination } =
           await specHelper.runAction<DestinationEdit>(
@@ -645,7 +645,7 @@ describe("actions/destinations", () => {
       connection.params = {
         csrfToken,
         id,
-        trackedGroupId: "_none",
+        groupId: "_none",
       };
       const { destination } = await specHelper.runAction<DestinationEdit>(
         "destination:edit",
--- a/core/__tests__/actions/destinations.ts
+++ b/core/__tests__/actions/destinations.ts
@@ -24,6 +24,7 @@ import {
   DestinationView,
 } from "../../src/actions/destinations";
 import { SessionCreate } from "../../src/actions/session";
+import { APIData } from "../../src/modules/apiData";
 
 describe("actions/destinations", () => {
   helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
@@ -492,7 +493,7 @@ describe("actions/destinations", () => {
         await colorProperty.destroy();
       });
 
-      test("an administrator can list and remove a tracked group", async () => {
+      test("an administrator can view a destination", async () => {
         connection.params = {
           csrfToken,
           id,
@@ -501,12 +502,37 @@ describe("actions/destinations", () => {
           "destination:view",
           connection
         );
+        expect(destination.id).toBe(id);
         expect(destination.group.id).toBe(group.id);
+      });
 
+      test("an administrator can track a model", async () => {
         connection.params = {
           csrfToken,
           id,
-          groupId: null,
+          collection: "model",
+        };
+        const { destination, oldRun, newRun, error } =
+          await specHelper.runAction<DestinationEdit>(
+            "destination:edit",
+            connection
+          );
+        expect(error).toBeFalsy();
+
+        expect(oldRun.creatorId).toBe(group.id);
+        expect(oldRun.force).toBe(false);
+        expect(oldRun.state).toBe("running");
+
+        expect(newRun.creatorId).toBe("mod_profiles");
+
+        expect(destination.collection).toBe("model");
+        expect(destination.group).toBe(null);
+      });
+
+      test("an administrator can remove a tracked model", async () => {
+        connection.params = {
+          csrfToken,
+          id,
           collection: null,
         };
         const {
@@ -520,10 +546,12 @@ describe("actions/destinations", () => {
         );
         expect(error).toBeFalsy();
         expect(newRun).toBeUndefined();
-        expect(oldRun.creatorId).toBe(group.id);
+        expect(oldRun.creatorId).toBe("mod_profiles");
         expect(oldRun.force).toBe(false);
         expect(oldRun.state).toBe("running");
+
         expect(updatedDestination.group).toBe(null);
+        expect(updatedDestination.collection).toBe(null);
       });
 
       test("update the tracked group", async () => {
@@ -597,7 +625,7 @@ describe("actions/destinations", () => {
         await runningRuns[0].stop();
       });
 
-      test("an administator will not trigger an export by default when updating a destination", async () => {
+      test("an administrator will not trigger an export by default when updating a destination", async () => {
         connection.params = {
           csrfToken,
           id,
@@ -656,7 +684,7 @@ describe("actions/destinations", () => {
       connection.params = {
         csrfToken,
         id,
-        groupId: "_none",
+        groupId: APIData.nullKey,
       };
       const { destination, error } =
         await specHelper.runAction<DestinationEdit>(
--- a/core/__tests__/actions/destinations.ts
+++ b/core/__tests__/actions/destinations.ts
@@ -332,18 +332,19 @@ describe("actions/destinations", () => {
           csrfToken,
           id,
           groupId: group.id,
+          collection: "group",
         };
 
-        const { destination, run, error } =
+        const { destination, newRun, error } =
           await specHelper.runAction<DestinationEdit>(
             "destination:edit",
             connection
           );
         expect(error).toBeFalsy();
-        expect(destination.destinationGroup.id).toBe(group.id);
-        expect(run.creatorId).toBe(group.id);
-        expect(run.force).toBe(true);
-        expect(run.state).toBe("running");
+        expect(destination.group.id).toBe(group.id);
+        expect(newRun.creatorId).toBe(group.id);
+        expect(newRun.force).toBe(true);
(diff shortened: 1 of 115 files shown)


plugins/@grouparoo/mysql/public/templates/destination/{{{id}}}.js.template
```diff
@@ -7,7 +7,9 @@ exports.default = async function buildConfig() {
       type: "mysql-export-records",
       appId: {{{appId}}}, // The ID of the App this Destination uses - e.g. `appId: "data_warehouse"`
       modelId: "mod_profiles", // The ID of the Grouparoo Model that this Destination will send
+      collection: "group", // What collection of Records should be sent?  "group" requires a groupId below, and "model" will sync all Records from "modelId"
       groupId: "...", // The ID of the group whose members you want to export - e.g. `groupId: "high_value_customers"`
+      syncMode: {{{syncMode}}}, // How should Grouparoo sync with this destination? Options: "sync", "additive", "enrich"
```
evantahler: Some destination templates were missing syncMode

evantahler:
https://github.com/grouparoo/grouparoo/pull/2394 will explain some testing weirdness, why runs were taking so long to complete...

core/src/migrations/000086-addDestinationRecordCollection.ts
```diff
@@ -0,0 +1,26 @@
+import Sequelize from "sequelize";
```
evantahler, line 1: Assuming https://github.com/grouparoo/grouparoo/pull/2341 comes in with migrations 84 and 85, this would be migration 86

core/__tests__/actions/destinations.ts
```diff
@@ -643,13 +681,16 @@ describe("actions/destinations", () => {
       connection.params = {
         csrfToken,
         id,
-        trackedGroupId: "_none",
+        groupId: "_none",
```
pedroslopez: should this instead be `collection: _none`?

core/src/migrations/000086-addDestinationRecordCollection.ts
```diff
@@ -0,0 +1,26 @@
...
+    });
+
+    await queryInterface.sequelize.query(
+      `UPDATE "destinations" SET "collection"='group' WHERE "groupId" IS NOT NULL AND state='ready'`
+    );
+
+    await queryInterface.addColumn("runs", "collection", {
+      type: DataTypes.STRING(191),
+      allowNull: true,
+    });
```
pedroslopez: Is this column being used? It looks like `creatorId` would be the equivalent on the `runs` table. Let's remove it if it's not needed!

core/src/modules/ops/destination.ts
```diff
@@ -40,53 +44,79 @@ export namespace DestinationOps {
...
-    if (group) return group.run(force, destination.id);
+  export async function exportMembers(destination: Destination, force = false) {
+    if (destination.collection === "group") {
+      const group = await destination.$get("group");
+      if (group) return group.run(force, destination.id);
+    } else if (destination.collection === "model") {
+      const model = await destination.$get("model");
+      if (model) return model.run(force, destination.id);
+    } else {
+      throw new Error(`cannot export members for a ${destination.collection}`);
```
pedroslopez, line 56: Notably this now throws an error if we're not tracking anything, let's make sure that we can handle this from wherever this is called. For example, when deleting the destination.

core/src/tasks/grouparooModel/run.ts
```diff
@@ -0,0 +1,98 @@
...
+  async runWithinTransaction(params) {
+    const run = await Run.scope(null).findOne({
+      where: { id: params.runId },
+    });
+
+    if (!run) return;
+    if (run.state === "stopped") return;
+
+    const offset: number = run.groupMemberOffset || 0;
+    const limit: number = run.groupMemberLimit || config.batchSize.imports;
```
pedroslopez: using `groupMember` for these fields now doesn't quite match up, maybe just `memberOffset` and `memberLimit`?

plugins/@grouparoo/mysql/public/templates/destination/{{{id}}}.js.template
pedroslopez: It doesn't look like `mysql` supports syncModes at all... 🤔

plugins/@grouparoo/mysql/src/initializers/plugin.ts
```diff
@@ -38,6 +38,9 @@ export class Plugins extends Initializer {
   }
 
   async initialize() {
+    const syncModes: DestinationSyncMode[] = ["sync", "additive", "enrich"];
```
pedroslopez: If we're adding these sync modes, some code changes are also needed in `exportRecord.ts` to handle the `syncOperations`. Maybe this is another story?

core/src/modules/ops/destination.ts
```diff
@@ -40,53 +44,79 @@ export namespace DestinationOps {
...
+      destination.collection === collection &&
+      destination.groupId === collectionId
+    ) {
+      return { oldRun, newRun }; // no changes
     }
-    if (destination.modelId !== group.modelId) {
-      throw new Error(
-        `destination ${destination.id} and group ${group.id} do not share the same modelId`
+
+    if (collection === "group" && collectionId && collectionId !== NullKey) {
```
pedroslopez: Curious as to why `_none` is being handled on the `collectionId` level instead of `collection`. It works, not a huge deal, but when calling this to stop tracking a model it would feel a bit odd that it would be calling the function with `group` collection

core/src/modules/ops/destination.ts
```diff
@@ -40,53 +44,79 @@ export namespace DestinationOps {
...
 
-  /**
-   * Track a Group
-   */
-  export async function trackGroup(
+  export async function updateTracking(
     destination: Destination,
-    group: Group,
-    force = true
+    collection: Destination["collection"] | typeof NullKey,
```
pedroslopez: Oh, I see `NullKey` is able to be passed for the `collection`, but I don't think it's being handled?

pedroslopez (changes requested):
Looking good! Just a few questions. This is definitely a nice change for those that dont really need groups.

plugins/@grouparoo/mysql/public/templates/destination/{{{id}}}.js.template
evantahler: Removing, we will do this in a new PR (https://www.pivotaltracker.com/story/show/179840171)

core/src/tasks/grouparooModel/run.ts
evantahler: Good call! https://github.com/grouparoo/grouparoo/pull/2353/commits/7d03b099fc71ce1021d9133918df91d6c40a1e0a updates all the columns on run that used to be about `group*`

core/src/migrations/000086-addDestinationRecordCollection.ts
evantahler: fixed in 53bc00eba0c7cfc989a5f1569ce99bef04bc4e6f

evantahler:
To @pedroslopez: Here are the updates since the last review:

- [x] Move the mysql and logger SyncModes to another PR
- [x] Remove the `NullKey` thing
- [x] Update the tracking API and Destination model to have non-null collection: `none`,`group`, or `model` are the 3 choices.  
- [x] Translate a `null` code-config destination.collection to "none"
- [x] migrate run#Group* to run#member* to be better named
- [x] Generally fixup the migration

core/src/models/Destination.ts
```diff
@@ -164,6 +167,11 @@ export class Destination extends LoggedModel<Destination> {
   @Column(DataType.ENUM(...SYNC_MODES))
   syncMode: DestinationSyncMode;
 
+  @AllowNull(false)
+  @Default("none")
+  @Column(DataType.ENUM(...DESTINATION_COLLECTIONS))
+  collection: DestinationCollection;
```
evantahler, line 173: Destination.collection cannot be null now

core/src/models/Destination.ts
```diff
@@ -52,6 +52,9 @@ export interface SimpleDestinationOptions extends OptionHelper.SimpleOptions {}
 const SYNC_MODES = ["sync", "additive", "enrich"] as const;
 export type DestinationSyncMode = typeof SYNC_MODES[number];
 
+const DESTINATION_COLLECTIONS = ["none", "group", "model"] as const;
```
evantahler, line 55: `"none"` is now a valid Destination.collection

core/src/migrations/000086-addDestinationRecordCollection.ts
```diff
@@ -0,0 +1,57 @@
...
+    queryInterface: Sequelize.QueryInterface,
+    DataTypes: typeof Sequelize
+  ) => {
+    await queryInterface.addColumn("destinations", "collection", {
+      type: DataTypes.STRING(191),
+      allowNull: true,
+    });
+
+    await queryInterface.sequelize.query(
+      `UPDATE "destinations" SET "collection"='group' WHERE "groupId" IS NOT NULL AND state='ready'`
```
pedroslopez: should we also set `collection` to `none` for those that don't have a group set on them?

core/src/migrations/000086-addDestinationRecordCollection.ts
```diff
@@ -0,0 +1,57 @@
...
+      "groupMemberLimit",
+      "memberLimit"
+    );
+    await queryInterface.renameColumn(
+      "runs",
+      "groupMemberOffset",
+      "memberOffset"
+    );
+    await queryInterface.renameColumn("runs", "groupMethod", "method");
+    await queryInterface.removeColumn("runs", "groupHighWaterMark");
```
pedroslopez, line 34: was this one no longer being used?

core/src/migrations/000086-addDestinationRecordCollection.ts
```diff
@@ -0,0 +1,57 @@
...
+  },
+
+  down: async (
+    queryInterface: Sequelize.QueryInterface,
+    DataTypes: typeof Sequelize
+  ) => {
+    await queryInterface.removeColumn("destinations", "recordCollection");
```
pedroslopez:
```suggestion
    await queryInterface.removeColumn("destinations", "collection");
```

core/src/modules/configLoaders/destination.ts
```diff
@@ -83,13 +84,14 @@ export async function loadDestination(
   }
   await destination.setDestinationGroupMemberships(destinationGroupMemberships);
 
-  if (group && destination.groupId !== group.id) {
-    await destination.trackGroup(group);
-  }
+  await destination.updateTracking(
+    configObject.collection ?? "none", // allow a "null" collection in config to be treated as a "none" collection
+    configObject.groupId
+  );
```
pedroslopez, line 90: This looks good, but it's interesting that people could have a `groupId` set on their config and no collection, and nothing would happen with that groupId. Should there be some kind of warning that says you have set a groupId but it has no effect because of not having `group` as the collection?

pedroslopez (changes requested):
Nice! setting the collection itself to `none` definitely feels nicer. I think there are just a few changes left on the migration, but otherwise looks good!

core/src/migrations/000086-addDestinationRecordCollection.ts
evantahler: Yeah - It was kind of silly to have something separate from the normal highwaterMark column, now we just use that

pedroslopez (approved):
woo! 🎉

evantahler:
Ha! You got your PR in first @pedroslopez, I've got one more rebase it seems :D

PR #2384, opened 2021-10-01
`CLSTasks` include the `QueueLock` Middleware

evantahler:
Most of Grouparoo's tasks that do not talk to an external system extend `CLSTask` rather than `RetryableTask`.  until this PR, `CLSTask` did not include any middleware.  However, we always want to include the `QueueLock` Middleware, which will prevent the same task with the same args being enqueued more than once.   This prevents double-enqueuing when more than one node boots up, things retry, etc.

### Development

### Impact

>

### Code review

### DIFF
--- a/core/src/classes/tasks/clsTask.ts
+++ b/core/src/classes/tasks/clsTask.ts
@@ -7,6 +7,7 @@ import { CLS } from "../../modules/cls";
 export abstract class CLSTask extends Task {
   constructor() {
     super();
+    this.plugins = ["QueueLock"];
   }
 
   async run(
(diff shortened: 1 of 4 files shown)


core/src/classes/tasks/clsTask.ts
```diff
@@ -7,6 +7,7 @@ import { CLS } from "../../modules/cls";
 export abstract class CLSTask extends Task {
   constructor() {
     super();
+    this.plugins = ["QueueLock"];
```
bleonard, line 10: This feels right. I'm not aware of any thing should exist twice with the same args. Any one them you are worried about?

bleonard (approved):

core/src/classes/tasks/clsTask.ts
evantahler: I don't think so!

PR #2387, opened 2021-10-02
Make it simpler to re-record Nocks with `NOCK=1`

bleonard:
## Change description

Trying to make it easier to record nocks.
Goal: `NOCK=1 jest`


plugins/@grouparoo/braze/__tests__/export/destination-mapping-options.ts
```diff
@@ -2,12 +2,7 @@ import "@grouparoo/spec-helper";
...
 import { destinationMappingOptions } from "../../src/lib/export/destinationMappingOptions";
 
-// these comments to use nock
-const newNock = false;
-// or these to make it true
-// const newNock = true;
-
-const appOptions = loadAppOptions(newNock);
+const appOptions = loadAppOptions(false);
 const appId = "app_mix-panel-09012378476890123-123903547";
```
evantahler, line 6: App name probably shouldn't be mixpanel

plugins/@grouparoo/onesignal/__tests__/fixtures/export/destination-mapping-options.js
```diff
@@ -0,0 +1 @@
+const nock = require("nock");
```
evantahler, line 1: This one is kind of weird - nothing happens here?

plugins/@grouparoo/snowflake/__tests__/fixtures/integration/cli-generate-from.js
```diff
@@ -0,0 +1 @@
+const nock = require("nock");
```
evantahler, line 1: Here's another one where nothing happens

evantahler (approved):
This is awesome!

I think it would helpful to document the NOCK=1 behavior somewhere... perhaps the README of `@grouparoo/spec-helper`?

plugins/@grouparoo/snowflake/__tests__/fixtures/integration/cli-generate-from.js
bleonard: I decided to leave them here because it prevents any calls. if we didn't do it, it might be talking to the remove server and we didn't know.

PR #2389, opened 2021-10-04
Global model selection in web interface

pedroslopez:
## Change description

Adds a model selection dropdown on the sidebar to indicate currently used model

image

Lots of URLs have been updated to be within a specific model, e.g. `/records => /model/drivers/records` and `/record/12345 => /model/drivers/record/12345`.

## Related issues

No
### Development

### Impact

>

### Code review


ui/ui-components/components/navigation/highlightingNavLink.tsx
```diff
@@ -4,16 +4,30 @@ import { useRouter } from "next/router";
...
   const [active, setActive] = useState(false);
 
   useEffect(() => {
-    const mainPathSection = router?.asPath.split("/")[1] || "";
-    const mainHrefSection = href.split("/")[1];
-    const active =
+    const pathParts = router?.asPath.split("/");
+    const hrefParts = href.split("/");
+    const mainPathSection = pathParts[mainPathSectionIdx] || "";
+    const mainHrefSection = hrefParts[mainPathSectionIdx];
```
pedroslopez: We were making some assumptions that we wanted to highlight links based on the first part of the url, but this meant that we were highlighting all of the URLs that started with `/model` in them when in any of those pages. This workaround passes a `mainPathSectionIdx` instead of assuming it's always `1` to consider a different part of the url as the "topic". For example, when 3 is passed and the url is `/model/abcd/records`, `records` will be used to determine if it's on that page and anything before that will be ignored.

ui/ui-components/components/visualizations/homepageWidgets.tsx
```diff
@@ -123,8 +123,8 @@ export function GroupsByNewestMember({
                 <tr key={`group-${group.id}`}>
                   <td>
                     <EnterpriseLink
-                      href="/group/[id]/rules"
-                      as={`/group/${group.id}/rules`}
+                      href="/object/[id]"
+                      as={`/object/${group.id}`}
```
pedroslopez, line 127: I'm not too happy about using `object` here, but in these dashboard widgets we don't have access to any other details to get the right `modelId`

ui/ui-components/pages/_app.tsx
```diff
@@ -48,11 +52,30 @@ const uploadHandler = new UploadHandler();
 export default function GrouparooWebApp(props) {
   const { Component, pageProps, err, hydrationError } = props;
 
+  const onChangeModelId = async (newModelId: string) => {
+    const cookies = new Cookies();
+    cookies.set("grouparooModelId", newModelId, { path: "/" });
```
pedroslopez, line 57: we're using cookies so that the currently selected model sticks around when navigating to pages that don't have models in their url. 

My initial approach was to use react context and/or state, but because of how we're loading the nav in `getInitialProps` on every page load, it proved difficult to do in a way that wasn't too hacky. Things get super weird when initializing state from props and it's in general a react anti-pattern, so using state to manage this would mean refactoring how nav works and potentially moving away from loading it on the server.

ui/ui-components/pages/dashboard.tsx
```diff
@@ -24,7 +28,6 @@ export default function Page(props) {
         <Col>
           <BigTotalNumber
             statusHandler={statusHandler}
-            href="/records"
```
pedroslopez, line 27: Removing link to model-scoped pages on the dashboard totals, since the numbers reflect total amount of records/groups and not for a certain model, though we no longer have pages that show them all. Probably something to look at when we update the dashboard for models.

ui/ui-enterprise/pages/model/[modelId]/group/[groupId]/destinations.tsx
```diff
@@ -9,11 +9,12 @@ import LoadingTable from "@grouparoo/ui-components/components/loadingTable";
 import AppIcon from "@grouparoo/ui-components/components/appIcon";
 import { Models, Actions } from "@grouparoo/ui-components/utils/apiData";
 import { ErrorHandler } from "@grouparoo/ui-components/utils/errorHandler";
-import { formatTimestamp } from "../../../../ui-components/utils/formatTimestamp";
+import { formatTimestamp } from "@grouparoo/ui-components/utils/formatTimestamp";
```
Comment (evantahler) [inline ui/ui-enterprise/pages/model/[modelId]/group/[groupId]/destinations.tsx:12]: Good fix!  - There were a few of these

core/src/actions/navigation.ts
```diff
@@ -46,6 +53,18 @@ export class NavigationList extends OptionallyAuthenticatedAction {
         ? "authenticated"
         : "unauthenticated";
 
+    const isAuthenticated =
+      navigationMode === "authenticated" ||
+      navigationMode === "config:authenticated";
+
+    const models = await GrouparooModel.findAll();
+    const currentModel = models.find((m) => m.id === params.modelId);
```
evantahler, line 63: It looks like you are loading up all the models here so you can have a fallback to use if `params.modelId` isn't provided. I'd suggest adding an `ORDER BY` here so the default model is deterministic.  Alphabetically by name or oldest might be a good choice

ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
```diff
@@ -4,6 +4,7 @@ import mockAxios from "jest-mock-axios";
 import { Actions } from "../../utils/apiData";
 import { UseApi } from "../../hooks/useApi";
 import { EventDispatcher } from "../../utils/eventDispatcher";
+import { NextPageContext } from "next";
```
evantahler, line 7: So that's you type `ctx` 🤯 !

ui/ui-components/components/import/list.tsx
```diff
@@ -100,8 +96,8 @@ export default function ImportList(props) {
                     <br /> Record:{" "}
                     {_import.recordId ? (
                       <Link
-                        href="/record/[id]/edit"
-                        as={`/record/${_import.recordId}/edit`}
+                        href="/object/[id]"
+                        as={`/object/${_import.recordId}`}
```
evantahler, line 100: Calling out another instance of relying on the `/object` page

ui/ui-components/components/log/list.tsx
```diff
@@ -10,6 +10,12 @@ import Pagination from "../pagination";
 import LoadingTable from "../loadingTable";
 import { Models, Actions } from "../../utils/apiData";
 import { ErrorHandler } from "../../utils/errorHandler";
+import { NextPageContext } from "next";
+
+const getOwnerId = (query: { [key: string]: string | string[] }) => {
+  const { id, recordId, propertyId, sourceId, destinationId, groupId } = query;
+  return id ?? recordId ?? propertyId ?? sourceId ?? destinationId ?? groupId;
+};
```
evantahler, line 18: That's a fascinating method, but it makes sense!

ui/ui-components/components/navigation/highlightingNavLink.tsx
```diff
@@ -4,16 +4,30 @@ import { useRouter } from "next/router";
...
+    let active =
       mainPathSection === mainHrefSection ||
       `${mainPathSection}s` === mainHrefSection;
+
+    // special case: models in sidebar
+    if (active && mainHrefSection === "models" && pathParts.length > 3) {
+      active = false;
+    }
```
evantahler, line 29: [Low Priority] There's something strange with the order-of-operations here - If you are on `/model/admins/groups` and then use the model drop down to change to `/model/users/groups`, the highlight flickers and then goes away.  I wonder if this `useEffect` needs to consider the `router.asPath` too

ui/ui-components/components/tabs.tsx
```diff
@@ -19,24 +19,31 @@ export default function GrouparooTabs({
   if (!router.pathname) return null;
 
   const parts = router.asPath.split("/");
-  const topic = parts[1];
-  const id = parts[2];
-  const verb = parts[3].split("?")[0];
+  const scope = parts.slice(0, parts.length - 3).join("/");
+  const topic = parts[parts.length - 3];
+  const id = parts[parts.length - 2];
+  const verb = parts[parts.length - 1].split("?")[0];
```
evantahler, line 25: 👍 still works dynamically with multiple route depths

ui/ui-components/pages/_app.tsx
```diff
@@ -86,23 +109,39 @@ export default function GrouparooWebApp(props) {
...
-GrouparooWebApp.getInitialProps = async (appContext) => {
+GrouparooWebApp.getInitialProps = async (appContext: AppContext) => {
   const { execApi } = UseApi(appContext.ctx);
   let currentTeamMember: Partial<Actions.SessionView["teamMember"]> = {
     firstName: "",
     id: null,
   };
 
+  let modelId: string;
+  if (appContext.ctx.pathname.match("/model/")) {
```
evantahler, line 120: maybe match `$\/model\/` to be sure that the URL starts with "model"

ui/ui-components/pages/_app.tsx
evantahler: [question] I agree with the need to store the `modelId` "outside" of React's state - why did you pick cookies vs localstorage?

ui/ui-components/pages/_app.tsx
```diff
@@ -48,11 +52,30 @@ const uploadHandler = new UploadHandler();
...
+        pathname: router.pathname,
+        query: {
+          ...router.query,
+          modelId: newModelId,
+        },
+      });
+    } else {
+      router.reload();
+    }
+  };
```
evantahler: [question] I don't think I understand what's happening here.

ui/ui-components/pages/dashboard.tsx
evantahler: Perhaps one day we show totals per-model

ui/ui-components/pages/model/[modelId]/destination/new/[appId].tsx
```diff
@@ -65,28 +62,6 @@ export default function Page(props) {
...
-              {models.map((model) => (
-                <option key={`model-${model.id}`} value={model.id}>
-                  {model.name}
-                </option>
-              ))}
-            </Form.Control>
-            <Form.Text className="text-muted">
-              This source will only contribute to the chosen Grouparoo Model
-            </Form.Text>
-          </Form.Group>
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/destination/new/[appId].tsx]: It's great to remove this form question!  

Screen Shot 2021-10-07 at 10 03 06 AM

I'd suggest adding the name of the model somewhere on the page to be really sure that folks know what model the destination belongs to, maybe in the title?

Comment (evantahler) [reply ui/ui-components/pages/model/[modelId]/destination/new/[appId].tsx]: The same goes for Groups and Sources

ui/ui-components/pages/object/[id].tsx
```diff
@@ -57,14 +67,75 @@ export default function FindObject(props) {
...
+
+      if (!response || !response[topic]) {
+        setError(`Cannot find object "${id}"`);
+        return;
+      }
+
+      router.replace(
+        `/model/[modelId]/${topic}/[${topic}Id]/${page}`,
+        `/model/${response[topic].modelId}/${topic}/${response[topic].id}/${page}`
+      );
```
Comment (evantahler) [inline ui/ui-components/pages/object/[id].tsx:121]: nice! `router.replace` doesn't include this page in the history, which is great

evantahler (commented):
This is one heck of a PR - 200 files changed!  Because there are lots of file, path, and import changes, this probably warrants a bug bash - we need to make sure that no links/imports are broken in all 3 of the UIs. 

👍 
* better names of the variables in the URLs
* The actions didn't change that much
* good types added when possible

Questions:
* What should happen if you are viewing an object owned by a model and then change the model.  `/model/users/group/all_emails/rules` and then change the dropdown to `/model/admins/group/all_emails/rules`.  The page still renders even though there is no group "all_emails" under the admin model.  
* should we put any effort into making the `object:find` API faster/smarter now that more pages are using it?

ui/ui-components/pages/_app.tsx
pedroslopez: I couldn't use localStorage because it's not available from the server when the first load happens

ui/ui-components/pages/_app.tsx
pedroslopez: if we're on a model page and the selected model is changed, it'll redirect you to the relevant page for the new model. So if you're viewing records for model A and switch to model B, you'll now be viewing records for model B.

if we aren't on a model page, we won't move to a new page but we still need to refresh the navigation to get links for the new model. Because we want to basically call `getInitialProps` again, I'm reloading the current page.

ui/ui-components/components/navigation/highlightingNavLink.tsx
pedroslopez: hmm, I tried adding a few dependencies to the `useEffect` hook but it didn't help with the flicker... 🤔
I did notice that pages with query strings removed the highlight, so addressed that.

ui/ui-components/pages/model/[modelId]/destination/new/[appId].tsx
Comment (pedroslopez) [reply ui/ui-components/pages/model/[modelId]/destination/new/[appId].tsx]: I agree!

How's this? Changed the view around a bit

image

pedroslopez:
> this probably warrants a bug bash - we need to make sure that no links/imports are broken in all 3 of the UIs.

Totally agree!

> * What should happen if you are viewing an object owned by a model and then change the model.  `/model/users/group/all_emails/rules` and then change the dropdown to `/model/admins/group/all_emails/rules`.  The page still renders even though there is no group "all_emails" under the admin model.

Good question! It's definitely a bit weird for the link to work even though it doesn't match up. In 98b51f0 I've added a check and will set an error if the models don't match up. I like giving an error when going directly to a page when models don't match up, but getting the error when already on a valid page and switching the model is a bit interesting. I think it makes sense though, since the URL did change. Open to suggestions!

> * should we put any effort into making the `object:find` API faster/smarter now that more pages are using it?

I think so! What's interesting about the instances where I added it in is that we actually know what type of object it is, we just don't know the exact url to get to it. So perhaps we can make `/objects/[type]/[id]` or something that does the lookup in the right table instead of overall.

ui/ui-components/pages/_app.tsx
krishnaglick: If we're looking to maybe consider a global-store solution, [redux-toolkit](https://redux-toolkit.js.org/introduction/getting-started) might be worth eyes. I'm still skeptical of redux in general, but they seem to have improved things greatly.

ui/ui-components/utils/ensureMatchingModel.ts
```diff
@@ -0,0 +1,17 @@
...
+) {
+  if (objModel !== expectedModel) {
+    const err = new Error(`Cannot find ${objType} in ${expectedModel}`);
+    if (errorHandler) {
+      errorHandler.set({ error: err });
+    } else {
+      throw err;
+    }
+  }
+}
```
evantahler, line 17: It's great that this is in a shared library - I think throwing is OK for now.  Perhaps in the future we decide to redirect to the list view or something like that... here's the one place to put that logic now.

evantahler (approved):

core/src/actions/navigation.ts
```diff
@@ -46,6 +53,20 @@ export class NavigationList extends OptionallyAuthenticatedAction {
...
+
+    const models = await GrouparooModel.findAll({
+      order: [["createdAt", "asc"]],
+    });
+    const currentModel = models.find((m) => m.id === params.modelId);
+    const currentModelId = currentModel
+      ? currentModel.id
+      : models.length > 0
+      ? models[0].id
+      : null;
```
krishnaglick, line 68:
```suggestion
    const currentModelId = currentModel?.id ?? models?.[0].id ?? null;
```

ui/ui-components/__tests__/components/setupStepsNavProgressBar.tsx
```diff
@@ -32,7 +33,7 @@ describe("setupStepsNavProgressBar", () => {
   });
 
   describe("on a logged in page", () => {
-    const { execApi } = UseApi({});
+    const { execApi } = UseApi({} as NextPageContext);
```
krishnaglick, line 36: Might be worth making UseApi default its param to `{}` if we need to do this though I don't know how widespread they may be.

ui/ui-components/components/record/list.tsx
```diff
@@ -416,18 +373,26 @@ export default function RecordsList(props) {
...
+    searchKey,
+    searchValue,
+    caseSensitive,
+  } = ctx.query;
 
   let groupId: string;
   if (id) {
     if (ctx.pathname.match("/group/")) {
-      groupId = id;
+      groupId = id as string;
```
krishnaglick, line 395:
```suggestion
      groupId = String(id);
```
Safer way to do this if you'd like

ui/ui-components/hooks/useApi.ts
```diff
@@ -1,14 +1,12 @@
+import { NextPageContext } from "next";
```
krishnaglick, line 1:
```suggestion
import type { NextPageContext } from "next";
```
It _might_ be worth using a type import. [Details here](https://stackoverflow.com/a/64243357)

ui/ui-components/pages/object/[id].tsx
```diff
@@ -57,14 +67,75 @@ export default function FindObject(props) {
...
+      `/model/[modelId]/property/[propertyId]/edit`,
+      `/model/${source.modelId}/property/${id}/edit`
+    );
+  }
+
+  function redirectTopicWithModel(topic: string, page: string = "edit") {
+    return async function (id: string) {
+      const response = await execApi("get", `/${topic}/${id}`);
+
+      if (!response || !response[topic]) {
```
Comment (krishnaglick) [inline ui/ui-components/pages/object/[id].tsx:113]: I think `!response?.[topic]` would work but that's minor.

ui/ui-config/pages/model/[modelId]/record/new.tsx
```diff
@@ -41,10 +41,19 @@ export default function NewRecord(props) {
...
       "post",
       `/record`,
-      Object.assign({}, { properties: { [data.uniqueProperty]: data.value } })
+      Object.assign(
+        {},
+        {
+          modelId: router.query.modelId,
+          properties: { [data.uniqueProperty]: data.value },
+        }
+      )
```
Comment (krishnaglick) [inline ui/ui-config/pages/model/[modelId]/record/new.tsx:50]: Why are we Object.assigning here?

krishnaglick (commented):
You merged it so this is too late, but oh well!

pedroslopez:
@krishnaglick oh no! haha, thanks for taking a look. Feel free to absolutely make these code quality changes if you'd like

PR #2390, opened 2021-10-04
Clarify auto migration logic in initializer

bleonard:
## Change description

clarifies logic for migrating.
Also adds development env

It was not migrating when `WORKERS=0` in local development/demo in `apps/staging-config`

### DIFF
--- a/core/src/config/sequelize.ts
+++ b/core/src/config/sequelize.ts
@@ -2,7 +2,7 @@ import fs from "fs";
 import { URL } from "url";
 import { join, isAbsolute } from "path";
 import { getParentPath, getPluginManifest } from "../modules/pluginDetails";
-import { log } from "actionhero";
+import { log, env } from "actionhero";
 
 import cls from "cls-hooked";
 import Sequelize from "sequelize";
@@ -75,11 +75,7 @@ export const DEFAULT = {
         storage = `${parsed.hostname}${parsed.pathname}`;
       }
 
-      if (
-        process.env.NODE_ENV === "test" &&
-        !parsed.hostname &&
-        !parsed.pathname
-      ) {
+      if (env === "test" && !parsed.hostname && !parsed.pathname) {
         storage = join(getParentPath(), `${database}.sqlite`);
       }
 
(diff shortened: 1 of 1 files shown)


core/src/config/sequelize.ts
```diff
@@ -2,7 +2,7 @@ import fs from "fs";
 import { URL } from "url";
 import { join, isAbsolute } from "path";
 import { getParentPath, getPluginManifest } from "../modules/pluginDetails";
-import { log } from "actionhero";
+import { log, env } from "actionhero";
```
evantahler: ... It might be that maybe using `{env}` was a bad idea - it might not be ready yet when the config is being calculated :/

evantahler (changes requested):

core/src/config/sequelize.ts
evantahler:
```
npm run dev

> @grouparoo/app-staging-enterprise@0.7.0-alpha.10 dev
> cd node_modules/@grouparoo/core && ./bin/dev

[nodemon] 2.0.12
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): src/**/* ../plugins/**/*.js
[nodemon] watching extensions: js,jsx,ts,tsx
[nodemon] starting `node_modules/.bin/ts-node --transpile-only --log-error src/grouparoo.ts`
{ env: undefined }
{ env: undefined }
{ env: undefined }
{ env: undefined }
{ env: undefined }
2021-10-04T22:44:55.905Z - notice: Starting Grouparoo v0.7.0-alpha.10 on node.js v16.6.1
{ env: 'development' }
{ env: 'development' }
```

core/src/config/sequelize.ts
```diff
@@ -18,6 +18,7 @@ require("pg").defaults.parseInt8 = true;
 
 export const DEFAULT = {
   sequelize: (config) => {
+    const env = process.env.NODE_ENV;
```
evantahler: On most developer machines, NODE_ENV will be `undefined`, so maybe

```suggestion
    const env = process.env.NODE_ENV ?? "development";
```

evantahler (approved):

PR #2397, opened 2021-10-05
Do not create default "mod_profiles" Model

evantahler:
This PR makes it so the Grouparoo app does not create a default Model in most cases.  Just like everything else, you should create your Models in the UI or via code config.  This prevents a class of bugs in which setting up code config was... odd, as a model you didn't create was already in the app. 

There's an exception for Grouparoo users migrating from an old version in which:
* They have created a Source or Destination
* The above Sources or Destinations were created via the enterprise UI (not code config)

In these cases, we include a default `mod_profiles` Model along with the migration. 

### Development

### Impact

> This is a rare case when we are editing an older migration.  This is OK because we haven't yet released v0.7

### Code review


core/src/migrations/000083-models.ts
```diff
@@ -58,6 +78,10 @@ export default {
   },
 
   down: async (queryInterface: Sequelize.QueryInterface) => {
-    await queryInterface.dropTable("models");
```
evantahler, line 61: This was a bad 'down' migration - fixed!

core/src/migrations/000083-models.ts
```diff
@@ -1,20 +1,40 @@
...
+      `SELECT * FROM "destinations" WHERE "locked" IS NOT NULL`
+    );
+    if (sources.length > 0 || destinations.length > 0) toCreteModel = true;
+
+    if (toCreteModel) {
+      await queryInterface.sequelize.query(`
+INSERT INTO "models" ("id", "name", "type", "createdAt", "updatedAt")
+VALUES ('${default_model.id}', '${default_model.name}', '${default_model.type}', NOW(), NOW());
+      `);
+    }
```
evantahler: This is the new interesting part of the migration

plugins/@grouparoo/spec-helper/src/lib/factories/properties.ts
```diff
@@ -15,7 +17,14 @@ export default async (
     purchaseAmounts: "float",
   }
 ) => {
-  const source = (await SourceFactory()) as Source;
+  if (!modelId) modelId = "mod_profiles";
+  const model =
+    (modelId
+      ? await GrouparooModel.findOne({ where: { id: modelId } })
+      : await GrouparooModel.findOne()) ??
+    ((await ModelFactory({ id: modelId })) as GrouparooModel);
```
evantahler: The goal of these factories is to set up everything you need for you, so if you ask for "properties", now we will make you a model, app, and source (previously just app and source).  You can also pass a modelId in now to use an existing model if you want.

most of our tests setup the world by calling `helper.factories.properties()` so this seems the most helpful.

plugins/@grouparoo/spec-helper/src/lib/factories/record.ts
```diff
@@ -1,14 +1,20 @@
...
+import ModelFactory from "./model";
 
-const data = async (props = {}) => {
+const data = async (props: { modelId?: string } = {}) => {
   const { GrouparooModel } = await import(`@grouparoo/core/${loadPath}`);
+  const model =
+    (props.modelId
+      ? await GrouparooModel.findOne({ where: { id: props.modelId } })
+      : await GrouparooModel.findOne()) ??
+    ((await ModelFactory({ id: props.modelId })) as GrouparooModel);
```
evantahler: For the other single-instance factories which require a modelId, it's now optionally passed in.

core/__tests__/actions/exportProcessors.ts
```diff
@@ -26,6 +26,7 @@ describe("actions/exports", () => {
     let id;
 
     beforeAll(async () => {
+      await helper.factories.model();
```
krishnaglick: Is this necessary if it's unused?

core/src/migrations/000083-models.ts
```diff
@@ -1,20 +1,40 @@
...
     queryInterface: Sequelize.QueryInterface,
     DataTypes: typeof Sequelize
   ) => {
+    let toCreteModel = false;
+    const [sources] = await queryInterface.sequelize.query(
+      `SELECT * FROM "sources" WHERE "locked" IS NOT NULL`
+    );
+    const [destinations] = await queryInterface.sequelize.query(
+      `SELECT * FROM "destinations" WHERE "locked" IS NOT NULL`
+    );
```
krishnaglick, line 20: There a reason these are raw sql over sequelize?

core/src/migrations/000083-models.ts
```diff
@@ -1,20 +1,40 @@
...
+  name: "Profiles",
+  type: "profile",
+};
 
 export default {
   up: async (
     queryInterface: Sequelize.QueryInterface,
     DataTypes: typeof Sequelize
   ) => {
+    let toCreteModel = false;
```
krishnaglick:
```suggestion
    let toCreateModel = false;
```

core/src/migrations/000083-models.ts
```diff
@@ -58,6 +78,10 @@ export default {
   },
 
   down: async (queryInterface: Sequelize.QueryInterface) => {
-    await queryInterface.dropTable("models");
+    await queryInterface.removeColumn("models", "locked");
+    await queryInterface.removeColumn("records", "modelId");
+    await queryInterface.removeColumn("sources", "modelId");
+    await queryInterface.removeColumn("groups", "modelId");
+    await queryInterface.removeColumn("destinations", "modelId");
```
krishnaglick, line 85: Should this still include the dropTable?

plugins/@grouparoo/spec-helper/src/lib/factories/destination.ts
```diff
@@ -1,15 +1,21 @@
...
+import { Destination, GrouparooModel } from "@grouparoo/core";
 
-const data = async (props = {}) => {
+const data = async (props: { modelId?: string } = {}) => {
   const { GrouparooModel } = await import(`@grouparoo/core/${loadPath}`);
+  const model =
+    (props.modelId
+      ? await GrouparooModel.findOne({ where: { id: props.modelId } })
+      : await GrouparooModel.findOne()) ??
+    ((await ModelFactory({ id: props.modelId })) as GrouparooModel);
```
krishnaglick:
```suggestion
  const model =
    await GrouparooModel.findOne(props.modelId ? { where: { id: props.modelId } } : undefined)) ??
    ((await ModelFactory({ id: props.modelId })) as GrouparooModel);
```
I think this works

plugins/@grouparoo/spec-helper/src/lib/factories/group.ts
```diff
@@ -1,14 +1,20 @@
...
+import ModelFactory from "./model";
 
-const data = async (props = {}) => {
+const data = async (props: { modelId?: string } = {}) => {
   const { GrouparooModel } = await import(`@grouparoo/core/${loadPath}`);
+  const model =
+    (props.modelId
+      ? await GrouparooModel.findOne({ where: { id: props.modelId } })
+      : await GrouparooModel.findOne()) ??
+    ((await ModelFactory({ id: props.modelId })) as GrouparooModel);
```
krishnaglick: See my comment in `plugins/@grouparoo/spec-helper/src/lib/factories/destination.ts`

plugins/@grouparoo/spec-helper/src/lib/factories/model.ts
```diff
@@ -3,8 +3,9 @@ import { GrouparooModel } from "@grouparoo/core";
 
 const data = async (props = {}) => {
   const defaultProps = {
-    name: "test-model",
+    id: undefined,
```
krishnaglick, line 6: Couldn't you just have this unset?

plugins/@grouparoo/spec-helper/src/lib/factories/record.ts
krishnaglick: See my comment in `plugins/@grouparoo/spec-helper/src/lib/factories/destination.ts`

plugins/@grouparoo/spec-helper/src/lib/factories/source.ts
```diff
@@ -1,15 +1,21 @@
...
+import { Source, GrouparooModel } from "@grouparoo/core";
+import ModelFactory from "./model";
 
-const data = async (props = {}) => {
+const data = async (props: { modelId?: string } = {}) => {
   const { GrouparooModel } = await import(`@grouparoo/core/${loadPath}`);
+  const model =
+    (props.modelId
+      ? await GrouparooModel.findOne({ where: { id: props.modelId } })
+      : await GrouparooModel.findOne()) ??
```
krishnaglick: See my comment in `plugins/@grouparoo/spec-helper/src/lib/factories/destination.ts`

core/src/migrations/000083-models.ts
evantahler: There isn't a table created in this migration, just columns added

core/src/migrations/000083-models.ts
evantahler: Good catch - fixed in https://github.com/grouparoo/grouparoo/pull/2397/commits/df43ae35786c9dd4e7a69d5aeeafcdb8c72eff4a

core/src/migrations/000083-models.ts
evantahler: Yeah! Because in the migrations the models aren't loaded/ready yet. We could make "stub" models against the table in question, but I've usually seen that backfire because the expected associations/validations won't be there because it's not a "real" model instance

core/__tests__/actions/exportProcessors.ts
evantahler: This factory creates and returns a model, app, source, and a few properties.  In this instance, yeah - we don't need it, but in most of the tests this setup is needed.  

I was able to find a few places where this setup wasn't needed via ec960258a560c24b443dbfd61c605f1ba34ff4ab

plugins/@grouparoo/spec-helper/src/lib/factories/destination.ts
evantahler: Nice!

plugins/@grouparoo/spec-helper/src/lib/factories/destination.ts
```diff
@@ -7,10 +7,9 @@ import { Destination, GrouparooModel } from "@grouparoo/core";
 const data = async (props: { modelId?: string } = {}) => {
   const { GrouparooModel } = await import(`@grouparoo/core/${loadPath}`);
   const model =
-    (props.modelId
-      ? await GrouparooModel.findOne({ where: { id: props.modelId } })
-      : await GrouparooModel.findOne()) ??
-    ((await ModelFactory({ id: props.modelId })) as GrouparooModel);
+    (await GrouparooModel.findOne(
+      props.modelId ? { where: { id: props.modelId } } : undefined
+    )) ?? ((await ModelFactory({ id: props.modelId })) as GrouparooModel);
```
krishnaglick, line 12: You might be able to get away with `await A ?? B` but this is fine!

krishnaglick (approved):

evantahler:
Thanks @krishnaglick!

PR #2398, opened 2021-10-05
`GrouparooModel` gets state, models can be destroyed

teallarson:
## Change description
Solves a bug where a user could delete a model in the UI even if there were sources/groups/destinations dependent upon it and another bug where `roo run` would crash if a model's config file was deleted.

- Adds a state column to GrouparooModel
- Adds a `model:destroy` task
- Adds a `model` task queue
- Adds a status metric to track deleted models
- Adds `model:destroy` to the system destroy sweep task

## Related issues

no

### Development

### Impact

> Migration included for adding `state` column to `GrouparooModel`.  Performance improved since `roo run` no longer crashes if a model is deleted in code config.

### Code review


core/__tests__/models/grouparooModel.ts
```diff
@@ -25,6 +25,14 @@ describe("models/grouparooModel", () => {
     expect(latestLog).toBeTruthy();
   });
 
+  test("models start in the ready state", async () => {
+    const model = GrouparooModel.scope(null).findOne({
+      where: { name: "test model", state: "ready" },
```
evantahler: If models start in the 'ready' state, does the 'draft' state matter?

core/__tests__/modules/status.ts
```diff
@@ -117,6 +117,19 @@ describe("modules/status", () => {
 
       const foundMetrics = await Status.get();
       expect(foundMetrics).toEqual({
+        Model: {
```
evantahler, line 120: Can you also please add Models Count to the stats that get reported to Telemetry?

core/__tests__/tasks/model/destroy.ts
```diff
@@ -0,0 +1,106 @@
+import { helper } from "@grouparoo/spec-helper";
```
evantahler, line 1: 👍 Good testing!

core/src/config/routes.ts
```diff
@@ -144,6 +144,7 @@ export const DEFAULT = {
         { path: "/v:apiVersion/record/:id", action: "record:destroy" },
         { path: "/v:apiVersion/property/:id", action: "property:destroy" },
         { path: "/v:apiVersion/app/:id", action: "app:destroy" },
+        { path: "/v:apiVersion/model/:id", action: "model:destroy" },
```
evantahler: We already have a model:destroy action (line 141)

evantahler (changes requested):
This is really good!  

There are some minor questions/comments below.  

Once https://github.com/grouparoo/grouparoo/pull/2397 is merged in, I'll bet your tests will need to change here - there won't be any `mod_profiles` by default any more

core/__tests__/models/grouparooModel.ts
evantahler: Maybe the UI should use `state:draft` when POSTING to the API to make a new model?

core/src/config/routes.ts
teallarson: Oops!  I'm going to organize these a bit so I don't do that again.

teallarson (commented):
@evantahler I've added the `grouparooModels` metric to the object that gets sent to Telemetry.  It looks like the way the `telemetryMetrics` table is set up, the metric will be read in as-is (no migration needed).

evantahler:
> @evantahler I've added the `grouparooModels` metric to the object that gets sent to Telemetry. It looks like the way the `telemetryMetrics` table is set up, the metric will be read in as-is (no migration needed).

Yep! The telemetry database is super-normalized, we can add more metrics without needing a migration

core/__tests__/models/grouparooModel.ts
teallarson: At what point would it change to `state:ready` then?  We do the behavior you mentioned when creating a group, for example, because there is additional information to gather on a separate page.

I started working to implement this, but now that I've started... I am thinking of going with `ready` and `deleted` as the only two states for now.  We can add `draft` when models are at a level of complexity that we actually need it for the UI.

evantahler: 👍

core/src/config/routes.ts
```diff
@@ -7,149 +7,150 @@ export const DEFAULT = {
...
-        { path: "/v:apiVersion/plugins/available", action: "plugins:available:list" },
-        { path: "/v:apiVersion/session", action: "session:view" },
-        { path: "/v:apiVersion/setupSteps", action: "setupSteps:list" },
-        { path: "/v:apiVersion/totals", action: "totals" },
-        { path: "/v:apiVersion/logs", action: "logs:list" },
-        { path: "/v:apiVersion/imports", action: "imports:list" },
-        { path: "/v:apiVersion/import/:id", action: "import:view" },
-        { path: "/v:apiVersion/exports", action: "exports:list" },
-        { path: "/v:apiVersion/exports/totals", action: "exports:totals" },
+        { path: "/v:apiVersion/account", action: "account:view" },
```
evantahler, line 10: ❤️ Alphabetical sorting

core/src/models/Group.ts
```diff
@@ -730,7 +730,7 @@ export class Group extends LoggedModel<Group> {
   @BeforeCreate
   @BeforeSave
   static async ensureModel(instance: Destination) {
-    const model = await GrouparooModel.findOne({
+    const model = await GrouparooModel.scope(null).findOne({
```
evantahler: This is interesting - this is saying that it's OK to save a Group under a deleted model.  I don't think that's right

core/__tests__/modules/status.ts
evantahler: This would need to be added to `src/modules/telemetry` to be sent, but I'll do this in another story https://www.pivotaltracker.com/story/show/179858322

core/src/models/Group.ts
teallarson: Good find, thank you!

evantahler (approved):

PR #2400, opened 2021-10-07
`record:import` action custom transaction isolation

evantahler:
In order to support the custom behavior of the `record:import` action, we need to remove transaction isolation from the `record.import()` step.  As we want to rely on a write to the database throwing to handle setting invalid values, we cannot then attempt to re-use that same transaction to import the next source/properties.  To that end, we will use custom transaction behaviror within this action. 

### Development

### Impact

>

### Code review

### DIFF
--- a/core/src/actions/records.ts
+++ b/core/src/actions/records.ts
@@ -1,7 +1,8 @@
-import { config } from "actionhero";
+import { Action, config } from "actionhero";
 import { AuthenticatedAction } from "../classes/actions/authenticatedAction";
 import { GrouparooRecord } from "../models/GrouparooRecord";
 import { RecordProperty } from "../models/RecordProperty";
+import { Group } from "../models/Group";
 import { internalRun } from "../modules/internalRun";
 import { Op } from "sequelize";
 import { ConfigWriter } from "../modules/configWriter";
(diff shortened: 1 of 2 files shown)


core/src/actions/records.ts
```diff
@@ -167,38 +170,48 @@ export class RecordImport extends CLSAction {
...
+    await CLS.wrap(async () => {
+      record = await GrouparooRecord.findById(params.id);
+      await record.buildNullProperties();
+      await record.markPending();
+    });
 
-    await record.buildNullProperties();
-    await record.markPending();
+    // this method should not be wrapped in a transaction because we want to allow multiple sources to throw and recover their imports
     await record.import();
```
bleonard, line 189: what was throwing before on invalid property? If it's in here, won't it still cause the whole action to throw?

evantahler: It wasn't the `throw` that was causing the problem, but re-using the same transaction for a subsequent query (the next source or property we were importing) that was causing the problem.  So this solution, which is the fast one but probably not the most safe, is to just not use a transaction - this way we can try/catch the update to recordProperties and not worry about what happens next

This isn't a problem when importing recordProperties in the task, because there's only ever one write with a collection of properties from one source.  Here we have multiple

bleonard (approved):

core/src/actions/records.ts
evantahler: There's probably a future solution where the import button fires of the tasks and polls until the record is ready again... but that won't work in an environment where there are no workers (ui:config, etc)

PR #2402, opened 2021-10-07
Config UI write correct model id on save

krishnaglick:
## Change description

> Model ID's on Source configs were not friendly.

## Related issues

### Development

### Impact

- [X] Code follows company security practices and guidelines
- [X] Security impact of change has been considered
- [X] Performance impact of change has been considered
- [X] Possible migration needs considered (model migrations, config migrations, etc.)

>

### Code review

- [X] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.

### DIFF
--- a/core/__tests__/modules/configWriter.ts
+++ b/core/__tests__/modules/configWriter.ts
@@ -163,7 +163,9 @@ describe("modules/configWriter", () => {
         id: "mod_profiles",
       });
       const app: App = await helper.factories.app();
-      const source: Source = await helper.factories.source(app);
+      const source: Source = await helper.factories.source(app, {
+        modelId: model.id,
+      });
       await source.setOptions({ table: "test-table-04" });
       await source.bootstrapUniqueProperty("userId_04", "integer", "id");
       await source.setMapping({ id: "userId_04" });
@@ -187,6 +189,14 @@ describe("modules/configWriter", () => {
         fs.readFileSync(expConfigFiles[0]).toString()
       );
       expect(appConfig).toEqual(await app.getConfigObject());
+
+      const sourceConfig = JSON.parse(
+        fs.readFileSync(
+          expConfigFiles.find((f) => f.includes("sources")),
+          { encoding: "utf8" }
+        )
+      );
+      expect(sourceConfig.modelId).toBe(model.id);
     });
 
     test("does not write objects that are locked", async () => {
--- a/core/__tests__/modules/configWriter.ts
+++ b/core/__tests__/modules/configWriter.ts
@@ -401,6 +401,7 @@ describe("modules/configWriter", () => {
       const model: GrouparooModel = await GrouparooModel.create({
         type: "profile",
         name: "Profiles",
+        id: "mod_" + uuid.v4(),
       });
       const app: App = await helper.factories.app();
       const source: Source = await helper.factories.source(app);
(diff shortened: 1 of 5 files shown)


evantahler (changes requested):
This is a good fix... but it implies that Group, GrouparooRecord, and Destination also likely have a similar problem.  Can you please check those too?

krishnaglick:
> This is a good fix... but it implies that Group, GrouparooRecord, and Destination also likely have a similar problem. Can you please check those too?
Will fix

pedroslopez:
@krishnaglick note that this only happens when the model you're assigning has been created in the same `roo config` session. If you exit out and start creating things with an existing model, you won't see it. I was trying out some things working on another story and did see the issue with groups as well

core/src/models/Source.ts
```diff
@@ -345,7 +346,7 @@ export class Source extends LoggedModel<Source> {
     let configObject: any = {
```
bleonard, line 349: Just above here it did `if (!appId || !name) return;`
Should that be `if (!appId || !modelId || !name) return;` ?

krishnaglick: I could have it fallback to `this.modelId`, or bail out. Whichever is fine with me.

core/__tests__/modules/configWriter.ts
```diff
@@ -160,10 +161,12 @@ describe("modules/configWriter", () => {
       const model: GrouparooModel = await GrouparooModel.create({
         type: "profile",
         name: "Profiles",
-        id: "mod_profiles",
+        id: "mod_" + uuid.v4(),
```
krishnaglick, line 164: This simulates config mode.

core/__tests__/modules/configWriter.ts
```diff
@@ -1033,6 +1046,7 @@ describe("modules/configWriter", () => {
           [bootstrapProperty.getConfigId()]: [12],
         },
       });
+      expect(record.modelId).toBe("mod_profiles");
```
bleonard: I'm suprised this isn't just `"profiles"`- unless we are in another block?
IIn that case, should we do the `uuid.v4()` thing in that block too to verify they are using the `getConfigId`?

core/src/models/Group.ts
```diff
@@ -682,7 +684,7 @@ export class Group extends LoggedModel<Group> {
     for (const rule of convenientRules) {
       const property = await Property.findOneWithCache(
         rule.key,
-        this.modelId,
+        this.model?.getConfigId(),
```
bleonard: just `modelId` (using the variable above)

core/src/models/Group.ts
```diff
@@ -695,7 +697,9 @@ export class Group extends LoggedModel<Group> {
       });
     }
 
-    if (!name) return;
+    if (!name || !modelId) {
```
bleonard: maybe move this up over the `const property` situation to save that work if there is `modelId`

bleonard (approved):
Few changes to be clearer and some test questions, but overall good to go.

core/__tests__/modules/configWriter.ts
krishnaglick: It doesn't seem to be in another block. `expect(record.model.id).toBe("mod_profiles");` is green.
I'm not entirely certain why this happens yet.

bleonard (approved):

evantahler (approved):

PR #2403, opened 2021-10-08
Set collection to none for demo destinations

bleonard:
From the new changes for model/group collections, we need to update the JSON for the seeded demo destinations.


pedroslopez:
Should the `nobody` group be removed too?

evantahler (approved):

bleonard:
> Should the `nobody` group be removed too?

Good catch! But I decided to leave it because sometimes I test with it.

PR #2405, opened 2021-10-11
Model scoping on Groups, model state in UI

teallarson:
## Change description

Fixes a bug introduced by #2398 where the search for models in the `@BeforeCreate`/`@BeforeSave` hook for Groups wasn't properly scoped.  It caused crashes when deleting a group because the group is updated/saved during the deletion process -- when its model may or may not also be in "deleted" state.  To solve this, I've split out the logic because I don't think they should be the same hook.

- When creating a Group, its Models can only be in `"ready"` state
- When saving a Group, if the Group is in the `"deleted"` state, then its Model may be in the `"deleted"` state.  This is required for final runs to process correctly for deletion.

Also adds a "state" column to `/models` in the ui.

## Related issues

no

### Development

### Impact

>

### Code review

### DIFF
--- a/core/__tests__/models/group/group.ts
+++ b/core/__tests__/models/group/group.ts
@@ -57,7 +57,9 @@ describe("models/group", () => {
       modelId: model.id,
     });
 
-    expect(group.save()).rejects.toThrow(/cannot find ready model with id/);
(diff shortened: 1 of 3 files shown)


evantahler (approved):
Good testing of this nuanced logic

PR #2408, opened 2021-10-11
migration and model for AppRefreshQuery

teallarson:
## Change description

App Refresh Query introduces a new feature on Apps.  Users can add a `"refreshQuery"` to an App's config that tells core where to scan for updated data in the App's database.  When an updated value is found, all Schedules for the App marked `"refreshEnabled: true"` (new default) will be run.  

This PR introduces the model (including migration) and codeConfig to implement this feature.

## Related issues

### Development

### Impact

database migration included
>

### Code review


core/src/migrations/000088-createAppDataRefreshes.ts
```diff
@@ -0,0 +1,86 @@
+import Sequelize from "sequelize";
+
+export default {
+  up: async (
+    queryInterface: Sequelize.QueryInterface,
+    DataTypes: typeof Sequelize
+  ) => {
+    await queryInterface.createTable("appDataRefreshes", {
+      id: {
+        type: DataTypes.STRING(40),
```
evantahler: 191 chars!

core/src/migrations/000088-createAppDataRefreshes.ts
```diff
@@ -0,0 +1,86 @@
...
+      },
+
+      appId: {
+        type: DataTypes.STRING(191),
+        allowNull: false,
+      },
+
+      refreshQuery: {
+        type: DataTypes.TEXT,
+        allowNull: false,
```
evantahler: allowNull should be true to allow for "draft state"

core/src/migrations/000088-createAppDataRefreshes.ts
```diff
@@ -0,0 +1,86 @@
...
+
+      lastConfirmedAt: {
+        type: DataTypes.DATE,
+      },
+    });
+
+    await queryInterface.changeColumn("appDataRefreshes", "state", {
+      type: DataTypes.STRING(191),
+      allowNull: false,
+    });
```
evantahler: This is a new table, so you can just say `allowNull=false` above

core/src/models/Schedule.ts
```diff
@@ -119,6 +119,11 @@ export class Schedule extends LoggedModel<Schedule> {
   @Column
   confirmRecords: boolean;
 
+  @AllowNull(false)
+  @Default(true)
+  @Column
+  refreshEnabled: boolean;
```
evantahler, line 125: Add this to `apiData()` as well

core/src/models/AppDataRefresh.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  @BeforeSave
+  static async updateState(instance: AppDataRefresh) {
+    await StateMachine.transition(instance, STATE_TRANSITIONS);
+  }
+
+  @AfterSave
+  static async runCheckIfNewQuery(instance: AppDataRefresh) {
+    //if instance.query was changed, run the check from appDataRefreshOps
```
evantahler: in a `@BeforeSave` you can check `bool: instance.changed('query')` to see what is new on the instances.  Enqueue the task if it changed.

core/src/models/AppDataRefresh.ts
```diff
@@ -0,0 +1,126 @@
...
+  AfterSave,
+} from "sequelize-typescript";
+import { LoggedModel } from "../classes/loggedModel";
+import { App } from "./App";
+import { StateMachine } from "./../modules/stateMachine";
+import { LockableHelper } from "../modules/lockableHelper";
+import { ConfigWriter } from "../modules/configWriter";
+import { APIData } from "../modules/apiData";
+
+const STATES = ["draft", "ready", "deleted"] as const;
```
evantahler: Probably don't need deleted state.

evantahler:
Task architecture overview https://jamboard.google.com/d/107uNQEhlgFBTLZOw6qN4RaiA54PjE7Vq9hp2CRpSoUY/viewer?f=0

core/src/models/AppDataRefresh.ts
```diff
@@ -0,0 +1,117 @@
...
+  static async updateState(instance: AppDataRefresh) {
+    await StateMachine.transition(instance, STATE_TRANSITIONS);
+  }
+
+  @BeforeSave
+  static async runCheckIfNewQuery(instance: AppDataRefresh) {
+    if (
+      instance.changed("refreshQuery") &&
+      !instance.changed("lastConfirmedAt") &&
+      instance.state === "ready"
```
teallarson: Don't validate or query anything for draft instances

core/src/classes/codeConfig.ts
```diff
@@ -44,6 +44,7 @@ export interface ModelConfigurationObject extends ConfigurationObject {
 export interface AppConfigurationObject extends ConfigurationObject {
   name: string;
   type: string;
+  refreshQuery?: string;
```
evantahler, line 47: I was under the impression this was going to be something like

```
refresh: {query: string, frequency: number}
```

Or is that a later story?

evantahler: NVM! looking at the model and migration, that's a later story :D

core/__tests__/fixtures/codeConfig/initial/config.js
```diff
@@ -16,6 +15,7 @@ module.exports = async function getConfig() {
       options: {
         fileId: "test-file-path.db",
       },
+      refreshQuery: "SELECT MAX(updated_at) FROM users",
     },
```
evantahler, line 20: There's probably a test to write that checks when the refreshQuery changes, that the plugin is checked for new data

core/src/migrations/000088-createAppDataRefreshes.ts
```diff
@@ -0,0 +1,75 @@
...
+      },
+
+      locked: {
+        type: DataTypes.STRING(191),
+      },
+
+      state: {
+        type: DataTypes.STRING(191),
+        allowNull: false,
+        defaultValue: "draft",
```
evantahler: no defaults in the DB please

core/src/modules/ops/appDataRefresh.ts
```diff
@@ -0,0 +1,58 @@
...
+    );
+    if (sampleValue !== value) {
+      appDataRefresh.value = sampleValue;
+      appDataRefresh.lastChangedAt = new Date();
+    }
+    appDataRefresh.lastConfirmedAt = new Date();
+    await appDataRefresh.save();
+
+    //return true, schedules will be enqueued
+    return true;
```
evantahler: It looks like you always return true.   What does this indicate?  Perhaps you might return `return sampleValue !== value` - which would indicate if there were changes or not?

core/src/models/AppDataRefresh.ts
```diff
@@ -0,0 +1,117 @@
...
+    ) {
+      const isUpdated = await AppDataRefreshOps.checkDataRefreshValue(instance);
+      if (isUpdated === true) AppDataRefreshOps.triggerSchedules(instance);
```
evantahler: If `AppDataRefreshOps.checkDataRefreshValue()` doesn't crash, it looks like it always returns true

core/src/modules/ops/appDataRefresh.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+    const schedulesToRun: Schedule[] = [];
+
+    for (const source of sources) {
+      schedulesToRun.push(
+        ...(await Schedule.findAll({
+          where: { sourceId: source.id },
+        }))
+      );
+    }
```
evantahler: This means you will run all schedules, even those that are not recurring or those with the new `refreshEnabled` column (which I don't think is used anywhere yet).

I'm guessing you might want to select all ready schedules for the source which have `refreshEnabled=true`.  If so, please add a test for that

core/__tests__/modules/codeConfig/codeConfig.ts
```diff
@@ -102,6 +103,11 @@ describe("modules/codeConfig", () => {
         expect(options).toEqual({ fileId: "test-file-path.db" });
       });
 
+      test("appDataRefresh is created", async () => {
+        const appDataRefreshes = await AppDataRefresh.findAll();
+        expect(appDataRefreshes.length).toBe(1);
+      });
```
evantahler: Test a little deeper - is the query what you expect?  Is the app it belongs to correct?

plugins/@grouparoo/bigquery/src/lib/appQuery.ts
```diff
@@ -0,0 +1,12 @@
+import { AppQueryMethod } from "@grouparoo/core";
+import { validateQuery } from "./validateQuery";
+
+export const appQuery: AppQueryMethod = async ({
+  connection,
+  refreshQuery,
+}) => {
+  validateQuery(refreshQuery);
+
+  const rows = await connection.query(refreshQuery); // return [];
```
evantahler: unnecessary comment

plugins/@grouparoo/bigquery/src/initializers/plugin.ts
```diff
@@ -80,7 +81,7 @@ export class Plugins extends Initializer {
               placeholder: "e.g. -----BEGIN PRIVATE KEY-----\nMII ...",
             },
           ],
-          methods: { test, connect, disconnect },
+          methods: { test, connect, disconnect, query: appQuery },
```
evantahler: Why the name change?  If the method is called `query`,  it would be clearer if the file was `src/lib/query.ts` returning a method called `query`.  Or, maybe the plugin method should be called `appQuery` - either way it should match I think.

plugins/@grouparoo/mysql/public/templates/app/{{{id}}}.js.template
```diff
@@ -5,6 +5,7 @@ exports.default = async function buildConfig() {
       id: {{{id}}},
       name: {{{id}}},
       type: "{{{__pluginName}}}",
+      refreshQuery: "...", // Optional: How will Grouparoo know to trigger a Data Refresh? ie: `SELECT MAX(updated_at) FROM users`;
       options: {
         host: "localhost",
         port: 3306,
```
evantahler, line 11: There's probably a change missing from the source/schedule template files to use the new column `refreshEnabled`

plugins/@grouparoo/mysql/src/initializers/plugin.ts
```diff
@@ -28,6 +29,7 @@ import {
   QuerySourceTemplate,
   QueryPropertyTemplate,
 } from "@grouparoo/app-templates/dist/source/query";
+import { validateQuery } from "../lib/validateQuery";
```
evantahler: Unused I think

plugins/@grouparoo/mysql/src/lib/appQuery.ts
```diff
@@ -0,0 +1,12 @@
+import { validateQuery } from "./validateQuery";
+import { AppQueryMethod } from "@grouparoo/core";
+
+export const appQuery: AppQueryMethod = async ({
+  connection,
+  refreshQuery,
+}) => {
+  validateQuery(refreshQuery);
+
+  const rows = await connection.asyncQuery(refreshQuery); // return [];
```
evantahler: unnecessary comment

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -285,6 +287,14 @@ export namespace helper {
                 password: { type: "password" },
               };
             },
+            getConnection: async () => {
+              return {
+                connection: "this is a connection",
+              };
+            },
```
evantahler: I don't think this exists on plugins?

evantahler (changes requested):
This is awesome work, and it's almost there!  Good work getting all those plugin tests to pass!

The main missing topics is around the new `schedule#refreshEnabled` option - it's not used yet or considered in code-config.

core/__tests__/fixtures/codeConfig/initial/config.js
teallarson: Yes.  I checked this at the model level, but it should probably be checked here as well.

plugins/@grouparoo/bigquery/src/initializers/plugin.ts
teallarson: To differentiate it from the `query-import` methods.  Though those are explicitly named `query-import`.

core/src/models/AppDataRefresh.ts
teallarson: updated `AppDataRefreshOps.checkDataRefreshValue` to return `value !== originalValue`.

plugins/@grouparoo/bigquery/src/initializers/plugin.ts
teallarson: renamed method `appQuery`

plugins/@grouparoo/mysql/src/lib/appQuery.ts
teallarson: removed from all plugins

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
teallarson: Correct.  Removed it.

core/__tests__/fixtures/codeConfig/initial/config.js
teallarson: @evantahler having trouble testing it at this level.  Edit: It's not the same issue I thought, but it there is definitely weirdness going on.

teallarson:
@evantahler Should [this](https://github.com/grouparoo/grouparoo/blob/a86a643a8df0a68bb9393a2072fd3109e550e986/core/src/modules/ops/appDataRefresh.ts#L56) be calling the `ScheduleRun` Action or Task instead of directly calling the method? 

I'm stuck on these failing tests.  
- Models failure: I can see the run being created in the database (🎉), but it the test returns before the run is there it seems.  That's not great.
- Modules failure: It seems like it's tracking one step behind what I want it to... not waiting for the steps to fully complete before returning.

core/src/modules/configLoaders/app.ts
```diff
@@ -62,7 +62,8 @@ export async function loadApp(
 
   logModel(app, validate ? "validated" : isNew ? "created" : "updated");
 
-  if (configObject.refreshQuery && configObject.refreshQuery !== "") {
+  if (configObject.refreshQuery && configObject.refreshQuery.length > 3) {
+    //template placeholder is "..."
```
evantahler: This is sort of a strange check.  We don't validate any of the other placeholders in config files (eg: password shouldn't be `"..."`) - https://github.com/grouparoo/grouparoo/blob/main/plugins/%40grouparoo/mysql/public/templates/app/%7B%7B%7Bid%7D%7D%7D.js.template#L11-L13

evantahler: Since you are running the query onSave, if SQL doesn't like the statement, it should fail, right? - that's a strong validation.

core/src/modules/configLoaders/schedule.ts
```diff
@@ -30,7 +30,7 @@ export async function loadSchedule(
     delete configObject["confirmProfiles"];
   }
 
-  validateConfigObjectKeys(Schedule, configObject);
+  validateConfigObjectKeys(Schedule, configObject, ["refreshEnabled"]);
```
evantahler: I don't think the extra `["refreshEnabled"]` is needed now that refreshEnabled is added to the Schedule model

core/src/modules/ops/appDataRefresh.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    for (const schedule of schedulesToRun) {
+      //stop any existing run on this schedule
+      const existingRuns = await Run.findAll({
+        where: { creatorId: schedule.id },
+      });
+      for (const run of existingRuns) {
+        await run.stop();
+      }
```
evantahler: This is going to load up and call `stop()` all all the runs for the schedule - you may want to add `state:'running'` to the where clause

core/src/migrations/000089-createAppDataRefreshes.ts
```diff
@@ -0,0 +1,74 @@
+import Sequelize from "sequelize";
+
+export default {
+  up: async (
+    queryInterface: Sequelize.QueryInterface,
+    DataTypes: typeof Sequelize
+  ) => {
+    await queryInterface.createTable("appDataRefreshes", {
```
evantahler: (comment, doesn't need change) - this is a bit of an awkward table name (and model name).  Now that we know a little more about what this model is doing, perhaps `AppRefreshQuery` and `table: appRefreshQueries` might be clearer?

evantahler (changes requested):

core/src/modules/configLoaders/app.ts
teallarson: I'll dig into that more!  There were a handful of plugins that had failing cli-generate tests without this check.  I'd had it as `!== ""` before, which made more sense.

core/src/migrations/000089-createAppDataRefreshes.ts
teallarson: That makes sense.  The data its storing is all centered on the query, not the actual refreshes.

core/src/modules/configLoaders/app.ts
```diff
@@ -10,14 +10,15 @@ import { App } from "../..";
...
+import { AppRefreshQuery } from "../../models/AppRefreshQuery";
 
 export async function loadApp(
   configObject: AppConfigurationObject,
   externallyValidate: boolean,
   validate = false
 ): Promise<IdsByClass> {
   let isNew = false;
-  validateConfigObjectKeys(App, configObject);
+  validateConfigObjectKeys(App, configObject, ["refreshQuery"]);
```
evantahler, line 21: Was `["refreshQuery"]` needed now that that's an attribute of the model?

core/src/modules/ops/appRefreshQuery.ts
```diff
@@ -0,0 +1,69 @@
...
+          where: { sourceId: source.id, refreshEnabled: true, state: "ready" },
+        }))
+      );
+    }
+
+    for (const schedule of schedulesToRun) {
+      //stop any existing run on this schedule
+      const existingRuns = await Run.findAll({
+        where: { creatorId: schedule.id, state: "running" },
+      });
```
evantahler, line 60: 👍 stopping only runs that are currently running

evantahler (approved):
🎉

core/src/modules/configLoaders/app.ts
teallarson: Yes. `refreshQuery` is an attribute of an `AppRefreshQuery` not an `App`.  Removing it causes this error on validation:
```    
"[ config ] error with App `Data Warehouse` (data_warehouse): refreshQuery is not a valid property of a App",
```

evantahler: Makes sense!

PR #2412, opened 2021-10-11
fix: only keep params in new path when switching models

pedroslopez:
## Change description

We were always keeping the params from the old route when switching to a new model, which may or may not be required. For example, `/model/admins/record/rec_6a0c9714-d18b-4fde-8f5d-879f819fc72f/edit` took you to `/model/users/records?recordId=rec_6a0c9714-d18b-4fde-8f5d-879f819fc72f` after switching because the param was kept. 

This filters out the params to only include those that are present in the new route. 

## Related issues

No

### Development

### Impact

>

### Code review

### DIFF
--- a/ui/ui-components/pages/_app.tsx
+++ b/ui/ui-components/pages/_app.tsx
@@ -1,6 +1,6 @@
 import App from "next/app";
-import Cookies from "universal-cookie";
-import { plural } from "pluralize";
+import { AxiosError } from "axios";
+import type { AppContext } from "next/app";
 
 import { UseApi } from "../hooks/useApi";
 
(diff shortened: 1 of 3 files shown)


ui/ui-components/pages/_app.tsx
```diff
@@ -67,12 +67,22 @@ export default function GrouparooWebApp(props) {
         newPath = `/model/[modelId]/${plural(topic)}`;
       }
 
+      // only keep path params for new url
```
evantahler: This is 👍 as is.

As `onChangeModelId` gets more complex, it's probably time to move this into it's own utility.  I also think it's only used by the navigation component, so it might not need to be in `_app`?

evantahler (approved):

ui/ui-components/pages/_app.tsx
pedroslopez: You're right! I had it in `_app` because I was doing some stuff with state initially, but that's not the case anymore. Moved it out

PR #2416, opened 2021-10-13
reload model edit page when selected model changes

pedroslopez:
## Change description

When on the model edit page, if we used the global model selection dropdown to change to another model, the url would be updated but the model we were looking at was still the old one.

Honestly I think the core issue is actually that in many places throughout the app we are initializing state from props, which gets us into issues with not updating things if props change after the page is rendered. But changing this is a greater refactor and also affects how we're using uncontrolled components in forms.

## Related issues

> Fix [#1]()

### Development

### Impact

>

### Code review

### DIFF
--- a/ui/ui-components/pages/model/[modelId]/edit.tsx
+++ b/ui/ui-components/pages/model/[modelId]/edit.tsx
@@ -29,20 +29,8 @@ export default function Page(props) {
   const { modelId } = router.query;
 
   useEffect(() => {
-    load();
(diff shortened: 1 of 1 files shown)


evantahler (approved):
This is a fine fix for now.

As you say in the description, there are a lot of edge cases with that dropdown.  Maybe changing the dropdown on some pages should be disabled - eg if you aren't on a page with `/model` in path.

PR #2417, opened 2021-10-13
Filter runs list for Groups

evantahler:
Fixes a bug when viewing the list of Runs created by a Group that was showing all the Runs

### Development

### Impact

>

### Code review

### DIFF
--- a/ui/ui-components/components/runs/list.tsx
+++ b/ui/ui-components/components/runs/list.tsx
@@ -40,8 +40,8 @@ export default function RunsList(props) {
 
   async function load() {
     const params = { limit, offset, topic };
-    if (router.query.sourceId) params["id"] = router.query.sourceId.toString();
-    if (router.query.groupId) params["id"] = router.query.groupId.toString();
+    params["creatorId"] =
+      router.query.sourceId ?? router.query.groupId ?? router.query.propertyId;
     if (stateFilter !== "") params["state"] = stateFilter;
     if (errorFilter !== "") params["hasError"] = errorFilter;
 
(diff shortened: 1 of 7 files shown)


ui/ui-components/components/runs/list.tsx
```diff
@@ -41,6 +41,7 @@ export default function RunsList(props) {
   async function load() {
     const params = { limit, offset, topic };
     if (router.query.sourceId) params["id"] = router.query.sourceId.toString();
+    if (router.query.groupId) params["id"] = router.query.groupId.toString();
```
pedroslopez: this `id` param is quite weird to me, when I did the source one I got confused. Don't have to change this now, but
I'm thinking maybe it should be more like `creatorId`

pedroslopez (approved):
This looks good, but it looks like the runs list is also included in a `property`'s page, should we do the same for them?

ui/ui-components/components/runs/list.tsx
evantahler: Good call! Updated via 816c9ea755da10460fc0638c502534574b58ae14

ui/ui-components/components/runs/list.tsx
```diff
@@ -40,7 +40,8 @@ export default function RunsList(props) {
 
   async function load() {
     const params = { limit, offset, topic };
-    if (router.query.sourceId) params["id"] = router.query.sourceId.toString();
+    params["creatorId"] =
```
bleonard, line 43: in general, it's a risk here by not including the `creatorType` of ids that match across these different models.

PR #2420, opened 2021-10-14
dont redirect to `new` page for sources and destinations

pedroslopez:
## Change description

This was causing inconsistent behavior (going directly to the page you would see an empty list, but if the model changed you would see the `new` page.  I'm not entirely sure why these redirects were in there, since before this model change you would never really hit it- it was only checking for empty results when updating the offset and limit params, which wouldnt normally cause a `length 0` 🤔 

We could've also just matched the behavior in `getIntitialProps` and redirect from there if the total was 0, but this doesn't really match what the other list pages do anyway, and might be confusing imo.

## Related issues

No

### Development

### Impact

>

### Code review


evantahler (approved):
👍 

We've got "add new" buttons on these pages, so removing the redirect it OK with me!

PR #2423, opened 2021-10-14
Run all Schedules for this Model

evantahler:
Updates the "Run all Schedules" button to be scoped by the Model currently selected.

Screen Shot 2021-10-14 at 10 49 31 AM

### Development

### Impact

>

### Code review

### DIFF
--- a/core/__tests__/actions/schedules.ts
+++ b/core/__tests__/actions/schedules.ts
@@ -311,6 +311,22 @@ describe("actions/schedules", () => {
         expect(runsAgain.length).toBe(0);
       });
 
+      test("an administrator can request to run some schedules by modelId", async () => {
+        connection.params = { csrfToken, modelId: model.id };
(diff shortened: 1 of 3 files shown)


pedroslopez (approved):
Looks good! The double plural on `Run all Admins Schedules` feels a bit odd, but at least it's clear what you're doing. 
Not sure if there should be a general `run all schedules` button somewhere else

evantahler:
> Looks good! The double plural on `Run all Admins Schedules` feels a bit odd, but at least it's clear what you're doing. Not sure if there should be a general `run all schedules` button somewhere else

The double plural comes from the model name, which happens to be `admins` in this case.  I didn't want to try to de-pluralize it if it was 🤷 

I left it so the action could still run all schedules (regardless of model) which we might want later for an integration with an ETL/ELT tool...

PR #2425, opened 2021-10-15
Improve error message collect for Zendesk plugin

pauloouriques:
- add better messages to Zendesk plugin retrieving the `result` field from the API client response.

### DIFF
--- a/plugins/@grouparoo/zendesk/src/lib/export/exportRecord.ts
+++ b/plugins/@grouparoo/zendesk/src/lib/export/exportRecord.ts
@@ -148,6 +148,8 @@ export const exportRecord: ExportRecordPluginMethod = async ({
 
     return { success: true };
   } catch (error) {
+    const errorMessage = error?.result?.toString();
+    error.message = errorMessage || error.message;
     // https://develop.zendesk.com/hc/en-us/articles/360001074328-Best-practices-for-avoiding-rate-limiting
 
     // headers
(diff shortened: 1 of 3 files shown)


plugins/@grouparoo/zendesk/src/lib/export/exportRecord.ts
```diff
@@ -148,6 +148,8 @@ export const exportRecord: ExportRecordPluginMethod = async ({
 
     return { success: true };
   } catch (error) {
+    const errorMessage = error?.result?.toString();
```
bleonard, line 151: what's an example of this? I'm curious if it's losing information or not. (like it used to say `422`)

bleonard (changes requested):
I'd expect to see some test changes on a failed one that is checking for a better message.

plugins/@grouparoo/zendesk/src/lib/export/exportRecord.ts
pauloouriques: The `errorMessage` for a 429 is: _You have exceeded the rate limit for updating a single unique user. user email: othergrouparoo@demo.com id:  external_id: grouparoo123_

bleonard: That's cool. Let's verify one for a `422` as well. I think you'd sync someone with a bad or duplicate email

bleonard (approved):
Thanks!

PR #2426, opened 2021-10-15
Do not use Sequelize's rename column method

evantahler:
This PR makes it clear that Grouparoo will not be migrating SQLite database that have Records and Properties already.  SQLite users are to delete their database and re-start the application. 
  https://github.com/grouparoo/grouparoo/discussions/2428 explains more.

This PR includes a number of migration utilities for the future.

### Development

### Impact

>

### Code review

### DIFF
--- a/core/src/migrations/000053-removeCompoundindex.ts
+++ b/core/src/migrations/000053-removeCompoundindex.ts
@@ -6,7 +6,7 @@ export default {
     queryInterface: Sequelize.QueryInterface,
     DataTypes: typeof Sequelize
   ) => {
-    const dialect = await MigrationUtils.getDialect(queryInterface);
+    const dialect = MigrationUtils.getDialect(queryInterface);
 
     if (dialect === "sqlite") {
       await queryInterface.removeIndex(
--- a/core/src/migrations/000053-removeCompoundindex.ts
+++ b/core/src/migrations/000053-removeCompoundindex.ts
@@ -1,5 +1,3 @@
-import { config } from "actionhero";
-
 import Sequelize from "sequelize";
 
 export default {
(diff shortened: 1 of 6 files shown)


core/src/migrations/000053-removeCompoundindex.ts
```diff
@@ -1,13 +1,13 @@
...
 
 export default {
   up: async (
     queryInterface: Sequelize.QueryInterface,
     DataTypes: typeof Sequelize
   ) => {
-    if (config.sequelize?.dialect === "sqlite") {
+    const dialect: string = queryInterface.sequelize["options"].dialect;
+
+    if (dialect === "sqlite") {
```
evantahler: This is a better way of getting the dialect in migrations - config might not be ready yet

core/src/migrations/000081-profilesToRecords.ts
```diff
@@ -91,7 +66,9 @@ export default {
     for (const [table, collection] of Object.entries(renames)) {
       for (const batch of collection) {
         const [oldName, newName] = batch;
-        await queryInterface.renameColumn(table, oldName, newName);
+        await queryInterface.sequelize.query(
+          `ALTER TABLE "${table}" RENAME COLUMN "${oldName}" TO "${newName}";`
+        );
```
evantahler: So I guess going forward, we can't rely on Sequelize's `renameColumn()` method?

bleonard: Are they doing something more exciting that this? Maybe we'll see something we are missing.
OR should re just override this method until they fix it?

evantahler: Yeah, they are making s tmp table, copying all data to it, updating the new table, and then replacing the old table with the new one.

the problem is (in the linked Bug) that they are making the indexes in the new table wrong.

this type of multi-phase migration is great in production systems that don’t want downtime… and we don’t care, especially with SQLite :/

evantahler: Maybe we monkey patch alterTable just for SQLite… there’s still something wrong with the indexes… it looks like we need to rename the constraints as the column names change (failing test).

evantahler: Note: this is really only a problem with RecordProperties and SQLite  - it’s the only table with compound/conditional indexes. 
1) conditionally unique per position x record x property
2) conditionally unique per property x value x isUnique

bleonard (approved):
Looks like a good solution given the messier alternatives.

PR #2430, opened 2021-10-19
dont try to export records after destroy

pedroslopez:
## Change description

In the normal record deletion workflow, the `record:destroy` task is already taking care of exporting the profile before destroying it. This hook only really had an effect in config ui when clicking the `delete` button. I don't quite understand why this was in an `@AfterDestroy` hook, since some of the export functions that are called later on require the record to be present, so it hasn't really been working, causing errors such as `Error: Errors returned without recordId - throw if unknown ` due to the record not existing.

This delete button on config ui should not export anyway, since we're just working with sample profiles and the intention of clicking that button probably means just remove the sample profile.

When checking this I did notice that the `record:destroy` task doesn't seem to be considering models, but that seems like another issue. Created a story on tracker https://www.pivotaltracker.com/story/show/179998326.

## Related issues

No

### Development

### Impact

>

### Code review

### DIFF
--- a/core/src/models/GrouparooRecord.ts
+++ b/core/src/models/GrouparooRecord.ts
@@ -279,11 +279,6 @@ export class GrouparooRecord extends LoggedModel<GrouparooRecord> {
   }
 
   @AfterDestroy
-  static async removeFromDestinations(instance: GrouparooRecord) {
-    await instance.export();
-  }
-
-  @AfterDestroy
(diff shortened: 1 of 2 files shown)


evantahler (approved):

pedroslopez:
Re-requested because of one more change: the delete button now says `remove sample record` to match the list's `add sample record` and be a bit clearer of what's going on

image

evantahler (approved):
That's a clearer button!

PR #2431, opened 2021-10-19
Include the email on the Zendesk find user function

pauloouriques:
Lookup users by email (in addition to external_id)


plugins/@grouparoo/zendesk/__tests__/export/export-record.ts
```diff
@@ -401,7 +401,7 @@ describe("zendesk/exportRecord", () => {
     expect(user.name).toBe(migratedName);
     expect(user.user_fields.checkbox_field).toBe(true);
     expect(user.user_fields.text_field).toBe("change");
-    expect(user.tags.sort()).toEqual([groupThree]);
+    expect(user.tags.sort()).toEqual([checkboxTag, groupThree]); // including the automatically added tag.
```
pauloouriques: Interesting that this test case, without the `checkboxTag` which is automatically added when the `checkbox_field = true`, start to fail after including the email on the search. So, I added the tag there.

plugins/@grouparoo/zendesk/src/lib/export/exportRecord.ts
```diff
@@ -195,14 +195,22 @@ export async function findUser(
...
   let found = null;
-  if (!found && newId) {
+  if (newId) {
     found = await searchForUser(client, { external_id: newId });
   }
   if (!found && oldId && oldId !== newId) {
     found = await searchForUser(client, { external_id: oldId });
   }
+  if (!found && newEmail && newEmail !== "") {
+    found = await searchForUser(client, { query: newEmail });
```
bleonard: it's interesting to me this is `query` here and not `email` specifically.
The above is `external_id` specifically, which suggests there is `email`

This docs has a customer field in search: https://developer.zendesk.com/api-reference/ticketing/ticket-management/search/

The issue with a generic `query` would be if that email address was just part of some text somewhere. It would still return that user even though it wasn't the value of their "email" field.

bleonard (changes requested):

plugins/@grouparoo/zendesk/src/lib/export/exportRecord.ts
pauloouriques: I didn't find a way to perform this via API client. Then I had to do it in code.

plugins/@grouparoo/zendesk/src/lib/export/exportRecord.ts
```diff
@@ -174,8 +174,9 @@ export const exportRecord: ExportRecordPluginMethod = async ({
 export async function searchForUser(client, findBy: any) {
   const response = await client.users.search(findBy);
   let found = null;
+  const email = findBy.query;
```
bleonard: if we do have to do it this way, let's still pass in `{email}` (and not `query`) into this function and build the `query` here.

plugins/@grouparoo/zendesk/src/lib/export/exportRecord.ts
```diff
@@ -172,11 +172,19 @@ export const exportRecord: ExportRecordPluginMethod = async ({
...
   for (const user of response) {
     if (user.active) {
-      if (!found) {
+      if (
+        !found &&
+        (!email ||
+          email.toLowerCase().trim() === user.email.toLowerCase().trim())
+      ) {
         // not deleted
         found = user;
```
bleonard: I think we have to get rid of this code below here that throws. there might actually be more than one and that's why you are filtering to match the email.

bleonard (approved):
Looks good.

bleonard:
Definitely some weird cases here.
It seems that Zendesk was already having the desired behavior. That is' if you sent an external id and an email and there was already a user with that email and no external id, it would add the external id to the existing user.
The bug that this fixes is that we would lose existing tags on that record. Now there are persisted.

There was some struggle with finding a user by email. We tried searching for `email:name@example.com` but it seemed unreliable. I believe that was because of indexing time. Maybe Zendesk is switching to using Elasticsearch or something in those cases. What we found to be more reliable was just searching for the email. However, that could return someone who wrote an email in some other field. So we add a check in post-processing to make sure it matches.

PR #2432, opened 2021-10-19
add model ids to imports and exports in ui

teallarson:
## Change description

Adds the model id and a link to the model on imports and exports.
Screen Shot 2021-10-19 at 3 08 04 PM

Screen Shot 2021-10-19 at 3 08 19 PM

Screen Shot 2021-10-19 at 3 08 43 PM

Screen Shot 2021-10-19 at 3 09 01 PM

Imports do not track and return their sources in `apiData` like exports track and return their destinations.  Instead, I've pulled the `modelId` from the associated record and return the id with the API Data.

## Related issues

no

### Development

### Impact

>

### Code review

### DIFF
--- a/core/src/actions/imports.ts
+++ b/core/src/actions/imports.ts
@@ -1,3 +1,5 @@
+import { Includeable } from "sequelize/types";
+import { GrouparooRecord } from "..";
 import { AuthenticatedAction } from "../classes/actions/authenticatedAction";
 import { Import } from "../models/Import";
 import { Property } from "../models/Property";
@@ -27,13 +29,24 @@ export class ImportsList extends AuthenticatedAction {
     const where = {};
     if (params.creatorId) where["creatorId"] = params.creatorId;
     if (params.recordId) where["recordId"] = params.recordId;
+    let include: Includeable[] = [];
+    if (params.recordId) {
+      include = [
+        {
+          model: GrouparooRecord,
+          where: { id: params.recordId },
+        },
+      ];
+    }
 
     const search = {
+      include,
       where,
       limit: params.limit,
       offset: params.offset,
       order: params.order,
     };
+
     const total = await Import.count({ where });
     const imports = await Import.findAll(search);
 
--- a/core/src/actions/imports.ts
+++ b/core/src/actions/imports.ts
@@ -1,8 +1,7 @@
 import { Includeable } from "sequelize/types";
-import { GrouparooRecord } from "..";
+import { GrouparooRecord } from "../models/GrouparooRecord";
 import { AuthenticatedAction } from "../classes/actions/authenticatedAction";
 import { Import } from "../models/Import";
-import { Property } from "../models/Property";
 import { APIData } from "../modules/apiData";
 
 export class ImportsList extends AuthenticatedAction {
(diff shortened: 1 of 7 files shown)


ui/ui-components/pages/import/[id]/edit.tsx
```diff
@@ -35,6 +38,11 @@ export default function Page(props) {
         <Card.Body>
           <h2>Details</h2>
           <p>
+            Model:{" "}
+            <Link href={`/model/${_import.modelId}/edit`}>
```
Comment (teallarson) [inline ui/ui-components/pages/import/[id]/edit.tsx:42]: intentionally not using the `as` here.  It was flashing a quick (~.25 second) next error with the routing when I did.

core/src/models/Import.ts
```diff
@@ -153,12 +154,15 @@ export class Import extends CommonModel<Import> {
     delete data._meta;
     delete rawData._meta;
 
+    const record = await this.$get("record");
```
evantahler, line 157: This means that we should probably pre-load the record when we get large lists of Imports, like the `imports:list` again - use an `include:[]`

ui/ui-components/pages/import/[id]/edit.tsx
Comment (evantahler) [reply ui/ui-components/pages/import/[id]/edit.tsx:42]: Huh! I would have expected the error flash to happen /without/ the as.  no need to change it I guess if it's working

evantahler (changes requested):
Please pre-load the Record in `imports:list` now that it will be used in `import.apiData()`
The `<Link>` comment is strange to me, but non-blocking.

core/src/models/Import.ts
teallarson: Added.  Ideas how to test that?  It looks like `properties:list` is the only other action using an `Includeable[]` and I don't see a test that demonstrates it.

core/src/actions/imports.ts
```diff
@@ -27,13 +28,24 @@ export class ImportsList extends AuthenticatedAction {
     const where = {};
     if (params.creatorId) where["creatorId"] = params.creatorId;
     if (params.recordId) where["recordId"] = params.recordId;
+    let include: Includeable[] = [];
+    if (params.recordId) {
+      include = [
+        {
+          model: GrouparooRecord,
+          where: { id: params.recordId },
+        },
```
evantahler: You actually don't the `Includeable` type - it's checked for you as part of the search clause.  I think you may have added an extra where clause.

```ts
const search = {
      include: [GrouparooRecord], // since Imports `@BelongTo` GrouparooRecord the association is already handled 
      where,
      limit: params.limit,
      offset: params.offset,
      order: params.order,
    };
```

evantahler: There are lots of other places using `include:[]`, like [here](https://github.com/grouparoo/grouparoo/blob/70c622aa791dfc32bb7daa78800da1a335b81815/core/src/modules/middleware/authentication.ts#L65-L68) - load teamMembers, but include the team and permissions so we can do `teamMember.team.permissions... `

evantahler (changes requested):

core/src/actions/imports.ts
teallarson: Ah ha.  Sounds like I was looking in the wrong place.  Thanks.

evantahler (approved):

PR #2436, opened 2021-10-20
Do not show Group Labels in UI if not present in plugin

evantahler:
Allows the labels a Destination Plugin provides not to include the "groups" collection.  If these labels are excluded, the UI for mapping Groups to tags/lists/etc will not be shown.

This PR does not remove the group labels from any of the plugins.

### Development

### Impact

>

### Code review

### DIFF
--- a/core/src/classes/plugin.ts
+++ b/core/src/classes/plugin.ts
@@ -601,7 +601,7 @@ export interface DestinationMappingOptionsMethodResponse {
       singular: string; // merge var
       plural: string; // merge vars
     };
-    group: {
+    group?: {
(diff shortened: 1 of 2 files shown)


teallarson (approved):
This will be useful for the destination changes we're making!  It looks like the plugin class isn't tested in core -- is there anywhere this should be tested?

evantahler:
> This will be useful for the destination changes we're making! It looks like the plugin class isn't tested in core -- is there anywhere this should be tested?

This should get tested when one of the plugins changes to not have a groups lablel.  I think Salesforce, in some circumstances, would be first

PR #2439, opened 2021-10-20
AppDataRefresh model test passing

evantahler:
A few fixes to unblock https://github.com/grouparoo/grouparoo/pull/2408

### DIFF
--- a/core/__tests__/models/appDataRefresh.ts
+++ b/core/__tests__/models/appDataRefresh.ts
@@ -1,29 +1,28 @@
 import { helper } from "@grouparoo/spec-helper";
-import { api, redis, utils } from "actionhero";
-import { plugin, App, AppDataRefresh, GrouparooModel, Log } from "../../src";
+import { App, AppDataRefresh, Log } from "../../src";
+import { AppDataRefreshOps } from "../../src/modules/ops/appDataRefresh";
 
 describe("appDataRefresh", () => {
+  let app: App;
   helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
 
   describe("with app", () => {
-    const app = new App({
-      name: "test app",
-      type: "test-plugin-app",
+    beforeAll(async () => {
+      app = await helper.factories.app();
     });
-    app.save();
-    beforeAll(async () => {});
 
     test("an app data refresh can be created with an app", async () => {
       const appDataRefresh = new AppDataRefresh({
         appId: app.id,
         refreshQuery: "test query",
       });
+
       await appDataRefresh.save();
 
       expect(appDataRefresh.id.length).toBe(40);
       expect(appDataRefresh.createdAt).toBeTruthy();
       expect(appDataRefresh.updatedAt).toBeTruthy();
-      expect(appDataRefresh.value).toBeFalsy();
+      expect(appDataRefresh.value).toBeTruthy();
     });
 
     test("creating an app data refresh creates a log entry", async () => {
@@ -41,13 +40,14 @@ describe("appDataRefresh", () => {
         refreshQuery: "test query",
       });
       await appDataRefresh.save();
+
+      const spy = jest.spyOn(AppDataRefreshOps, "checkDataRefreshValue");
       await appDataRefresh.update({
         refreshQuery: "SELECT MAX(updated_at) FROM users;",
       });
-      await appDataRefresh.save();
-      expect(AppDataRefresh.runCheckIfNewQuery).toHaveBeenCalledWith(
-        appDataRefresh
-      );
+      expect(spy).toHaveBeenCalledWith(appDataRefresh);
+
+      spy.mockRestore();
     });
   });
 });
--- a/core/src/classes/plugin.ts
+++ b/core/src/classes/plugin.ts
@@ -47,7 +47,7 @@ export interface PluginApp {
     connect?: ConnectPluginAppMethod;
     disconnect?: DisconnectPluginAppMethod;
     test: TestPluginMethod;
-    query: AppQueryMethod;
+    query?: AppQueryMethod;
     appOptions?: AppOptionsMethod;
     parallelism?: AppParallelismMethod;
   };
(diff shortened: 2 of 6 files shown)


core/__tests__/models/appDataRefresh.ts
```diff
@@ -1,29 +1,28 @@
...
 describe("appDataRefresh", () => {
+  let app: App;
   helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
 
   describe("with app", () => {
-    const app = new App({
-      name: "test app",
-      type: "test-plugin-app",
+    beforeAll(async () => {
+      app = await helper.factories.app();
```
evantahler, line 11: In the tests, it's usually easier to use the factories.  The reason that `app` wasn't being found in AppDataRefreshOps was that the app wasn't ready yet.  To become ready, it would have had to have options set and then changed into the ready state.  The factory does this all for you

core/__tests__/models/appDataRefresh.ts
```diff
@@ -41,13 +40,14 @@ describe("appDataRefresh", () => {
         refreshQuery: "test query",
       });
       await appDataRefresh.save();
+
+      const spy = jest.spyOn(AppDataRefreshOps, "checkDataRefreshValue");
```
evantahler, line 44: You can't call `.toHaveBeenCalledWith` on a regular method, you need to "spy" on it with jest.  
If you spy on the actual model's method, you end up removing the `@beforeSave` hooks, so I spied on the Ops instead to confirm that it was called

core/__tests__/models/appDataRefresh.ts
```diff
@@ -41,13 +40,14 @@ describe("appDataRefresh", () => {
...
       await appDataRefresh.update({
         refreshQuery: "SELECT MAX(updated_at) FROM users;",
       });
-      await appDataRefresh.save();
-      expect(AppDataRefresh.runCheckIfNewQuery).toHaveBeenCalledWith(
-        appDataRefresh
-      );
+      expect(spy).toHaveBeenCalledWith(appDataRefresh);
+
+      spy.mockRestore();
```
evantahler, line 50: Always remember to un-spy, or else the method won't don anything for the rest of your test suite

core/src/classes/plugin.ts
```diff
@@ -371,6 +372,18 @@ export interface TestPluginMethod {
...
+ * Method is used to run arbitrary queries against the app.  AppDataRefresh uses this.  It returns arrays of "rows", whatever that means to the App
+ */
+export interface AppQueryMethod {
+  (argument: {
+    app: App;
+    appId: string;
+    appOptions: SimpleAppOptions;
+    connection: any;
+  }): Promise<Array<unknown>>;
+}
```
evantahler, line 385: I made up a generic `query` method that plugins can opt-into.  They get the same arguments as `test`

core/src/models/AppDataRefresh.ts
```diff
@@ -99,8 +100,11 @@ export class AppDataRefresh extends LoggedModel<AppDataRefresh> {
 
   @BeforeSave
   static async runCheckIfNewQuery(instance: AppDataRefresh) {
-    if (instance.changed("refreshQuery")) {
-      AppDataRefreshOps.checkDataRefreshValue(instance);
+    if (
+      instance.changed("refreshQuery") &&
+      !instance.changed("lastConfirmedAt")
```
evantahler, line 105: You'll have an infinite loop if you don't check that you've already run `checkDataRefreshValue()` once before in this save (because you update the value in the Op, so it calls '@beforeSave' again).  I used the `lastConfirmedAt` timestmp as a way to check

core/src/modules/ops/appDataRefresh.ts
```diff
@@ -1,22 +1,33 @@
...
+    }
 
-    const sampleValue = JSON.stringify(response.rows);
-    console.log("sample value: " + sampleValue);
+    const responseRows = await pluginApp.methods.query({
+      app,
+      appId: app.id,
+      appOptions: options,
+      connection,
+    });
```
evantahler, line 28: Here we are calling the plugin's new `query` method, if it exists

core/src/modules/plugin.ts
```diff
@@ -36,6 +37,7 @@ import { ExportProcessor } from "../models/ExportProcessor";
 // the order matters here - the children need to come before the parents (destinationGroup -> destination)
 const models = [
   App,
+  AppDataRefresh,
```
evantahler, line 40: This is needed so plugins can 'boot up' and connect to all the models

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -292,6 +292,9 @@ export namespace helper {
                 query: "hi!",
               };
             },
+            query: async () => {
+              return [new Date().getTime()];
+            },
```
evantahler, line 297: Here's a mock implementation of the new `query` method on the plugin that just returns the time... this way in your tests, it will have a new value every time you call it

core/src/classes/plugin.ts
evantahler: You make want to make something more specific to appDataRefresh, but this shows the pattern of how to add a new method to the plugins

teallarson (approved):
Thanks!  I had a feeling the way I was building the connection was incomplete.  Looks like there were multiple moving parts I missed.  

EDIT: The force-pushes are resolving issues that popped up merging as I'd made some changes on my branch while you were working on these.

core/src/classes/plugin.ts
teallarson: Super helpful.  I'll tweak it a bit once it's merged in.

core/src/models/AppDataRefresh.ts
teallarson: The query itself (`appDataRefresh.refreshQuery`) doesn't get updated in `AppDataRefreshOps.checkDataRefreshValue()`, just `appDataRefresh.value`, so I'm not sure I follow here.

PR #2440, opened 2021-10-20
Hubspot custom objects plugin

pauloouriques:
Hubspot custom objects destination pluginin


plugins/@grouparoo/hubspot/__tests__/export-objects/destination-mapping-options.ts
```diff
@@ -0,0 +1,50 @@
...
+    destinationId: null,
+  });
+}
+
+describe("hubspot/destinationMappingOptions", () => {
+  test("can load all destinationMappingOptions", async () => {
+    const destinationOptions = {
+      schemaId: "CONTACT",
+      primaryKey: "email",
+    };
```
evantahler, line 26: That's cool - using Contacts as if they were custom objects

plugins/@grouparoo/hubspot/src/initializers/plugin.ts
```diff
@@ -33,9 +36,15 @@ export class Plugins extends Initializer {
...
+          [path.join(templateRoot, "destination", "contacts", "*.template")],
+          contactsSupportedSyncModes,
+          defaultSyncMode
+        ),
+        new DestinationTemplate(
+          "hubspot:objects",
+          [path.join(templateRoot, "destination", "objects", "*.template")],
+          objectsSupportedSyncModes,
           defaultSyncMode
         ),
```
evantahler, line 49: 2 templates 👍

plugins/@grouparoo/hubspot/src/lib/client/client.ts
```diff
@@ -95,6 +85,15 @@ class HubspotClient {
...
   }
+
+  async _request(config: AxiosRequestConfig): Promise<any> {
+    config["baseURL"] = "https://api.hubapi.com";
+    config["params"] = Object.assign({}, config.params, {
+      hapikey: this.hapikey,
+    });
+    const { data = {} } = await Axios(config);
+    return data;
+  }
```
evantahler, line 96: 👍 helper

plugins/@grouparoo/hubspot/src/lib/client/client.ts
```diff
@@ -1,66 +1,56 @@
...
-    const { data }: { data: HubspotDataResponse } = await Axios({
+    const data = await this._request({
       method: "GET",
-      url,
-      headers: { "Content-Type": "application/json" },
+      url: `/contacts/v1/lists/static`,
+      params: { count: 999 },
     });
     return data.lists;
   }
```
evantahler, line 25: Now that this client is getting rather meta - it's probably time to type some of these responses, even partially.  like, in this case we know that `data.lists` is an array of objects, and we probably know /some/ of the keys of those objects

plugins/@grouparoo/hubspot/src/lib/client/objects.ts
```diff
@@ -0,0 +1,89 @@
...
+    const data = {
+      filterGroups,
+      properties,
+    };
+    const response = await this.client._request({
+      data,
+      method: "POST",
+      url: `/crm/v3/objects/${schemaId}/search`,
+    });
+    return response?.results || [];
```
evantahler, line 55: Some comment about typing as above - we don't know all the properties of this object, but we do know that 

```ts
return (response?.results || []) as Array<Record<string, string | number | date>>; // I think that Hubspot objects are all 1-dimensional objects (no nested objects)
```

plugins/@grouparoo/hubspot/src/lib/export-contacts/connection.ts
```diff
@@ -0,0 +1,29 @@
...
+  "sync",
+  "additive",
+  "enrich",
+];
+
+export const contactsDestinationConnection: PluginConnection = {
+  name: "hubspot-export-contacts",
+  displayName: "Hubspot Export Contacts",
+  direction: "export",
+  description: "Export Records to Hubspot and add them to Contact Lists.",
```
evantahler:
```suggestion
  description: "Export Records as Hubspot Contacts and add them to Contact Lists.",
```

plugins/@grouparoo/hubspot/src/lib/export-contacts/exportRecord.ts
```diff
@@ -140,6 +140,7 @@ export const exportRecord: ExportRecordPluginMethod = async ({
 
     return { success: true };
   } catch (error) {
+    console.log(error.response?.data);
```
evantahler: unnecessary?

plugins/@grouparoo/hubspot/src/lib/export-objects/connection.ts
```diff
@@ -0,0 +1,44 @@
...
+import { destinationOptions } from "./destinationOptions";
+import { destinationMappingOptions } from "./destinationMappingOptions";
+import { exportArrayProperties } from "../shared/exportArrayProperties";
+
+export const objectsSupportedSyncModes: DestinationSyncMode[] = [
+  "sync",
+  "additive",
+  "enrich",
+];
+
```
evantahler, line 13: Nice! All 3 sync modes work on custom objects

plugins/@grouparoo/hubspot/src/lib/export-objects/connection.ts
```diff
@@ -0,0 +1,44 @@
...
+export const objectsDestinationConnection: PluginConnection = {
+  name: "hubspot-export-objects",
+  displayName: "Hubspot Export Objects",
+  direction: "export",
+  description:
+    "Export Records to Hubspot Custom Objects and associate them with Hubspot standard objects or others custom objects.",
```
evantahler:
```suggestion
    "Export Records as Hubspot Custom Objects and associate them with Hubspot standard objects or other custom objects.",
```

evantahler: What do you mean by `and associate them` in this case?

plugins/@grouparoo/hubspot/src/lib/export/options.ts
```diff
@@ -0,0 +1,119 @@
...
+  public async getDestinationOptions({
+    destinationOptions,
+  }): Promise<DestinationOptionsMethodResponse> {
+    const { appOptions } = this.cacheData;
+    this.client = await connect(appOptions);
+    this.customObjects = await this.getCustomObjects();
+    const out: DestinationOptionsMethodResponse = {};
+    Object.assign(out, await this.getRecordOptions(destinationOptions));
+    return out;
+  }
```
evantahler, line 34: Nice! this will make it easy to choose the custom objects that exist

evantahler (changes requested):
Strong work & good tests!

I think that it's prudent to add some better types where you can for the API client response methods - things are getting complex.

Thanks for adding `destinationOptions` too! The UI looks good.
Screen Shot 2021-10-27 at 6 14 01 PM

evantahler:
There's something not quite right with the `destinationOptions` - it always look like its returning the options for a `QUOTE` object.  I would expect contacts to have things like "email address" 

Screen Shot 2021-10-27 at 6 18 06 PM

plugins/@grouparoo/hubspot/src/lib/export-objects/destinationMappingOptions.ts
```diff
@@ -0,0 +1,116 @@
...
+  return null;
+};
+
+export const getRequiredFields = (
+  customObject: CustomObjectHandler,
+  primaryKey: string
+): Array<{
+  key: string;
+  type: DestinationMappingOptionsResponseType;
+}> => {
```
bleonard, line 71: Is it possible they have fields that are required for a schema?
those would go here too.

bleonard (approved):

plugins/@grouparoo/hubspot/src/lib/client/objects.ts
pauloouriques: I did it, but we could save it to do when we're moving to the new API client. What do you think?

evantahler (approved):
💥

PR #2441, opened 2021-10-20
A few fixes for when Records and Models are deleted

evantahler:
A few fixes for when Records and Models are deleted:

* When a Record is destroyed, any final Exports will explicitly set `toDelete=true`
* Implement GrouparooModel state machine
* Prevent situations where we delete a model (and source, properties, etc) from prematurely deleting Grouparoo's local data before sending the final export to any destination that had previously been tracking the model

## TODO:
- [ ] Test all the new behavior
- [ ] Sort out why the final export's properties aren't getting the right data on that last export.  
   * Or, maybe the mailchimp plugin doesn't know what to do when there's no new email but there is an old one?
- [ ] It's time to refactor DestinationOps

### Development

### Impact

>

### Code review

### DIFF
--- a/core/src/models/GrouparooModel.ts
+++ b/core/src/models/GrouparooModel.ts
@@ -18,14 +18,17 @@ import { APIData } from "../modules/apiData";
 import { ConfigWriter } from "../modules/configWriter";
 import { LockableHelper } from "../modules/lockableHelper";
 import { Destination } from "./Destination";
+import { GrouparooRecord } from "./GrouparooRecord";
 import { Group } from "./Group";
 import { RunOps } from "../modules/ops/runs";
+import { StateMachine } from "../modules/stateMachine";
 
 export const ModelTypes = ["profile"] as const;
 export type ModelType = typeof ModelTypes[number];
 
 const STATES = ["ready", "deleted"] as const;
 const STATE_TRANSITIONS = [
+  { from: "draft", to: "ready", checks: [] },
   { from: "ready", to: "deleted", checks: [] },
   {
     from: "deleted",
@@ -137,6 +140,11 @@ export class GrouparooModel extends LoggedModel<GrouparooModel> {
     await LockableHelper.beforeSave(instance);
   }
 
+  @BeforeSave
+  static async updateState(instance: Group) {
+    await StateMachine.transition(instance, STATE_TRANSITIONS);
+  }
+
   @BeforeDestroy
   static async noDestroyIfLocked(instance) {
     await LockableHelper.beforeDestroy(instance);
@@ -166,6 +174,14 @@ export class GrouparooModel extends LoggedModel<GrouparooModel> {
     if (groups.length > 0) {
       throw new Error(
         `cannot delete this model, group ${groups[0].id} relies on it`
+      );
+    }
+    const records = await GrouparooRecord.scope(null).count({
+      where: { modelId: instance.id },
+    });
+    if (records > 0) {
+      throw new Error(
+        `cannot delete this model, ${records} records rely on it`
       );
     }
   }
--- a/core/src/models/Property.ts
+++ b/core/src/models/Property.ts
@@ -40,6 +40,7 @@ import { GrouparooRecord } from "./GrouparooRecord";
 import { RecordProperty } from "./RecordProperty";
 import { Run } from "./Run";
 import { Source } from "./Source";
+import { Destination } from "./Destination";
 
 export function propertyJSToSQLType(jsType: string) {
   const map = {
@@ -583,6 +584,25 @@ export class Property extends LoggedModel<Property> {
       throw new Error(
         `cannot delete property "${instance.key}" as ${mapping.ownerId} is using it in a mapping`
       );
+    }
+
+    if (instance.directlyMapped) {
+      const source = await Source.scope(null).findOne({
+        where: { id: instance.sourceId },
+        logging: true,
+      });
+      if (source) {
+        // We also call this method from Source@beforeDestroy hook ensureDirectlyMappedPropertyNotInUse - the source will be null when trying to be deleted
+        const model = await source.$get("model", { scope: null });
+        const destinationsCount = await Destination.scope(null).count({
+          where: { modelId: model.id },
+        });
+        if (destinationsCount > 0) {
+          throw new Error(
+            `cannot delete property "${instance.key}" as this property is directly mapped and there is still a destination`
+          );
+        }
+      }
     }
   }
 
--- a/core/src/modules/ops/destination.ts
+++ b/core/src/modules/ops/destination.ts
@@ -328,7 +328,8 @@ export namespace DestinationOps {
     record: GrouparooRecord,
     synchronous = false,
     force = false,
-    saveExports = true
+    saveExports = true,
+    toDelete?: boolean
   ) {
     if (destination.modelId !== record.modelId) {
       throw new Error(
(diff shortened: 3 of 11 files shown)


core/src/tasks/record/destroy.ts
```diff
@@ -34,7 +34,7 @@ export class RecordDestroy extends CLSTask {
       // clear groups and export
       // when the export is done, this task will be enqueued again to destroy it
       await GrouparooRecord.destroyGroupMembers(record);
-      await record.export(false, oldGroups, true, false);
+      await record.export(false, oldGroups, true, false, true);
```
evantahler, line 37: When we destroy a record, we pass `toDelete=true` all the way though the plumbing so we know that any eventual exports we make should try and delete this record from the destination.

core/src/models/GrouparooModel.ts
```diff
@@ -137,6 +140,11 @@ export class GrouparooModel extends LoggedModel<GrouparooModel> {
     await LockableHelper.beforeSave(instance);
   }
 
+  @BeforeSave
+  static async updateState(instance: Group) {
+    await StateMachine.transition(instance, STATE_TRANSITIONS);
+  }
+
```
evantahler, line 147: Looks like GrouparooModel wasn't actually using the state machine logic

core/src/models/GrouparooModel.ts
```diff
@@ -168,5 +176,13 @@ export class GrouparooModel extends LoggedModel<GrouparooModel> {
...
       );
     }
+    const records = await GrouparooRecord.scope(null).count({
+      where: { modelId: instance.id },
+    });
+    if (records > 0) {
+      throw new Error(
+        `cannot delete this model, ${records} records rely on it`
+      );
+    }
```
evantahler, line 186: [Delete Check] - We shouldn't delete a Model if it still has records

core/src/modules/ops/destination.ts
```diff
@@ -399,7 +403,13 @@ export namespace DestinationOps {
...
     for (const k in mapping) {
       const property = properties.find((r) => r.key === mapping[k]);
-      if (!property) throw new Error(`cannot find rule for ${mapping[k]}`);
+      if (!property) {
+        if (destination.state === "ready") {
+          throw new Error(`cannot find rule for ${mapping[k]}`);
+        } else {
+          continue;
+        }
+      }
```
evantahler, line 412: In the case we are deleting an exported record, we don't really care what the old property values are, so we shouldn't throw.  Otherwise, we would need to consider Properties with state=deleted and not use the cache

core/src/modules/ops/record.ts
```diff
@@ -818,9 +818,13 @@ export namespace RecordOps {
     let models: GrouparooModel[] = await GrouparooModel.scope(null).findAll();
     let modelIdsToClear: string[] = [];
 
+    const properties = await Property.scope(null).findAll({
+      include: [{ model: Source.unscoped() }],
+    });
```
evantahler, line 823: We need deleted properties, so don't use the cache

core/src/tasks/grouparooModel/run.ts
```diff
@@ -46,10 +46,12 @@ export class RunInternalRun extends CLSTask {
...
-      await RecordProperty.update(
-        { state: "pending" },
-        { where: { recordId: { [Op.in]: records.map((p) => p.id) } } }
-      );
+      if (model.state === "ready") {
+        await RecordProperty.update(
+          { state: "pending" },
+          { where: { recordId: { [Op.in]: records.map((p) => p.id) } } }
+        );
+      }
```
evantahler, line 54: If the model is being deleted, we shouldn't update the record properties to pending - as they won't be re-imported again

core/src/modules/ops/record.ts
```diff
@@ -67,10 +67,8 @@ export namespace RecordOps {
       const property = properties.find(
         (r) => r.id === recordProperties[i].propertyId
       );
-      if (!property) {
-        await recordProperties[i].destroy();
-        continue;
-      }
```
evantahler, line 73: This one was interesting - previously, `record.getProperties()` would remove any record properties that shouldn't be there any more (maybe because a property was deleted...).  

I think this is extraneous because we'll remove these properties when:
* The record is deleted
* the property is deleted
* we've got a sweeper looking for orphan properties that would delete these too: `recordProperties:sweep`

core/src/models/Property.ts
```diff
@@ -584,6 +585,25 @@ export class Property extends LoggedModel<Property> {
...
+        const destinationsCount = await Destination.scope(null).count({
+          where: { modelId: model.id },
+        });
+        if (destinationsCount > 0) {
+          throw new Error(
+            `cannot delete property "${instance.key}" as this property is directly mapped and there is still a destination`
+          );
+        }
+      }
+    }
```
evantahler, line 606: We don't want to allow a directly-mapped Property to be deleted if there is *any* destination still around.  We might still be working on exporting those records.   If we were to remove the directlyMapped property, we would then sweep away the records once they had no properties

evantahler (commented):
TODO: There's something fishy with `DestinationOps.buildExportedRecords()` when everything is in the deleted state - we aren't getting the correct 'oldProperty' email address to Mailchimp to delete the profile

core/src/modules/ops/record.ts
```diff
@@ -818,9 +818,13 @@ export namespace RecordOps {
...
 
+    const properties = await Property.scope(null).findAll({
+      include: [{ model: Source.unscoped() }],
+    });
+
     for (const model of models) {
-      const propertiesByModel: Property[] = await Property.findAllWithCache(
-        model.id
+      const propertiesByModel = properties.filter(
+        (property) => property.source.modelId === model.id
```
pedroslopez, line 827: can probably combine this with the filter right below that looks for directlyMapped properties

core/src/tasks/record/destroy.ts
pedroslopez, line 37: Note that all of this is happening _if the record has any groups_. If it's not in any groups but being sent to a destination that is tracking a model, the final export would not be triggered. A better check may be "if it's in any destinations" instead of "if it's in any groups", but I don't think we can get that in a relatively straightforward way at the moment.

core/src/models/GrouparooModel.ts
```diff
@@ -137,6 +140,11 @@ export class GrouparooModel extends LoggedModel<GrouparooModel> {
     await LockableHelper.beforeSave(instance);
   }
 
+  @BeforeSave
+  static async updateState(instance: Group) {
```
pedroslopez, line 144:
```suggestion
  static async updateState(instance: GrouparooModel) {
```

core/src/models/Property.ts
pedroslopez: This is interesting- when clearing things out, the task that gets enqueued to destroy records only happens if there are no directly mapped properties for the model. If this prevents the directlyMapped property from getting deleted, does it mean that records wont ever get their final export?

core/src/modules/ops/destination.ts
pedroslopez: I'm not sure the destination state is what would need to be checked here then - I believe the condition would be `if(!toDelete)` if that's the intention

core/src/modules/ops/record.ts
pedroslopez: I can't find the original story that stated this issue, what was the reasoning for why we needed to include deleted properties?

pedroslopez (commented):
Lots of interesting stuff!

I believe there are still some things to look into for the record deletion...

evantahler:
Replaced by https://github.com/grouparoo/grouparoo/pull/2443 and https://github.com/grouparoo/grouparoo/pull/2445

PR #2442, opened 2021-10-21
dont force import all properties when updating destination tracking

pedroslopez:
## Change description

In general, we no longer want to import all record properties when exporting / changing tracked object for destinations. This was being handled by the `force` option on Model and Group runs.

`force` was actually being used for two things in these runs:
- control whether we should import all properties or not
- (for groups) control wether we should import all records or records that are not already in this group

For the second point, we were already doing this if a `destinationId` was passed to the run, which covers all the cases where we were passing `force=true` anyway. 

Because we're no longer using `force` and it had a confusing dual-meaning, I've removed the concept from model/group runs.

## Related issues

No

### Development

### Impact

>

### Code review

### DIFF
--- a/core/src/modules/ops/group.ts
+++ b/core/src/modules/ops/group.ts
@@ -16,7 +16,6 @@ export namespace GroupOps {
     recordIds: string[],
     creatorType: string,
     creatorId: string,
-    force: boolean,
     destinationId?: string
   ) {
     const bulkData = [];
@@ -35,7 +34,7 @@ export namespace GroupOps {
 
     const _imports = await Import.bulkCreate(bulkData);
 
-    await RecordOps.markPendingByIds(recordIds, force);
+    await RecordOps.markPendingByIds(recordIds, false);
 
     return _imports;
   }
@@ -277,7 +276,6 @@ export namespace GroupOps {
       recordsNeedingGroupMembership.map((p) => p.id),
       "run",
       run.id,
-      force,
       destinationId
     );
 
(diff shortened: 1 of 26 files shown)


core/src/tasks/grouparooModel/run.ts
```diff
@@ -46,14 +46,9 @@ export class RunInternalRun extends CLSTask {
...
-        { state: "pending" },
-        { where: { recordId: { [Op.in]: records.map((p) => p.id) } } }
-      );
-
-      await GrouparooRecord.update(
-        { state: "pending" },
-        { where: { id: { [Op.in]: records.map((p) => p.id) } } }
+      await RecordOps.markPendingByIds(
+        records.map((r) => r.id),
+        false
```
evantahler: Well that's a lot nicer!

core/src/modules/ops/group.ts
```diff
@@ -35,7 +34,7 @@ export namespace GroupOps {
 
     const _imports = await Import.bulkCreate(bulkData);
```
evantahler, line 34: 👍 We will still create an Import for each record so we can use this to determine what's changed (group membership maybe) and determine when it is time to export

core/src/tasks/grouparooModel/run.ts
```diff
@@ -46,14 +46,9 @@ export class RunInternalRun extends CLSTask {
...
       );
 
       // create imports to track the lineage of the record property values
```
evantahler: The section below is a lot like what happens in `ops/group#updateRecords` - can these be combined?

evantahler (approved):
What you've got here already is a great improvement - removing `force` in these cases is 👍 

This is 🚢 able as-is.  There might be a place to refactor in the notes if you wanted.

PR #2443, opened 2021-10-21
Remove previously sync'd Records from Destinations when they are deleted

evantahler:
Updates GrouparooRecords to have a `deleted` state.  This allows the `record:destroy` task to enqueue one final export for any Destinations that this Record had been previously synced too.  These final exports will be `toDelete=true` explicitly.

This PR also locks in that GrouparooModels cannot be destroyed if there are still GrouparooRecords that use the model.

More or less replaces https://github.com/grouparoo/grouparoo/pull/2441

### Development

### Impact

>

### Code review

### DIFF
--- a/core/__tests__/tasks/grouparooModel/destroy.ts
+++ b/core/__tests__/tasks/grouparooModel/destroy.ts
@@ -1,7 +1,6 @@
 import { helper } from "@grouparoo/spec-helper";
 import { api, task, specHelper } from "actionhero";
-import { Group } from "../../../dist";
-import { App, GrouparooModel, Destination, Source } from "../../../src";
+import { GrouparooModel, Destination } from "../../../src";
 
 describe("tasks/model:destroy", () => {
   helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
@@ -19,7 +18,7 @@ describe("tasks/model:destroy", () => {
     });
 
     test("will delete model immediately if it's not being used", async () => {
-      const model: GrouparooModel = await helper.factories.model();
+      const model = await helper.factories.model();
 
       await model.update({ state: "deleted" });
       await task.enqueue("model:destroy", { modelId: model.id });
@@ -33,9 +32,9 @@ describe("tasks/model:destroy", () => {
     });
 
     test("will wait for source to be deleted if it uses the model", async () => {
-      const model: GrouparooModel = await helper.factories.model();
-      const app: App = await helper.factories.app();
-      const source: Source = await helper.factories.source(app, {
+      const model = await helper.factories.model();
+      const app = await helper.factories.app();
+      const source = await helper.factories.source(app, {
         modelId: model.id,
       });
       await model.update({ state: "deleted" });
@@ -57,8 +56,8 @@ describe("tasks/model:destroy", () => {
     });
 
     test("will wait for destination to be deleted if it uses the model", async () => {
-      const model: GrouparooModel = await helper.factories.model();
-      const app: App = await helper.factories.app();
+      const model = await helper.factories.model();
+      const app = await helper.factories.app();
       const destination: Destination = await helper.factories.destination(app, {
         modelId: model.id,
       });
@@ -81,8 +80,8 @@ describe("tasks/model:destroy", () => {
     });
 
     test("will wait for group to be deleted if it uses the model", async () => {
-      const model: GrouparooModel = await helper.factories.model();
-      const group: Group = await helper.factories.group();
+      const model = await helper.factories.model();
+      const group = await helper.factories.group();
       await group.update({ modelId: model.id });
 
       await model.update({ state: "deleted" });
@@ -97,6 +96,28 @@ describe("tasks/model:destroy", () => {
 
       // remove group
       await group.destroy();
+
+      // run it again
+      await specHelper.runTask("model:destroy", foundTasks[0].args[0]);
+      await expect(model.reload()).rejects.toThrow(/does not exist anymore/);
+    });
+
+    test("will wait for record to be deleted if it uses the model", async () => {
+      const model = await helper.factories.model();
+      const record = await helper.factories.record({ modelId: model.id });
+
+      await model.update({ state: "deleted" });
+      await task.enqueue("model:destroy", { modelId: model.id });
+
+      const foundTasks = await specHelper.findEnqueuedTasks("model:destroy");
+      expect(foundTasks.length).toBe(1);
+      await specHelper.runTask("model:destroy", foundTasks[0].args[0]);
+
+      const reloadedModel = await GrouparooModel.findById(model.id);
+      expect(reloadedModel.state).toBe("deleted");
+
+      // remove group
+      await record.destroy();
 
       // run it again
       await specHelper.runTask("model:destroy", foundTasks[0].args[0]);
--- a/core/__tests__/tasks/grouparooModel/destroy.ts
+++ b/core/__tests__/tasks/grouparooModel/destroy.ts
@@ -116,7 +116,7 @@ describe("tasks/model:destroy", () => {
       const reloadedModel = await GrouparooModel.findById(model.id);
       expect(reloadedModel.state).toBe("deleted");
 
-      // remove group
+      // remove record
       await record.destroy();
 
       // run it again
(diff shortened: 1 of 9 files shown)


core/__tests__/tasks/record/destroy.ts
```diff
@@ -43,28 +42,28 @@ describe("tasks/record:destroy", () => {
...
 
   test("does not delete records that aren't ready", async () => {
-    const record: GrouparooRecord = await helper.factories.record();
+    const record = await helper.factories.record();
     await record.update({ state: "pending" });
 
     await specHelper.runTask("record:destroy", { recordId: record.id });
     await record.reload();
-    expect(record.state).toBe("pending");
+    expect(record.state).toBe("deleted");
```
evantahler: Only the "record:destroy" task should be setting the state to be deleted

core/src/models/GrouparooModel.ts
```diff
@@ -133,10 +136,15 @@ export class GrouparooModel extends LoggedModel<GrouparooModel> {
...
   @BeforeSave
-  static async noUpdateIfLocked(instance) {
+  static async noUpdateIfLocked(instance: GrouparooModel) {
     await LockableHelper.beforeSave(instance);
   }
 
+  @BeforeSave
+  static async updateState(instance: GrouparooModel) {
+    await StateMachine.transition(instance, STATE_TRANSITIONS);
+  }
```
evantahler, line 146: GrouparooModel wasn't really using the state machine before

core/src/models/GrouparooModel.ts
```diff
@@ -168,5 +176,13 @@ export class GrouparooModel extends LoggedModel<GrouparooModel> {
...
       );
     }
+    const records = await GrouparooRecord.unscoped().count({
+      where: { modelId: instance.id },
+    });
+    if (records > 0) {
+      throw new Error(
+        `cannot delete this model, ${records} records rely on it`
+      );
+    }
```
evantahler, line 186: We should be checking that there are no records left before destroying

core/__tests__/tasks/grouparooModel/destroy.ts
```diff
@@ -57,8 +56,8 @@ describe("tasks/model:destroy", () => {
     });
 
     test("will wait for destination to be deleted if it uses the model", async () => {
-      const model: GrouparooModel = await helper.factories.model();
-      const app: App = await helper.factories.app();
```
pedroslopez, line 61: curious about the removal of these typings... iirc the factories return `any`, is that true?

core/__tests__/tasks/grouparooModel/destroy.ts
```diff
@@ -102,5 +101,27 @@ describe("tasks/model:destroy", () => {
...
+      await task.enqueue("model:destroy", { modelId: model.id });
+
+      const foundTasks = await specHelper.findEnqueuedTasks("model:destroy");
+      expect(foundTasks.length).toBe(1);
+      await specHelper.runTask("model:destroy", foundTasks[0].args[0]);
+
+      const reloadedModel = await GrouparooModel.findById(model.id);
+      expect(reloadedModel.state).toBe("deleted");
+
+      // remove group
```
pedroslopez:
```suggestion
      // remove record
```

core/__tests__/tasks/record/destroy.ts
```diff
@@ -144,14 +143,82 @@ describe("tasks/record:destroy", () => {
...
+      where: { recordId: record.id, state: { [Op.ne]: "complete" } },
+    });
+    expect(exportCount).toBe(0);
+    await record.reload();
+    expect(record.state).toBe("ready");
+
+    // try to delete
+    await specHelper.runTask("record:destroy", { recordId: record.id });
+
+    // record is still there, but is now being exported
```
pedroslopez, line 183: we could also add a check to ensure that the export is a `toDelete: true` both here and on the group destination test

core/src/models/GrouparooModel.ts
```diff
@@ -168,5 +176,13 @@ export class GrouparooModel extends LoggedModel<GrouparooModel> {
         `cannot delete this model, group ${groups[0].id} relies on it`
       );
     }
+    const records = await GrouparooRecord.unscoped().count({
```
pedroslopez, line 179: oh! never seen this `unscoped()` thing before

core/src/tasks/record/destroy.ts
```diff
@@ -29,12 +27,13 @@ export class RecordDestroy extends CLSTask {
     });
     if (pendingExports > 0) return;
 
-    const oldGroups = await record.$get("groups");
-    if (oldGroups.length > 0) {
+    if (["draft", "pending", "ready"].includes(record.state)) {
```
pedroslopez: if the record is `pending`, should we wait until it's done importing?

pedroslopez (changes requested):
Nice, this `deleted` state makes a lot of sense. At the moment I don't think the `record:destroy` task will be enqueued again for deleted records though, so let's check that.

core/src/tasks/record/destroy.ts
```diff
@@ -29,12 +27,13 @@ export class RecordDestroy extends CLSTask {
...
       // clear groups and export
       // when the export is done, this task will be enqueued again to destroy it
+      const oldGroups = await record.$get("groups");
       await GrouparooRecord.destroyGroupMembers(record);
-      await record.export(false, oldGroups, true, false);
+      await record.update({ state: "deleted" });
```
pedroslopez, line 37: whoops, didnt submit this with the review:

I believe this `record:destroy` task is only enqueued if the records are `ready` and match a few conditions - if we now have to enqueue it for `deleted` records as well, the `getRecordsToDestroy` function will need to be updated, but I'm not seeing any changes to it in this PR.

pedroslopez (changes requested):

core/__tests__/tasks/grouparooModel/destroy.ts
evantahler: As of about a week ago, the factories now properly return the types!  
I haven't gone through the test suite and updated them all, I'm just removing extraneous types as I find them as it's not a problem since the types match

pedroslopez: oh that's great!

core/src/tasks/record/destroy.ts
evantahler: OH... thanks!

Added.

core/src/tasks/record/destroy.ts
evantahler: `getRecordsToDestroy()` won't include `pending` or `draft` records... so I guess not!

core/src/models/GrouparooRecord.ts
```diff
@@ -140,6 +143,10 @@ export class GrouparooRecord extends LoggedModel<GrouparooRecord> {
   }
 
   async buildNullProperties(state: GrouparooRecord["state"] = "pending") {
```
bleonard, line 145: when does this get called? What does throwing here end up doing?

core/src/models/GrouparooRecord.ts
```diff
@@ -179,9 +186,17 @@ export class GrouparooRecord extends LoggedModel<GrouparooRecord> {
...
     oldGroupsOverride?: Group[],
     saveExports = true,
-    sync = true
+    sync = true,
+    toDelete?: boolean
   ) {
-    return RecordOps._export(this, force, oldGroupsOverride, saveExports, sync);
+    return RecordOps._export(
+      this,
+      force,
```
bleonard, line 192: @pedroslopez using force here, so might have a conflict with your work.

bleonard (approved):

core/src/models/GrouparooRecord.ts
pedroslopez: Should be ok, this is forcing the exports, not the group/model runs

core/src/models/GrouparooRecord.ts
evantahler: This gets called by a few places: on creating the record, after an import, when new properties are created, etc.
While it seems save in the test, I'll remove the throw here, it's fine to make null properties again if the profile is deleted

pedroslopez (approved):

PR #2445, opened 2021-10-21
Deleting a Destination will not remove Records previously exported

evantahler:
When a Destination is deleted, we no longer attempt to remove all the Records that were previously exported.  Now, we just delete the Destination.

### Development

### Impact

>

### Code review

### DIFF
--- a/core/src/tasks/destination/destroy.ts
+++ b/core/src/tasks/destination/destroy.ts
@@ -25,22 +25,11 @@ export class DestinationDestroy extends CLSTask {
     // the destination may have been force-deleted
     if (!destination) return;
 
-    let run: Run;
-    // untrack the group, if we are still tracking one
-    // this will trigger a run to export all group members one last time
-    if (
-      (destination.groupId && destination.collection === "group") ||
-      destination.collection === "model"
-    ) {
-      const unTrackResponse = await destination.updateTracking("none");
-      run = unTrackResponse.oldRun;
-    } else {
-      run = await Run.scope(null).findOne({
-        where: { destinationId: destination.id },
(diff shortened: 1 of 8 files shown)


evantahler:
Upgrade guide notes via https://github.com/grouparoo/www.grouparoo.com/pull/732

core/src/tasks/destination/destroy.ts
```diff
@@ -25,22 +25,11 @@ export class DestinationDestroy extends CLSTask {
...
-      const unTrackResponse = await destination.updateTracking("none");
-      run = unTrackResponse.oldRun;
-    } else {
-      run = await Run.scope(null).findOne({
-        where: { destinationId: destination.id },
-        order: [["updatedAt", "desc"]],
-        limit: 1,
-      });
-    }
+    const run = await Run.scope(null).findOne({
```
bleonard, line 28: it feels like this would be a little safer if you asked for the states you wanted below and then see if anything is returned if > 0.
As is, if there are timing issues, there might be a second run that is in those states and this one is canceled.

bleonard (approved):

core/src/tasks/destination/destroy.ts
evantahler: I don't think that's quite what's going on here - this query asks for the most recently updated run - regardless of state.  Then, we see what state that run is in.   We don't want to pre-filter based on a run in a particular state.

bleonard: My read is that the goal is to bail out (and not delete) if there are ongoing runs. Is that correct?

PR #2460, opened 2021-10-25
Also migrate `csv-remote-import` old plugin names

evantahler:
Update the first CSV name migration to include a change from https://github.com/grouparoo/grouparoo/pull/2293 that didn't have a migration.  

### Impact

### Code review

### DIFF
--- a/plugins/@grouparoo/csv/src/migrations/000001-plugin-csv-import-name.ts
+++ b/plugins/@grouparoo/csv/src/migrations/000001-plugin-csv-import-name.ts
@@ -8,6 +8,11 @@ export default {
       { type: "csv-import-table" },
       { type: "csv-import" }
     );
+    await queryInterface.bulkUpdate(
+      "sources",
+      { type: "csv-import-table" },
+      { type: "csv-remote-import" }
+    );
   },
 
   down: async (queryInterface: Sequelize.QueryInterface) => {


pedroslopez (approved):
Note: if someone had upgraded to v0.7.1 before this fix and are affected, they'll need to manually patch their csv sources or delete the `000001-plugin-csv-import-name` row from SequelizeMeta and restart.

evantahler:
> Note: if someone had upgraded to v0.7.1 before this fix and are affected, they'll need to manually patch their csv sources or delete the `000001-plugin-csv-import-name` row from SequelizeMeta and restart.

Yeah - that's why I want to get this out fast!

plugins/@grouparoo/csv/src/migrations/000001-plugin-csv-import-name.ts
```diff
@@ -8,6 +8,11 @@ export default {
...
       { type: "csv-import" }
     );
+    await queryInterface.bulkUpdate(
+      "sources",
+      { type: "csv-import-table" },
+      { type: "csv-remote-import" }
+    );
   },
 
   down: async (queryInterface: Sequelize.QueryInterface) => {
```
pedroslopez, line 18: oop, should we add a down migration?

evantahler: I don't think so... 
Ideally, this migration would be 0002 and there would have been a 0001 that change the name a first time, but that was so long ago, I'm worried what a "down" would do in that case.  

This "down" leaves you where you should be at the point this migration was introduced, which was "ahead" of the missing first migration.

PR #2463, opened 2021-10-25
Don't crash when viewing an Import that has not yet been associated to a Record

evantahler:
Fixes an API bug which was trying to load the Import's Record's modelId... but the Import might not yet have a Model early in it's lifecyle

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/import.ts
+++ b/core/__tests__/models/import.ts
@@ -34,7 +34,7 @@ describe("models/import", () => {
   });
 
   test("an error can be set", async () => {
-    const _import: Import = await helper.factories.import();
+    const _import = await helper.factories.import();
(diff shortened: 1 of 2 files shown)


teallarson (commented):
Awesome!  

Does the `$get` use the default scope?  Looking at sequelize docs it seems like it may.  If so, I would change the test description to `"apiData works without a ready Record`, or add an additional test for that behavior.

evantahler:
> Awesome!
> 
> Does the `$get` use the default scope? Looking at sequelize docs it seems like it may. If so, I would change the test description to `"apiData works without a ready Record`, or add an additional test for that behavior.

I don't believe we have a default scope on GrouparooRecord like we do on some of the other models

teallarson (approved):

PR #2465, opened 2021-10-26
Strings can be mapped to any Destination type

pedroslopez:
## Change description

Strings can now be mapped to: `url`, `email`, `boolean`, `date`, `float`, `integer`, `number` and `phoneNumber` destination types. Note that no conversion is being done for these strings - it is left up to the destination to interpret whether it's actually valid or not.

### Development

### Impact

### Code review


evantahler (approved):
👍 I'm glad it wasn't that deep of a change.

I wonder about not doing any conversion... we could `value.toString()`.  I don't really know if we should or not.  Perhaps some of the API clients might get mad if we don't provide strings?

PR #2476, opened 2021-10-30
appRefreshQuery Task and Ops update

teallarson:
## Change description

Currently, the `appQueryRefresh` is querying and updating its values as part of the `@AfterSave` hook.  While we do want this behavior during a run, we don't want it during the validate/apply stages.  

Now:

- @BeforeSave hook runs the query, but doesn't save or run anything else (will throw if plugin validation on the query fails)
- `AppRefreshQueries:check` enqueues an `AppRefreshQuery:check` task for all ready appRefreshQueries with ready apps every five minutes
- The `appRefreshQuery:check`:
  - Queries the app, updates `lastConfirmedAt`
  - If the value returned is different than the currently saved value:
    - Save new value, update `lastChangedAt`
    - run `AppRefreshQueryOps.triggerSchedules()` to trigger all relevant schedules.

### Development

### Impact

### Code review


tealjulia:
While this is a simple solution, what's happening may be more clear if I exposed the `validateQuery()` method from the plugin, added a separate `AppRefreshQueryOps` method for validation, and called that instead during validate/apply.  Open to going that route instead.

core/__tests__/tasks/appRefreshQuery/check.ts
```diff
@@ -76,19 +92,31 @@ describe("tasks/appRefreshQuery:check", () => {
...
+      await schedule.update({ refreshEnabled: false });
+
+      await task.enqueue("appRefreshQuery:check", { appRefreshQuery }); //does not throw
+
+      const enqueuedRuns = await Run.findAll({
+        where: { creatorId: schedule.id, state: "running" },
+      });
+
+      expect(enqueuedRuns.length).toBe(0);
+    });
```
evantahler: There isn't a test that actually enqueues any runs.  Please add a test like this one, but that starts with `await schedule.update({ refreshEnabled: true });` and see that there are enqueued runs

core/__tests__/tasks/appRefreshQuery/checkAll.ts
```diff
@@ -0,0 +1,92 @@
...
+      await appRefreshQuery.save();
+    });
+
+    test("can be enqueued and run", async () => {
+      await task.enqueue("appRefreshQueries:check", {}); //does not throw
+
+      await specHelper.runTask("appRefreshQueries:check", {});
+      const found = await specHelper.findEnqueuedTasks(
+        "appRefreshQueries:check"
+      );
```
evantahler: These tests may be testing the wrong task (`appRefreshQueries:checkAll`) .  These tests also probably don't care that any runs were enqueued, but that the next task, `appRefreshQueries:check` was.

evantahler: This might be a hint that the names of these tasks aren't quite descriptive enough... `:check` vs `:query` might help?

core/src/models/AppRefreshQuery.ts
```diff
@@ -103,19 +103,10 @@ export class AppRefreshQuery extends LoggedModel<AppRefreshQuery> {
...
-      const isUpdated = await AppRefreshQueryOps.checkRefreshQueryValue(
-        instance
-      );
-      if (isUpdated === true) {
-      }
-      await AppRefreshQueryOps.triggerSchedules(instance);
+  @BeforeSave
+  static async appQuery(instance: AppRefreshQuery) {
+    if (instance.state === "ready") {
+      const sampleValue = await AppRefreshQueryOps.runAppQuery(instance);
```
evantahler: The return value doesn't matter since it's not being used

core/src/tasks/appRefreshQuery/check.ts
```diff
@@ -7,28 +7,28 @@ export class AppRefreshQueryCheck extends CLSTask {
...
     this.name = "appRefreshQuery:check";
-    this.description = "check all appRefreshQueries and run them if it is time";
-    this.frequency =
-      process.env.GROUPAROO_RUN_MODE === "cli:run" ? 0 : 1000 * 60 * 5; // Run every 5 minutes
+    this.description =
+      "run a single appRefreshQuery and update values/trigger schedules if needed";
+    this.frequency = 0;
     this.queue = "apps";
-    this.inputs = {};
+    this.inputs = { appRefreshQuery: { required: true } };
```
evantahler:
```suggestion
this.inputs = { appRefreshQueryId: { required: true } };
```

I think this implies that the task expects an actual instance of `appRefreshQuery` to be passed as an Input to the task.  That doesn't really work - task inputs should only be IDs or strings (just like actions).  This /might/ work in test environments, but it won't work for real.  All that gets stored in redis for the task is JSON, so it won't be able to run any of the methods on it. 

The first thing a task should do is load up the appRefreshQuery from the provided appRefreshQueryId, and check if it exists and is ready.  Then, the logic below can continue.

core/src/tasks/appRefreshQuery/check.ts
```diff
@@ -7,28 +7,28 @@ export class AppRefreshQueryCheck extends CLSTask {
...
   }
 
-  async runWithinTransaction() {
-    const appRefreshQueries = await AppRefreshQuery.findAll();
+  async runWithinTransaction(params) {
```
evantahler: You can type/destructure params to make it clear that only IDs are getting passed in: `runWithinTransaction({appRefreshQueryId}: {appRefreshQueryId:string})`

evantahler (changes requested):

core/__tests__/tasks/appRefreshQuery/checkAll.ts
teallarson: `:query` or `:run`?  I've put `query` for now, but this task queries _and_ enqueues the schedule runs.

core/__tests__/tasks/appRefreshQuery/query.ts
```diff
@@ -0,0 +1,154 @@
...
+      await specHelper.runTask("appRefreshQuery:query", {
+        appRefreshQueryId: appRefreshQuery.id,
+      });
+
+      const enqueuedRuns = await Run.findAll({
+        where: { creatorId: schedule.id, state: "running" },
+      });
+      expect(enqueuedRuns.length).toBe(1);
+
+      Run.truncate();
```
evantahler: [suggestion] - you might want to move `Run.truncate()` into a beforeAll or afterAll to ensure the tests always start in a cleaned-up state

evantahler (approved):
Thanks for making those changes to make the task names more clear!

core/__tests__/tasks/appRefreshQuery/query.ts
teallarson: Do you mean a beforeEach since I'm calling it at the end of multiple tests?

evantahler: oops! yep - `beforeEach`

PR #2515, opened 2021-11-02
Synced Records per Month

evantahler:
To simplify Grouparoo Cloud's billing, we will be tracking the metric "Synced Records per Month".

1. Keep at least one Export per Destination + Record for 90 days
2. When a Record or Destination is deleted, we will still keep the most recent related Exports, to satisfy the above
3. Display Synced Exports for the last 30 days on the dashboard
4. Send Synced Exports for the last 30 days to Telemetry
 
Screen Shot 2021-11-02 at 12 40 39 PM

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/ops/setting.ts
+++ b/core/src/modules/ops/setting.ts
@@ -161,14 +161,6 @@ export namespace SettingOps {
       type: "number",
     },
     {
-      key: "sweeper-delete-old-exports-days",
-      title: "Sweeper: Delete Old Exports Days",
-      defaultValue: 31,
-      description:
-        "How many days should we keep Export records for on the server?  We will keep the most recent export for each Record & Destination.",
-      type: "number",
-    },
-    {
       key: "sweeper-delete-old-logs-days",
       title: "Sweeper: Delete Old Logs Days",
       defaultValue: 7,
--- a/core/src/modules/statusReporters.ts
+++ b/core/src/modules/statusReporters.ts
@@ -1,6 +1,7 @@
 import os from "os";
-import { api, config, task, env as nodeEnv } from "actionhero";
+import { api, config, task } from "actionhero";
 import Sequelize from "sequelize";
+import Moment from "moment";
 import { getCoreVersion, getPluginManifest } from "../modules/pluginDetails";
 import { Op } from "sequelize";
 import { App } from "../models/App";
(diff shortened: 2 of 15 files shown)


core/src/modules/ops/setting.ts
```diff
@@ -160,14 +160,6 @@ export namespace SettingOps {
...
       type: "number",
     },
-    {
-      key: "sweeper-delete-old-exports-days",
-      title: "Sweeper: Delete Old Exports Days",
-      defaultValue: 31,
-      description:
-        "How many days should we keep Export records for on the server?  We will keep the most recent export for each Record & Destination.",
-      type: "number",
-    },
```
evantahler, line 170: This setting is removed now - we keep the most recent record-destination export for 90 days.

core/src/modules/statusReporters.ts
```diff
@@ -239,6 +240,34 @@ export namespace StatusReporters {
...
+      const days = [1, 7, 30];
+
+      for (const dayCount of days) {
+        const count = await Export.count({
+          distinct: true,
+          col: "recordId",
+          where: {
+            state: "complete",
+            createdAt: {
+              [Op.gte]: Moment().subtract(dayCount, "days").toDate(),
```
krishnaglick, line 255: Moment <3

ui/ui-components/components/visualizations/homepageWidgets.tsx
```diff
@@ -51,7 +53,55 @@ export function BigTotalNumber({
           )}
         </Card.Title>
 
-        <span style={{ fontSize: 25 }}>{total >= 0 ? total : <Loading />}</span>
+        <h3 style={{ fontWeight: "normal" }}>
+          {total >= 0 ? total : <Loading />}
```
krishnaglick: If you set total to `null` or `undefined` on init this could be `{total ?? <Loading />}`

evantahler:
@krishnaglick thanks for the initialization suggestion!

krishnaglick (approved):

core/src/modules/ops/setting.ts
bleonard: Does it now clear out the other ones at some point?

bleonard (commented):
There's something really weird (in code config situation) if a destination id deleted and comes _back_ with the same `id` as one that was deleted. Then it would match the exports.
not sure how important that is, but maybe it should clear out the destinationid? hard to say.

krishnaglick:
> There's something really weird (in code config situation) if a destination id deleted and comes _back_ with the same `id` as one that was deleted. Then it would match the exports. not sure how important that is, but maybe it should clear out the destinationid? hard to say.

Is the `id` in this case a GUID? Or is it some identifier from the destination? I'm curious how likely this is to occur.

bleonard:
> Is the `id` in this case a GUID? Or is it some identifier from the destination? I'm curious how likely this is to occur.

In code-config, the Destination ID is set in the JSON, so it's possible they choose the same name as before (`mailchimp`)

krishnaglick:
> > Is the `id` in this case a GUID? Or is it some identifier from the destination? I'm curious how likely this is to occur.
> 
> In code-config, the Destination ID is set in the JSON, so it's possible they choose the same name as before (`mailchimp`)

I'd say we should clear out the destination id then. I think it's extremely possible for someone to reuse a name-like ID.

evantahler:
Oh yeah - this is a strange place to be... but it might be OK? 
* If the previous delete of the Destination was fully drained, and the sync mode was Sync, the last exports would have been `toDelete=true`, so I think any new exports would be considered a fresh sync.  If the destination was not fully drained, then we might have some problems :/
* If the previous delete of the destination was not syncMode Sync, then we might export the next sync because there are no changes from the previous export... but that's more or less right?  Maybe?

We could do a thing where if a Destination is deleted, we keep the exports but assign them a new `destinationId` like `${destinationId}-deleted` or something like that

Story for this - https://www.pivotaltracker.com/story/show/180200308

PR #2519, opened 2021-11-02
Snowflake not reading primary key when "always uppercase" setting is on.

bleonard:
## Change description

Use uppercase keys when querying snowflake because there is a setting that forces that. 
So let's just always do it.

### Development

### Impact

### Code review


plugins/@grouparoo/snowflake/src/lib/table-import/getChangedRows.ts
```diff
@@ -34,7 +35,9 @@ export const getChangedRows: GetChangedRowsMethod = async ({
   validateQuery(query);
 
   const rows = await connection.execute(query, params);
-  const results: DataResponseRow[] = rows.map((row) => castRow(row));
+  const results: DataResponseRow[] = rows.map((row) =>
+    castRowWithHighWaterMark(row, tempHighWaterMarkKey, highWaterMarkKey)
+  );
```
evantahler, line 40: I don't think I understand what's going on here - it looks like the results are cast to strings and then cast back again?

bleonard: The API from app-templates requires a field called `_hwm` to be returned, but Snowflake refuses to return anything in lowercase. So this is the same as the old method but takes the uppercase one and set it to what the caller expects. I'm not sure what you mean by "cast to strings"

evantahler (approved):

PR #2520, opened 2021-11-02
Hubspot Custom Objects Plugin Adjusts

pauloouriques:
- Dropdowns order;
- Include all writable (non readOnly) properties;
- Add name as important field;
- Handle 429 responses;
- Filter required fields from the known list.


plugins/@grouparoo/hubspot/src/lib/export-objects/destinationMappingOptions.ts
```diff
@@ -81,11 +82,15 @@ export const getRequiredFields = (
...
+  return requiredFieldsWithType.sort((a, b) => (a.key > b.key ? 1 : -1));
 };
 
-const isImportant = (key): Boolean => {
-  return importantFields.includes(key.toLowerCase());
+const isImportant = (customObject: CustomObjectHandler, key): Boolean => {
+  const searchableProperties = customObject.getSearchableProperties();
+  return (
+    importantFields.includes(key.toLowerCase()) ||
+    searchableProperties.includes(key.toLowerCase())
```
bleonard: probably shouldn't lower case this because it's from them

plugins/@grouparoo/hubspot/src/lib/export-objects/destinationMappingOptions.ts
```diff
@@ -97,23 +102,27 @@ export const getObjectFields = (
...
+  const requiredFields = customObject.getRequiredProperties();
   const out = [];
   for (const property of properties) {
     if (
       property["name"] !== primaryKey &&
-      !property["archived"] &&
-      !property["calculated"]
+      !property.archived &&
+      !property.calculated &&
+      !property.modificationMetadata.readOnlyValue === true &&
```
bleonard: this look weird to me.
does `!property.modificationMetadata.readOnlyValue` work?
same below with `!property.modificationMetadata.readOnlyOptions`

plugins/@grouparoo/hubspot/src/lib/export-objects/exportRecords.ts
```diff
@@ -221,10 +212,18 @@ export const exportRecords: ExportRecordsPluginMethod = async ({
...
-  });
+  try {
+    return await exportBatch({
+      appOptions,
+      destinationOptions,
+      syncOperations,
+      exports: batchExports,
+    });
+  } catch (error) {
+    if (error?.response?.status === 429) {
```
bleonard, line 223: Hmm, this might happen from any of the APIs including search, etc.
Maybe this should be in the client?
Ideally, there might also be headers to read to know when to retry.

plugins/@grouparoo/hubspot/src/lib/export/options.ts
```diff
@@ -45,9 +45,9 @@ class OptionsHandler {
     );
   }
 
-  private async getCustomObject(customObjectIdentifier: string) {
+  private async getSchema(customObjectIdentifier: string) {
     const { appId } = this.cacheData;
-    const cacheKey = "HubspotGetSchema";
+    const cacheKey = `hubspot_get_schema_${customObjectIdentifier.toLowerCase()}`;
```
bleonard, line 50: 👍

plugins/@grouparoo/hubspot/src/lib/export/options.ts
```diff
@@ -63,10 +63,20 @@ class OptionsHandler {
...
     objects.map((object) => {
-      out.schemaId.options.push(object.objectTypeId);
-      out.schemaId.descriptions.push(object.name);
+      objectsToSort.push({
+        option: object.objectTypeId,
+        description: object.name,
+      });
     });
+    objectsToSort
+      .sort((a, b) => (a.description > b.description ? 1 : -1))
```
bleonard: Here (and in the other one), it would be better to do 
`.sort((a, b) => a.description.localeCompare(b.description))`
This handles the `equal -> 0` case

plugins/@grouparoo/hubspot/src/lib/export/options.ts
```diff
@@ -94,18 +104,21 @@ class OptionsHandler {
     const names = [];
     if (customObject?.properties) {
       for (const object of customObject.properties) {
-        if (object.name.startsWith("hs_")) {
+        if (
+          object.modificationMetadata.readOnlyValue === true ||
```
bleonard: above you are also skipping archived and calculated. Should we do that here too?

plugins/@grouparoo/hubspot/src/lib/export-objects/exportRecords.ts
pauloouriques: I got the header.

plugins/@grouparoo/hubspot/src/lib/export-objects/destinationMappingOptions.ts
```diff
@@ -97,23 +101,27 @@ export const getObjectFields = (
   important?: boolean;
 }> => {
   const properties = customObject.getProperties();
+  const requiredFields = customObject.getRequiredProperties();
```
bleonard: in the future, it might be better to pass in `requiredFields` above.
```
    const required = getRequiredFields(customObject, primaryKey as string);
    const known = getObjectFields(customObject, required, primaryKey as string);
```
Then do `!requiredFields.includes(property["name"])`. That allows us to remove the knowledge of what we chose as `required` into just the `getRequiredFields` function and makes it so getObjectFields will never duplicate. For example, then you'd remove `property["name"] !== primaryKey`

plugins/@grouparoo/hubspot/src/lib/export/options.ts
```diff
@@ -94,18 +104,23 @@ class OptionsHandler {
     const names = [];
     if (customObject?.properties) {
       for (const object of customObject.properties) {
-        if (object.name.startsWith("hs_")) {
+        if (
+          object.archived ||
```
bleonard: In the future, maybe a standardized `shouldShowProperty` that we us in both places.

plugins/@grouparoo/hubspot/src/lib/export/customObject.ts
```diff
@@ -25,15 +25,24 @@ class CustomObjectHandler {
     return null;
   }
 
-  public getProperties() {
+  public shouldShowProperty(property: any): boolean {
```
bleonard, line 28: 👍

bleonard (approved):

PR #2521, opened 2021-11-02
Postgres gets less group and more distinct SQL

bleonard:
## Change description

Better SQL for performance.

### Development

### Impact

### Code review

### DIFF
--- a/plugins/@grouparoo/postgres/src/lib/table-import/getPropertyValues.ts
+++ b/plugins/@grouparoo/postgres/src/lib/table-import/getPropertyValues.ts
@@ -21,55 +21,68 @@ export const getPropertyValues: GetPropertyValuesMethod = async ({
   primaryKeys,
 }) => {
   let responses: { [key: string]: { [column: string]: DataResponse[] } } = {};
-  let aggSelect = columnNames.map((col) => `"${col}"`).join(", ");
-  let orderBy = "";
-  let groupByColumns = [tablePrimaryKeyCol];
+  const columnList = columnNames.map((col) => `"${col}"`).join(", ");
+  const orderBys = [];
+  const groupByColumns = [];
+  let aggFunc = null;
 
   if (primaryKeys.length === 0) return responses;
 
   switch (aggregationMethod) {
     case AggregationMethod.Exact:
-      groupByColumns.push(...columnNames);
       if (sortColumn) {
-        orderBy = `"${sortColumn}" ASC`;
-        groupByColumns.push(sortColumn);
+        orderBys.push(`"${sortColumn}" ASC`);
       }
       break;
     case AggregationMethod.Average:
-      aggSelect = `COALESCE(AVG(${aggSelect}), 0) as ${aggSelect}`;
+      aggFunc = `COALESCE(AVG(${columnList}), 0) as ${columnList}`;
       break;
     case AggregationMethod.Count:
-      aggSelect = `COUNT(${aggSelect})::integer as ${aggSelect}`;
+      aggFunc = `COUNT(${columnList})::integer as ${columnList}`;
       break;
     case AggregationMethod.Sum:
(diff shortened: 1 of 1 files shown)


plugins/@grouparoo/postgres/src/lib/table-import/getPropertyValues.ts
```diff
@@ -21,55 +21,68 @@ export const getPropertyValues: GetPropertyValuesMethod = async ({
...
-  let query = `SELECT ${aggSelect}, "${tablePrimaryKeyCol}" as __pk FROM %I WHERE`;
-  let addAnd = false;
+  const params: Array<any> = [];
+
+  let query = `SELECT`;
+
+  if (aggFunc || isArray) {
+    query += ` "${tablePrimaryKeyCol}" as __pk`;
+  } else {
+    query += ` DISTINCT ON (__pk) "${tablePrimaryKeyCol}" as __pk`;
```
evantahler, line 71: Postgres magic - https://www.geekytidbits.com/postgres-distinct-on/

I don't personally have any knowledge of if `DISTINCT ON` is faster or slower than GROUP BY.

evantahler (commented):
Assuming this is part of https://www.pivotaltracker.com/story/show/180125788, I'm guessing this PR will gain peers with similar updates for all the table plugins?

I feel like there should be some test here to make sure that the eventual queries produced look like we expect. It's good all the data loading tests we already have pass... but there's got to be some way to prevent a regression in this.  Perhaps just testing that there's no `group by` when selecting the happy-path exact properties from a table would be enough?

evantahler:
I don’t believe Redshift supports `distinct on` queries. This opens up the can of worms avoided until now about potentially replacing/updating some methods for the redshift plugin that may be different than Postgres.  This would be a similar pattern to what @edmundito is doing with MySQL -> ClickHouse.  This would also mean we need to add tests for Redshift for whatever ends up being different. 


* https://stackoverflow.com/questions/35511803/redshift-postgresql-distinct-on-operator 
* https://dba.stackexchange.com/questions/270155/substitute-distinct-on-for-query-transformation-postgresql-to-redshift

plugins/@grouparoo/postgres/src/lib/table-import/getPropertyValues.ts
bleonard: I'm fairly sure you can't just GROUP BY `user_id` 

For example, this query (most recent purchase dates) is invalid.

> SELECT "user_id" as __pk, "created_at" FROM demo.purchases WHERE  "state" = 'successful' AND  "user_id" IN ('1','2','3','4') GROUP BY "user_id" ORDER BY "created_at" DESC 

> ERROR:  column "purchases.created_at" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: SELECT "user_id" as __pk, "created_at" FROM demo.purchases W...

So you have to add `created_at` to the `GROUP BY`. This defeats the purpose as they end up not being grouped.
I believe this works in MySQL and, even better, makes the one that's selected the first in the list.

bleonard: It seems something like this is the way to do the above query without Postgres magic
> SELECT * FROM
>   (SELECT  "user_id" as __pk, "created_at", 
>               ROW_NUMBER() OVER (PARTITION BY "user_id" ORDER BY "created_at" DESC) AS __rownum
>        FROM demo.purchases
>        WHERE  "state" = 'successful' AND  "user_id" IN ('1','2','3','4')
>   ) AS __ranked
> WHERE __ranked.__rownum = 1;

bleonard: Ok, doing that in https://github.com/grouparoo/grouparoo/pull/2522

bleonard:
Replaced by https://github.com/grouparoo/grouparoo/pull/2522

PR #2522, opened 2021-11-03
Sql updates for performance

bleonard:
## Change description

* Less `GROUP BY`
* More efficient by just returning one row per result in "has many" queries

Updates all the SQL-based plugins
* postgres
* sqlite
* mysql
* big query
* snowflake

### Development

### Impact

### Code review


plugins/@grouparoo/mysql/src/lib/table-import/getPropertyValues.ts
```diff
@@ -20,53 +20,65 @@ export const getPropertyValues: GetPropertyValuesMethod = async ({
...
+  let query = `SELECT \`${tablePrimaryKeyCol}\` as __pk`;
+
+  if (aggFunc) {
+    query += `, ${aggFunc}`;
+    groupByColumns.push(tablePrimaryKeyCol);
+  } else {
+    query += `, ${columnList}`;
+    if (!isArray && orderBys.length > 0) {
+      const order = `ORDER BY ${orderBys.join(", ")}`;
+      query += `, ROW_NUMBER() OVER (PARTITION BY \`${tablePrimaryKeyCol}\` ${order}) AS __rownum`;
```
bleonard, line 79: Tests were failing. it seems this Windowing stuff (OVER) is only on MySQL 8+.

plugins/@grouparoo/postgres/src/lib/table-import/getPropertyValues.ts
```diff
@@ -21,53 +21,65 @@ export const getPropertyValues: GetPropertyValuesMethod = async ({
...
+  const params: Array<any> = [];
+  let ranked = false;
+  let query = `SELECT "${tablePrimaryKeyCol}" as __pk`;
+
+  if (aggFunc) {
+    query += `, ${aggFunc}`;
+    groupByColumns.push(tablePrimaryKeyCol);
+  } else {
+    query += `, ${columnList}`;
+    if (!isArray && orderBys.length > 0) {
```
bleonard, line 73: to keep the most common case (exact queries where there is only one thing with the primary key) very simple I decided to only for this in the "least/most recent" cases.

plugins/@grouparoo/mysql/src/lib/table-import/getPropertyValues.ts
bleonard: Now limited to 8+

plugins/@grouparoo/mysql/__tests__/table-import/import-properties.ts
```diff
@@ -361,20 +361,26 @@ describe("mysql/table/recordProperties", () => {
           aggregationMethod: AggregationMethod.Exact,
         });
         expect(values[record.id][properties[0].id]).toEqual([
+          "Apple",
+          "Orange",
+          "Blueberry",
```
evantahler, line 366: 👍 - Please update the postgres tests as well

plugins/@grouparoo/postgres/src/lib/table-import/getPropertyValues.ts
```diff
@@ -93,8 +105,21 @@ export const getPropertyValues: GetPropertyValuesMethod = async ({
...
+
+  if (ranked) {
+    // -- Ranked example
+    // SELECT * FROM
+    //  (SELECT  "user_id" as __pk, "created_at",
+    //              ROW_NUMBER() OVER (PARTITION BY "user_id" ORDER BY "created_at" DESC) AS __rownum
+    //        FROM demo.purchases
+    //        WHERE  "state" = 'successful' AND "user_id" IN ('1','2','3','4')
+    //   ) AS __ranked
+    // WHERE __ranked.__rownum = 1;
```
evantahler, line 120: Looks like this should work in Redshift too! https://docs.aws.amazon.com/redshift/latest/dg/r_WF_ROW_NUMBER.html

plugins/@grouparoo/mysql/__tests__/table-import/import-properties.ts
evantahler: I wonder if it is time for `app-templates` to build this test suite for us automatically for all the SQL plugins

plugins/@grouparoo/postgres/src/lib/table-import/getPropertyValues.ts
evantahler: I really want to test against redshift, but there are no mocking libraries I can find, nor will it work for Nock.  It might be time to leave a cluster up and running for test purposes.

evantahler (commented):
Nice work!  It looks like the ROW_NUMBER approach will work for all the sources we have ([snowflake](https://docs.snowflake.com/en/sql-reference/functions/row_number.html), [big query](https://cloud.google.com/bigquery/docs/reference/standard-sql/numbering_functions), [redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_WF_ROW_NUMBER.html), [sqlite](https://www.sqlitetutorial.net/sqlite-window-functions/sqlite-row_number/)).  I'm curious what's happening with Mongo... I don't know enough about it yet to know if we have the same problem or not.

In the PR, for posterity, please include an old problematic query that this replaces

bleonard:
>  I'm curious what's happening with Mongo... I don't know enough about it yet to know if we have the same problem or not.

The mongo tests are actually correct now!

bleonard:
On an most recent:

Before:
> SELECT "category", "user_id" as __pk FROM purchases WHERE "user_id" IN ('1','2','3','4') ORDER BY "created_at" DESC GROUP BY user_id,category

(returns many rows for each user id) 

Now: 
> SELECT * FROM
>  (SELECT  "user_id" as __pk, "category",
>              ROW_NUMBER() OVER (PARTITION BY "user_id" ORDER BY "created_at" DESC) AS __rownum
>        FROM .purchases
>        WHERE  "user_id" IN ('1','2','3','4')
>   ) AS __ranked
> WHERE __ranked.__rownum = 1;

(returns 1 row for each user_id)

bleonard:
on an array, exact

Before:
> SELECT "category", "user_id" as __pk FROM purchases WHERE "user_id" IN ('1','2','3','4') GROUP BY user_id,category

(did not have duplicates in array)

Now:
> SELECT "category", "user_id" as __pk FROM purchases WHERE "user_id" IN ('1','2','3','4')

(correctly has duplicates)

plugins/@grouparoo/snowflake/__tests__/table-import/import-properties.ts
```diff
@@ -366,27 +366,35 @@ describe("snowflake/table/recordProperties", () => {
         );
       });
 
-      test("Exact + Array works as intended", async () => {
+      test.only("Exact + Array works as intended", async () => {
```
evantahler: Errant `.only`

evantahler (changes requested):

evantahler:
Other than that `.only`, this is great!

evantahler (approved):
🥳

PR #2523, opened 2021-11-03
Hubspot Companies export tests

pauloouriques:
Tests for the companies export using the Custom Objects destination plugin.


bleonard (approved):
Thanks!

PR #2527, opened 2021-11-03
UI for AppRefreshQuery

teallarson:
## Change description
- Actions/API routes/UI for App Refresh Query  -- create/edit/test/run/delete an app refresh query
- Model change: ensure only one App Refresh Query per app (handful of related testing updates)

Empty state in an app with Refresh Query enabled.
Screen Shot 2021-11-11 at 5 25 42 PM

Default: View state
Screen Shot 2021-11-11 at 5 26 16 PM

Edit state
Screen Shot 2021-11-11 at 5 26 45 PM

Apps without this feature have no refresh tab
Screen Shot 2021-11-11 at 5 27 14 PM

### Development

### Impact

### Code review


core/__tests__/actions/appRefreshQuery.ts
```diff
@@ -0,0 +1,195 @@
...
+            "appRefreshQuery:query",
+            connection
+          );
+        expect(error).toBeUndefined();
+        expect(appRefreshQuery.id).toBeTruthy();
+        expect(appRefreshQuery.refreshQuery).toBe("SELECT 'hello' AS name");
+        expect(appRefreshQuery.value.length).toBe(13);
+        expect(valueUpdated).toBeTruthy();
+      });
+      test("querying with an appRefreshQuery triggers associated schedule runs", async () => {
```
teallarson: If it seems valuable to have a reverse test here (that schedules are not run if the value is the same as before), I'll need to update what the test-plugin does for its `pluginMethods.query()`... it currently sends back a NOW() timestamp, so it is always different.

evantahler:
In the future, since this is a UI story - please add some screenshots to the PR!

core/src/models/App.ts
```diff
@@ -210,6 +215,7 @@ export class App extends LoggedModel<App> {
       pluginApp,
       createdAt: APIData.formatDate(this.createdAt),
       updatedAt: APIData.formatDate(this.updatedAt),
+      appRefreshQuery,
```
evantahler: This should be serialized:

```suggestion
      appRefreshQuery: await appRefreshQuery.apiData(),
```

core/src/modules/ops/appRefreshQuery.ts
```diff
@@ -61,4 +61,29 @@ export namespace AppRefreshQueryOps {
...
+
+  export async function test(
+    appRefreshQuery: AppRefreshQuery,
+    refreshQuery?: string
+  ) {
+    let success = false;
+    let message: string;
+    let error;
+
+    if (refreshQuery) appRefreshQuery.refreshQuery = refreshQuery;
```
evantahler: This pattern is a bit dangerous.  The database object is being modified with a temporary value provided by ab optional `refreshQuery`.  This implies that the `AppRefreshQueryOps.runAppQuery()` method should be changed to also allow for an optional refreshQuery to the object doesn't need to be modified.

evantahler: If you then go on to save or update the database object after this method is used, the wrong value for refreshQuery could be saved

core/src/modules/ops/appRefreshQuery.ts
```diff
@@ -61,4 +61,29 @@ export namespace AppRefreshQueryOps {
...
+    //does this need to be a try/catch to catch the error?
```
evantahler: I think this is a good pattern, sure!

core/src/tasks/appRefreshQuery/query.ts
```diff
@@ -18,7 +18,7 @@ export class AppRefreshQueryQuery extends CLSTask {
   }: {
     appRefreshQueryId: string;
   }) {
-    const appRefreshQuery = await AppRefreshQuery.findOne({
+    const appRefreshQuery = await AppRefreshQuery.scope(null).findOne({
```
evantahler: Why does this task want to run non-ready AppRefreshQueries?

plugins/@grouparoo/bigquery/__tests__/integration/app-refresh-query.ts
```diff
@@ -25,6 +25,10 @@ describe("integration/runs/bigquery/appRefreshQuery", () => {
     await app.update({ state: "ready" });
   });
 
+  afterEach(async () => {
+    await AppRefreshQuery.truncate();
+  });
```
evantahler, line 30: Not a problem, but the fixtures /probably/ didn't need to be re-recorded since all that changed in these tests was this (local only) cleanup method.

core/src/models/AppRefreshQuery.ts
```diff
@@ -73,21 +73,38 @@ export class AppRefreshQuery extends LoggedModel<AppRefreshQuery> {
...
+    const existingCount = await AppRefreshQuery.scope(null).count({
+      where: {
+        appId: instance.appId,
+      },
+    });
+
+    if (existingCount > 0) {
+      throw new Error(`app ${instance.appId} already has an app refresh query`);
+    }
+  }
```
evantahler, line 107: Because you added a unique index in `core/src/migrations/000090-appRefreshQueryUniqueAppId.ts`, you don't need this check.  The database will enforce that there's only one AppRefreshQuery per AppId.  This can be removed.  The specific error message will be different, but the logic is the same.

ui/ui-components/components/tabs/app.tsx
```diff
@@ -3,5 +3,12 @@ import { Models } from "../../utils/apiData";
...
 export default function AppTabs({ app }: { app: Models.AppType }) {
   const tabs = ["edit"];
-  return <Tabs name={app.name} draftType={app.type} tabs={tabs} />;
+  if (app.appRefreshQuery !== null) {
+    tabs.push("refresh");
+  }
+  return (
+    <div>
+      <Tabs name={app.name} draftType={app.type} tabs={tabs} />
+    </div>
```
evantahler: I don't think this should be wrapped in a `<div>`

ui/ui-components/utils/apiData.ts
```diff
@@ -269,6 +282,7 @@ import {
 } from "@grouparoo/core/src/actions/teams";
 import { TotalsAction } from "@grouparoo/core/src/actions/totals";
 import { ObjectFind } from "@grouparoo/core/src/actions/object";
+import { version } from "prettier";
```
evantahler: Probably not needed

ui/ui-components/utils/apiData.ts
```diff
@@ -107,6 +111,15 @@ import {
...
 } from "@grouparoo/core/src/actions/apps";
+import {
+  AppRefreshQueriesList,
+  AppRefreshQueryCreate,
+  AppRefreshQueryDestroy,
+  AppRefreshQueryEdit,
+  AppRefreshQueryQuery,
+  AppRefreshQueryTest,
+  AppRefreshQueryView,
+} from "@grouparoo/core/src/actions/appRefreshQuery";
```
evantahler: Thanks for bringing the types over to the UI!

core/src/actions/appRefreshQuery.ts
```diff
@@ -0,0 +1,205 @@
...
+import { AppRefreshQuery } from "../models/AppRefreshQuery";
+import { ConfigWriter } from "../modules/configWriter";
+import { APIData } from "../modules/apiData";
+import { AppRefreshQueryOps } from "../modules/ops/appRefreshQuery";
+import { CLS } from "../modules/cls";
+
+export class AppRefreshQueriesList extends AuthenticatedAction {
+  constructor() {
+    super();
+    this.name = "appRefreshQueries:list";
```
evantahler: I don't think there is a list action in the UI in use here?  The AppRefreshQueries are linked to from the app, and there can be only one

ui/ui-components/utils/appRefreshQueryHandler.ts
```diff
@@ -0,0 +1,4 @@
+import { Models } from "./apiData";
+import { EventDispatcher } from "./eventDispatcher";
+
+export class AppRefreshQueryHandler extends EventDispatcher<Models.AppType> {}
```
evantahler: I don't think this is used?  These `*Handlers` are for when more than one component on a page needs to /subscribe/ to changes in the UI.  The typical example is how a change on the `/account` page to the user's name should also update the `hello, {name}` in the navigation.  I think there's only one page/component at a time that would be displaying an AppDataRefresh

ui/ui-enterprise/pages/app/[id]/refresh.tsx
```diff
@@ -0,0 +1 @@
+export { default } from "@grouparoo/ui-components/pages/app/[id]/refresh";
```
Comment (evantahler) [inline ui/ui-enterprise/pages/app/[id]/refresh.tsx:1]: I think this same page should exist in `ui-config` too?

evantahler (changes requested):
There are some page-specific comments below.

Overall, I'm not really sure how I might add a new AppRefreshQuery.
* If I don't have a refresh query for the app yet, the tab doesn't appear, so I can't go to the page
* If I cheat and make a refresh query in the DB, I can't press any of the buttons (they are all "loading") (see screenshot).

I think the UI pattern we have elsewhere would be to:
* If the app supports it, always show the `refresh` query tab, and let me make a new one if I don't have one.  This is sort of how we treat schedules - the tab's visibility is based on the capability of the app, not if you've made a schedule yet or not.
* To determine if the app supports having a refresh query, you could inspect the plugin that created the app, and add `app.refreshQueryAvailalble` to `App.apiData()`.  We do something very similar with Sources and Schedules ([here](https://github.com/grouparoo/grouparoo/blob/main/core/src/models/Source.ts#L220))

Screen Shot 2021-11-11 at 11 50 32 AM

core/src/tasks/appRefreshQuery/query.ts
teallarson: it should not!

cli/templates/.env
```diff
@@ -0,0 +1,48 @@
+# Learn more about the Grouparoo environment variables @ https://www.grouparoo.com/docs/support/environment
```
evantahler: Why was this needed?  That's surprising

core/__tests__/actions/appRefreshQuery.ts
evantahler: I think that logic is already tested in the model tests

core/src/actions/appRefreshQuery.ts
```diff
@@ -0,0 +1,156 @@
...
+      id: { required: true },
+      refreshQuery: { required: false },
+    };
+  }
+
+  async runWithinTransaction({ params }) {
+    const appRefreshQuery = await AppRefreshQuery.findById(params.id);
+    if (!appRefreshQuery) {
+      throw new Error(`cannot find an appRefreshQuery ${params.id}`);
+    }
```
evantahler: `.findById()` already throws if the object can't be found

ui/ui-community/pages/logs/list.tsx
```diff
@@ -1 +0,0 @@
-export { default } from "@grouparoo/ui-components/pages/logs/list";
```
evantahler: These files are disappearing again...

ui/ui-community/pages/logs/stream.tsx
```diff
@@ -1 +0,0 @@
-export { default } from "@grouparoo/ui-components/pages/logs/stream";
```
evantahler: These files are disappearing again...

ui/ui-components/components/tabs/App.tsx
```diff
@@ -3,5 +3,14 @@ import { Models } from "../../utils/apiData";
 
 export default function AppTabs({ app }: { app: Models.AppType }) {
   const tabs = ["edit"];
-  return <Tabs name={app.name} draftType={app.type} tabs={tabs} />;
+
+  if (app.refreshQueryAvailable == true) {
```
evantahler: don't use `==` in javascript.  Either `===` or just `if (app.refreshQueryAvailable) {` because you know it's already a boolean

ui/ui-components/pages/logs/list.tsx
```diff
@@ -1,20 +0,0 @@
-import Head from "next/head";
```
evantahler: Missing page?

ui/ui-components/pages/logs/stream.tsx
```diff
@@ -1,20 +0,0 @@
-import Head from "next/head";
```
evantahler: Missing page?

ui/ui-enterprise/pages/logs/list.tsx
```diff
@@ -1 +0,0 @@
-export { default } from "@grouparoo/ui-components/pages/logs/list";
```
evantahler: Missing page?

ui/ui-enterprise/pages/logs/stream.tsx
```diff
@@ -1 +0,0 @@
-export { default } from "@grouparoo/ui-components/pages/logs/stream";
```
evantahler: Missing page?

evantahler (changes requested):
Alright!  Now this is getting close.  I can use the UI and it's very nice!  You designed a very functional page.  Some fixes needed below, but nothing major!

Screen Shot 2021-11-11 at 2 57 19 PM

cli/templates/.env
teallarson: That stack overflow solution for case sensitive commits was the culprit for files going missing on this PR and the last!  Definitely going to watch the commit changes more closely if I use it again.

core/src/actions/schedules.ts
```diff
@@ -201,6 +201,7 @@ export class ScheduleEdit extends AuthenticatedAction {
       name: { required: false },
       sourceId: { required: false },
       recurring: { required: false, formatter: APIData.ensureBoolean },
+      refreshEnabled: { required: false, formatter: APIData.ensureBoolean },
```
evantahler, line 207: This probably should be added to `schedule:create` too

ui/ui-components/pages/app/[id]/refresh.tsx
```diff
@@ -0,0 +1,400 @@
...
+      `/appRefreshQuery/`,
+      appRefreshQuery
+    );
+
+    if (response?.appRefreshQuery) {
+      setAppRefreshQuery(response.appRefreshQuery);
+
+      router.push(
+        "/app/[id]/refresh",
+        `/app/${response.appRefreshQuery.appId}/refresh`
```
Comment (evantahler) [inline ui/ui-components/pages/app/[id]/refresh.tsx]: FYI you don't need both now, just `/app/${response.appRefreshQuery.appId}/refresh` works

ui/ui-components/pages/app/[id]/refresh.tsx
```diff
@@ -0,0 +1,400 @@
...
+      );
+    }
+    setLoading(false);
+  }
+
+  async function update(event) {
+    const _appRefreshQuery = Object.assign({}, appRefreshQuery);
+    _appRefreshQuery[event.target.id] = event.target.value;
+    setAppRefreshQuery(_appRefreshQuery);
+  }
```
Comment (evantahler) [inline ui/ui-components/pages/app/[id]/refresh.tsx]: FYI @krishnaglick taught me a much tighter way to do these kinds of partial-object state updates:

```ts
setAppRefreshQuery({...appRefreshQuery, [event.target.id]: event.target.value});
```

ui/ui-components/pages/app/[id]/refresh.tsx
```diff
@@ -0,0 +1,400 @@
...
+          icon={app.icon}
+          title={app.name}
+          badges={[
+            <SourceBadge object={app} />,
+            <LockedBadge object={app} />,
+            <StateBadge state={app.state} />,
+          ]}
+        />
+        <Container>
+          <p>{app.name} has no App Refresh Query.</p>
```
Comment (evantahler) [inline ui/ui-components/pages/app/[id]/refresh.tsx]: You may want to put this inside of a `<Row><Col>` to help with the spacing

Comment (evantahler) [reply ui/ui-components/pages/app/[id]/refresh.tsx]: and the `<container>` here is probably not needed (you don't use it below)

ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
```diff
@@ -58,6 +58,11 @@ export default function Page(props) {
       _schedule.recurringFrequency = recurringFrequencyMinutes * (60 * 1000);
     }
 
+    let thing = Object.assign({}, _schedule, {
+      filters: localFilters,
+      state: "ready",
+    });
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx]: what's *the thing*?!

evantahler (commented):
Just a few more things to cleanup and this is ready to go!

ui/ui-components/pages/app/[id]/refresh.tsx
Comment (teallarson) [reply ui/ui-components/pages/app/[id]/refresh.tsx]: Cleaner!

ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
Comment (teallarson) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx]: You know... The _thing_.

(A line I used for debugging that I forgot to delete)

core/src/actions/schedules.ts
```diff
@@ -157,6 +157,7 @@ export class ScheduleCreate extends AuthenticatedAction {
         formatter: APIData.ensureBoolean,
       },
       state: { required: false },
+      refreshEnabled: { required: false, formatter: APIData.ensureBoolean },
```
evantahler, line 160: The option is added but it's not used with `Schedule.create` below

evantahler (changes requested):

evantahler (approved):
💥

PR #2530, opened 2021-11-05
Snowflake: schema name to introspect columns

bleonard:
## Change description

If there is the same table name in two schemas, then the column metadata could be wrong!

### Development


bleonard:
Thanks for the help tracking this down @nyccto

PR #2531, opened 2021-11-08
Test demo command via creating and deleting a config directory

evantahler:
This PR adds a new integration test which tests the `grouparoo demo` command that runs within `apps/staging-enterprise`.  

* Then we run `grouparoo demo -cl` to generate a default project + a Logger destination.  
* Next we run `grouparoo run` to completion, ensuring there are no errors
* Then we delete the entire config directory
* And finally, we run `grouparoo run` again to ensure that all models previously added by config are removed and the `grouparoo run` completes

This PR:
* Adds `logger` as a demo option (`grouparoo demo -c --logger`)
* Fixes a bug in which a deleting a Record that had previously been exported to a Destination would crash if that destination was deleted in the middle of that final export process

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/ops/destination.ts
+++ b/core/src/modules/ops/destination.ts
@@ -24,12 +24,10 @@ import {
 import { destinationTypeConversions } from "../destinationTypeConversions";
 import { GroupMember } from "../../models/GroupMember";
 import { ExportProcessor } from "../../models/ExportProcessor";
-import { GrouparooModel } from "../../models/GrouparooModel";
 import { Run } from "../../models/Run";
 import { MappingHelper } from "../mappingHelper";
 import { RecordPropertyOps } from "./recordProperty";
 import { Option } from "../../models/Option";
-import { RunOps } from "./runs";
 
 function deepStrictEqualBoolean(a: any, b: any): boolean {
   try {
@@ -103,30 +101,14 @@ export namespace DestinationOps {
       }
     }
 
-    oldRun = await runDestinationCollection(destination); // old collection
+    oldRun = await exportMembers(destination); // old collection
     await destination.update({
       collection,
       groupId: collection !== "group" ? null : collectionId,
     });
-    newRun = await runDestinationCollection(destination); // new collection
+    newRun = await exportMembers(destination); // new collection
 
     return { oldRun, newRun };
-  }
-
-  async function runDestinationCollection(destination: Destination) {
-    if (destination.collection === "none") {
-      // nothing to do
-    } else if (destination.collection === "group" && destination.groupId) {
-      const group = await Group.findById(destination.groupId);
-      if (group) return RunOps.run(group, destination.id);
-    } else if (destination.collection === "model") {
-      const model = await GrouparooModel.findById(destination.modelId);
-      if (model) return RunOps.run(model, destination.id);
-    } else {
-      throw new Error(
-        `unknown destination collection ${destination.collection}`
-      );
-    }
   }
 
   /**
--- a/core/src/modules/ops/record.ts
+++ b/core/src/modules/ops/record.ts
@@ -643,8 +643,11 @@ export namespace RecordOps {
 
     for (const _export of existingExportNotDeleted) {
       if (!destinations.map((d) => d.id).includes(_export.destinationId)) {
-        const destination = await Destination.findById(_export.destinationId);
-        if (destination.state === "ready") destinations.push(destination);
+        const destination = await Destination.findOne({
+          where: { id: _export.destinationId, state: "ready" },
+        });
+        // the destination may have been deleted or become not-ready
+        if (destination) destinations.push(destination);
       }
     }
 
--- a/plugins/@grouparoo/demo/__tests__/integration/megaDelete.ts
+++ b/plugins/@grouparoo/demo/__tests__/integration/megaDelete.ts
@@ -144,7 +144,7 @@ describe("create and delete", () => {
     await spawnPromise(bin, ["demo", "-c", "--scale", "0.1"]);
   });
 
-  it(
+  it.only(
     "can run 'grouparoo run' to completion with config",
     async () => {
       const { stdout } = await spawnPromise(bin, ["run"]);
@@ -158,7 +158,7 @@ describe("create and delete", () => {
     await spawnPromise("rm", ["-rf", "config"]);
   });
 
-  it.only(
+  it(
     "can run 'grouparoo run' to completion without config",
     async () => {
       const { stdout } = await spawnPromise(bin, ["run"]);
@@ -181,20 +181,26 @@ export async function spawnPromise(
 
     if (typeof args === "string") args = [args];
 
-    // console.log(`--> `, `${command} ${args.join(" ")}`);
+    console.log(`--> `, `${command} ${args.join(" ")}`);
     const spawnProcess = spawn(command, args, {
       cwd,
-      env: { ...process.env, NODE_ENV: undefined },
+      env: {
+        ...process.env,
+        NODE_ENV: undefined,
+        PWD: undefined,
+        ACTIONHERO_CONFIG: undefined,
+        GROUPAROO_SPEC_HELPER: undefined,
+      },
     });
 
     spawnProcess.stdout.on("data", (data) => {
       stdout += String(data);
-      // console.log(data.toString());
+      console.log(data.toString());
     });
 
     spawnProcess.stderr.on("data", (data) => {
       stderr += String(data);
-      // console.error(data.toString());
+      console.error(data.toString());
     });
 
     spawnProcess.on("close", (code) => {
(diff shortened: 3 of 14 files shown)


plugins/@grouparoo/demo/src/util/shared.ts
```diff
@@ -9,15 +11,17 @@ export async function execSync(command) {
...
+  const stats = fs.statSync(fileOrDirPath);
+  if (stats.isDirectory) {
+    const files = glob.sync(`${fileOrDirPath}/**/*.{json,js}`);
+    for (const file of files) await prettier(file);
+  } else {
+    const text = fs.readFileSync(fileOrDirPath).toString();
+    const pOptions = await Prettier.resolveConfig(process.cwd());
+    const formatted = Prettier.format(text, pOptions);
+    fs.writeFileSync(fileOrDirPath, formatted);
+  }
```
evantahler, line 24: Rather than using the prettier binary, we can use it pragmatically.  This allows us to not worry about the install path - loading prettier as a dependency just works!

plugins/@grouparoo/demo/src/util/shared.ts
```diff
@@ -9,15 +11,17 @@ export async function execSync(command) {
...
-  );
-  const pConfig = path.resolve(
-    path.join(__dirname, "..", "..", "..", "..", "..", ".prettierrc")
-  );
-  await execSync(`'${pCmd}' --config '${pConfig}' --write '${fileOrDirPath}'`);
+export async function prettier(fileOrDirPath: string) {
+  const stats = fs.statSync(fileOrDirPath);
+  if (stats.isDirectory) {
+    const files = glob.sync(`${fileOrDirPath}/**/*.{json,js}`);
+    for (const file of files) await prettier(file);
```
evantahler, line 18: If we are prettier-ing a directory, we recurse.

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -10,6 +10,7 @@ const {
 const corePath: string = getCoreRootPath();
 process.chdir(corePath);
 process.env.ACTIONHERO_CONFIG = `${corePath}/${loadPath}/config`;
+process.env.GROUPAROO_SPEC_HELPER = "true";
```
evantahler, line 13: While not functional, this will serve as a reminder that the specHelper has modified the ENV and CWD

plugins/@grouparoo/demo/__tests__/integration/megaDelete.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    console.log(`--> `, `${command} ${args.join(" ")}`);
+    const spawnProcess = spawn(command, args, {
+      cwd,
+      env: {
+        ...process.env,
+        NODE_ENV: undefined,
+        PWD: undefined,
+        ACTIONHERO_CONFIG: undefined,
+        GROUPAROO_SPEC_HELPER: undefined,
```
evantahler, line 81: Unloading some (but not all) of the ENV changes from the jest environment

plugins/@grouparoo/demo/__tests__/integration/megaDelete.ts
```diff
@@ -0,0 +1,211 @@
...
+  "cli",
+  "dist",
+  "grouparoo.js"
+);
+const tmpDir = path.join(tmpdir(), "grouparoo", "mega-delete-test");
+const version: string = JSON.parse(
+  fs.readFileSync(path.join(__dirname, "..", "..", "package.json")).toString()
+).version;
+
+const packageJSON = `
```
evantahler: The test `package.json` ends up like:


```json
{
  "author": "Your Name <email@example.com>",
  "name": "mega-delete-test",
  "description": "A Grouparoo Deployment",
  "version": "0.0.1",
  "engines": {
    "node": ">=12.0.0 <17.0.0"
  },
  "dependencies": {
    "@grouparoo/core": "file:/Users/evan/workspace/grouparoo/grouparoo/core/grouparoo-core-0.7.3-alpha.1.tgz",
    "@grouparoo/demo": "file:/Users/evan/workspace/grouparoo/grouparoo/plugins/@grouparoo/demo/grouparoo-demo-0.7.3-alpha.1.tgz",
    "@grouparoo/calculated-property": "file:/Users/evan/workspace/grouparoo/grouparoo/plugins/@grouparoo/calculated-property/grouparoo-calculated-property-0.7.3-alpha.1.tgz",
    "@grouparoo/postgres": "file:/Users/evan/workspace/grouparoo/grouparoo/plugins/@grouparoo/postgres/grouparoo-postgres-0.7.3-alpha.1.tgz"
  },
  "scripts": {
    "start": "cd node_modules/@grouparoo/core && ./bin/start"
  },
  "grouparoo": {
    "plugins": [
      "@grouparoo/demo",
      "@grouparoo/calculated-property",
      "@grouparoo/postgres"
    ]
  }
}

```

plugins/@grouparoo/demo/src/util/shared.ts
bleonard: seemed pretty predictable before and I trusted their recursiveness more than my own, but open

evantahler: The path to the `prettierrc` doesn't work if you are outside of the monorepo (eg you installed this from NPM) - that was the main change

plugins/@grouparoo/demo/__tests__/integration/megaDelete.ts
```diff
@@ -0,0 +1,101 @@
+import { spawn } from "child_process";
+import path from "path";
+import fs from "fs";
+
+const monorepoRoot = path.join(__dirname, "..", "..", "..", "..", "..");
+const bin = path.join(monorepoRoot, "cli", "dist", "grouparoo.js");
+const dir = path.normalize(
+  path.join(monorepoRoot, "apps", "staging-enterprise")
+);
```
evantahler, line 9: Testing within the `staging-enterprise` project

core/src/modules/ops/destination.ts
```diff
@@ -103,32 +101,16 @@ export namespace DestinationOps {
...
-      if (group) return RunOps.run(group, destination.id);
-    } else if (destination.collection === "model") {
-      const model = await GrouparooModel.findById(destination.modelId);
-      if (model) return RunOps.run(model, destination.id);
-    } else {
-      throw new Error(
-        `unknown destination collection ${destination.collection}`
-      );
-    }
-  }
```
evantahler, line 130: This was the same as `DestinationOps.exportMembers`

core/src/modules/ops/record.ts
```diff
@@ -643,8 +643,11 @@ export namespace RecordOps {
 
     for (const _export of existingExportNotDeleted) {
       if (!destinations.map((d) => d.id).includes(_export.destinationId)) {
-        const destination = await Destination.findById(_export.destinationId);
-        if (destination.state === "ready") destinations.push(destination);
+        const destination = await Destination.findOne({
+          where: { id: _export.destinationId, state: "ready" },
+        });
+        // the destination may have been deleted or become not-ready
+        if (destination) destinations.push(destination);
```
evantahler, line 650: This was failing in the new test!  Now we don't throw if the destination is deleted

plugins/@grouparoo/demo/config/logger/users/destinations/logger.json
```diff
@@ -0,0 +1,15 @@
+{
+  "class": "Destination",
+  "id": "logger_destination",
+  "modelId": "users",
+  "name": "Logger Destination",
+  "type": "logger-export-records",
+  "appId": "logger",
+  "collection": "model",
```
evantahler, line 8: This destination starts tracking a model

core/src/modules/ops/destination.ts
pedroslopez: nice catch

core/src/modules/ops/record.ts
pedroslopez: Hm.. why was this failing? The `findById` method on Destinations should clear the default scope and find the deleted one anyway (and then checked that it's ready by the conditional) 🤔

pedroslopez (approved):

core/src/modules/ops/record.ts
evantahler: We no longer send that final export when a destination is deleted.  So, it's possible that at the time of the mega-delete, there are no pending exports or running runs for the destination - so the Destination would be deleted before the records.  We also now keep Exports for 3 months, even past the Destination's deletion... so this method would find Exports that used to belong to the now-deleted destination...

pedroslopez: I guess I don't quite understand how this specific code change addresses that since it seems functionally equivalent to me- it's still looking for that destination id and pushing it if it's ready. There's probably something else I'm missing, but it looks fine either way

evantahler: Oh! `.findById` throws if the item can't be found - this where clause won't throw, and might find `undefined`

evantahler: There's also probably an optimization to INNER JOIN with destinations in the query above to skip this second lookup too

pedroslopez: oh derp, for some reason I thought the destination would be there but in the `deleted` state. Got it!

PR #2536, opened 2021-11-10
Plugin Connections can use multiple Apps

evantahler:
Now, Plugin Connections (import and export) can use more than one app type.  This opens the door for multiple ways of connecting to apps - perhaps there's an oAuth app and a user/password app, or a PEM vs user/password way to connect.  More to come!

### Development

### Impact

### Code review

### DIFF
--- a/core/src/initializers/plugins.ts
+++ b/core/src/initializers/plugins.ts
@@ -164,7 +164,7 @@ export class Plugins extends Initializer {
             .includes(connection.name)
         ) {
           errors.push(
-            `a plugin connection named ${connection.name} is already registered (${connection.name})`
+            `a plugin connection named ${connection.name} is already registered`
           );
         }
         if (!connection.apps || connection.apps.length < 1) {
--- a/core/src/initializers/plugins.ts
+++ b/core/src/initializers/plugins.ts
@@ -164,32 +164,32 @@ export class Plugins extends Initializer {
             .includes(connection.name)
         ) {
           errors.push(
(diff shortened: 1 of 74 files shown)


core/src/initializers/plugins.ts
```diff
@@ -164,32 +164,32 @@ export class Plugins extends Initializer {
             .includes(connection.name)
         ) {
           errors.push(
-            `a plugin connection named ${connection.name} is already registered`
+            `a plugin connection named ${connection.name} is already registered (${connection.name})`
```
pedroslopez: `connection.name` is already in the log?

pedroslopez (approved):
This seems fine, though I'm interested to see how this ends up looking like for the users. Would they see multiple apps in the Apps list for a same app/product? I feel like they should be grouped together somehow, though maybe that's a more UI/presentational thing.

I wonder if apps would become more like what we do with destinations, where there can be a few types within.

evantahler:
> This seems fine, though I'm interested to see how this ends up looking like for the users. Would they see multiple apps in the Apps list for a same app/product? I feel like they should be grouped together somehow, though maybe that's a more UI/presentational thing.
> 
> I wonder if apps would become more like what we do with destinations, where there can be a few types within.

For sure! I wanted to separate this work into 3 PRs: 
* Update the plugin structure/types (do this fast so I don't block other people working on plugins)
* UI Updates
* Make the first plugin with 2 apps - it's going to be a new way to connect to Snowflake via PEMs rather than user/pass

PR #2537, opened 2021-11-10
Snowflake Keypair App

### LINKED ISSUE #2532: Snowflake Key Pair Authentication
Snowflake supports key pair authentication
https://docs.snowflake.com/en/user-guide/key-pair-auth.html#configuring-key-pair-authentication
https://docs.snowflake.com/en/user-guide/nodejs-driver-use.html#label-nodejs-key-pair-authentication

Grouparoo should consider supporting this method of authentication since some organizations have strict security requirements that prohibit user/password based authentication.

evantahler:
Creates a new `snowflake-keypair` app to allow connecting to Snowflake databases with keypair authentication rather than password authentication.  

* We still assume that password-based authentication is most common. To that end, the existing password-based app is still called `snowflake` and this new one is `snowflake-keypair`
* Providing the private key in the property format is tricky - notes on how to do so are in the README for this plugin

---

This PR also makes the selection process better when adding a new App, Source, or Destination.  If there is more than one option for the Plugin or App you selected, you'll be taken to a chooser page.  If not, we'll make the item for you and take you to the edit page directly

apps

Closes #2532

### Development

### Impact

### Code review


plugins/@grouparoo/snowflake/README.md
```diff
@@ -8,6 +8,19 @@ Learn more about Grouparoo Plugins at [www.grouparoo.com/docs/plugins](https://w
...
+
+```bash
+# If rsa_key.p8 is the private key you generated for snowflake auth, per https://docs.snowflake.com/en/user-guide/key-pair-auth.html#configuring-key-pair-authentication
+# The following command will output a string Grouparoo can use for the `Private Key` connection option
+
+awk 'NF {sub(/\n/, ""); printf "%s\\n",$0;}' rsa_key.p8
+
+# note: some operating systems will include a final `%` at the end of the line.  Do not not include it
+```
+
```
evantahler, line 23: This was a whole thing :/

ui/ui-config/pages/app/new/index.tsx
```diff
@@ -35,17 +33,19 @@ class CustomErrorHandler extends EventDispatcher<{ error: string }> {
...
-    plugins,
+    availablePlugins,
+    installedPlugins,
   }: {
     errorHandler: ErrorHandler;
-    apps: Actions.AppOptions["types"];
-    plugins: Actions.PluginsAvailableList["plugins"];
+    installedPlugins: Actions.AppOptions["plugins"];
+    availablePlugins: Actions.PluginsAvailableList["plugins"];
   } = props;
```
evantahler: I think this change helps clarify what's happening on this page - we actually listing plugins that provide apps (some installed, some that could be installed).  We aren't listing apps.

core/src/actions/properties.ts
```diff
@@ -371,12 +371,13 @@ export class PropertyRecordPreview extends AuthenticatedAction {
...
-        params.options,
-        params.filters
-      );
+      newPropertyValues =
+        (await source.importRecordProperty(
+          record,
+          property,
+          params.options,
+          params.filters
+        )) ?? [];
```
evantahler, line 380: Adding some more saftey around a preview not finding any values - this and `ui/ui-components/components/property/RecordPreview.tsx` could crash without any values returned prior to this

core/bin/dev
```diff
@@ -12,7 +12,7 @@ else
 fi
 
 export NEXT_DEVELOPMENT_MODE=true
-export GROUPAROO_RUN_MODE="cli:start"
+export GROUPAROO_RUN_MODE="${GROUPAROO_RUN_MODE:='cli:start'}"
```
evantahler, line 15: This allows `npm run dev` to work in `apps/staging-config` again, and load the UI properly

core/src/actions/apps.ts
```diff
@@ -59,62 +59,12 @@ export class AppOptions extends AuthenticatedAction {
...
-            )
-          )
-            ? true
-            : false;
-
-          const appsOfThisType = await App.count({ where: { type: app.name } });
-          let addible = true;
-          if (app.maxInstances && app.maxInstances === appsOfThisType) {
-            addible = false;
-          }
```
pedroslopez, line 98: Noting that this `maxInstances` check seems to be gone

plugins/@grouparoo/snowflake/README.md
pedroslopez: Unfortunate, but makes sense. In general it's not great to log things out in a way that would be kept in your shell history. Should we recommend directing the output to a file instead?

ui/ui-components/pages/app/new/[...plugin].tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const create = async (
+    app: Actions.AppOptions["plugins"][number]["apps"][number]
+  ) => {
+    setLoading(true);
+    const response: Actions.AppCreate = await execApi("post", `/app`, {
+      type: app.name,
+    });
+    if (response?.app) {
+      return router.push("/app/[id]/edit", `/app/${response.app.id}/edit`);
```
Comment (pedroslopez) [inline ui/ui-components/pages/app/new/[...plugin].tsx]: is the `as` stuff required when pushing to routes still or can we use the dynamic path directy?

ui/ui-components/pages/app/new/[...plugin].tsx
```diff
@@ -0,0 +1,120 @@
...
+              >
+                Create App
+              </LoadingButton>
+            </Card.Body>
+          </Card>
+          <br />
+        </Fragment>
+      ))}
+
+      <Form id="form"></Form>
```
Comment (pedroslopez) [inline ui/ui-components/pages/app/new/[...plugin].tsx]:
```suggestion
```

plugins/@grouparoo/snowflake/src/lib/query-import/connection.ts
```diff
@@ -4,7 +4,7 @@ import { getChangedRows } from "./getChangedRows";
 
 export function getConnection() {
   return buildConnection({
-    apps: ["snowflake"],
+    apps: ["snowflake", "snowflake-keypair"],
```
pedroslopez, line 7: smooth!

plugins/@grouparoo/snowflake/src/initializers/plugin.ts
```diff
@@ -95,6 +95,56 @@ export class Plugins extends Initializer {
           ],
           methods: { test, connect, disconnect, appQuery },
         },
+        {
+          name: "snowflake-keypair",
```
pedroslopez, line 99: aha, so that's what this looks like! The fact that plugins could already register multiple apps makes this quite nice

pedroslopez (commented):
Nice! Adding new apps with different options seems pretty smooth!

One thing that I kept thinking about when going through this is the fact that we have two actions that return available plugins, and they are oftentimes combined: `plugins:available:list` and now `app:options`. Not exactly sure if there's anything to do about it yet since they do return different things for each plugin, but maybe something to keep an eye out for. 

On the Snowflake side: do we need to add some tests for this new app auth method?

ui/ui-components/pages/app/new/[...plugin].tsx
Comment (evantahler) [reply ui/ui-components/pages/app/new/[...plugin].tsx]: I think the `as` part of the route is no longer required!

core/src/actions/apps.ts
evantahler: Yeah, we don't have any apps that are limited in this way any more (like `events` or `manual`).  I was hesitant to remove that from the plugin type or App model at this point, but we might in the future

plugins/@grouparoo/snowflake/README.md
evantahler: eh... I'm not too worried about it.  The /value/ won't be stored in your shell history.  And if we wrote to a file, the next step would be to `cat` it anyway so 🤷

evantahler:
https://github.com/grouparoo/grouparoo/pull/2543 will be merged here

ui/ui-components/pages/app/new/[...plugin].tsx
Comment (pedroslopez) [reply ui/ui-components/pages/app/new/[...plugin].tsx:32]: could remove it from here in that case, but it's fine either way

Comment (pedroslopez) [reply ui/ui-components/pages/app/new/[...plugin].tsx]: Could remove it from here in that case, but fine either way

pedroslopez (approved):
🎉

PR #2540, opened 2021-11-11
Replace directly mapped property per source with a single primary key property for all models

edmundito:
## Change description

This change replaces the idea of a directly mapped property for each source with with a single primary key property for the model. This primary key is defined by the mapping set in the first source added to the model. You can only have one primary key per model and for other sources that key must be chosen.

When the first source is added to the model:
- This source is treated as the primary source
- When mapping the property, this property will become the primary key
- The first source mapped property can only be set to properties created for the first source

When other sources are added to the model:
- For mapping, they can select the property from any source
- The primary key of the first source will remain the primary key

Adds a primary identifier to the source and properties tables:
image
image
image

### Development

### Impact

Migration created to rename the `directlyMapped` column in the properties table to `isPrimaryKey`

### Code review


core/src/migrations/000090-renamePropertyDirectlyMappedToIsPrimaryKey.ts
```diff
@@ -0,0 +1,20 @@
+import Sequelize from "sequelize";
```
evantahler: Note on Coordination - looks like https://github.com/grouparoo/grouparoo/pull/2527 has migration 90 so this should be 91

core/src/migrations/000090-renamePropertyDirectlyMappedToIsPrimaryKey.ts
```diff
@@ -0,0 +1,20 @@
...
+
+// TODO: Type as migration?
```
evantahler: I don't think I understand this TODO.  This migration looks good!

ui/ui-components/pages/model/[modelId]/properties.tsx
```diff
@@ -147,6 +147,7 @@ export default function Page(props) {
                       </strong>
                     </a>
                   </GrouparooLink>
+                  {rule.isPrimaryKey && " (primary)"}
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/properties.tsx]: This is a nice touch.  Perhaps we can use a `<Badge size='small' variant='info' >primary</Badge>`?  This seems like the type of thing we badge on the source and destination already

evantahler (commented):
Nice!  The only blocking change is the migration name below.  The rest are optional.

Part 2 of this story (removing `*bootstrappedProperty*`) can be done in another PR.

pedroslopez (approved):
Assuming the "bootstrapped property" changes are coming later, looks good!
Like Evan mentioned, definitely keep an eye out for the migration number

evantahler:
There was part of the story to add a validation that there could only be one `primaryKey` as well - that could go here if you wanted

edmundito:
Paired with @evantahler to update how we determine the Grouparoo model's primary key across multiple sources and updated the existing tests to work with the changes. I still need to add new tests to verify this functionality.

core/src/models/Source.ts
```diff
@@ -428,13 +428,43 @@ export class Source extends LoggedModel<Source> {
   }
 
   @AfterSave
```
edmundito: I was thinking about removing the after save hook and just let this be triggered by the `afterSetMapping` hook instead. I don't think this needs to be updated every time the model is saved only when the mapping is set.

What do y'all think?

core/__tests__/models/source/source.ts
```diff
@@ -618,22 +695,42 @@ describe("models/source", () => {
...
+
+      const myEmailProperty = await Property.findOne({
+        where: { key: "myEmail" },
+      });
+      expect(myEmailProperty.isPrimaryKey).toBe(false);
 
-      await firstSource.setMapping({ userId: "userId" });
+      await source.setMapping({ userId: "userId" });
     });
   });
```
evantahler, line 743: I think we need a test for the workflow when we want to change the mapped property on the primary source - this should update the isPrimaryKey for both the old and new property.  we'll need to un-primary the old property first so the validation of 'only one primary property' isn't violated

ui/ui-components/pages/model/[modelId]/source/[sourceId]/mapping.tsx
```diff
@@ -16,48 +16,53 @@ import ModelBadge from "../../../../../components/badges/ModelBadge";
...
+  scheduleCount: number;
+  hydrationError: Error;
+  source: Models.SourceType;
+  properties: Models.PropertyType[];
+  preview: Actions.SourcePreview["preview"];
+  propertyExamples: Actions.PropertiesList["examples"];
+}
+
+export default function Page(props: Props & NextPageContext) {
+  const { errorHandler, successHandler, types, scheduleCount, hydrationError } =
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/mapping.tsx:32]: nice!

ui/ui-components/hooks/useApi.ts
```diff
@@ -10,14 +10,14 @@ export function UseApi(
   errorHandler?: ErrorHandler,
   uploadHandler?: UploadHandler
 ) {
-  async function execApi(
+  async function execApi<T extends {} = any>(
```
evantahler, line 13: I'm looking forward to the day that T is required - that will ensure our API usage is typed!

evantahler (approved):

PR #2541, opened 2021-11-11
Fix `APIData.ensureObject` optional recusion flag

evantahler:
Fixes a bug in our API's input formatter - the second option was always `api` when called from an action.

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/apiData.ts
+++ b/core/src/modules/apiData.ts
@@ -1,13 +1,16 @@
+import { Api } from "actionhero";
+
 export namespace APIData {
   export function ensureObject(
     param: { [key: string]: any } | string,
+    api: Api,
     recursing = false
   ) {
     if (!param) {
       return null;
     } else if (Array.isArray(param)) {
       try {
-        return param.map((row) => APIData.ensureObject(row, true));
+        return param.map((row) => APIData.ensureObject(row, api, true));
       } catch (error) {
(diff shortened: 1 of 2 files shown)


evantahler:
Found via https://github.com/grouparoo/www.grouparoo.com/pull/780

core/src/modules/apiData.ts
```diff
@@ -1,13 +1,16 @@
...
     param: { [key: string]: any } | string,
+    api?: Api,
     recursing = false
   ) {
     if (!param) {
       return null;
     } else if (Array.isArray(param)) {
       try {
-        return param.map((row) => APIData.ensureObject(row, true));
+        return param.map((row) => APIData.ensureObject(row, api, true));
```
krishnaglick, line 13: So is this now including something that was missed?

krishnaglick (approved):

core/src/modules/apiData.ts
evantahler: Yeah - the test I added was failing before because there already was a second param being passed into all these formatters - the `api` object.  That should be changed in actionhero...

evantahler: https://github.com/actionhero/actionhero/pull/2012 is the change in actionhero to prevent this kind of thing in the future

PR #2547, opened 2021-11-12
Disable Telemetry in Tests

evantahler:
More throughly disabled sending telemetry in tests.  We'll do this by setting the ENV variable `GROUPAROO_TELEMETRY_DISABLED` at the system level

### Development

### Impact

### Code review

### DIFF
--- a/apps/staging-enterprise/__tests__/integration
+++ b/apps/staging-enterprise/__tests__/integration
@@ -5,7 +5,6 @@ set -e
 cd "`dirname "$0"`/.."
 
 # Set needed ENV
-export GROUPAROO_TELEMETRY_ENABLED=false
 export REDIS_URL="redis://mock"
 export DATABASE_URL="sqlite://grouparoo_test.sqlite"
 export SERVER_TOKEN="ABC123"
--- a/plugins/@grouparoo/spec-helper/src/lib/cliSpecHelper.ts
+++ b/plugins/@grouparoo/spec-helper/src/lib/cliSpecHelper.ts
@@ -42,7 +42,11 @@ export namespace CLISpecHelper {
 
       const spawnProcess = spawn(command, args, {
         cwd,
-        env,
+        env: {
+          ...env,
+          GROUPAROO_TELEMETRY_DISABLED:
+            process.env.GROUPAROO_TELEMETRY_DISABLED,
+        },
       });
 
       spawnProcess.stdout.on("data", (data) => {
@@ -68,7 +72,10 @@ export namespace CLISpecHelper {
   ): { exitCode: number; stderr: string; stdout: string } {
     const syncResponse = spawnSync(command, args, {
       cwd,
-      env,
+      env: {
+        ...env,
+        GROUPAROO_TELEMETRY_DISABLED: process.env.GROUPAROO_TELEMETRY_DISABLED,
+      },
     });
     const stdout = syncResponse.stdout.toString();
     const stderr = syncResponse.stderr.toString();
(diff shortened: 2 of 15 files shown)


plugins/@grouparoo/spec-helper/src/lib/cliSpecHelper.ts
```diff
@@ -42,7 +42,11 @@ export namespace CLISpecHelper {
 
       const spawnProcess = spawn(command, args, {
         cwd,
-        env,
+        env: {
+          ...env,
+          GROUPAROO_TELEMETRY_DISABLED:
+            process.env.GROUPAROO_TELEMETRY_DISABLED,
+        },
       });
```
evantahler: Sometimes in tests we create totally fresh ENV (not relying on system env).  In these cases, we need to be sure to forward the previous value of `GROUPAROO_TELEMETRY_DISABLED`

cc @bleonard

bleonard: Why not just always "false" ? I think this context is always in test land

plugins/@grouparoo/spec-helper/src/lib/cliSpecHelper.ts
```diff
@@ -68,7 +71,10 @@ export namespace CLISpecHelper {
   ): { exitCode: number; stderr: string; stdout: string } {
     const syncResponse = spawnSync(command, args, {
       cwd,
-      env,
+      env: {
+        ...env,
+        GROUPAROO_TELEMETRY_ENABLED: process.env.GROUPAROO_TELEMETRY_ENABLED,
```
bleonard: Maybe just always "false" bc this is always in the test context? oR is there a reason for it to be true?

andyjih (commented):
Defer to Brian's review.

plugins/@grouparoo/spec-helper/src/lib/cliSpecHelper.ts
evantahler: probably not in this test!

evantahler:
CircleCI doesn't seem to print ENV set at the container or step level, but you can confirm that it is set via SSH:

```
➜  ~ ssh -p xxx xxx.xxx.xxx.xxx
...
circleci@4ba16be2c599:~$ echo $GROUPAROO_TELEMETRY_ENABLED
false
```

apps/staging-enterprise/__tests__/integration
```diff
@@ -5,7 +5,6 @@ set -e
 cd "`dirname "$0"`/.."
 
 # Set needed ENV
-export GROUPAROO_TELEMETRY_ENABLED=false
```
pedroslopez, line 8: Why remove these from the bash tests? I'm assuming we're now relying CircleCI to set it for us, but that wouldn't take effect when running locally

pedroslopez: although, I guess everyone running tests locally should have telemetry disabled already in their env?

pedroslopez (approved):

apps/staging-enterprise/__tests__/integration
evantahler: Yeah - we all should have telemetry disabled.  And you shouldn't need to mess with the tests to re-record the nocks - just change your ENV.

I was most concerned with our developer experience going forward.  You shouldn't need to remember to change your environment in any future tests; the OS/platform should handle this for you.

PR #2548, opened 2021-11-12
AppRefreshQuery tasks/actions update

teallarson:
## Change description

App Refresh Query should act like existing schedule triggers -- Tasks should not cancel runs, but Actions should.

- If `AppRefreshQueryOps.query()` returns a new value, `AppRefreshQueryOps.triggerSchedules()` is called.  If it's called from an Action, a new parameter, `shouldCancel: Boolean = true` is passed.
- If the value is new, it saves the new value, updates the timestamps, and enqueues all schedules for the app.  This triggers runs if the schedule doesn't already have one in progress.  If `shouldCancel === true`, then existing runs are canceled and _all_ schedules are enqueued.

This is similar to the pattern used by the existing tasks and actions for enqueueing schedules by timer.

I've also updated the task and action names to `appRefreshQuery:run` because they do more than just query, they also trigger the schedules.  And I've refactored across the board to use the `AppRefreshQueryOps` functions as methods on the model rather than in isolation to match patterns in place elsewhere in the codebase.

Finally, this PR adds code config and UI for a configurable `recurringFrequency` on the appRefreshQuery.
Screen Shot 2021-11-18 at 4 28 21 PM

### Development

### Impact

### Code review


teallarson:
@evantahler this works for a long run, which was the primary concern, but I don't know that it does anything for a long-running query.

teallarson:
@evantahler before I start writing tests for this behavior can you take a look at the solution I've implemented?

evantahler:
The description says that all Schedules for Sources that belong to this App have to have no runs for new Runs to be Scheduled.  That means that one slow Run could block all the rest.  Perhaps the logic could be changed to:

* `AppRefreshQueryOps.query()` runs and stores the new value each time, as it does now.  
* Then, `triggerSchedules()`only triggers Schedules for sources that don't already have a running run started by a previous appRefreshQuery check.

core/src/models/Schedule.ts
```diff
@@ -224,11 +224,12 @@ export class Schedule extends LoggedModel<Schedule> {
...
   }
 
-  async enqueueRun() {
+  async enqueueRun(triggeredBy?: string) {
     const run = await Run.create({
       creatorId: this.id,
       creatorType: "schedule",
       state: "running",
+      triggeredBy: triggeredBy,
     });
```
evantahler: I like this pattern a lot!
It might be interesting to to make `triggeredBy` be required.  Other options could be: 
* `manual` (user hit the button in the UI)
* `time` or `schcedule:check` (timer went off)

core/src/modules/ops/appRefreshQuery.ts
```diff
@@ -31,6 +31,14 @@ export namespace AppRefreshQueryOps {
   }
 
   export async function triggerSchedules(appRefreshQuery: AppRefreshQuery) {
+    //if there are runs that were triggered by this that are still running, don't save anything and don't trigger anything
+    const activeRuns = await Run.findAll({
+      where: { state: "running", triggeredBy: appRefreshQuery.id },
+    });
```
evantahler: You don't need to load up all the runs (findAll), you really just want to see if any are running - try `.count()`

teallarson:
> The description says that all Schedules for Sources that belong to this App have to have no runs for new Runs to be Scheduled. That means that one slow Run could block all the rest. Perhaps the logic could be changed to:
> 
> * `AppRefreshQueryOps.query()` runs and stores the new value each time, as it does now.
> * Then, `triggerSchedules()`only triggers Schedules for sources that don't already have a running run started by a previous appRefreshQuery check.

If the purpose is an app-wide refresh, I don't think that we should run some schedules and not others, which I think was what your suggestion was.  

I also don't think we'd want it to save a new value if it didn't trigger a run.  That means if the app's data updates during a run, the new value would be saved, but no runs would ever be triggered by that data change.  Wouldn't we want it to catch that data change after the run it's in the middle of finishes?

Open to discussion on any of this, but saving a value and triggering _some_ schedules while one is still mid-run seems to go against the idea of an app-wide update.

I did move this logic to the task -- we probably just want to avoid enqueueing an `appRefreshQuery:query` task at all if there are associated runs still in progress for that particular query.


EDIT: After some discussion, a lot of the above was misguided.  Moved forward with the idea that Tasks should not cancel existing runs ever, but Actions should.

core/src/models/Schedule.ts
teallarson: Interesting!  I added "API" and "UI" versus manual.

core/src/actions/schedules.ts
```diff
@@ -99,13 +99,14 @@ export class SchedulesRun extends AuthenticatedAction {
...
       modelId: { required: false },
+      triggeredBy: { required: true, default: "API" },
     };
   }
 
   async runWithinTransaction({
     params,
   }: {
-    params: { scheduleIds?: string[]; modelId?: string };
+    params: { scheduleIds?: string[]; modelId?: string; triggeredBy: string };
```
evantahler: I don't think `triggeredBy` is used in the action

evantahler: The action `schedule:run` (singular) should also get `triggeredBy`

evantahler: I'd suggest the name of the action rather than `"API"` (e.g. `this.name`) for the default value

core/src/migrations/000092-appRefreshQueryInterval.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export default {
+  up: async (
+    queryInterface: Sequelize.QueryInterface,
+    DataTypes: typeof Sequelize
+  ) => {
+    await queryInterface.addColumn("appRefreshQueries", "recurringFrequency", {
+      type: DataTypes.BIGINT,
+      allowNull: true,
+    });
```
evantahler: You may want to backfill `recurringFrequency` with the previous value for those `appRefreshQueries` that already exist at the time of the migration

Or...

Since this feature isn't released yet, we it's probably OK to skip the backfill

core/src/models/AppRefreshQuery.ts
```diff
@@ -127,6 +143,18 @@ export class AppRefreshQuery extends LoggedModel<AppRefreshQuery> {
...
+  static async checkRecurringFrequency(instance: AppRefreshQuery) {
+    // we cannot use the @Min validator as null is also allowed
+    if (instance.recurringFrequency) {
+      if (instance.recurringFrequency < 1000 * 60) {
+        throw new Error(
+          "recurring frequency is required to be one minute or greater"
+        );
+      }
+    }
+  }
```
evantahler, line 155: Please add a model test for this

core/src/models/AppRefreshQuery.ts
```diff
@@ -77,13 +80,26 @@ export class AppRefreshQuery extends LoggedModel<AppRefreshQuery> {
       state: this.state,
       lastChangedAt: APIData.formatDate(this.lastChangedAt),
       lastConfirmedAt: APIData.formatDate(this.lastConfirmedAt),
+      recurringFrequency: this.recurringFrequency,
```
evantahler, line 82: Please test that `recurringFrequency` is included in an Action's response

core/src/modules/configLoaders/app.ts
```diff
@@ -74,15 +74,17 @@ export async function loadApp(
     if (!appRefreshQuery) {
       appRefreshQuery = await AppRefreshQuery.create({
         appId: configObject.id,
-        refreshQuery: configObject.refreshQuery,
+        refreshQuery: configObject.refreshQuery.query,
+        recurringFrequency: configObject.refreshQuery.recurringFrequency,
```
evantahler: is it possible that `configObject.refreshQuery` is undefined?  This might be safer:

```suggestion
        refreshQuery: configObject.?refreshQuery.query,
        recurringFrequency: configObject.?refreshQuery.recurringFrequency,
```

evantahler: Same comment in the `else` case below

evantahler (changes requested):
Structurally, this is looking much better!

There are some things to fix in the actions (input not being used or tested), and a few more tests to add.

core/src/actions/schedules.ts
teallarson: I removed `triggeredBy` entirely as it isn't needed as things stand... I missed one during find and replace!

core/src/modules/configLoaders/app.ts
teallarson: I think it should actually check that `configObject.refreshQuery?.length > 1` at the top of the entire if block (line 65)... otherwise if someone leaves the defaults in the config file it would create an empty appRefreshQuery for them, which doesn't make sense.  

Adding both.

evantahler (approved):
🥳

PR #2552, opened 2021-11-17
Treating options that are not required anymore.

pauloouriques:
Treating the options:
- groupObject
- groupNameField
- membershipObject
- membershipRecordField
- membershipGroupField


plugins/@grouparoo/salesforce/src/lib/export/model.ts
```diff
@@ -32,8 +32,7 @@ export function getSalesforceModel(
 
   const modelKeys = Object.keys(model);
   const refKeys = modelKeys.filter((k) => k.indexOf("Reference") > 0);
-  // refKeys are optional
-  const requiredKeys = modelKeys.filter((k) => !refKeys.includes(k));
+  const requiredKeys = ["recordObject", "recordMatchField"];
```
bleonard: There is something to do in this file like we did with references where it says `// needs either zero or all refKeys`
And test(s)  to that effect as well.
So if they have 1 of the group ones, they need to have them all.

plugins/@grouparoo/salesforce/src/lib/export/model.ts
```diff
@@ -32,8 +32,17 @@ export function getSalesforceModel(
 
   const modelKeys = Object.keys(model);
   const refKeys = modelKeys.filter((k) => k.indexOf("Reference") > 0);
```
bleonard: Might as well just make these `refKeys` explicit, too then (in an array of the 3)

plugins/@grouparoo/salesforce/src/lib/export/model.ts
```diff
@@ -60,10 +74,12 @@ export function getSalesforceModel(
     }
   }
   if (count > 0) {
-    if (refKeys.length !== count) {
-      throw new Error(`All Salesforce reference model data is required`);
+    if (keys.length !== count) {
+      const message =
+        keys[0].indexOf("Reference") > 0
+          ? `All Salesforce reference model data is required`
```
bleonard: And let's switch this one to: `To enable reference data syncing, all related options must be set.`

bleonard: Then maybe we should pass in the word `Group` or `reference` into this function from its caller instead of trying to guess.

bleonard (commented):
I'd expect to see some checks in the `exportRecords` code that bailed out of the group work if the options are not set.

plugins/@grouparoo/salesforce/__tests__/export-objects/destination-options.ts
```diff
@@ -310,4 +316,68 @@ describe("salesforce/sales-cloud/destinationOptions", () => {
...
+        exports: [
+          {
+            recordId: id1,
+            oldRecordProperties: {},
+            newRecordProperties: {
+              Custom_External_ID__c: custom1,
+              LastName: "Smith",
+            },
+            oldGroups: [],
+            newGroups: [],
```
bleonard: let's have a test where we send in some groups (new and old) here and the code shouldn't do anything because the group options are not set. i'm still expecting some changes to `SalesforceExporter` code like `addToGroups` and anywhere there is code like `{ membershipObject, membershipRecordField, membershipGroupField }` or `{ groupObject, groupNameField }`, etc

I think maybe a new test file is a good idea.

plugins/@grouparoo/salesforce/jest.config.js
```diff
@@ -2,7 +2,7 @@ const { helper } = require("@grouparoo/spec-helper");
 
 module.exports = {
   testEnvironment: "node",
-  testTimeout: helper.defaultTime,
+  testTimeout: helper.hugeTime,
```
bleonard: let's not check this in. assuming the nock version is fast.

bleonard (approved):

PR #2555, opened 2021-11-18
Empty query crashes "run query"

teallarson:
## Change description

Check that a query is at least 1 character long and is in the "ready state" to display the "run refresh query" button.  Also triggers a run when the query is created.

before adding query: 
Screen Shot 2021-11-18 at 1 10 46 PM

after adding query:
Screen Shot 2021-11-18 at 1 15 19 PM

Also, add bootstrap class `text-break` to result so results don't overflow the box

Screen Shot 2021-11-18 at 1 10 13 PM

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/components/alerts/Success.tsx
+++ b/ui/ui-components/components/alerts/Success.tsx
@@ -42,6 +42,7 @@ export default function SuccessAlert({
       show={show}
       onClose={() => setShow(false)}
       variant="success"
+      className="text-break"
     >
       {message}
     </Alert>
--- a/ui/ui-components/pages/app/[id]/refresh.tsx
+++ b/ui/ui-components/pages/app/[id]/refresh.tsx
@@ -304,7 +304,7 @@ export default function Page(props) {
                   testResult.success !== false &&
                   testResult.success !== undefined &&
                   !testResult.error ? (
-                    <Alert variant="success">
+                    <Alert variant="success" className="text-break">
                       <strong>Test Passed. </strong>Sample Value ={" "}
                       {testResult.message}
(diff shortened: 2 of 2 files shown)


ui/ui-components/components/alerts/Success.tsx
```diff
@@ -42,6 +42,7 @@ export default function SuccessAlert({
       show={show}
       onClose={() => setShow(false)}
       variant="success"
+      className="text-break"
```
evantahler, line 45: This is new to me and very powerful! https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

ui/ui-components/pages/app/[id]/refresh.tsx
```diff
@@ -88,8 +88,26 @@ export default function Page(props) {
...
+          });
+        } else {
+          successHandler.set({
+            message: `App Refresh Query updated. Query returned ${response.appRefreshQuery.value}. No schedules enqueued.`,
+          });
+        }
+        setAppRefreshQuery(response.appRefreshQuery);
+      } finally {
+        setLoading(false);
+      }
```
Comment (evantahler) [inline ui/ui-components/pages/app/[id]/refresh.tsx:110]: This is new too - looks like after editing the query, you will try to run it now.  Makes sense!

evantahler (approved):

ui/ui-components/components/alerts/Success.tsx
teallarson: Yep! And bootstrap's utility class adds support for IE/Edge as well https://getbootstrap.com/docs/4.4/utilities/text/#word-break

PR #2560, opened 2021-11-19
When an export is rate-limited, do not count that as an error

evantahler:
This PR updates how we handle rate-limiting errors with a `retryDelay`.  If there's a `retryDelay`, the error will not count against the 5 retires (configurable via Setting) that each Export has to be sent.

The UI is also adjusted to make it more clear that this export is waiting for the right time to be sent

Screen Shot 2021-11-18 at 5 49 55 PM

## Change description

Description here

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/components/export/List.tsx
+++ b/ui/ui-components/components/export/List.tsx
@@ -94,6 +94,22 @@ export default function ExportsList(props) {
     );
   }
 
+  function getExportDelayRow(_export: Models.ExportType) {
+    if (!_export.sendAt || _export.sendAt <= new Date().getTime()) {
+      return null;
+    }
+
+    return (
+      <tr>
+        <td colSpan={7} style={{ border: 0 }}>
+          <Alert variant={"info"}>
+            Export delayed until {formatTimestamp(_export.sendAt)}
+          </Alert>
+        </td>
+      </tr>
+    );
+  }
+
   return (
     <>
       {props.header ? props.header : <h1>Exports</h1>}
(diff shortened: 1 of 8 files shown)


ui/ui-components/components/export/List.tsx
```diff
@@ -77,18 +77,31 @@ export default function ExportsList(props) {
...
 
   function getErrorRow(_export: Models.ExportType) {
-    if (!_export.errorMessage) {
-      return null;
-    }
+    if (!_export.errorMessage) return null;
+
+    const delayMessage =
+      ["pending", "processing"].includes(_export.state) &&
+      _export.sendAt > new Date().getTime()
```
evantahler, line 84: There are some potential shenanigans here about time drift between the server and browser... but I don't think it will be that bad, as this is just for the UI and not processing...

ui/ui-components/components/export/List.tsx
```diff
@@ -77,18 +77,31 @@ export default function ExportsList(props) {
...
+        ? `Export will be retried after ${formatTimestamp(_export.sendAt)}`
+        : null;
 
     let level = "warning";
-    if (_export.errorLevel === "info") {
+    if (_export.errorLevel === "info" || delayMessage) {
       level = "info";
     }
```
evantahler, line 91: I decided to keep the error message the plugin returns along with the retryDelay, but make it less scary looking

pedroslopez (approved):
Looks good! seems like we have tests already for when there is no retryDelay and the retryCount is incremented, so 👍🏼

PR #2561, opened 2021-11-19
Only show properties for this Model in mustache builder

evantahler:
Fixes a bug in which Grouparoo was displaying properties from other models as options for the Mustache property builder

Screen Shot 2021-11-01 at 2 00 45 PM

This PR also now shows Properties that are mapped to a Source's column that are not directly mapped

Screen Shot 2021-11-19 at 10 54 01 AM

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/modules/plugin.ts
+++ b/core/__tests__/modules/plugin.ts
@@ -279,6 +279,35 @@ describe("modules/plugin", () => {
           )
         ).toEqual(initialString);
       });
+
+      test("only properties from the same model can be used in mustache replacement", async () => {
+        const otherModel = await helper.factories.model({ name: "otherModel" });
+        const otherApp = await helper.factories.app();
+        const otherSource = await helper.factories.source(otherApp, {
+          modelId: otherModel.id,
+        });
+        await otherSource.setOptions({ table: "foo" });
+        const otherProperty = await otherSource.bootstrapUniqueProperty(
+          "otherUserId",
+          "integer",
+          "otherId"
+        );
+        await otherSource.setMapping({ otherUserId: "otherUserId" });
+        await otherSource.update({ state: "ready" });
+
+        const property = await Property.findOne({
+          where: { key: "userId" },
+        });
+        const source = await property.$get("source");
+        const initialString = `select * from users where id = {{ ${otherProperty.key} }}`;
+
+        await expect(
+          plugin.replaceTemplateRecordPropertyKeysWithRecordPropertyId(
+            initialString,
+            source.modelId
+          )
+        ).rejects.toThrow('missing mustache key "otherUserId"');
+      });
     });
 
     describe("createImport", () => {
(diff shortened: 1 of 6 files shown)


ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx
```diff
@@ -178,65 +182,85 @@ export default function Page(props) {
...
+              </LoadingButton>
+            )}
+          </td>
+        </tr>
+        {otherProperties.map((otherProperty) => (
+          <tr key={`otherProperty-${otherProperty.id}`}>
+            <td style={{ border: 0 }} />
+            <td style={{ border: 0 }}>➡</td>
+            <td style={{ border: 0 }}>
+              <strong>{otherProperty.key}</strong>
```
Comment (pedroslopez) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx:247]: I don't quite understand where these other properties are coming from, could you give an example?

If it's from another source, maybe we should show where this other property is coming from in parenthesis next to the key or something to avoid confusion... 🤔

core/__tests__/modules/plugin.ts
```diff
@@ -279,6 +279,35 @@ describe("modules/plugin", () => {
...
+        });
+        const source = await property.$get("source");
+        const initialString = `select * from users where id = {{ ${otherProperty.key} }}`;
+
+        await expect(
+          plugin.replaceTemplateRecordPropertyKeysWithRecordPropertyId(
+            initialString,
+            source.modelId
+          )
+        ).rejects.toThrow('missing mustache key "otherUserId"');
```
evantahler, line 309: We were already enforcing that you couldn't use a Properties from another model, so this PR is mostly about updating the UI. This test just confirms that behavior more explicitly.

pedroslopez (commented):
The mustache filtering is 👍🏼 , I just mainly have a question about the change on the multiple properties view

ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx
Comment (evantahler) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx:247]: Ah! The properties are from the same source, but perhaps they aren't the ones with `exact` aggregation methods.  I renamed things so we have `exactProperty` and `otherProperties` now. Before this PR, we would only show you those with `exact` - how it's displaying all the properties that have something to do with that column.  

Using the `demo` command, here's the purchases table - now we show that we have a property for price (`ltv`) even though it's not the `exact` one, along with lots of others


Screen Shot 2021-11-23 at 10 01 47 AM

Comment (pedroslopez) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx:247]: Oh! That makes way more sense now. Thanks for the example

pedroslopez (approved):

PR #2564, opened 2021-11-20
First-time User Improvements

evantahler:
* Allows Properties to have the key `id`
  * There was an explicit check that Grouparoo would not set Record Properties with a key of `id` that should have been removed as part of https://github.com/grouparoo/grouparoo/pull/1996
  * Added a test that locks in `id` is a valid Property Key
* Fixes a bug when determining a Property's scope if the Model is not ready yet
  * This manifested itself in strange ways, like when bootstrapping the first property for a source/model Grouparoo was unable to determine if the Source was ready to have a schedule or not
* Prevents the saving of Sample Records when there are no non-null directlyMapped Property Values
  * The fixes above should prevent users from getting into this state, but we'll be extra safe about this in the future
* If there are no Teams yet on the Grouparoo instance, and an authorization/authentication error occurs, redirect the user to `/` to start the team creation process
  * This happens commonly when someone transitions from `grouparoo config` to `grouparoo start` for the first time and they use the same browser tab

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/property/property.ts
+++ b/core/__tests__/models/property/property.ts
@@ -164,7 +164,7 @@ describe("models/property", () => {
     });
 
     test("keys cannot be from the reserved list of keys", async () => {
-      const reservedKeys = ["grouparooId", "grouparooCreatedAt"];
+      const reservedKeys = ["grouparooId", "grouparooCreatedAt", "_meta"];
       for (const i in reservedKeys) {
         const key = reservedKeys[i];
         await expect(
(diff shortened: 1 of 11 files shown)


core/__tests__/models/property/property.ts
```diff
@@ -164,7 +164,7 @@ describe("models/property", () => {
     });
 
     test("keys cannot be from the reserved list of keys", async () => {
-      const reservedKeys = ["grouparooId", "grouparooCreatedAt"];
+      const reservedKeys = ["grouparooId", "grouparooCreatedAt", "_meta"];
```
bleonard, line 167: I can't quite figure out what `_meta` is for. Is this new?

evantahler: That's been there for a long time - check out https://github.com/grouparoo/grouparoo/blob/21ce6fc83cca60911b87b6d5d4824be43180e959/core/src/tasks/record/export.ts#L67-L80.  We use this to pass atypical information though an Import back out to the next time it's exported.  For example, if a Record is no longer supposed to be sent to a Destination (maybe it was removed from a Group), `import.data._meta.destinationId` would tell us that there should be one more export made to that destination

bleonard (approved):

PR #2570, opened 2021-11-23
Config & Config UI for App Refresh Query

teallarson:
## Change description
- config files now use the structure: 
```
  "refresh" : {
    "query": "...",
    "recurringFrequency": 60000
  }
```
- Config UI fixes (don't show the currently saved values or "run" button, query gets saved to file correctly now)
- Fixed UI to default to "Edit" state upon creating a new App Refresh Query.  Also fixed issue with React intermittently locking the object for "undefined" values for `appRefreshQuery.locked` and `app.locked`
- `appRefreshQuery:run` task is enqueued immediately if the query is updated on the model

New Config UI for App Refresh Query:
Screen Shot 2021-11-24 at 10 46 41 AM

### Development

### Impact

### Code review


core/__tests__/models/appRefreshQuery.ts
```diff
@@ -78,6 +77,28 @@ describe("appRefreshQuery", () => {
...
+    test("an updated query on a ready appRefreshQuery enqueues a run", async () => {
+      await api.resque.queue.connection.redis.flushdb();
+
+      const appRefreshQuery = new AppRefreshQuery({
+        appId: app.id,
+        refreshQuery: "SELECT MAX(updated_at) FROM users;",
+        state: "ready",
+      });
+      await appRefreshQuery.save();
+
```
evantahler, line 87: I think here there should be check that there wasn't a task enqueued the first time when the  AppRefreshQuery was created.

core/src/models/App.ts
```diff
@@ -261,6 +261,10 @@ export class App extends LoggedModel<App> {
 
     const options = await this.getOptions(false);
 
+    const appRefreshQuery: AppRefreshQuery = await AppRefreshQuery.findOne({
+      where: { appId: this.id },
+    });
```
evantahler: This could be 

```suggestion
const appRefreshQuery = await this.$get('appRefreshQuery')
```

ui/ui-components/pages/app/[id]/refresh.tsx
```diff
@@ -45,7 +46,8 @@ export default function Page(props) {
   );
   const { id } = router.query;
   const { schedules, runs, sources } = props;
-  const disabled = app?.locked !== null || appRefreshQuery?.locked !== null;
+  const disabled =
+    app?.locked?.length > 1 || appRefreshQuery?.locked?.length > 1;
```
Comment (evantahler) [inline ui/ui-components/pages/app/[id]/refresh.tsx]: Was this part of the code still causing problems?  Was there still a `undefined` somewhere?

Comment (evantahler) [reply ui/ui-components/pages/app/[id]/refresh.tsx]: I'd checkout this part of the code for maying making the thing undefined:

```
    setAppRefreshQuery({
      ...appRefreshQuery,
      [event.target.id]: event.target.value,
```

Comment (teallarson) [reply ui/ui-components/pages/app/[id]/refresh.tsx]: yep. I'll look there.

core/__tests__/models/appRefreshQuery.ts
teallarson: I'm not sure why we wouldn't want it to enqueue then if the state is ready and there's a query.

But I do think we should flush redis before the update in that case.

ui/ui-components/pages/app/[id]/refresh.tsx
```diff
@@ -61,10 +63,10 @@ export default function Page(props) {
 
     if (response?.appRefreshQuery) {
       setAppRefreshQuery(response.appRefreshQuery);
```
Comment (teallarson) [inline ui/ui-components/pages/app/[id]/refresh.tsx:67]: @evantahler RE: `locked=undefined`...it doesn't even get as far as the place you mentioned (in the `edit` function).  If I `console.log()` the response here as well as what it sets the `appRefreshQuery` variable as, they both also have `locked = undefined`

Comment (evantahler) [reply ui/ui-components/pages/app/[id]/refresh.tsx:67]: Can you check in the `getInitialProps` - to make sure we are starting in a good place?

Comment (teallarson) [reply ui/ui-components/pages/app/[id]/refresh.tsx:67]: yep it's reading as `null` there

Comment (teallarson) [reply ui/ui-components/pages/app/[id]/refresh.tsx:67]: I paired with @krishnaglick  on it and couldn't get it to fail for some reason.  We did update `disabled` to be a `useMemo`.  Seems to be working great.

core/src/models/App.ts
```diff
@@ -269,6 +271,12 @@ export class App extends LoggedModel<App> {
       name,
       type,
       options,
+      refresh: appRefreshQuery
+        ? {
+            query: appRefreshQuery.refreshQuery,
+            recurringFrequency: appRefreshQuery.recurringFrequency,
```
krishnaglick, line 277: Is there more on `appRefreshQuery` than these two fields? Just curious.

ui/ui-components/pages/app/[id]/refresh.tsx
```diff
@@ -60,11 +64,11 @@ export default function Page(props) {
     );
 
     if (response?.appRefreshQuery) {
-      setAppRefreshQuery(response.appRefreshQuery);
-
+      setAppRefreshQuery({ ...response.appRefreshQuery });
```
Comment (krishnaglick) [inline ui/ui-components/pages/app/[id]/refresh.tsx]: Not sure this spread is required.

ui/ui-components/pages/app/[id]/refresh.tsx
```diff
@@ -113,11 +118,17 @@ export default function Page(props) {
         );
         if (response?.valueUpdated == true) {
           successHandler.set({
-            message: `App Refresh Query updated. Returned ${response.appRefreshQuery.value}. Enqueueing Schedules.`,
+            message: `App Refresh Query updated. Returned ${
+              response.appRefreshQuery.value
+            }.${grouparooUiEdition() !== "config" ?? " Enqueueing Schedules"}`,
```
Comment (krishnaglick) [inline ui/ui-components/pages/app/[id]/refresh.tsx:123]: This may have two spaces in the text if we're not in config mode looks like.

ui/ui-components/pages/app/[id]/refresh.tsx
```diff
@@ -337,26 +347,28 @@ export default function Page(props) {
...
+                      {testResult.success && !testResult.error && ranTest ? (
+                        <Alert variant="success" className="text-break">
+                          <strong>Test Passed. </strong>Sample Value ={" "}
+                          {testResult.message}
+                        </Alert>
+                      ) : ranTest ? (
+                        <Alert variant="warning">
+                          <strong>Test Failed</strong> {testResult.error}
+                        </Alert>
+                      ) : null}
```
Comment (krishnaglick) [inline ui/ui-components/pages/app/[id]/refresh.tsx:361]: Probably not necessary, but if this gets any bigger would be good as its own component.

core/src/models/App.ts
teallarson: On the model, yes.  On the configObject (what this is building), no.

We do some similar things for schedules and sources [here](https://github.com/grouparoo/grouparoo/blob/0fb5e1761f691671f8d70fe0e8b4698b2de0e15c/core/src/models/Source.ts#L350).

ui/ui-components/pages/app/[id]/refresh.tsx
Comment (teallarson) [reply ui/ui-components/pages/app/[id]/refresh.tsx]: Ah, ty!  This was trying to troubleshoot the mysterious null problem.

core/src/models/App.ts
krishnaglick: Could just do `refresh: appRefreshQuery` then. If it's not there it should be undefined. If it returns null, can do: `refresh: appRefreshQuery ?? undefined`

ui/ui-components/pages/app/[id]/refresh.tsx
Comment (teallarson) [reply ui/ui-components/pages/app/[id]/refresh.tsx:123]: Can you help me understand that?  My intention was that config mode would return something like
`App Refresh Query updated. Returned {date: 1234567890}.`
Whereas non-config modes would return 
`App Refresh Query updated. Returned {date: 1234567890}. Enqueueing schedules.` (or not enqueueing schedules).

ui/ui-components/pages/app/[id]/refresh.tsx
Comment (teallarson) [reply ui/ui-components/pages/app/[id]/refresh.tsx:361]: Sounds right! This is currently used here and on `/app/[id]/edit` (though the implementation there is slightly different -- the result never disappears there), so if it gets bigger or gets used a third place I agree.

core/src/models/App.ts
teallarson: Ah, `appRefreshQuery` here has all of the fields from the model here (appId, timestamps, value, etc.), that's why I spelled out the two that should be in the App config object being built.

krishnaglick: Ah, alright! All good then :)

teallarson:
@evantahler one question on this PR... I'm not sure that there are customers using an app refresh query in production, but if there were, this would be breaking I believe because of the rename.  How do we do migrations for code config?

evantahler:
I think that for expediency's sake, it's a safe assumption that folks are not yet using code-config (as we haven't announced it yet).  I wouldn't focus on a code-config migration at this time.

@bleonard - does that work for you?

krishnaglick (approved):

evantahler (approved):

PR #2572, opened 2021-11-23
Fix property selection in source mapping page to be based on which source has the primary key property

edmundito:
## Change description
This change fixes the property selection in the source mapping page so that:

- When there is no primary key property defined in the model, the user can add a unique property on the mapping page
- When the source owns the primary key property, the user can update the mapping to properties only available in that source
- When the source does not own the primary key property, mapping only shows the properties from other sources

Bonus: It updates the "Unique" checkmark in the mapping table with ✅  match the rest of the app
image

### Development

### Impact

### Code review


evantahler (approved):
👍 Good filtering (and memo!)

PR #2573, opened 2021-11-24
Remove `Property`.`identifying` and validate config object keys

evantahler:
Now that we have a single Primary Property per model, this is what we should show in the UI to identity Records.  We no longer need the optional, UI-Only concept of `Property`.`identifying`.   Identifying also didn't quite make sense in a multi-model world as it was currently implemented. 

* Removed Setting UI to change the Identifying Property
* Removed the column
* Removed from config files & show a deprecation warning
* Removed code usage

Screen Shot 2021-11-24 at 12 22 12 PM

This PR also:
* Removes the Record ID from the list of records
* Ensures that there are no errant properties/attributes on grouparoo config files
* Removes the previously deprecated `source.bootstrapProperty` config option

### Development

### Impact

### Code review


core/src/modules/configLoaders/property.ts
```diff
@@ -55,8 +56,8 @@ export async function loadProperty(
     await property.setFilters(configObject.filters, externallyValidate);
   }
 
-  if (configObject.identifying === true) {
-    await property.makeIdentifying();
+  if (configObject["identifying"] !== null) {
```
edmundito:
```suggestion
  if (!!configObject["identifying"]) {
```

This change would ensure that the value is nil (either `undefined` or `null`)

ui/ui-components/components/record/List.tsx
```diff
@@ -111,15 +111,10 @@ export default function RecordsList(props) {
...
-  )[0];
-
+  const primaryProperty = properties.find((property) => property.isPrimaryKey);
   const uniqueRecordPropertyKeys = properties
-    .filter((rule) =>
-      identifyingRecordProperty
-        ? rule.id !== identifyingRecordProperty.id
-        : rule
+    .filter((property) =>
+      primaryProperty ? property.id !== primaryProperty.id : true
```
edmundito:
```suggestion
      !primaryProperty || propery.id !== primaryProperty.id
```

ui/ui-components/pages/settings/[tab].tsx
```diff
@@ -7,13 +7,14 @@ import { useRouter } from "next/router";
...
 import SettingCard from "../../components/settings/SettingCard";
 import ImportAndUpdateAllRecords from "../../components/settings/ImportAndUpdate";
-import IdentifyingProperty from "../../components/settings/IdentifyingProperty";
 import ResetCluster from "../../components/settings/ResetCluster";
 import ResetData from "../../components/settings/ResetData";
 import ResetCache from "../../components/settings/ResetCache";
 import { ErrorHandler } from "../../utils/errorHandler";
 import { SuccessHandler } from "../../utils/successHandler";
 
+const settingsWhichTriggerInterfaceReload = [["core", "cluster-name"]];
```
Comment (edmundito) [inline ui/ui-components/pages/settings/[tab].tsx:16]: I typically wrap these in an `Object.freeze()` so that they're read only and use `ALL_CAPS` to identify as a constant.

edmundito (approved):
Looks great and good to merge. I left some trivial notes.

ui/ui-components/pages/settings/[tab].tsx
Comment (evantahler) [reply ui/ui-components/pages/settings/[tab].tsx:16]: I saw that! Can you share more about the benefit?  Does it help with memory usage or something like that?

evantahler:
@edmundito pinging you again because this PR got a little bigger, and I ended up solving https://www.pivotaltracker.com/story/show/180404106 at the same time 🤷

core/src/modules/configLoaders/destination.ts
```diff
@@ -21,7 +21,11 @@ export async function loadDestination(
 
   const app: App = await getParentByName(App, configObject.appId);
 
-  validateConfigObjectKeys(Destination, configObject);
+  validateConfigObjectKeys(Destination, configObject, [
+    "options",
+    "mapping",
+    "destinationGroupMemberships",
+  ]);
```
evantahler, line 28: We now tightly validate the properties from config objects.  Either they are top-level properties we know about (name, type, etc) or this explicit list of extra objects

core/src/modules/configLoaders/property.ts
```diff
@@ -56,7 +56,10 @@ export async function loadProperty(
     await property.setFilters(configObject.filters, externallyValidate);
   }
 
-  if (configObject["identifying"] !== null) {
+  if (
+    configObject["identifying"] !== null &&
+    configObject["identifying"] !== undefined
+  ) {
```
edmundito: Long term, it might be good for us to introduce some sort of utilities repo that can help us with check for these things. for example in this case: `isNil`

edmundito (approved):

ui/ui-components/pages/settings/[tab].tsx
Comment (edmundito) [reply ui/ui-components/pages/settings/[tab].tsx:16]: The main benefit is type safety. The TypeScript compiler will prevent compilation but even then the javascript will throw an error when you try to add to an object or push to an array.

You could also force it at compile time by using:
```ts
const settingsWhichTriggerInterfaceReload: readonly string[][] = ....
```

PR #2574, opened 2021-11-24
Throw error when mapping to source's own property while primary key belongs to another source

edmundito:
This change ensures that it's not possible to set a mapping to a property in the source when another source owns the primary key by throwing an error.

When running `grouparoo config` where two different sources (`PG users` and `PG purchases`) in the same model point to their own properties in the mapping, we see an error such as:

```
warning: [ config ] error with Source `PG users` (pg_users): 'PG users' cannot map 'id' to own
Property 'userId'. 'PG users' must map to a Property from primary Source 'Pg purchases'. 
```

Related: #2572 

## Change description

Description here

### Development

### Impact

In the case that a customer had already setup a configuration where two sources were mapping to their own keys, they would need to update the mapping in the config files.

### Code review

### DIFF
--- a/core/__tests__/models/property/property.ts
+++ b/core/__tests__/models/property/property.ts
@@ -634,7 +634,6 @@ describe("models/property", () => {
   });
 
   describe("with plugin", () => {
-    // let model: GrouparooModel;
     let app: App;
     let source: Source;
     let queryCounter = 0;
--- a/core/__tests__/models/property/property.ts
+++ b/core/__tests__/models/property/property.ts
@@ -634,9 +634,18 @@ describe("models/property", () => {
   });
 
   describe("with plugin", () => {
+    // let model: GrouparooModel;
     let app: App;
     let source: Source;
     let queryCounter = 0;
+
+    const propertiesToMoveKeys = Object.freeze([
+      "userId",
+      "firstName",
+      "lastName",
+    ]);
+    const prevSourceIds: string[] = [];
+
     beforeAll(async () => {
       plugin.registerPlugin({
         name: "test-plugin",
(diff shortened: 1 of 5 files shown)


core/src/modules/mappingHelper.ts
```diff
@@ -97,6 +115,12 @@ export namespace MappingHelper {
         );
       }
 
+      if (otherSourcePrimaryKey && property.sourceId === instance.id) {
+        throw new Error(
+          `'${instance.name}' cannot map '${remoteKey}' to own Property '${key}'. '${source.name}' must map to a Property from primary Source '${otherSourcePrimaryKey.source.name}'.`
+        );
+      }
```
edmundito: I'm still thinking about the wording. I think in the case this happen the primary source is ambiguous. For example if the Users and Purchases source point to their own properties, it's not clear which is the primary source. I'm open for feedback.

Maybe it should just say while {sourceB} is mapping to its own property {propertyB} ?

core/__tests__/models/property/property.ts
```diff
@@ -634,9 +634,18 @@ describe("models/property", () => {
   });
 
   describe("with plugin", () => {
+    // let model: GrouparooModel;
```
evantahler: can remove

core/__tests__/models/source/source.ts
```diff
@@ -740,6 +740,12 @@ describe("models/source", () => {
 
       await source.setMapping({ userId: "userId" });
     });
+
+    test("will throw error when mapping to own property and primary key is owned by other source", async () => {
+      await expect(source.setMapping({ email: "myEmail" })).rejects.toThrow(
+        /cannot map 'email' to own Property 'myEmail'/
+      );
+    });
```
evantahler, line 748: This is the new test - the other changes are cleanup / clarifications based on this new behavior.  Initially I would have guessed this change would have required a more complex series of tests... but... it doesn't! We've got other tests confirming that the first property (bootstrap) still works like expect

core/src/modules/mappingHelper.ts
```diff
@@ -74,6 +73,25 @@ export namespace MappingHelper {
...
+                modelId: instance.modelId,
+              },
+            },
+            where: {
+              isPrimaryKey: true,
+              sourceId: {
+                [Op.ne]: instance.id,
+              },
+            },
+          })
```
evantahler, line 92: 👍 great single query to find this!

core/src/modules/mappingHelper.ts
evantahler: What about something like

```ts
`${otherSourcePrimaryKey.soruce.name} is the primary source for the ${instance.model.name} Model - ${instance.name} requires a mapping back to ${otherSourcePrimaryKey.soruce.name}`
```

> users-table is the primary source for the profiles - purchases-table requires a mapping back to users-table`

evantahler (approved):

PR #2576, opened 2021-11-26
Salesforce float recordMatchField issue

pauloouriques:
## Change description

This PR inserts the code responsible for parsing to Number the Salesforce foreignKeys if the recordMatchField was float, integer, or double.

### Development

### Impact

### Code review


plugins/@grouparoo/salesforce/src/lib/export/salesforceExporter.ts
```diff
@@ -338,6 +342,25 @@ function buildUserPayload(
   return { row, referenceData };
 }
 
+function getForeignKeyType(users: BatchExport[], recordMatchField: string) {
+  if (users.length > 0) {
```
bleonard: this could be `null` value. It's probably better to get it from the (cache, if not already) field definition itself: the same one used in `destinationMappings`.

plugins/@grouparoo/salesforce/src/lib/export/mapping.ts
```diff
@@ -86,6 +86,7 @@ const supportedTypeMap = {
   id: "string",
   picklist: "string",
   currency: "float",
+  number: "float",
```
bleonard, line 93: is this a real type we missed before?

plugins/@grouparoo/salesforce/src/lib/export/salesforceExporter.ts
```diff
@@ -338,6 +348,40 @@ function buildUserPayload(
...
+  client: any,
+  cacheData: SalesforceCacheData,
+  recordObject: string,
+  recordMatchField: string
+) {
+  const recordInfo = await describeObject(
+    client,
+    cacheData,
+    recordObject,
+    true
```
bleonard: I think you want `false` here so it actually does the caching.
`passthru?: boolean; // send true to not read or write or lock`

bleonard: It does `true` in the other spot because the mapping page should always be up to date. But this will get called a lot, so it should use the cache.

bleonard (changes requested):

plugins/@grouparoo/salesforce/src/lib/export/mapping.ts
pauloouriques: yes, I saw it here: 
image

bleonard: I think that might be "integer" in the code. Did you see it for real in the types returned? I guess it doens't hurt anyone if it's extra

bleonard (approved):

PR #2596, opened 2021-12-01
Add `location` as App option for BigQuery

teallarson:
## Change description

Previously, because BigQuery defaults to US, we could only support US datasets.  This PR adds an optional `appOption` to BigQuery plugin for the region or multi-region code of a BigQuery dataset, passes that as part of the connection.  This is only required for users with a dataset outside the United States.

Screen Shot 2021-12-01 at 10 42 35 AM

### Development

### Impact

### Code review

### DIFF
--- a/plugins/@grouparoo/bigquery/public/templates/app/{{{id}}}.js.template
+++ b/plugins/@grouparoo/bigquery/public/templates/app/{{{id}}}.js.template
@@ -14,6 +14,7 @@ exports.default = async function buildConfig() {
         dataset: "...", // Default dataset ID to use for sources - e.g. `dataset: "dataset"`
         client_email: "xxx@xxx.iam.gserviceaccount.com", // Client Email Address
         private_key: "-----BEGIN PRIVATE KEY-----\n...", // Private key of service account
+        //location: "..." // Optional: If your dataset is located outside of the US, you'll need to provide the region or multi-region location: https://cloud.google.com/bigquery/docs/locations 
       }
     },
   ];
--- a/plugins/@grouparoo/bigquery/src/initializers/plugin.ts
+++ b/plugins/@grouparoo/bigquery/src/initializers/plugin.ts
@@ -85,7 +85,7 @@ export class Plugins extends Initializer {
               displayName: "Location",
               required: false,
               description:
-                "Region or multi-region code of your dataset.  Required for datasets outside the US.",
+                'Region or multi-region code of your dataset. ("US", "EU", "asia-east1", etc.) Required for datasets outside the US.',
               placeholder: "e.g. EU",
             },
           ],
--- a/plugins/@grouparoo/bigquery/src/initializers/plugin.ts
+++ b/plugins/@grouparoo/bigquery/src/initializers/plugin.ts
@@ -80,6 +80,14 @@ export class Plugins extends Initializer {
               description: "Private key of service account.",
               placeholder: "e.g. -----BEGIN PRIVATE KEY-----\nMII ...",
             },
+            {
+              key: "location",
+              displayName: "Location",
+              required: false,
+              description:
+                "Region or multi-region code of your dataset.  Required for datasets outside the US.",
+              placeholder: "e.g. EU",
+            },
           ],
           methods: { test, connect, disconnect, appQuery },
         },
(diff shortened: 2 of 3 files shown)


teallarson:
Wasn't sure on testing here.  It looks like App Options aren't widely tested.  In the UI, if someone puts in a bogus region code, the worst that happens is it returns "0 tables found", same as if they enter a bogus dataset name.

plugins/@grouparoo/bigquery/src/initializers/plugin.ts
```diff
@@ -80,6 +80,14 @@ export class Plugins extends Initializer {
...
               placeholder: "e.g. -----BEGIN PRIVATE KEY-----\nMII ...",
             },
+            {
+              key: "location",
+              displayName: "Location",
+              required: false,
+              description:
+                "Region or multi-region code of your dataset.  Required for datasets outside the US.",
+              placeholder: "e.g. EU",
+            },
```
evantahler: I would propose making this required, but having a default value: 

```suggestion
            {
              key: "location",
              displayName: "Location",
              required: true,
              description:
                "Region or multi-region code of your dataset",
              defaultValue: "US",
            },
````

This is how we handle things like the postgres port - https://github.com/grouparoo/grouparoo/blob/3e4bfda676026b34ec6a4285abb519db596f67ba/plugins/%40grouparoo/postgres/src/initializers/plugin.ts#L76-L80

File: plugins/@grouparoo/bigquery/src/lib/connect.ts [hunk 2]
```diff
@@ -6,6 +6,9 @@ export const connect: ConnectPluginAppMethod = async ({ appOptions }) => {
   const projectId = appOptions.project_id?.toString() || "";
   const dataset = appOptions.dataset?.toString() || "";
   const client_email = appOptions.client_email?.toString() || "";
+  const datasetOptions = appOptions.location
+    ? { location: appOptions.location?.toString() }
+    : {};
```
evantahler, line 11: If the location is now required, this could become:

```suggestion
  const datasetOptions = { location: appOptions.location.toString() };
```

plugins/@grouparoo/bigquery/public/templates/app/{{{id}}}.js.template
```diff
@@ -14,6 +14,7 @@ exports.default = async function buildConfig() {
         dataset: "...", // Default dataset ID to use for sources - e.g. `dataset: "dataset"`
         client_email: "xxx@xxx.iam.gserviceaccount.com", // Client Email Address
         private_key: "-----BEGIN PRIVATE KEY-----\n...", // Private key of service account
+        //location: "..." // Optional: If your dataset is located outside of the US, you'll need to provide the region or multi-region location: https://cloud.google.com/bigquery/docs/locations
```
evantahler, line 17: Becomes required, so not commented out

plugins/@grouparoo/bigquery/src/initializers/plugin.ts
teallarson: If we require location, having the default value will negate migration issues, correct?

evantahler: I *think* so.  I would try it out to be sure

teallarson:
We decided to not move forward with making the option required with a default as it caused unwanted side effects and BigQuery uses US as a default already.  I reset the commit where we tried it out.

plugins/@grouparoo/bigquery/src/initializers/plugin.ts
```diff
@@ -80,6 +80,14 @@ export class Plugins extends Initializer {
               description: "Private key of service account.",
               placeholder: "e.g. -----BEGIN PRIVATE KEY-----\nMII ...",
             },
+            {
+              key: "location",
+              displayName: "Location",
+              required: false,
+              description:
+                "Region or multi-region code of your dataset.  Required for datasets outside the US.",
```
bleonard: maybe put some expected values in this description (Us, EU, et) to help people.

tealjulia: in addition to the placeholder?

evantahler (approved):

plugins/@grouparoo/bigquery/src/lib/connect.ts
98aho47fati: Issuse

PR #2616, opened 2021-12-01
Add Sample Record Card to Model Overview Page

edmundito:
## Change description

This change adds the sample record card to the model overview UI. It includes all the different states (disabled, empty, with record), functionality to define a sample record and remember it in the local browser, show groups associated with the record.

Screen Shot 2021-12-06 at 1 38 07 PM

Screen Shot 2021-12-06 at 1 32 03 PM

Screen Shot 2021-12-06 at 1 32 19 PM

Screen Shot 2021-12-06 at 2 01 05 PM

Screen Shot 2021-12-06 at 2 01 13 PM

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-community/next-env.d.ts
+++ b/ui/ui-community/next-env.d.ts
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
--- a/ui/ui-components/components/record/AddSampleRecordForm.tsx
+++ b/ui/ui-components/components/record/AddSampleRecordForm.tsx
@@ -1,0 +1,77 @@
+import { useMemo, useState } from "react";
+import { Form } from "react-bootstrap";
+import { useForm } from "react-hook-form";
+import { useGrouparooModelContext } from "../../contexts/grouparooModel";
+import { ApiHook } from "../../hooks/useApi";
+import { Actions, Models } from "../../utils/apiData";
+import LoadingButton from "../LoadingButton";
+
+interface Props {
+  properties: Models.PropertyType[];
+  onSubmitComplete: (record?: any) => void;
+  execApi: ApiHook["execApi"];
+}
+
+const AddSampleRecordForm: React.FC<Props> = ({
+  execApi,
+  onSubmitComplete,
+  properties,
+}) => {
+  const model = useGrouparooModelContext();
+  const { handleSubmit, register } = useForm();
+  const [submitting, setSubmitting] = useState(false);
+
+  const propertiesWithPrimaryKey = useMemo(
+    () =>
+      properties.filter((property) => property.unique && property.isPrimaryKey),
+    [properties]
+  );
+
+  const onSubmit: Parameters<typeof handleSubmit>[0] = async (data) => {
+    setSubmitting(true);
+    const response = await execApi<Actions.RecordCreate>("post", `/record`, {
+      modelId: model.id,
+      properties: { [data.uniqueProperty]: data.value },
+    });
+    setSubmitting(false);
+    onSubmitComplete(response?.record);
+  };
+
+  return (
+    <Form id="form" onSubmit={handleSubmit(onSubmit)}>
+      <Form.Group>
+        <Form.Label>Unique Property</Form.Label>
+        <Form.Control
+          name="uniqueProperty"
+          as="select"
+          ref={register}
+          disabled={submitting}
+        >
+          {propertiesWithPrimaryKey.map((rule) => (
+            <option key={`rule-${rule.key}`}>{rule.key}</option>
+          ))}
+        </Form.Control>
+      </Form.Group>
+      <Form.Group>
+        <Form.Label>Value</Form.Label>
+        <Form.Control
+          autoFocus
+          required
+          type="text"
+          name="value"
+          ref={register}
+          disabled={submitting}
+        />
+        <Form.Control.Feedback type="invalid">
+          Value is required
+        </Form.Control.Feedback>
+      </Form.Group>
+
+      <LoadingButton variant="primary" type="submit" disabled={submitting}>
+        Submit
+      </LoadingButton>
+    </Form>
+  );
+};
+
+export default AddSampleRecordForm;
--- a/ui/ui-components/components/record/AddSampleRecordForm.tsx
+++ b/ui/ui-components/components/record/AddSampleRecordForm.tsx
@@ -6,6 +6,16 @@ import { ApiHook } from "../../hooks/useApi";
 import { Actions, Models } from "../../utils/apiData";
 import LoadingButton from "../LoadingButton";
 
+const getInputType = (type?: Models.PropertyType["type"]): string => {
+  switch (type) {
+    case "float":
+    case "integer":
+      return "number";
+    default:
+      return "text";
+  }
+};
+
 interface Props {
   properties: Models.PropertyType[];
   onSubmitComplete: (record?: Models.GrouparooRecordType) => void;
@@ -27,6 +37,23 @@ const AddSampleRecordForm: React.FC<Props> = ({
     [properties]
   );
 
+  const [selectedUniquePropertyValue, setSelectedUniquePropertyValue] =
+    useState(() => {
+      return propertiesWithPrimaryKey?.[0]?.key || "";
+    });
+
+  const selectedUniqueProperty = useMemo(() => {
+    return propertiesWithPrimaryKey?.find(
+      ({ key }) => key === selectedUniquePropertyValue
+    );
+  }, [selectedUniquePropertyValue]);
+
+  const onSelectUniqueProperty: React.ChangeEventHandler<HTMLInputElement> = (
+    event
+  ) => {
+    setSelectedUniquePropertyValue(event.target.value);
+  };
+
   const onSubmit: Parameters<typeof handleSubmit>[0] = async (data) => {
     setSubmitting(true);
     const response = await execApi<Actions.RecordCreate>("post", `/record`, {
@@ -46,9 +73,13 @@ const AddSampleRecordForm: React.FC<Props> = ({
           as="select"
           ref={register}
           disabled={submitting}
+          value={selectedUniquePropertyValue}
+          onChange={onSelectUniqueProperty}
         >
           {propertiesWithPrimaryKey.map((rule) => (
-            <option key={`rule-${rule.key}`}>{rule.key}</option>
+            <option key={`rule-${rule.key}`} value={rule.key}>
+              {rule.key}
+            </option>
           ))}
         </Form.Control>
       </Form.Group>
@@ -57,7 +88,7 @@ const AddSampleRecordForm: React.FC<Props> = ({
         <Form.Control
           autoFocus
           required
-          type="text"
+          type={getInputType(selectedUniqueProperty?.type)}
           name="value"
           ref={register}
           disabled={submitting}
--- a/ui/ui-components/components/record/AddSampleRecordForm.tsx
+++ b/ui/ui-components/components/record/AddSampleRecordForm.tsx
@@ -8,7 +8,7 @@ import LoadingButton from "../LoadingButton";
 
 interface Props {
   properties: Models.PropertyType[];
-  onSubmitComplete: (record?: any) => void;
+  onSubmitComplete: (record?: Models.GrouparooRecordType) => void;
   execApi: ApiHook["execApi"];
 }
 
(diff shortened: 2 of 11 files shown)


ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,231 @@
...
+  const [importing, setImporting] = useState(false);
+  const [addingRecord, setAddingRecord] = useState(false);
+  const [recordId, setRecordId] = useState(() => {
+    if (typeof window !== "undefined") {
+      return window.localStorage.getItem(`sampleRecord:${model.id}`);
+    }
+    return undefined;
+  });
+
+  const { data: record, mutate: mutateRecord } = useSWR(
```
evantahler: Let's talk more about SWR! I don't think I see implicitly what it gains us here.  Our `Client` already has some caching mechanisms built in.  perhaps we could replace those with SWR?  The rough API was that all the components could call GET whenever they wanted, but the client would be smart and returned cached data if it already had it based on timing (I think ~5s).  All other HTTP verbs were not cached (POST, DELETE, etc) 

The cache area in the Client is here https://github.com/grouparoo/grouparoo/blob/main/ui/ui-components/client/client.ts#L180-L190

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,231 @@
...
+        modelId: model.id,
+        recordId,
+      }).then((data) => data?.records?.[0]);
+    }
+  );
+
+  useEffect(() => {
+    if (record) {
+      window.localStorage.setItem(`sampleRecord:${model.id}`, record.id);
+    }
```
evantahler: 👍 local storage seems like the right way to store this.  I imagine setters and getters for sampleRecords will be abstracted once this is shared on more pages

ui/ui-components/components/record/SampleRecordCard.tsx
edmundito: I didn't realize that the client had a caching mechanism. I will roll back the changes for now but I can see how SWR could simplify the client while at the same time providing some extra niceties. But if we were to adopt it we should also revisit the client.

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,340 @@
...
+      setCachedSampleRecordId(model.id, record.id);
+    },
+    [model]
+  );
+
+  const loadRecord = useCallback(async () => {
+    setLoading(true);
+
+    let record: Actions.RecordView["record"];
+    let groups: Actions.RecordView["groups"];
```
edmundito: The types for these is hard to resolve and I'd like to figure out a better way to do it. I need to revisit this to see if it's fixable.

evantahler: Those should be the same types as `Models.GroupType` if that's easier

ui/ui-components/components/record/AddSampleRecordForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <option key={`rule-${rule.key}`}>{rule.key}</option>
+          ))}
+        </Form.Control>
+      </Form.Group>
+      <Form.Group>
+        <Form.Label>Value</Form.Label>
+        <Form.Control
+          autoFocus
+          required
+          type="text"
```
edmundito: I need to improve the validation for this because I noticed that I can create really bad data. I could probably take a look at the property type itself and determine what type of validation should happen.

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+          if (randomRecord?.id) {
+            return execApi<Actions.RecordView>(
+              "get",
+              `/record/${randomRecord.id}`
+            );
+          }
+        }
+
+        return { record: undefined, groups: undefined };
```
edmundito: Create a new API for getting a sample record or keep this logic in the UI side?

edmundito:
Added destinations, fixed some validations, and improved how groups render, and fixed a few issues. This is ready for another review pass.

ui/ui-components/components/record/AddSampleRecordForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const propertiesWithPrimaryKey = useMemo(
+    () =>
+      properties.filter((property) => property.unique && property.isPrimaryKey),
+    [properties]
+  );
+
+  const [selectedUniquePropertyValue, setSelectedUniquePropertyValue] =
+    useState(() => {
+      return propertiesWithPrimaryKey?.[0]?.key || "";
+    });
```
krishnaglick: Why not just
```suggestion
    useState(propertiesWithPrimaryKey?.[0]?.key || "");
```

ui/ui-components/components/record/AddSampleRecordForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+    return propertiesWithPrimaryKey?.find(
+      ({ key }) => key === selectedUniquePropertyValue
+    );
+  }, [selectedUniquePropertyValue]);
+
+  const onSelectUniqueProperty: React.ChangeEventHandler<HTMLInputElement> = (
+    event
+  ) => {
+    setSelectedUniquePropertyValue(event.target.value);
+  };
```
krishnaglick, line 53: I think this is a case for useCallback

ui/ui-components/components/record/AddSampleRecordForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const onSubmit: Parameters<typeof handleSubmit>[0] = async (data) => {
+    setSubmitting(true);
+    const response = await execApi<Actions.RecordCreate>("post", `/record`, {
+      modelId: model.id,
+      properties: { [data.uniqueProperty]: data.value },
+    });
+    setSubmitting(false);
+    onSubmitComplete(response?.record);
+  };
```
krishnaglick: useCallback as well

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,357 @@
...
+
+const isConfigUI = grouparooUiEdition() === "config";
+
+const getCachedSampleRecordId = (modelId: string): string => {
+  return globalThis.localStorage?.getItem(`sampleRecord:${modelId}`);
+};
+
+const setCachedSampleRecordId = (modelId: string, recordId: string): void => {
+  recordId &&
+    globalThis.localStorage?.setItem(`sampleRecord:${modelId}`, recordId);
```
krishnaglick: As a more stylistic decision, while I get javascript lets us do `thing && action` maybe we should enforce
```
if (thing) {
    action();
}
```

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,357 @@
...
+};
+
+const clearCachedSampleRecordId = (modelId: string): void => {
```
krishnaglick, line 37: Not sure how helpful the void return types are _here_. I know this is something we want to discuss!

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,357 @@
...
+  const [loading, setLoading] = useState(false);
+  const [importing, setImporting] = useState(false);
+  const [addingRecord, setAddingRecord] = useState(false);
+  const [hasRecords, setHasRecords] = useState(true);
+  const [record, setRecord] = useState<Models.GrouparooRecordType>();
+  const [groups, setGroups] = useState<Models.GroupType[]>();
+  const [destinations, setDestinations] = useState<Models.DestinationType[]>();
+  const [recordId, setRecordId] = useState(() =>
+    getCachedSampleRecordId(model.id)
+  );
```
krishnaglick, line 58: Lazy initial state cool

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,357 @@
...
+      setGroups(response.groups);
+      successHandler.set({ message: "Import Complete!" });
+    } else {
+      loadRecord(); // we may have done a partial import
+    }
+    setImporting(false);
+  };
+
+  const actions = [
+    <LoadingButton
```
krishnaglick, line 169: React says these need `key`'s

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,357 @@
...
+        </p>
+      </Col>
+    </Row>
+  );
+
+  return (
+    <ManagedCard
+      title="Sample Record"
+      disabled={disabled}
+      actions={[
```
krishnaglick, line 316: React says these need `key`'s. Potentially a case for a `useMemo`

ui/ui-components/hooks/usePrevious.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useRef } from "react";
+
+export const usePrevious = <T>(value: T): T => {
+  const ref = useRef<T>();
+
+  useEffect(() => {
+    ref.current = value;
```
krishnaglick, line 7: Cool!

ui/ui-components/components/record/AddSampleRecordForm.tsx
edmundito: Not necessarily because this function does not really have any prop/hook dependencies.

ui/ui-components/components/record/SampleRecordCard.tsx
edmundito: This is a really simple funciton but in general it guarantees that the implementation complies with the definition of the function instead of leaving it to the implementation to define the return type.

ui/ui-components/components/record/SampleRecordCard.tsx
edmundito: This is just an array of elements that do eventually get assigned a key by the component that renders them. No errors in the console for this.

ui/ui-community/next-env.d.ts
```diff
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
```
edmundito, line 2: Hmm, this changed with the latest rebase. I wonder why 🤔

ui/ui-components/components/record/AddSampleRecordForm.tsx
krishnaglick: Based on my understanding of `useCallback`, if you pass it an empty array for deps it will get created once and then never again. So it takes up a little more memory, but requires less time during rendering.

I guess the call would be based on how often this component re-renders.

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,358 @@
...
+  const [hasRecords, setHasRecords] = useState(true);
+  const [record, setRecord] = useState<Models.GrouparooRecordType>();
+  const [groups, setGroups] = useState<Models.GroupType[]>();
+  const [destinations, setDestinations] = useState<Models.DestinationType[]>();
+  const [recordId, setRecordId] = useState(() =>
+    getCachedSampleRecordId(model.id)
+  );
+
+  const saveRecord = useCallback(
+    (record, groups = undefined, destinations = undefined) => {
```
krishnaglick, line 61: These could use some types!

ui/ui-community/next-env.d.ts
evantahler: 🤷 Next changes these things automatically from time to time...

evantahler (approved):
This is 👍 from me!

We are having a conversation in this PR about the the appropriateness useMemo and useCallback.  For the Grouparoo user - a desktop-only user with a powerful enough work computer - we don't need to write /perfect/ react code.  We are more concerned with feature velocity right now.  Let's keep having the conversation about the right places to useMemo/useCallback, but not prevent work from getting delivered.

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -0,0 +1,358 @@
...
+        }
+      ).then((response) => {
+        if (response.total === 0) {
+          setHasRecords(false);
+        } else {
+          const randomRecord = response?.records?.length
+            ? response.records[
+                Math.floor(Math.random() * response.records.length)
+              ]
+            : undefined;
```
evantahler, line 104: Calling out that for now the "random" button will re a random selection of the first 25 records

PR #2631, opened 2021-12-02
Salesforce destination split out

pauloouriques:
## Change description

This PR contains the most common usages of the Salesforce destination split out into separated destinations.

### Development

### Impact

### Code review


plugins/@grouparoo/salesforce/src/initializers/plugin.ts
```diff
@@ -74,93 +91,9 @@ export class Plugins extends Initializer {
...
-            },
-          ],
-          methods: {
-            exportRecords,
-            destinationOptions,
-            destinationMappingOptions,
-            exportArrayProperties,
-          },
-        },
+        accountsDestinationConnection,
```
bleonard, line 94: 👍

plugins/@grouparoo/salesforce/src/lib/export-account/exportRecords.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export const exportBatch: MyBatchMethod = async ({
+  appId,
+  appOptions,
+  destinationOptions,
+  syncOperations,
+  exports,
+}) => {
+  exports.map((currentExport) => {
```
bleonard: this should probably move to `exportSalesforceBatch` and removed from each.

bleonard: Then you can probably just do it all in the one method.

plugins/@grouparoo/salesforce/src/lib/export/model.ts
```diff
@@ -30,6 +37,19 @@ export function getSalesforceModel(
     recordReferenceMatchField: null,
   };
 
+  let destinationOptionsMapping = destinationOptions;
+  switch (destination) {
+    case SALES_FORCE_DESTINATIONS.object:
+      // explicit default
+      break;
+    case SALES_FORCE_DESTINATIONS.contact:
+      destinationOptionsMapping = getContactsMapping(destinationOptions);
```
bleonard: Instead of calling these here and sharing the defintion so much, what if we passed in the defaults to this method (and not the destination name)? That keeps the dependency chain less tangled.

bleonard: That would remove the methods below and do the merge in this one.

plugins/@grouparoo/salesforce/src/lib/export-account/options.ts
```diff
@@ -0,0 +1,60 @@
+import {
```
bleonard: This method is almost exactly the same as the `-contacts` (and `-objects`?) one. That's true for many of these files.
I feel like there is an oppurtunity have a meta concept where you pass what's important (model defaults, known special fields to a single method (or maybe one each) and get back all the methods that implement the destination. TSo the desitnation-specific code would all be in `connection`.

See how `app-templates` `buildConnection` is used here: 
https://github.com/grouparoo/grouparoo/blob/main/plugins/%40grouparoo/postgres/src/lib/table-import/connection.ts

```
export function getConnection() {
  return buildConnection({
    apps: ["postgres"],
    name: "postgres-import-table",
    displayName: "Postgres Table Import",
    description: "Import or update Records from a Postgres database table.",
    tableOptionDescription: "The table to scan",
    getSampleRows,
    getColumns,
    getTables,
    getChangedRows,
    getPropertyValue,
    getPropertyValues,
    getChangedRowCount,
  });
}
```

bleonard (commented):
Looking good. I think there's some tactics to shred more code and make it more declarative as noted.

plugins/@grouparoo/salesforce/src/lib/export/model.ts
pauloouriques: Excellent!

plugins/@grouparoo/salesforce/src/lib/export-contacts/model.ts
```diff
@@ -0,0 +1,22 @@
...
+  destinationOptions: SimpleDestinationOptions
+): SalesforceModel {
+  const mapping: { [key: string]: string } = {};
+  if (destinationOptions.primaryKey) {
+    Object.assign(mapping, {
+      recordObject: "Contact",
+      recordMatchField: destinationOptions.primaryKey as string,
+    });
+  }
+  if (destinationOptions.accountKey) {
```
bleonard, line 14: This is great.

bleonard (approved):
Good work!

PR #2633, opened 2021-12-02
Ignore null/undefined config objects and return a useful error about the file with the problem

evantahler:
1. If there are null or undefined config objects, we now return an error with the filename that has the problem (rather than mysteriously crashing)
2. If there's a problem saving the Sample Record, don't write `null` to the config file

This PR also changes config object writing (specifically for records, but it may apply in the future to other config types). This allows for merging config objects that share a destination filepath.

What this allows us to do is break out `Record` config files on a per-model basis

Before:
```
config/development/records.json
```

After:
```
config/development/customers/records.json
config/development/accounts/records.json
```

Thanks to @rwfeather!

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/configWriter.ts
+++ b/core/src/modules/configWriter.ts
@@ -98,7 +98,7 @@ export namespace ConfigWriter {
   }
 
   export async function getConfigObjects(): Promise<WritableConfigObject[]> {
-    let objects = [];
+    let objects: Array<{ filePath: string; object: any }> = [];
 
     const queryParams = { where: { locked: null } };
     let queries = {
@@ -108,6 +108,7 @@ export namespace ConfigWriter {
       properties: await Property.findAll(queryParams),
       groups: await Group.findAll(queryParams),
       destinations: await Destination.findAll(queryParams),
+      records: await GrouparooRecord.findAll({ include: [GrouparooModel] }),
     };
 
     const clusterNameSetting: Setting = await Setting.findOne({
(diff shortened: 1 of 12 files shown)


core/src/modules/configWriter.ts
```diff
@@ -126,23 +127,21 @@ export namespace ConfigWriter {
...
-    if (recordObjects.length > 0) {
-      objects.push({
-        filePath: "development/records.json",
-        object: recordObjects,
-      });
+        const filePath =
+          instance instanceof GrouparooRecord
+            ? `development/${(
+                await instance.$get("model")
+              ).getConfigId()}/records.json`
```
pedroslopez, line 135: I feel like this could possibly live in `generateFilePath`, but here is also fine

core/src/modules/configWriter.ts
```diff
@@ -126,23 +127,21 @@ export namespace ConfigWriter {
...
+            : generateFilePath(object as AnyConfigurationObject, type);
+
+        const index = objects.findIndex((obj) => obj.filePath === filePath);
+        if (index >= 0) {
+          objects[index].object = [objects[index].object, object].flat();
+        } else {
+          objects.push({ filePath, object });
+        }
+      }
```
pedroslopez, line 144: 👍🏼 merge objects with the same path

core/src/modules/configWriter.ts
```diff
@@ -108,6 +108,7 @@ export namespace ConfigWriter {
       properties: await Property.findAll(queryParams),
       groups: await Group.findAll(queryParams),
       destinations: await Destination.findAll(queryParams),
+      records: await GrouparooRecord.findAll({ include: [GrouparooModel] }),
```
pedroslopez, line 111: Where were the `null` objects coming from? Did `getConfigObject()` for an invalid record return `null`? 

I'm guessing this fixes the `null` or `undefined` objects because the loop that goes over these `queries` and calls `getConfigObject()` does some validations?

pedroslopez (approved):
Few questions mainly to understand what was causing the null objects and to make sure it was being addressed since, while I see errors being handled on the config loader, the generation of record config objects seems like it hasn't changed.

If my assumptions are correct, looks good!

core/src/modules/configWriter.ts
evantahler: Yep! The checks below filter away the null objects.  Before this PR, the records writer wasn't using those checks - so you could get null sample records - eg, with `user_id=-1` (when the primary key goes null)

core/src/modules/configWriter.ts
evantahler: we went back and forth too - we left `generateFilePath ()` as a method that used the instance's ID

PR #2634, opened 2021-12-03
Import all Sample Profiles

evantahler:
Adds a Button in `ui-config` to import all sample Records.
Also automatically tries to pre-load all the existing sample records at boot when using `ui-cofig`

Screen Shot 2021-12-02 at 6 18 45 PM
Screen Shot 2021-12-02 at 6 17 22 PM

### Development

### Impact

### Code review


core/src/initializers/codeConfig.ts
```diff
@@ -11,30 +14,57 @@ declare module "actionhero" {
...
+      const configDir = await getConfigDir(
+        process.env.GROUPAROO_RUN_MODE === "cli:config"
+      );
+      const { errors } = await loadConfigDirectory(configDir);
+      if (errors.length > 0) throw new Error("code config error");
+    });
 
-    if (errors.length > 0) throw new Error("code config error");
     api.codeConfig.allowLockedModelChanges = false; // after this point in the Actionhero boot lifecycle, locked models cannot be changed
+    await loadSampleProfiles();
```
evantahler, line 42: The story called for this to **not** be awaited and run in the "background".  However, I'm concerned about creating parallel queries when using SQLite (which ui-config is guaranteed to be using).  

At the very least, we now are using a "regular" Initializer, not CLSInitializer so failures are OK and the transaction won't be polluted by an import failure here.

bleonard: That's a good point but I'm also worried it might take a while to launch.
At a minimum, let's log before we do that so it's clear what is happening on the console.

evantahler: Screen Shot 2021-12-02 at 6 41 08 PM

added in f89ddaa66

core/src/actions/records.ts
```diff
@@ -206,6 +201,40 @@ export class RecordImport extends Action {
...
+  }
+
+  async run({ params }: { params: { modelId: string } }) {
+    let model: GrouparooModel;
+    let records: GrouparooRecord[];
+
+    await CLS.wrap(async () => {
+      model = await GrouparooModel.findById(params.modelId);
+      records = await GrouparooRecord.findAll({
+        where: { modelId: model.id },
```
krishnaglick, line 229: Do we even need to load the model if the modelId is in the params?

evantahler: Yeah - we wanted to validate that the modelId you provided exists.  Otherwise, if you sent `modelId=foo` you would get no error, and an empty list of Records back

krishnaglick (approved):

krishnaglick:
This has conflicts.

PR #2635, opened 2021-12-03
Actionhero v28

evantahler:
### Development

### Impact

### Code review


core/__tests__/modules/codeConfig/codeConfig.ts
```diff
@@ -711,9 +712,12 @@ describe("modules/codeConfig", () => {
...
 
-      previousGroupRun = await specHelper.runTask("group:destroy", {
-        groupId: "high_value",
-      });
+      previousGroupRun = await specHelper.runTask<GroupDestroy>(
+        "group:destroy",
+        {
+          groupId: "high_value",
+        }
+      );
```
evantahler, line 720: `specHelper.runTask` has return types now!

core/src/index.ts
```diff
@@ -58,3 +59,20 @@ export { waitForLock } from "./modules/locks";
...
+import "./config/next";
+import "./config/pluginInjection";
+import "./config/plugins";
+import "./config/redis";
+import "./config/routes";
+import "./config/sequelize";
+import "./config/tasks";
+import "./config/telemetry";
+import "./config/web";
+import "./config/websocket";
```
evantahler, line 78: This was an interesting trick - to force consumers (aka plugins) to be aware of the config types, they need to source the module overrides in the config files.  Adding them here in the main export file makes that happen!

core/src/actions/swagger.ts
```diff
@@ -54,76 +49,91 @@ export class Swagger extends Action {
...
+        if (!action) return;
+
+        const tag = action.name.split(":")[0];
+        const formattedPath = route.path
+          .replace("/v:apiVersion", "")
+          .replace(/\/:(\w*)/, "/{$1}")
+          .replace(/\/:(\w*)/, "/{$1}")
+          .replace(/\/:(\w*)/, "/{$1}")
+          .replace(/\/:(\w*)/, "/{$1}")
+          .replace(/\/:(\w*)/, "/{$1}");
```
pedroslopez:
```suggestion
          .replace(/\/:(\w*)/g, "/{$1}")
```

Use global flag to replace all?

core/src/config/api.ts
```diff
@@ -105,24 +113,21 @@ export const DEFAULT = {
...
     };
   },
 };
 
 export const production = {
-  general: (config) => {
+  [namespace]: () => {
     return {
       fileRequestLogLevel: "debug",
-      developmentMode: false,
```
pedroslopez, line 125: No idea what this is for, but noting that it's been removed from `production`

core/src/config/errors.ts
```diff
@@ -16,48 +26,57 @@ export const DEFAULT = {
...
 
       // action not useable by this client/server type
-      unsupportedServerType: (data) => {
+      unsupportedServerType: (data: ActionProcessor<any>) => {
         return `this action does not support the ${data.connection.type} connection type`;
       },
 
       // action failed because server is mid-shutdown
-      serverShuttingDown: (data) => {
+      serverShuttingDown: (data: ActionProcessor<any>) => {
```
pedroslopez: [suggestion] these unused params can probably be removed

core/src/config/logger.ts
```diff
@@ -1,15 +1,34 @@
...
 
 export const DEFAULT = {
-  logger: (config) => {
-    const loggers = [];
-    loggers.push(buildConsoleLogger(process.env.GROUPAROO_LOG_LEVEL));
-    config.general.paths.log.forEach((p) => {
-      loggers.push(buildFileLogger(p));
+  [namespace]: (config: ActionheroConfigInterface) => {
+    const loggers: ActionheroConfigLoggerBuilderArray = [];
+    loggers.push(buildConsoleLogger(process.env.LOG_LEVEL));
```
pedroslopez: Not sure if this was intentional, but this is changing the log level env var from `GROUPAROO_LOG_LEVEL` to `LOG_LEVEL`, which may have an impact in other places. I would suggest keeping it as `GROUPAROO_LOG_LEVEL` if possible.

plugins/@grouparoo/demo/src/util/runAction.ts
```diff
@@ -1,6 +1,3 @@
-//import { Connection, Action, config, log, utils, api, action } from 'actionhero';
-//import * as dotProp from "dot-prop";
```
pedroslopez, line 2: it looks like this `dotProp` thing was being used in the commented out code below - should all that commented code be removed too?

pedroslopez (changes requested):
Nice! I think the main thing to look at is the `LOG_LEVEL` environment var change that may potentially affect users and is referenced in multiple other places in the app.

core/src/config/api.ts
evantahler: I think that was a mistake - we don't want it (it's how the hot-reloading works)

core/src/config/errors.ts
evantahler: Yeah - this one was tricky! In the templates Actionhero generates, I went with generating these methods with all possible args... so now we have them :/

evantahler: I'll remove

plugins/@grouparoo/demo/src/util/runAction.ts
evantahler: I think I'll leave it for now... not really sure what @bleonard was up to here :D

core/src/actions/swagger.ts
evantahler: I *think* this was a problem with the `$1` replacement value - we don't know which index to use to replace things :/   
There's probably a loop we can use here to get them *all* - want to give it a try in the Actionhero repo? 

(edit) made an issue in the Actionhero repo - https://github.com/actionhero/actionhero/issues/2038

core/src/config/logger.ts
evantahler: Good catch!

evantahler:
@pedroslopez proved that https://github.com/grouparoo/grouparoo/pull/2635/commits/02cf895bcd8ef89776a89f3e09b0c42bc52fb949 was a good idea via https://github.com/actionhero/actionhero/pull/2039

core/src/actions/swagger.ts
```diff
@@ -82,13 +82,7 @@ export class Swagger extends Action {
         if (!action) return;
 
         const tag = action.name.split(":")[0];
-        const formattedPath = route.path
-          .replace("/v:apiVersion", "")
```
pedroslopez: A bit too eager on the deletion there, I believe`.replace("/v:apiVersion", "")` should still be present 😛

pedroslopez (changes requested):

core/src/actions/swagger.ts
evantahler: oops! updated now

pedroslopez (approved):

PR #2639, opened 2021-12-04
Clarify UI disabled logic when object is locked

evantahler:
Clarifies the UI logic when an object is disabled.  We cast the string to a Boolean which handles `null` and `undefined` as we expect

```
Welcome to Node.js v16.13.1.
Type ".help" for more information.
> Boolean('')
false
> Boolean(null)
false
> Boolean(undefined)
false
> Boolean('foo')
true
```

---

When using in-memory SQLite, the creation of new objects may have `undefined` default values when the object is initially created.  Subsequent reading of that object will return `null` values.  Rather than re-read each object when fetching `apiData()`, we'll be safer in the UI. 

This PR also clarifies some older migration behavior:
* No default values (a default value of `NULL` is the default behavior anyway - this is a safe change).  Default values belong in the model, [per our convention](https://www.grouparoo.com/docs/development/databases)
* Clarify `allowNull=true` (this is the default behavior anyway - this is a safe change)

### Development

### Impact

### Code review

### DIFF
--- a/core/src/migrations/000054-lockedString.ts
+++ b/core/src/migrations/000054-lockedString.ts
@@ -23,6 +23,7 @@ export default {
       const table = tables[i];
       await queryInterface.changeColumn(table, "locked", {
         type: DataTypes.STRING(191),
+        defaultValue: null, // normally the default value of a string column is null, but since we are migrating from a boolean, we need to re-set the default
         allowNull: true,
       });
       if (table !== "teams") {
--- a/core/src/migrations/000054-lockedString.ts
+++ b/core/src/migrations/000054-lockedString.ts
@@ -24,7 +24,6 @@ export default {
       await queryInterface.changeColumn(table, "locked", {
         type: DataTypes.STRING(191),
         allowNull: true,
-        defaultValue: null,
       });
       if (table !== "teams") {
         await queryInterface.sequelize.query(
--- a/core/src/migrations/000055-errorLevelForExports.ts
+++ b/core/src/migrations/000055-errorLevelForExports.ts
@@ -8,7 +8,6 @@ export default {
     await queryInterface.addColumn("exports", "errorLevel", {
       type: DataTypes.STRING(191),
       allowNull: true,
-      defaultValue: null,
     });
 
     await queryInterface.sequelize.query(
--- a/core/src/migrations/000058-runDestinationId.ts
+++ b/core/src/migrations/000058-runDestinationId.ts
@@ -8,7 +8,6 @@ export default {
     await queryInterface.addColumn("runs", "destinationId", {
       type: DataTypes.STRING(40),
       allowNull: true,
-      defaultValue: null,
     });
   },
 
--- a/core/__tests__/integration/errors.ts
+++ b/core/__tests__/integration/errors.ts
@@ -48,7 +48,7 @@ describe("errors", () => {
       message: "name must be unique",
       code: "unique violation",
       fields: ["name"],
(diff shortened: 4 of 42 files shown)


core/src/migrations/000055-errorLevelForExports.ts
```diff
@@ -8,7 +8,6 @@ export default {
     await queryInterface.addColumn("exports", "errorLevel", {
       type: DataTypes.STRING(191),
       allowNull: true,
-      defaultValue: null,
```
teallarson, line 11: I noticed that most of these tables you moved the `default(null)` to the model, but not here.  Was that intentional because we allow `null` on that column?

core/src/migrations/000058-runDestinationId.ts
```diff
@@ -8,7 +8,6 @@ export default {
     await queryInterface.addColumn("runs", "destinationId", {
       type: DataTypes.STRING(40),
       allowNull: true,
-      defaultValue: null,
```
teallarson, line 11: Same question I had about the `exports.errorLevel`

teallarson (changes requested):
UI bug when creating an `appRefreshQuery` is fixed 🎉
Screen Shot 2021-12-06 at 8 49 30 AM

Adding `@allowNull(true)` and `@defaultValue(null)` to the `AppRefreshQuery`'s `locked` column like was added to the other models should fix this issue that is still remaining when an app refresh query is being created in the UI:
Screen Shot 2021-12-06 at 8 52 52 AM

evantahler:
Whoops, thanks for catching those @tealjulia!

core/src/migrations/000055-errorLevelForExports.ts
evantahler: This change here is to remove the default(null) from the migration.  This isn't needed because a default of null is the default behavior in postgres.

teallarson: Makes sense.  I see that many of the other models got the `@AllowNull(true)` and `@DefaultValue(null)` decorators on columns (specifically on the `locked` column), but I don't see that on the `errorLevel` column in `exports`.  Just clarifying, your comment is saying that exports.errorLevel doesn't have those decorators on purpose, correct?

evantahler:
Hmm - perhaps the best way to go about this is to go relax the model constraints.  
* all `@AllowNull(true)` are removed, because that's the default
* all `@Default(null)` are removed, because that's the default
When in doubt, I'd prefer the models to be simplest/clearest.

But we still need to consider in the UI both the null & undefined cases, and the cast to `Boolean()` should be a good pattern for that regardless. 

What do you think about this PR now?

teallarson:
That makes sense as long as the models are treated in the same way.  We should pop a story into tracker to update our development docs for models to include this.  

Should we check other areas of the codebase where we might be depending on these columns to be `columnValue !== null` and instead use `Boolean(columnValue)`?  It sounds like it will _almost_ always be correctly `null`, but there's a small window of error with sqlite.  Correct?

evantahler:
Doc update: https://github.com/grouparoo/www.grouparoo.com/pull/819

I don't think that we need to scour the codebase for additional null checks at the moment, although this class of bug may certainly pop up again.  We are usually pretty good at checking `falsey` values `if(var)` rather than checking a specific value of a falsey value (`if var === null` or `if var === false`).   

One more review please @tealjulia!

core/src/migrations/000054-lockedString.ts
```diff
@@ -23,8 +23,8 @@ export default {
       const table = tables[i];
       await queryInterface.changeColumn(table, "locked", {
         type: DataTypes.STRING(191),
+        defaultValue: null, // normally the default value of a string column is null, but since we are migrating from a boolean, we need to re-set the default
```
evantahler, line 26: Comment added because this adding a `defaultValue=null` is not the normal pattern

teallarson (approved):
Looks good!

PR #2646, opened 2021-12-07
Test sample records when created by trying an import

evantahler:
This PR validates Sample Records when they are added by checking that it exists within the Primary Source

Screen Shot 2021-12-07 at 11 43 59 AM

This PR also has the side effect of automatically importing as much as we can about the Sample Record when it is created.

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/actions/records/records.ts
+++ b/core/__tests__/actions/records/records.ts
@@ -99,7 +99,7 @@ describe("actions/records", () => {
 
       expect(record.properties["userId"].state).toBe("ready");
       expect(record.properties["email"].state).toBe("ready");
-      expect(record.properties["purchases"].state).toBe("pending");
+      expect(record.properties["purchases"].state).toBe("ready");
 
       expect(groups).toEqual([]);
       expect(destinations).toEqual([]);
--- a/core/__tests__/actions/records/records.ts
+++ b/core/__tests__/actions/records/records.ts
@@ -84,7 +84,7 @@ describe("actions/records", () => {
         await specHelper.runAction<RecordCreate>("record:create", connection);
       expect(error).toBeUndefined();
       expect(record.id).toBeTruthy();
-      expect(record.state).toBe("pending");
+      expect(record.state).toBe("ready");
       expect(simpleRecordValues(record.properties)).toEqual({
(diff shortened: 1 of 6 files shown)


core/__tests__/actions/records/records.ts
```diff
@@ -99,7 +99,7 @@ describe("actions/records", () => {
 
       expect(record.properties["userId"].state).toBe("ready");
       expect(record.properties["email"].state).toBe("ready");
-      expect(record.properties["purchases"].state).toBe("pending");
+      expect(record.properties["purchases"].state).toBe("ready");
```
evantahler, line 102: New records that succeed will start as "ready" now, because they have been imported

core/src/actions/records.ts
```diff
@@ -143,6 +143,24 @@ export class RecordCreate extends AuthenticatedAction {
     if (params.properties) {
       await record.addOrUpdateProperties(params.properties);
     }
+
+    await record.markPending();
+    await record.import();
```
evantahler, line 148: This will also catch any other errors that would be found on import, like a a secondary property not being unique.  Whatever error is thrown will be shown to the user

core/src/actions/records.ts
```diff
@@ -143,6 +143,26 @@ export class RecordCreate extends AuthenticatedAction {
...
+    await record.updateGroupMembership();
+    await record.update({ state: "ready" });
```
pedroslopez, line 150: Is this a use case for `RecordOps.importAndUpdateInline`?

pedroslopez (approved):
Nice! Just one suggestion so these type of imports go through the same codepath

core/src/actions/records.ts
evantahler: I gave it a shot, but the behavior is a little different:
1. we can't use a transaction, which we want to do
2. we do want to throw if something goes wrong with the import (`record.import()` throws).  `importAndUpdateInline` does not throw and tries to keep going

evantahler: I did a little rename of the method to make that more clear -> `importAndUpdateInline -> opportunisticallyImportAndUpdateInline`

PR #2647, opened 2021-12-07
Add eslint and noImplicitAny to Postgres Plugin

krishnaglick:
## Change description

Enabled eslint in the Postgres plugin, fixed things.

### Development

### Impact

### Code review


plugins/@grouparoo/postgres/src/lib/connect.ts
```diff
@@ -1,11 +1,23 @@
-import { ConnectPluginAppMethod } from "@grouparoo/core";
+import { ConnectPluginAppMethod, SimpleAppOptions } from "@grouparoo/core";
 import { log } from "actionhero";
 import { Pool, types } from "pg";
+import { TypeId } from "pg-types";
 import format from "pg-format";
 import parseDate from "postgres-date";
 
+type PoolClient = Parameters<Parameters<InstanceType<typeof Pool>["on"]>[1]>[0];
```
krishnaglick: Too difficult for them to export this type.

plugins/@grouparoo/postgres/src/lib/export/exportRecord.ts
```diff
@@ -29,14 +27,86 @@ export const exportRecord: ExportRecordPluginMethod = async ({
     );
   }
 
-  try {
-    // --- Records --- //
-    if (toDelete) {
-      if (!syncOperations.delete) {
+  // --- Records --- //
+  if (toDelete) {
```
krishnaglick, line 38: The linter noticed this: https://eslint.org/docs/rules/no-unsafe-finally
Since we were throwing the error if it happened we didn't need a try/catch.
Functionality doesn't seem to have changed but I imagine this fixes a weird edge case.

plugins/@grouparoo/postgres/tsconfig.json
```diff
@@ -1,13 +1,14 @@
 {
   "compilerOptions": {
-    "rootDir": "./src",
+    "rootDir": ".",
```
krishnaglick: To allow importing packageJson

plugins/@grouparoo/postgres/.eslintrc.js
```diff
@@ -0,0 +1,17 @@
...
+  parserOptions: {
+    ecmaVersion: 12,
+    sourceType: "module",
+  },
+  plugins: ["@typescript-eslint"],
+  rules: {
+    "@typescript-eslint/consistent-indexed-object-style": ["warn", "record"],
+  },
+  ignorePatterns: ["/dist/", "*.js"],
+};
```
krishnaglick, line 25: Initial eslint config idea.

plugins/@grouparoo/postgres/tsconfig.json
evantahler: Does this change the folder structure within the `/dist` output?  If it does, that's going to mess with how the plugin is loaded.  We expect `<pluginRoot>/dist/initializers/stuff`.  I'll bet we have this now: `<pluginRoot>/postgres/dist/initializers/stuff`

evantahler: That would explain why the build fails.

evantahler: This *might* be rectified with changes to `core/config/plugins.ts` to set a new path for the plugin roots

plugins/@grouparoo/postgres/__tests__/export/export-records.ts
```diff
@@ -11,11 +11,12 @@ import {
...
 import { helper } from "@grouparoo/spec-helper";
 import { App } from "@grouparoo/core";
+import { PostgresPoolClient } from "../../src/lib/connect";
 
 let app: App;
 let destination: Destination;
-let client: any;
-let user: any;
+let client: PostgresPoolClient;
+let user: unknown;
```
evantahler: I think the user would at least be a `Record<>`

plugins/@grouparoo/postgres/__tests__/integration/app-refresh-query.ts
```diff
@@ -34,11 +34,13 @@ describe("integration/refreshQuery/postgres", () => {
...
-      refreshQuery: "SELECT 'HI' as name",
-      state: "ready",
-    }); //does not throw
+    expect(
+      AppRefreshQuery.create({
+        appId: app.id,
+        refreshQuery: "SELECT 'HI' as name",
+        state: "ready",
+      })
+    ).resolves.toBeTruthy();
```
evantahler, line 43: That's a nice way of doing these tests!

plugins/@grouparoo/postgres/__tests__/integration/log-checking.ts
```diff
@@ -39,6 +39,7 @@ async function getPropertyValue(query: string) {
 
 describe("postgres/integration/log-checking", () => {
   helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   const actionhero = require("actionhero");
   const logMock = jest.fn();
   Object.defineProperty(actionhero, "log", { value: logMock, writable: false });
```
evantahler, line 46: Ohh.... we are hacking the logger here :/

plugins/@grouparoo/postgres/package.json
```diff
@@ -40,8 +40,11 @@
     "@types/jest": "*",
     "@types/node": "*",
     "@types/pg": "*",
+    "@typescript-eslint/eslint-plugin": "^5.6.0",
+    "@typescript-eslint/parser": "^5.6.0",
```
evantahler: exact versions please

plugins/@grouparoo/postgres/src/lib/connect.ts
evantahler: That's a crazy way to get that type!

evantahler:
Overall, that didn't seem too terrible!

evantahler:
Could `noImplicitAny` also be turned on?

plugins/@grouparoo/postgres/tsconfig.json
krishnaglick: This causes too many issues right now, for simply getting around a require import. Maybe worth revisiting later but reverting for now.

plugins/@grouparoo/postgres/src/lib/connect.ts
krishnaglick: Thanks I Hate It 😂

krishnaglick:
> 

Lemme try!

plugins/@grouparoo/postgres/__tests__/integration/log-checking.ts
krishnaglick: This was some of my first work here. If there was a better way to do it, I was unaware at the time.

krishnaglick:
> Could `noImplicitAny` also be turned on?

Due to imports from core, no. It looks like we may need to pull off the **massive** bandaid that is core first :/
We could turn on `skipLibCheck` but since we import from `core/.../dist` directly instead of from the core package, that doesn't do anything.

Looks like we have really inconsistent import patterns:
`import { SessionCreate } from "@grouparoo/core/src/actions/session";`
`import { GrouparooRecord, Property } from "@grouparoo/core";`
```ts
import {
  Destination,
  DestinationSyncModeData,
} from "@grouparoo/core/dist/models/Destination";
```

If we did everything like `from "@grouparoo/core";` I think this would be doable.

plugins/@grouparoo/postgres/__tests__/query-import/getChangedRows.ts
```diff
@@ -3,7 +3,7 @@ import { getChangedRows } from "../../src/lib/query-import/getChangedRows";
 describe("getChangedRows", () => {
   let loggedQuery = "";
 
-  const commonArgs = {
+  const commonArgs: any = {
```
teallarson: why use `any` here instead of `unknown`?

plugins/@grouparoo/postgres/__tests__/table-import/get-changed-row-count.ts
```diff
@@ -7,7 +7,7 @@ import { App, Source, Schedule } from "@grouparoo/core";
 import { helper } from "@grouparoo/spec-helper";
 import { FilterOperation } from "@grouparoo/app-templates/dist/source/table";
 import { getChangedRowCount } from "../../src/lib/table-import/getChangedRowCount";
-import { beforeData, afterData, getConfig } from "../utils/data";
```
teallarson, line 10: I would think we'd want to use `afterData` to clean up after this test suite.  I am curious why that wasn't being used before!

plugins/@grouparoo/postgres/__tests__/table-import/import-property.ts
```diff
@@ -760,6 +761,9 @@ describe("postgres/table/recordProperty", () => {
   });
 });
 
-function fixedLengthFloat(value: any, decimalDigits = 2) {
+function fixedLengthFloat(
+  value: Awaited<ReturnType<typeof getPropertyValue>>,
```
teallarson: fancy type! 🎩

plugins/@grouparoo/postgres/__tests__/integration/log-checking.ts
pedroslopez: Something similar had to be done in core for testing export logs 

https://github.com/grouparoo/grouparoo/blob/a9ce0f6c516dffa8e6d54659eb153696f6a91039/core/__tests__/models/export.ts#L576-L585

plugins/@grouparoo/postgres/__tests__/utils/data.ts
```diff
@@ -71,6 +71,7 @@ CREATE TABLE ${groupsDestinationTableName}(
 };
 
 export const appOptions = {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   user: config.sequelize.username || require("os").userInfo().username,
```
pedroslopez: Is this necessary? Could we just import `os` at the top like normal?

plugins/@grouparoo/postgres/src/lib/connect.ts
```diff
@@ -24,19 +36,17 @@ export const connect: ConnectPluginAppMethod = async ({ appOptions }) => {
   sslOptions.forEach((opt) => {
     if (formattedOptions[opt] !== null && formattedOptions[opt] !== undefined) {
       if (!formattedOptions["ssl"]) formattedOptions["ssl"] = {};
+      // @ts-ignore I think this is wrong, but don't want to mess with it
```
pedroslopez, line 38: Why is it screaming at you? 

I would guess something about the `ssl` obj within the `formattedOptions` type above not having `cert`, `key` and `ca`? Would adding those to the type and doing `const sslOptions = ["ssl_cert", "ssl_key", "ssl_ca"] as const` work? not sure if it's smart enough to work out the `.replace()`

plugins/@grouparoo/postgres/__tests__/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "noImplicitAny": false
```
pedroslopez: such a shame :(

pedroslopez (commented):
lookin good!

plugins/@grouparoo/postgres/__tests__/tsconfig.json
krishnaglick: I can try a pass into the tests, see if I can get rid of this. This was a stopgap.

plugins/@grouparoo/postgres/src/lib/export/exportRecord.ts
```diff
@@ -48,144 +125,63 @@ export const exportRecord: ExportRecordPluginMethod = async ({
...
 
 const insert = async (
-  connection,
-  table,
-  syncOperations,
-  newRecordProperties
+  connection: PostgresPoolClient,
+  table: string,
+  syncOperations: DestinationSyncOperations,
+  newRecordProperties: Record<string, any>
```
teallarson: Another `any` I'm wondering about. Is it a record of a string and an array of recordProperties?

plugins/@grouparoo/postgres/src/lib/table-import/util.ts
```diff
@@ -1,12 +1,16 @@
...
 
-export function makeWhereClause(matchCondition: MatchCondition, params: any[]) {
+export function makeWhereClause(
+  matchCondition: MatchCondition,
+  params: (DataResponse | DataResponse[])[]
+) {
   const { columnName, filterOperation, value, values } = matchCondition;
-  let op;
-  let match = values || value || null;
+  let op: string;
```
teallarson, line 12: I think `op: SourceFilterMethodResponseRow["ops"][number]` can be used here

teallarson (changes requested):
Nice work!  One plugin (almost) down ✅

A few places I think we could upgrade types a little.

core/src/index.ts
```diff
@@ -79,3 +79,4 @@ import "./config/tasks";
 import "./config/telemetry";
 import "./config/web";
 import "./config/websocket";
+import "./initializers/plugins";
```
evantahler, line 82: Not needed?

plugins/@grouparoo/postgres/.eslintrc.js
evantahler: Perhaps this config moves into `@grouparoo/app-templates` and we can synlink it into all future plugins

plugins/@grouparoo/postgres/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  env: {
+    es2021: true,
+    node: true,
+  },
```
evantahler, line 5: We still need to support node.js v12... does this work?

evantahler: I guess it's just for the linter... so it's OK

plugins/@grouparoo/postgres/__tests__/table-import/import-properties.ts
```diff
@@ -1166,6 +1166,9 @@ describe("postgres/table/recordProperties", () => {
   });
 });
 
-function fixedLengthFloat(value: any, decimalDigits = 2) {
+function fixedLengthFloat(
+  value: Property[] | RecordPropertiesPluginMethodResponse,
+  decimalDigits = 2
+) {
   return parseFloat(parseFloat(value.toString()).toFixed(decimalDigits));
 }
```
evantahler: I wonder if this common method should be shared in `@grouparoo/app-templates` or `@grouparoo/spec-helper` too

plugins/@grouparoo/postgres/__tests__/table-import/import-property.ts
evantahler: Yeah... let's move this to `@gropuaroo/spec-helper`

plugins/@grouparoo/postgres/__tests__/tsconfig.json
evantahler: I think that `noImplicitAny: true` is more valuable than getting linting to work.  Ideally, both can be enabled, but if I had to choose, I would prefer to enable `noImplicitAny` first.

plugins/@grouparoo/postgres/package.json
```diff
@@ -22,9 +22,9 @@
   },
   "scripts": {
     "prepare": "rm -rf dist && tsc --declaration",
-    "test": "jest",
+    "test": "jest ./__tests__",
```
evantahler: Isn't this the default?  I'd rather keep the `jest` command working out of the box.  If you really need to ignore a file, you can do that in `jest.config.js`

plugins/@grouparoo/postgres/package.json
```diff
@@ -40,8 +40,12 @@
     "@types/jest": "*",
     "@types/node": "16.*.*",
     "@types/pg": "*",
+    "@types/pg-format": "1.0.2",
+    "@typescript-eslint/eslint-plugin": "5.6.0",
+    "@typescript-eslint/parser": "5.6.0",
     "actionhero": "28.1.4",
     "csv-parse": "4.16.3",
+    "eslint": "8.6.0",
```
evantahler, line 48: 👍 strict versions and just @types packages added

plugins/@grouparoo/postgres/src/initializers/plugin.ts
```diff
@@ -31,6 +31,7 @@ import {
   QueryPropertyTemplate,
 } from "@grouparoo/app-templates/dist/source/query";
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
 const packageJSON = require("./../../package.json");
```
evantahler, line 35: We could also use `fs` to load the file if it's better

plugins/@grouparoo/postgres/__tests__/tsconfig.json
evantahler: oh wait! this is a new `tsconfig.json` just for the tests!

plugins/@grouparoo/postgres/tsconfig.json
```diff
@@ -7,7 +7,9 @@
     "target": "es2019",
     "esModuleInterop": true,
     "experimentalDecorators": true,
-    "emitDecoratorMetadata": true
+    "emitDecoratorMetadata": true,
+    "resolveJsonModule": true,
+    "noImplicitAny": true
```
evantahler, line 12: Woo!

plugins/@grouparoo/postgres/src/lib/connect.ts
krishnaglick: `ssl` is not given a type it expects. And then it gets deeper from there.
Screen Shot 2022-01-18 at 10 16 21 AM

plugins/@grouparoo/postgres/__tests__/tsconfig.json
krishnaglick: I was able to remove this!

plugins/@grouparoo/postgres/__tests__/query-import/getChangedRows.ts
krishnaglick: This type was slighly obnoxious, but I went ahead and typed it :)

plugins/@grouparoo/postgres/__tests__/table-import/get-changed-row-count.ts
krishnaglick: They weren't in use. 🤷

plugins/@grouparoo/postgres/src/lib/table-import/util.ts
krishnaglick: This doesn't seem to match 1:1

core/src/index.ts
krishnaglick: Very needed. Plugins need this import to know the type for the api object.

plugins/@grouparoo/postgres/.eslintrc.js
krishnaglick: Typescript should transpile everything down so it doesn't matter I'd think. Either way we could always change this so we don't use 2021 features.

plugins/@grouparoo/postgres/package.json
krishnaglick: Reverting

plugins/@grouparoo/postgres/src/initializers/plugin.ts
krishnaglick: Not sure it is, but an easy change.

krishnaglick: Breaks things

core/src/index.ts
evantahler: Ah ha!  Better than every plugin needing to require the initializer. 

To be complete, we probably should also import/export any other initializer that adds something to the global `api` object.  

* https://github.com/grouparoo/grouparoo/blob/main/core/src/initializers/apm.ts
* https://github.com/grouparoo/grouparoo/blob/main/core/src/initializers/codeConfig.ts
* https://github.com/grouparoo/grouparoo/blob/main/core/src/initializers/next.ts
* https://github.com/grouparoo/grouparoo/blob/main/core/src/initializers/notifiers.ts
* https://github.com/grouparoo/grouparoo/blob/main/core/src/initializers/plugins.ts
* https://github.com/grouparoo/grouparoo/blob/main/core/src/initializers/session.ts

plugins/@grouparoo/postgres/src/lib/table-import/util.ts
teallarson: Ah, does that give all the Filter options not just the ones Postgres can use?  I don't suppose it's blocking anything... I just feel like there _has to_ be a way to make it smarter!

plugins/@grouparoo/postgres/__tests__/integration/postgres-table-import.ts
```diff
@@ -42,14 +40,14 @@ describe("integration/runs/postgres", () => {
...
   let model;
-  let app;
-  let source;
-  let schedule;
-  let destination;
-  let group;
+  let app: Awaited<ReturnType<AppCreate["run"]>>["app"];
+  let source: Awaited<ReturnType<SourceCreate["run"]>>["source"];
+  let schedule: Awaited<ReturnType<ScheduleCreate["run"]>>["schedule"];
+  let destination: Awaited<ReturnType<DestinationCreate["run"]>>["destination"];
```
evantahler, line 49: Presumably these could also be `typeof App.apiData()`, `typeof Source.apiData()`, etc maybe?  Either way, no need to change them

evantahler (approved):
Well done!

I know this was a slog, but now we know how to strongly type plugins!

plugins/@grouparoo/postgres/__tests__/table-import/import-properties.ts
```diff
@@ -84,8 +88,8 @@ async function getPropertyArrays(
     };
     counter++;
   }
-
-  const propertyFilters = usePropertyFilters
+  // TODO: This type
```
krishnaglick, line 91: @evantahler Some types in here don't line up with what I expect

krishnaglick:
@evantahler Not sure you had any more comments on this, or if it's good to merge.

plugins/@grouparoo/postgres/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+  rules: {
+    "@typescript-eslint/consistent-indexed-object-style": ["warn", "record"],
+    "@typescript-eslint/array-type": "error",
+    "object-shorthand": ["error", "always"],
+    "no-useless-computed-key": "error",
+    "prefer-arrow-callback": "error",
+    "prefer-object-spread": "error",
+    "no-useless-computed-key": "error",
+    "@typescript-eslint/no-explicit-any": "off", // Remove this in some Ivory Tower future.
+    "@typescript-eslint/ban-ts-comment": "off", // Remove this in some Ivory Tower future.
```
edmundito, line 22: What's the ivory tower future?

plugins/@grouparoo/postgres/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+  plugins: ["@typescript-eslint"],
+  rules: {
+    "@typescript-eslint/consistent-indexed-object-style": ["warn", "record"],
+    "@typescript-eslint/array-type": "error",
+    "object-shorthand": ["error", "always"],
+    "no-useless-computed-key": "error",
+    "prefer-arrow-callback": "error",
+    "prefer-object-spread": "error",
+    "no-useless-computed-key": "error",
```
edmundito, line 20: One concern about using error is when using the editor there's just a lot of red for something that maybe should be a suggestion. You can always bump up the level in cI using `--max-warnings=0`

plugins/@grouparoo/postgres/.eslintrc.js
edmundito: Yeah I'd like to see how we can share these core rules. Could be a separate PR.

edmundito:
It's probably worth updating the title and description at this point. Sure this is a POC, but this change is specifically adding Eslint rules to the Postgres plugin.

plugins/@grouparoo/postgres/.eslintrc.js
krishnaglick: Where all our props/interfaces are accurate. We have `// @ts-ignore` in a number of places rather than sorting out complex types.

plugins/@grouparoo/postgres/.eslintrc.js
krishnaglick: An error says "Don't do this"
Most of these aren't that big a deal in either direction, but more just enforcing opinions.

edmundito (approved):

krishnaglick:
Will merge this after we cut 0.8

plugins/@grouparoo/postgres/.eslintrc.js
edmundito: I'm ok starting with error level and see how we feel about it.

teallarson (approved):

PR #2648, opened 2021-12-07
Update `apply` to pull plugin default values when missing from app config

edmundito:
## Change description

This change updates the `apply` command to resolve app default values when applying a config with missing values. Previously, any missing options with default values were not getting stored in the db because it assumed it was.

---

Example:
1. Reset your config/db and start the config UI in `staging-config`
1. Create a basic config for a postgres app
2. Exit the config UI and remove the `port` option from the app config file
3. Run `grouparoo apply`

Before:
- The port is not stored in the options table

After:
- The default port is pulled from the plugin options and stored in the options table

---

TODO
- [ ] Tests (Currently there are no tests for the apply command?)
- [ ] Figure out if this change impacts other areas

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/configLoaders/app.ts
+++ b/core/src/modules/configLoaders/app.ts
@@ -6,7 +6,7 @@ import {
   validateConfigObjectKeys,
   IdsByClass,
 } from "../../classes/codeConfig";
-import { App } from "../..";
+import { App, SimpleAppOptions } from "../..";
 import { Op } from "sequelize";
 
 import { ConfigWriter } from "../configWriter";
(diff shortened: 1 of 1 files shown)


evantahler (commented):
I think I'd prefer this change happen within the model itself, and not the configLoader.  Also, the same problem may exist for Sources and Destinations as well.

The logic you added could go in an `@AfterCreate()` hook, and we could start the instance of the App/Source/Destination with the default option values. This could be a new shared utility like `OptionHelper.createWithDefaultOptions(instance)`

There's a second part of this work that also needs to consider `OptionHelper.setOptions()` - If we are setting an App's options and the hash we pass in doesn't contain something that wants a default value, we probably need to re-insert it into the hash - just like you've done here.  `app.setOptions({host:'localhost'})` should also end up with `port:5432` for a postgres app.

evantahler:
@pedroslopez since you picked up the story, feel free to close out this PR if you don't need it

PR #2664, opened 2021-12-09
Use unescaped values for mustache strings

teallarson:
## Change description

Initially, we thought there was something wrong with SQLite's escaping.  It turns out, it was mustachejs returning escaped strings in Calculated Property functions.

Mustache variables automatically escape HTML. However, with calculated properties, we actually want to be using the unescaped HTML value of a string.  That way, if a property value has an HTML-escapable character like an apostrophe, it gets rendered correctly.

Because of how our system parses in the values for mustache strings, we'll start by defaulting to triple brackets (ie: `{{{ firstName }}}` everywhere.

I believe we will want to revisit this and simply keep whatever bracket notation a user inputs by making our parser more sophisticated.  For now, we'll start here.

O'Donoghue in `fullName` property used to be one of the culprits, now renders fine:

Screen Shot 2021-12-10 at 4 55 39 PM

### Development

### Impact

- Users who have properties that have an option with mustache variables will be replaced by the triple bracket version (both in the database when Grouparoo is running and within the actual config files if Config mode is run)
- All property options (calculated property or query source property) with mustache variables will now be saved as the triple bracket notation version of whatever is entered (double or triple)

### Code review

### DIFF
--- a/core/src/modules/ops/property.ts
+++ b/core/src/modules/ops/property.ts
@@ -134,7 +134,7 @@ export namespace PropertyOps {
     for (const key in ruleOptions) {
       const mustacheString = ruleOptions[key];
       const mustacheVariables: string[] = Mustache.parse(mustacheString)
-        .filter((chunk) => chunk[0] === "name")
+        .filter((chunk) => chunk[0] === "&")
         .map((chunk) => chunk[1]);
       properties
         .filter((p) => mustacheVariables.includes(p.key))
--- a/plugins/@grouparoo/calculated-property/__tests__/import/import-property.ts
+++ b/plugins/@grouparoo/calculated-property/__tests__/import/import-property.ts
@@ -44,7 +44,7 @@ describe("calculated-property/recordProperty", () => {
       userId: [1],
       email: ["ejervois0@example.com"],
       firstName: ["Mario"],
-      lastName: ["Jones"],
+      lastName: [`O'Donnell`],
       ltv: [390.42],
       lastLoginAt: ["2021-08-23 15:02:39.297-07"],
       isVIP: [true],
@@ -59,6 +59,13 @@ describe("calculated-property/recordProperty", () => {
     }`;
     const value = await getPropertyValue(fn);
     expect(value[0]).toEqual(`hi Mario`);
+  });
+  test("it evaluates string properties with escapable characters as expected", async () => {
+    const fn = `() => {
+        return "hi {{{firstName}}} {{{ lastName }}}"
+    }`;
+    const value = await getPropertyValue(fn);
+    expect(value[0]).toEqual(`hi Mario O'Donnell`);
   });
   test("it evaluates float properties as expected", async () => {
     const fn = `() => {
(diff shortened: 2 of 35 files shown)


core/src/modules/ops/property.ts
```diff
@@ -134,7 +134,7 @@ export namespace PropertyOps {
     for (const key in ruleOptions) {
       const mustacheString = ruleOptions[key];
       const mustacheVariables: string[] = Mustache.parse(mustacheString)
-        .filter((chunk) => chunk[0] === "name")
+        .filter((chunk) => chunk[0] === "&")
```
teallarson, line 137: This was the missing piece that had broken the DAG this morning!  When one uses the triple bracket, `Mustache.parse`'s response looks different.  The ampersand tells mustache to render the value unescaped.

evantahler: That's 🍌s

Some notes here - https://stackoverflow.com/questions/22910428/mustache-globally-disable-html-escaping/22910523

plugins/@grouparoo/calculated-property/__tests__/import/import-property.ts
```diff
@@ -55,28 +55,35 @@ describe("calculated-property/recordProperty", () => {
...
     }`;
     const value = await getPropertyValue(fn);
     expect(value[0]).toEqual(`hi Mario`);
   });
+  test("it evaluates string properties with escapable characters as expected", async () => {
+    const fn = `() => {
+        return "hi {{{firstName}}} {{{ lastName }}}"
+    }`;
+    const value = await getPropertyValue(fn);
+    expect(value[0]).toEqual(`hi Mario O'Donnell`);
```
evantahler, line 68: 👍 good testing

plugins/@grouparoo/snowflake/__tests__/query-import/import-property.ts
```diff
@@ -50,43 +50,43 @@ describe("snowflake/query/recordProperty", () => {
...
   test("can run a integer query to get a string", async () => {
-    const sql = "SELECT first_name FROM profiles WHERE id = {{ userId }}";
+    const sql = "SELECT first_name FROM profiles WHERE id = {{{ userId }}}";
     const value = await getPropertyValue(sql);
     expect(value).toEqual(["Erie"]);
   });
 
   test("can run a integer query to get a float", async () => {
-    const sql = "SELECT ltv FROM profiles WHERE id = {{ userId }}";
+    const sql = "SELECT ltv FROM profiles WHERE id = {{{ userId }}}";
```
evantahler, line 59: I'm curious - would these queries have still worked with the `{{}}`?

evantahler (approved):

core/src/modules/ops/property.ts
teallarson: Interesting!  I had to dig around in their github repo to find this rule 😅

plugins/@grouparoo/snowflake/__tests__/query-import/import-property.ts
teallarson: It does!  Just checked.  Core converts it all correctly and figures it out.

PR #2665, opened 2021-12-09
Create Klaviyo Plugin

aelafifi:
## Change description

Create Klaviyo plugin with export profiles destination

### Development

### Impact

### Code review


plugins/@grouparoo/klaviyo/src/lib/client.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  async getListMapByName() {
+    const allLists = await this.lists.getLists();
+    return Object.fromEntries(
+      allLists.map(({ list_id, list_name }) => [list_name, list_id])
+    );
+  }
+
+  async isMemberOfList(listId: string, email: string) {
```
bleonard, line 41: I don't see this being used. It feels like a lot of api calls if it was, though.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
```diff
@@ -0,0 +1,159 @@
+import { getKnownFieldMap } from "./destinationMappingOptions";
+import { Errors, ExportRecordPluginMethod } from "@grouparoo/core";
+import { connect } from "../connect";
+
+async function handleGroupsChanges(client, email, oldGroups, newGroups) {
+  const listMapByName = await client.getListMapByName();
```
bleonard: let's do some caching of this. see other plugins `listMethods` approach.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  // Unsubscribe from old groups
+  for (const groupName of oldGroups) {
```
bleonard: This isn't quite right. see how there is some "set" work done in other plugins
`oldGroups` is what it was before (so it could be ['a', 'b'])
`newGroups` is what it is now (so it could be ['b','c'])
The user is still in `b` so we don't want to remove them.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
```diff
@@ -0,0 +1,159 @@
...
+        emails: [email],
+      });
+
+      const result = await client.lists.getAllMembers({
+        groupId: listMapByName[groupName],
+      });
+      if (result.records.length === 0) {
+        await client.lists.deleteList(listMapByName[groupName]);
+      }
+    }
```
bleonard: We don't have to do this call and we never have to delete a list. This is too expensive to do this often from an API/time perspective. So we can delete these lines.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  // Subscribe to new groups
+  for (const groupName of newGroups) {
+    // Create list if not exists
+    if (!listMapByName.hasOwnProperty(groupName)) {
```
bleonard: This would be moved to `listMethods`

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
```diff
@@ -0,0 +1,159 @@
...
+  const allFields = new Set([...newFields, ...oldFields]);
+
+  // get fields
+  const fieldKeys = await getKnownFieldMap();
+  for (const fieldName of allFields) {
+    const value = newRecordProperties[fieldName];
+    const klaviyoKey = fieldKeys[fieldName];
+
+    payload[klaviyoKey ? klaviyoKey : fieldName] =
+      formatVar(value) ?? "[deleted]";
```
bleonard: maybe set to null or empty string if that works.

aelafifi: It doesn't work, any `falsy` value will be ignored!
for them, null or empty string means keep it untouched!!
Basically, they don't have such "clear field" functionality. and that was the best solution I've thought about to deal with that weird behavior.

plugins/@grouparoo/klaviyo/src/lib/client.ts
aelafifi: It's used in the tests to check that the user is successfully added to the list.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
aelafifi: I thought that every call for this function could affect the values there.
Let me check how the caching mechanism really works in other plugins, and how to deal with the demand of force reloading the data even if the cache key is present but the requested list ID isn't.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
aelafifi: Ahaaa, good catch.
Gonna handle it.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
aelafifi: You mean the whole function? or these lines?

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
bleonard: I bet it will fail on date and integers, though. 🤔

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
```diff
@@ -0,0 +1,159 @@
...
+  const oldFields = Object.keys(oldRecordProperties);
+  const allFields = new Set([...newFields, ...oldFields]);
+
+  // get fields
+  const fieldKeys = await getKnownFieldMap();
+  for (const fieldName of allFields) {
+    const value = newRecordProperties[fieldName];
+    const klaviyoKey = fieldKeys[fieldName];
+
+    payload[klaviyoKey ? klaviyoKey : fieldName] =
```
bleonard: what's the case here where there is no `klaviyoKey` ?

aelafifi: All other custom fields. You can add any key/value on the fly without being predefined.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
aelafifi: No dates or integers, everything is string here. they don't have different data types! Only strings.
and `formatVar` returns `null` only when the field is set `undefined` which means "erase the value of it".

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
bleonard: creating the list if it didn't exist is in th `listMethods` in other plugins.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
bleonard: I see. Is that just also the `filedName`, though?

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
bleonard: I'm pretty sure `formatVar` is never going to return `undefined` here, so this will never be `[deleted]`.
I think you want something like:
```
if (value===undefined) value = "[deleted]"; // because it's not in newProperties
payload[fieldName] = formatVar(value);
```

or otherwise handling `undefined` before formatting.

aelafifi: Yup, it will never return `undefined`, it returns `null` instead if the value isn't set in the new data.
That's why I'm using `?? (Nullish Coalescing)` to convert `null` to `"[deleted]"`

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
aelafifi: Yup, you just send the field name as is, no need to make any changes to it.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
bleonard: I think you want to have `"deleted"` if it was truly deleted though. If someone's last name (or whatever setting) is `null` in grouparoo, it's probably not right that it ends up in Klaviyo as `"deleted"`.

aelafifi: Aha, I got the point

bleonard: Let's try this `$unset` approach and see if it works. https://community.klaviyo.com/apis-40/set-custom-property-to-null-blank-or-remove-it-1501

bleonard: Random thought: I guess we should also use `$unset` if it's falsy at all inlcuding empty string if that's how Klaviyo works. so maybe `formatVar` should trim and then return null if it's empty.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
```diff
@@ -0,0 +1,140 @@
...
+  const allFields = new Set([...newFields, ...oldFields]);
+
+  // get fields
+  const fieldKeys = await getKnownFieldMap();
+  for (const fieldName of allFields) {
+    const value = newRecordProperties[fieldName];
+    const klaviyoKey = fieldKeys[fieldName];
+    const key = klaviyoKey ? klaviyoKey : fieldName;
+
+    if (value === undefined) {
```
bleonard: If I understand what you said correctly, we still need to `unset` this if the variable is `null` or `""`
So maybe this whole section needs to be 
```
const value = formatVar(value);
if(!value) {
  payload.$unset ??= [];
   payload.$unset.push(key);
}
```

to handle both deleted properties and ones that are now empty.

aelafifi: Yeah, I handled this in the body of `formatVar` below
```
if (value === undefined) {
    return null;
  }
  if (typeof value === "string" && value.trim() === "") {
    return null;
  }
```

bleonard: But that will send `null` to Klaviyo, which will have no effect. So if the "first name" is now null, wil won't be set to null or unset wit hthe current code. that right?

It would be good to have a test that shows this works.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
```diff
@@ -0,0 +1,141 @@
...
+    const value = newRecordProperties[fieldName];
+    const klaviyoKey = fieldKeys[fieldName];
+    const key = klaviyoKey ? klaviyoKey : fieldName;
+
+    const formattedValue = formatVar(value);
+    if (!formattedValue) {
+      payload.$unset ??= [];
+      payload.$unset.push(key);
+    } else {
+      payload[key] = formatVar(formattedValue);
```
bleonard: I think this is already formatted!
`payload[key] = formattedValue;` should work

bleonard (approved):
One more change and should be good to go.

plugins/@grouparoo/klaviyo/src/lib/export/exportRecord.ts
aelafifi: Oh yeah!

PR #2672, opened 2021-12-10
Property and Schedule Filter Ops to match Group Rule Ops

teallarson:
## Change description

Property and Schedule filters now use the same terminology as Group rules.  There are also new filters for `Exists` and `NotExists` for `null`/`not null` values. 

Models: Models have a migration to new filter values

Code Config: Old filter operators are deprecated and will throw with a message that directs users to (a) the new filter operator they should use instead and (b) a link to our code config filter docs.

Plugins: SQL plugins' filter operators use the new terminology and have a new Exists and Not Exists filter operators.  Mongo does not get this feature in this PR.

UI:

Operators use human-friendly language in the UI on both the Property Edit and Schedule Edit pages:
Screen Shot 2022-01-03 at 11 12 12 AM

Box to input a match only shows up on Operators that use a match (like in the existing group rules builder).  So "exists with any value" does not have one:
Screen Shot 2022-01-03 at 11 11 54 AM
but "greater than" does:
Screen Shot 2022-01-03 at 11 12 00 AM

### Development

### Impact

- Migration for models: old filter op values will be replaced by new filter op values (ie: "equals" --> "eq")
- Code config using old filter op values will show a deprecation warning and fail.

### Code review


core/src/classes/plugin.ts
```diff
@@ -111,15 +111,17 @@ export enum AggregationMethod {
 }
 
 export enum FilterOperation {
-  Equal = "equals",
```
bleonard, line 114: Just in case you haven't considered it, we have some work to do to maintain compatibility here.
* db migrations?
* deprecations for code config?
* still handle old value?

teallarson: Appreciate it.  I do have those written down as items to check for here.

plugins/@grouparoo/clickhouse/__tests__/table-import/import-properties.ts
```diff
@@ -362,6 +362,7 @@ describe("clickhouse/table/recordProperties", () => {
           aggregationMethod: AggregationMethod.Exact,
         });
         expect(values[record.id][properties[0].id]).toEqual([
+          null,
```
teallarson: Shenanigans going on with the order of the responses when one of the rows contains a`null` in this Clickhouse test... is there a looser way I can match the array (IE: by contents but not necessarily order?)

evantahler: We do this in some of other tests with a `sort()` -> `expect(names.stort()).toEqual(['a','b'].sort())`

core/src/migrations/000094-changePropertyFilterNotation.ts
```diff
@@ -0,0 +1,20 @@
+import Sequelize from "sequelize";
+import { filterOpsDescriptions } from "../modules/filterOpsDictionary";
+
+export default {
+  up: async (queryInterface: Sequelize.QueryInterface) => {
+    for (const op in filterOpsDescriptions) {
+      await queryInterface.sequelize.query(
+        `UPDATE "filters" SET "op"='${op}' WHERE "op"='${filterOpsDescriptions[op]}'`
+      );
```
evantahler, line 9: [note] This will update both schedule filters and property filters

core/src/modules/filterHelper.ts
```diff
@@ -90,10 +98,25 @@ export namespace FilterHelper {
...
+    const deprecatedFilters = {
+      equals: "eq",
+      "does not equal": "ne",
+      "greater than": "gt",
+      "greater than or equal to": "gte",
+      "less than": "lt",
+      "less than or equal to": "lte",
+      contains: "substring",
+      "does not contain": "notSubstring",
+    };
```
evantahler: I wonder if this check belongs here in the filterHelper, or in the configLoaders.  After the migration, the app shouldn't boot up with any of these old filters in place.  I think we would be more concerned with catching them at load time from a config file rather than holistically in the whole app (here)

core/src/modules/filterOpsDictionary.ts
```diff
@@ -0,0 +1,28 @@
...
+  ne: "does not equal",
+  exists: "exists with any value",
+  notExists: "does not exist",
+  gt: "greater than",
+  gte: "greater than or equal to",
+  lt: "less than",
+  lte: "less than or equal to",
+  substring: "contains",
+  notSubstring: "does not contain",
+};
```
evantahler: Add a `as const` here to get this to be strongly typed

plugins/@grouparoo/bigquery/__tests__/table-import/import-property.ts
```diff
@@ -189,31 +189,31 @@ describe("bigquery/table/recordProperty", () => {
           sourceMapping,
           aggregationMethod: "average",
         });
-        expect(value).toEqual([1.73]);
+        expect(value).toEqual([1.6257142857142857]);
```
evantahler: If you are having trouble with rounding/avg aggregation numbers like this in tests, we sometimes use a "float helper"

https://github.com/grouparoo/grouparoo/blob/fbc78c3ba43fd1974a5c6bdff6cabb33a78126c7/plugins/%40grouparoo/mysql/__tests__/table-import/import-properties.ts#L1151-L1153

plugins/@grouparoo/bigquery/src/lib/table-import/getPropertyValue.ts
```diff
@@ -91,7 +91,7 @@ export const getPropertyValue: GetPropertyValueMethod = async ({
     }
   } catch (error) {
     throw new Error(
-      `Error with BigQuery SQL Statement: Query - \`${options}\`, Error - ${error}`
+      `Error with BigQuery SQL Statement: Query - \`${options.query}\`, Error - ${error}`
```
evantahler: I think we did want `options` here, maybe this should be `JSON.stringify(options)` - this way you can see the query and the ? params

plugins/@grouparoo/clickhouse/__tests__/utils/data.ts
```diff
@@ -120,29 +119,39 @@ async function createTables() {
...
+        ? 1
+        : curr === "false"
+        ? 0
+        : !isNaN(Number(curr))
+        ? Number(curr)
+        : `'${curr}'`) +
+      (idx === values.length - 1 ? "" : ", ")
+    );
+  }, "");
+}
```
evantahler, line 138: 👍 fancy.

plugins/@grouparoo/demo/src/connections/clickhouse.ts
```diff
@@ -130,6 +130,8 @@ export default class ClickHouse extends Connection {
           nextColumnType = nextColumnType.replace("PRIMARY KEY", "");
         }
 
+        columnType = `NULLABLE(${columnType})`;
```
evantahler, line 133: interesting clickhouse-specific column type

plugins/@grouparoo/mysql/__tests__/table-import/import-properties.ts
```diff
@@ -331,7 +331,7 @@ describe("mysql/table/recordProperties", () => {
           sourceMapping,
           aggregationMethod: AggregationMethod.MostRecentValue,
         });
-        expect(values[record.id][properties[0].id][0]).toEqual("Orange");
+        expect(values[record.id][properties[0].id][0]).toEqual(null);
```
evantahler, line 334: This test (in all plugins) went from testing a value (Orange) to finding a null.  It's OK because the test below (least recent value) still checks that non-null value (Apple) is found

plugins/@grouparoo/postgres/__tests__/table-import/import-properties.ts
```diff
@@ -548,7 +549,7 @@ describe("postgres/table/recordProperties", () => {
   describe("filters", () => {
     const sourceMapping = { record_id: "userId" };
     const columns = ["amount"];
-    const aggregationMethod = "count";
+    const aggregationMethod = "count"; // select count(*) from purchases where amount IS NOT NULL AND userID = ${}
```
evantahler: Comment doesn't need to stay :D

plugins/@grouparoo/postgres/__tests__/table-import/import-properties.ts
```diff
@@ -882,96 +918,14 @@ describe("postgres/table/recordProperties", () => {
...
         );
-        expect(values[record.id][properties[0].id]).toEqual([4]);
+        expect(values[record.id][properties[0].id]).toEqual([5]);
         expect(values[otherRecord.id][properties[0].id]).toEqual([4]);
         expect(values[thirdRecord.id][properties[0].id]).toEqual([0]);
       });
     });
 
-    describe("equals", () => {
-      const op = "equals";
```
evantahler, line 892: Why was this section removed?

plugins/@grouparoo/postgres/__tests__/table-import/import-property.ts
```diff
@@ -307,58 +307,28 @@ describe("postgres/table/recordProperty", () => {
...
-        expect(value).toEqual(0);
-      });
-      test("date", async () => {
-        const value = await getPropertyValue(
-          {
-            column,
-            sourceMapping,
-            aggregationMethod,
-          },
-          [{ op, key: "date", match: "2020-02-15" }]
```
evantahler: some more curious removals

plugins/@grouparoo/postgres/__tests__/table-import/import-property.ts
```diff
@@ -307,58 +307,28 @@ describe("postgres/table/recordProperty", () => {
...
-      test("float", async () => {
-        const value = await getPropertyValue(
-          {
-            column,
-            sourceMapping,
-            aggregationMethod,
-          },
-          [{ op, key: "amount", match: "1.54" }]
-        );
-        expect(value).toEqual(2);
```
evantahler: another lost test maybe

plugins/@grouparoo/snowflake/__tests__/fixtures/table-import/schedule-options.js
```diff
@@ -7,7 +7,7 @@ nock("https://xyz12345.us-east-1.snowflakecomputing.com:443", {
   .query({
     requestId: /.+/g,
     warehouse: "COMPUTE_WH",
-    databaseName: "SAMPLE_SOURCES",
+    databaseName: "SAMPLE_SOURCES_001",
```
evantahler, line 10: 👍 making a new test database

plugins/@grouparoo/snowflake/__tests__/table-import/import-property.ts
```diff
@@ -283,8 +283,40 @@ describe("snowflake/table/recordProperty", () => {
     //   relativeMatchUnit?: string;
     //   relativeMatchDirection?: string;
     // }
+
+    //TO DO: EXISTS/NOTEXISTS
```
evantahler: TODO is done!

ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
```diff
@@ -49,6 +51,8 @@ export default function Page(props) {
     schedule.recurringFrequency / (60 * 1000)
   );
 
+  console.log(filterOptionDescriptions);
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx]: errant log

ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
```diff
@@ -488,7 +492,7 @@ export default function Page(props) {
                                             <option
                                               key={`op-opt-${localFilter.key}-${op}`}
                                             >
-                                              {op}
+                                              {filterOptionDescriptions[op]}
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx:494]: I think this will change how this form works. In the `onChange()` handler, we are using `e.target.value`, which is now the human-readable name from the dictionary.  I think the option also needs `value=${op}`

ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx
```diff
@@ -626,61 +628,72 @@ export default function Page(props) {
...
-                                          localFilter.match &&
-                                          localFilter.match !== "null"
-                                            ? new Date(
-                                                parseInt(
-                                                  localFilter.match.toString()
+                                    {/* match */}
+                                    {!localFilter.op
+                                      .toLowerCase()
+                                      .includes("exist") &&
+                                      (localFilter.key === "occurredAt" ? (
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx:646]: Can you please add a screenshot or video of how the form works now to the PR?

evantahler (changes requested):
This is a huge PR with really great work!  Figuring out how to build the queries right (`cleanedValues()`) for each plugin is 👍!

There are a few mysterious things going on in some of the tests below and a few questions about the UI, but overall, this is looking really good!

plugins/@grouparoo/postgres/__tests__/table-import/import-properties.ts
teallarson: It is a duplicate of the section starting on line 599

plugins/@grouparoo/postgres/__tests__/table-import/import-property.ts
teallarson: The diff on this is more confusing than helpful for some reason.  I added tests for "exists" and am only testing if against strings.  These tests are still there for "equals" as they should be.

plugins/@grouparoo/postgres/__tests__/table-import/import-property.ts
teallarson: The diff on this is more confusing than helpful for some reason. I added tests for "exists" and am only testing if against strings. These tests are still there for "equals" as they should be.

tealjulia:
@evantahler This is close, but Group Rules use the config object shape
```
 rules: [{
   propertyId: "email",
   operation: {"op": "exists"}
 }]
```
whereas Property and Schedule filters use the shape:
```
 filters: [{
   key: "email",
   "op": "exists"
 }]
```
Should group rules be un-nested?  Or filters be nested?  It seems like they should match up better.

teallarson:
Significant change since the last review: The shape of group rules now more closely matches the shape of property/schedule filters.  

Before, group rules were:
```
[{
  "propertyId": firstName,
  "operation": { "op": "eq"} //core needs this nested in this shape, however it doesn't have any meaning for the user
  "match": "John"
  }]
```
whereas property filters were:
```
[{
  "key": first_name, // should still be key, not propertyId, as we're referencing a source column not a grouparoo property
  "op": "eq",
  "match": "John"
}]
```

Since the purpose of this story is to make these two objects as identical as possible, I've unnested the group rule operation and instead have the configLoader make the operation object into the right shape for core to utilize.

evantahler:
I like the un-nesting you did for group rules - the config objects look simpler now, and I don't think we ever had anything besides `op` within `operation` anyway.  The changes you made were for the code-config files - do you want to go further and change the internal (model) representation of Group Rules as well to bring `op` up out of `operation`?  It noticed it most within the config loader that you were re-building the nested rule structure after loading the new simpler object from the config files: https://github.com/grouparoo/grouparoo/blob/d710d617ed2147b9f6db608e9b3e423d54154d7d/core/src/modules/configLoaders/group.ts#L81-L82.  This would make the internal (model) and external (config file) representations of the object the same.   This could be part of this work, another PR, or we could skip it.

teallarson:
> I like the un-nesting you did for group rules - the config objects look simpler now, and I don't think we ever had anything besides `op` within `operation` anyway. The changes you made were for the code-config files - do you want to go further and change the internal (model) representation of Group Rules as well to bring `op` up out of `operation`? It noticed it most within the config loader that you were re-building the nested rule structure after loading the new simpler object from the config files:
> 
> https://github.com/grouparoo/grouparoo/blob/d710d617ed2147b9f6db608e9b3e423d54154d7d/core/src/modules/configLoaders/group.ts#L81-L82
> 
> . This would make the internal (model) and external (config file) representations of the object the same. This could be part of this work, another PR, or we could skip it.

In other parts of core, the `operation` object is part of the `GroupRuleWithKey` object and has other properties in addition to `op`, such as `description`.  If we do want to change the object's shape, it seems like it may be a good idea to make it a PR of its own.  I looked into whether or not to do that with this PR, however because it seemed like there were quite a few layers of the codebase that would be affected by restructuring the group rule object (model, modules, actions, etc.) I opted not to at this time.  Open to it, just unsure whether we want this (already pretty hefty) PR to keep growing.

evantahler:
> In other parts of core, the operation object is part of the GroupRuleWithKey object and has other properties in addition to op, such as description. If we do want to change the object's shape, it seems like it may be a good idea to make it a PR of its own. I looked into whether or not to do that with this PR, however because it seemed like there were quite a few layers of the codebase that would be affected by restructuring the group rule object (model, modules, actions, etc.) I opted not to at this time. Open to it, just unsure whether we want this (already pretty hefty) PR to keep growing.

Makes sense... And I guess that explains the history of this nested `Operation -> Op` object.  This PR is 👍 and we can talk about changing the internals of core later then.

evantahler (approved):

core/src/classes/codeConfig.ts
```diff
@@ -61,10 +65,21 @@ export interface DestinationConfigurationObject extends ConfigurationObject {
   destinationGroupMemberships?: { [key: string]: string };
 }
 
+export interface GroupRuleConfigurationObject {
```
evantahler, line 68: 👍 new type

evantahler (approved):

PR #2675, opened 2021-12-12
Use Toasts for Success and Error Alerts

evantahler:
This PR swaps our rather aggressive ErrorAlert and SuccessAlerts for Toasts!    These toasts are 'pinned' to the UI  (`position:fixed`) so you see them no matter where you are in the UI.  Like before, they fade away after 4 seconds.

Screen Shot 2021-12-12 at 2 47 30 PM
Screen Shot 2021-12-12 at 2 47 34 PM
Screen Shot 2021-12-12 at 2 47 52 PM

This PR collapsed the APIs of `SucessHandler` and `ErrorHandler` to be the same, accepting the "message" to show to the user.  This PR does not modify any of the content of these messages, except for removing a starting `Error: ` on error messages

---

For Reference, here's what things used to look like:

 
Screen Shot 2021-12-12 at 2 56 14 PM

### Development

### Impact

### Code review


ui/ui-components/components/alerts/Toast.tsx
```diff
@@ -0,0 +1,62 @@
...
+    };
+  }, []);
+
+  function subscription({ message }) {
+    setMessage(formatMessage(message));
+    setShow(true);
+  }
+
+  function formatMessage(message: unknown) {
+    let formattedMessage = String(message);
```
edmundito: I'm concerned that we don't know what the message is. Casting as String may throw an error at some point. Is there a way the type of message can be more specific?

ui/ui-components/components/alerts/Toast.tsx
```diff
@@ -0,0 +1,62 @@
...
+  variant,
+}: {
+  handler: SuccessHandler | ErrorHandler;
+  variant: "success" | "danger";
+}) {
+  const [message, setMessage] = useState("");
+  const [show, setShow] = useState(false);
+
+  useEffect(() => {
+    handler.subscribe(subscriptionKey, subscription.bind(this));
```
edmundito: There's no need to bind the subscription function if it's is an arrow function.

ui/ui-components/components/alerts/Toast.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    return () => {
+      handler.unsubscribe(subscriptionKey);
+    };
+  }, []);
+
+  function subscription({ message }) {
```
edmundito: Because useEffect depends on subscription, this should be a callback hook or a lambda inside the useEffect itself.

ui/ui-components/components/alerts/Toast.tsx
```diff
@@ -0,0 +1,62 @@
...
+    setMessage(formatMessage(message));
+    setShow(true);
+  }
+
+  function formatMessage(message: unknown) {
```
edmundito: this function can live outside the component function because it does not depend on anything related to the component.

edmundito (commented):
Great to see this, looking much better than before.

evantahler:
Updated in 2610ca09b1f7f1f9e31a72e4567147e09fc6859c with some better types and hook logic

ui/ui-components/components/alerts/Toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+function formatMessage(message: string | Error) {
+  let formattedMessage = String(message);
+  if (formattedMessage.startsWith("Error: ")) {
+    formattedMessage = formattedMessage.replace("Error: ", "");
+  }
+
+  return formattedMessage;
+}
```
krishnaglick, line 63: Nitpick. This could be:
`const formatMessage = (message: string | Error) => String(message).replace("Error: ", "");`

krishnaglick (approved):
Seems fine, I'd be inclined to stringify the errors on the way in but we might want to keep them intact in case we need them later.

ui/ui-components/components/alerts/Toast.tsx
evantahler: The `startsWith` check seemed important, as I didn't want to mess with `There was an Error: you broke it` (unlikely with that capital E in there, but you never know... it's an error).

PR #2677, opened 2021-12-13
Replace Record Preview in Property Editor with Sample Record Card

edmundito:
## Change description

This change replaces the record preview in the property editor with a reusable sample record card.

The sample record card component has been updated to be able to fetch data from different sources, highlight a property, and reload when a key is invalidated. Both the model overview and the property editor use the same sample record card with small customization.

image

Error state:
Screen Shot 2021-12-13 at 4 27 55 PM

TODO:
- [x] Verify that all the functionality from the record preview is available from the sample record
- [x] Show record preview error message
- [x] Remove `RecordPreview` component
- [x] Resolve minor layout issues

### Development

### Impact

### Code review


ui/ui-components/components/property/PropertySampleRecord.tsx
```diff
@@ -0,0 +1,87 @@
...
+    ]);
+  }, [
+    debouncedProperty.id,
+    debouncedProperty.type,
+    debouncedProperty.unique,
+    debouncedProperty.isArray,
+    JSON.stringify(debouncedProperty.options),
+    JSON.stringify(debouncedProperty.filters),
+    JSON.stringify(localFilters),
+  ]);
```
edmundito: This logic comes from the `RecordPreview` component. There's a couple of options to improve this:

a. Use https://www.npmjs.com/package/fast-json-stringify over `JSON.stringify`
b. Introduce https://www.npmjs.com/package/use-deep-compare

Thoughts?

core/src/actions/properties.ts
```diff
@@ -381,7 +382,19 @@ export class PropertyRecordPreview extends AuthenticatedAction {
       updatedAt: null,
     };
 
-    return { errorMessage, record: apiData };
+    const groups = await record.$get("groups");
+    const destinations = await DestinationOps.relevantFor(record, groups);
+
+    await ConfigWriter.run();
```
evantahler: We should not write config files when previewing a record

core/src/actions/properties.ts
```diff
@@ -381,7 +382,19 @@ export class PropertyRecordPreview extends AuthenticatedAction {
       updatedAt: null,
     };
 
-    return { errorMessage, record: apiData };
+    const groups = await record.$get("groups");
+    const destinations = await DestinationOps.relevantFor(record, groups);
```
evantahler, line 386: 👍 Makes sense that we need more data to fill out the preview card

ui/ui-components/components/property/PropertySampleRecord.tsx
```diff
@@ -0,0 +1,87 @@
...
+          recordId,
+        }
+      );
+
+      if (response?.record) {
+        setErrorMessage(
+          response.errorMessage
+            ? response.errorMessage.match(
+                /is required for a property of type/ // ignore errors about missing options
+              )
```
evantahler: I'll be this happens when you are filling out a new property and you haven't provided all the options yet.  Is there maybe a way not to hit the API until all the (required) options have a value in the UI?

ui/ui-components/hooks/useDebouncedValue.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from "react";
```
evantahler, line 1: Nice - this generic debounce component will be helpful!

ui/ui-components/components/property/PropertySampleRecord.tsx
evantahler: Yeah... the need to stringify these objects to notice changes took a long time to debug, and this solution certainly isn't great.  

As an alternative approach, what if we had a new `lastChangeAt` timestamp useState variable on the parent form.  We could bind to any change on the form (`form.addEventListener('change',() => setLastChangeAd(new Date())`) and then we could just use this timestamp to know it was time to reload

evantahler: Or... I guess we could just stringify the whole thing -> `JSON.stringify(debouncedProperty)`.  That would probably also get the job done

core/__tests__/actions/properties.ts
```diff
@@ -373,12 +373,15 @@ describe("actions/properties", () => {
...
-      const { error, record } =
+      const { error, record, groups, destinations } =
         await specHelper.runAction<PropertyRecordPreview>(
           "property:recordPreview",
           connection
         );
+
       expect(error).toBeUndefined();
+      expect(Array.isArray(groups)).toBeTruthy();
+      expect(Array.isArray(destinations)).toBeDefined();
```
evantahler: Probably better to test the structure here, even if there is no data:

```ts
expect(groups).toEqual([])
expect(destinations).toEqual([])
```

evantahler (changes requested):

core/src/actions/properties.ts
edmundito: Whoops, fix coming up. I pulled that out of records:create by mistake.

ui/ui-components/components/property/PropertySampleRecord.tsx
edmundito: I looked into this and it's fairly challenging to get it right because we'd have to do some client-side validation that matches the form/submission. Besides, the rest of the sample record (other props, groups, dests) still get populated by this call so it's worth making the request.

edmundito:
Addressed @evantahler 's comments, fixed a few layout issues and this should be ready for review.

ui/ui-components/__tests__/components/recordPreview.tsx
```diff
@@ -1,165 +0,0 @@
-import RecordPreview from "../../components/property/RecordPreview";
```
evantahler, line 1: 👋 Bye!

evantahler (approved):

PR #2681, opened 2021-12-14
Remove Manual Groups

evantahler:
To simplify & clarify Grouparoo, this PR removes the concept of "manual" Groups.  From this point forward, all Grouparoo Groups are "calculated", meaning that they dynamically determine which Records are members based on the Properties of those Records.  This has the effect of simplifying many things from the UI, configuration files, and internal logic.  This is a breaking change.

After the application of this PR, the following breaking changes will happen:
* The migration will fail if there are manual Groups in the database.  These will need to be removed before the migration can be applied
* If there are manual Groups in code config, Grouparoo will fail to validate/apply/start/run with an error message
* If there are calculated Groups in code config, Grouparoo will boot, and provide a deprecation warning that `group.type` is no longer a property of a Grouparoo group.

Docs @ https://github.com/grouparoo/www.grouparoo.com/pull/828

### Development

### Impact

### Code review


core/src/models/GroupRule.ts
```diff
@@ -37,17 +43,17 @@ export class GroupRule extends CommonModel<GroupRule> {
...
 
-  @Column
-  relativeMatchUnit: string;
+  @Column(DataType.STRING)
+  relativeMatchUnit: RelativeMatchUnitType;
 
-  @Column
-  relativeMatchDirection: string;
+  @Column(DataType.STRING)
+  relativeMatchDirection: RelativeMatchDirectionType;
```
evantahler, line 56: Adds better TS Types

core/src/models/Group.ts
```diff
@@ -50,14 +55,13 @@ export interface GroupRuleWithKey {
...
   type?: string;
   topLevel?: boolean;
-  operation: { op: string; description?: string };
+  operation: { op: GroupRuleOpType; description?: string };
   match?: string | number | boolean;
   relativeMatchNumber?: number;
-  relativeMatchUnit?: string;
-  relativeMatchDirection?: string;
+  relativeMatchUnit?: RelativeMatchUnitType;
+  relativeMatchDirection?: RelativeMatchDirectionType;
```
evantahler, line 62: Adds better TS Types

core/src/modules/configLoaders/group.ts
```diff
@@ -11,14 +11,15 @@ import { Group } from "../../models/Group";
...
+import { Deprecation } from "../deprecation";
 
 export async function loadGroup(
   configObject: GroupConfigurationObject,
   externallyValidate: boolean,
   validate = false
 ): Promise<IdsByClass> {
   let isNew = false;
-  validateConfigObjectKeys(Group, configObject, ["rules"]);
+  validateConfigObjectKeys(Group, configObject, ["rules", "type"]);
```
evantahler, line 22: For now, we won't crash if `group.type` is provided on a config object, but we check it below

core/src/modules/ruleOpsDictionary.ts
```diff
@@ -85,3 +99,12 @@ export function buildPropertyOpsDictionary(cfg?: any) {
...
 export const PropertyOpsDictionary = buildPropertyOpsDictionary(config || {});
+
+export type GroupRuleOpType =
+  | typeof _boolean_ops[number]["op"]
+  | typeof _number_ops[number]["op"]
+  | typeof _string_ops_sqlite[number]["op"]
+  | typeof _string_ops_postgres[number]["op"]
+  | typeof _date_ops[number]["op"];
+export type RelativeMatchUnitType = typeof _relativeMatchUnits[number];
+export type RelativeMatchDirectionType = "add" | "subtract";
```
evantahler, line 110: Finally providing types for the Group Rules

core/__tests__/integration/happyPath.ts
```diff
@@ -259,90 +255,10 @@ describe("integration/happyPath", () => {
...
-      expect(simpleRecordValues(records[0].properties).email).toEqual([
-        "luigi@example.com",
-      ]);
-    });
-  });
-
-  describe("calculated group", () => {
-    let groupId: string;
-
     test("a calculated group can be created", async () => {
```
teallarson:
```suggestion
    test("a group can be created", async () => {
```

pedroslopez (approved):
This is a big one! Looks good though and I think the migration path is clear.

teallarson (approved):
Bye Manual Groups!  I found one spot that still used "Calculated Group" instead of "Group" in a test, but other than that looks good.

PR #2686, opened 2021-12-15
Pipedrive add parallelism method to the app

aelafifi:
## Change description

Add parallelism method to the app

### Development

### Impact

### Code review

### DIFF
--- a/plugins/@grouparoo/pipedrive/src/initializers/plugin.ts
+++ b/plugins/@grouparoo/pipedrive/src/initializers/plugin.ts
@@ -3,6 +3,7 @@ import { Initializer } from "actionhero";
 import { DestinationSyncMode, plugin } from "@grouparoo/core";
 
 import { test } from "../lib/test";
+import { parallelism } from "../lib/parallelism";
 import { AppTemplate } from "@grouparoo/app-templates/dist/app";
 import { DestinationTemplate } from "@grouparoo/app-templates/dist/destination/templates";
 
@@ -53,7 +54,7 @@ export class Plugins extends Initializer {
               description: "Pipedrive API token",
             },
           ],
-          methods: { test },
+          methods: { test, parallelism },
         },
       ],
       connections: [
--- a/plugins/@grouparoo/pipedrive/src/lib/client.ts
+++ b/plugins/@grouparoo/pipedrive/src/lib/client.ts
@@ -1,4 +1,6 @@
 import axios, { AxiosInstance } from "axios";
+import { objectCache } from "@grouparoo/core";
+import { PipedriveCacheData } from "./common/destinationMappingOptions";
 
 type Entity = "persons" | "organizations";
 type EntityField = "personFields" | "organizationFields";
@@ -151,7 +153,7 @@ export class PipedriveClient {
   persons: EntityCalls;
   organizations: EntityCalls;
 
-  constructor(apiToken: string) {
+  constructor(apiToken: string, cacheData: PipedriveCacheData) {
     this.request = axios.create({
       baseURL: "https://api.pipedrive.com/v1",
       params: {
@@ -160,7 +162,18 @@ export class PipedriveClient {
     });
 
     this.request.interceptors.response.use(
-      (response) => response,
+      async (response) => {
+        await objectCache(
+          {
+            objectId: cacheData.appId,
+            cacheKey: ["pipedrive:parallelism", cacheData.appOptions],
+            cacheDurationMs: 2000, // 2 seconds
+            read: false, // Write-only
+          },
+          async () => parseInt(response.headers["x-ratelimit-remaining"])
+        );
+        return response;
+      },
       (error) => {
         if (error.code === "ECONNRESET") {
           // console.log("ECONNRESET, retry after 2 seconds...");
--- a/plugins/@grouparoo/pipedrive/src/lib/client.ts
+++ b/plugins/@grouparoo/pipedrive/src/lib/client.ts
@@ -167,7 +167,8 @@ export class PipedriveClient {
           {
             objectId: cacheData.appId,
             cacheKey: ["pipedrive:parallelism", cacheData.appOptions],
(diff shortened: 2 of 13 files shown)


plugins/@grouparoo/pipedrive/src/lib/client.ts
```diff
@@ -160,7 +162,18 @@ export class PipedriveClient {
     });
 
     this.request.interceptors.response.use(
-      (response) => response,
+      async (response) => {
+        await objectCache(
+          {
+            objectId: cacheData.appId,
+            cacheKey: ["pipedrive:parallelism", cacheData.appOptions],
+            cacheDurationMs: 2000, // 2 seconds
```
bleonard: one of the headers actually says how many seconds. Maybe we should use that.

plugins/@grouparoo/pipedrive/src/lib/parallelism.ts
```diff
@@ -0,0 +1,24 @@
+import { AppParallelismMethod, objectCache } from "@grouparoo/core";
+
+const DEFAULT_VALUE = 10;
+
+export const parallelism: AppParallelismMethod = async ({
+  app,
+  appOptions,
+}) => {
+  const cachedValue = await objectCache(
```
bleonard, line 9: Let's move the read of this value to the PipedriveClient, so it's more likely to share the same key and access method.

plugins/@grouparoo/pipedrive/src/lib/client.ts
```diff
@@ -160,7 +162,18 @@ export class PipedriveClient {
...
+            read: false, // Write-only
+          },
+          async () => parseInt(response.headers["x-ratelimit-remaining"])
```
bleonard, line 174: Note to myself: how many remaining requests is the value.

plugins/@grouparoo/pipedrive/src/lib/parallelism.ts
```diff
@@ -0,0 +1,24 @@
...
+      write: false, // Read-only
+    },
+    async () => null
+  );
+
+  if (cachedValue === null) {
+    return DEFAULT_VALUE;
+  }
+
+  return parseInt(cachedValue);
```
bleonard: So this will be the number of remaining requests in our period. I think we want to divide that my the number of requests that the `exportRecord` method will do.

1. findPersonIdByEmail
2. delete/update/create
3. Maybe a few to handle group creation

So let's divide by `3` here to be on the safe side.

plugins/@grouparoo/pipedrive/src/lib/parallelism.ts
```diff
@@ -0,0 +1,24 @@
+import { AppParallelismMethod, objectCache } from "@grouparoo/core";
+
+const DEFAULT_VALUE = 10;
```
bleonard: Let's make this `7` here. Then we will fan out to more once we learn the plan they are on.
This is the lowest plan (20) divided by 3.

plugins/@grouparoo/pipedrive/src/lib/parallelism.ts
bleonard: And floor it.
`return Math.floor(parseInt(cachedValue) / 3.0)` for safety

bleonard:
Let's also do this to handle the the daily limit cutoff in `getExportRecordWithErrorHandling`

```
    if (error?.response?.status === 429) {
      const retryIn = Math.floor(Math.random() * 10) + 1;
      // update this to look for header `x-daily-requests-left` 
      // to see if we are done for the day
      // set retryIn = 24 * 60 * 60
      return { error, success: false, retryDelay: 1000 * retryIn };
    }
```

plugins/@grouparoo/pipedrive/src/lib/parallelism.ts
aelafifi: But the `parallelism` method doesn't receive client instance as one of its parameters!
How could we call it from there?

plugins/@grouparoo/pipedrive/src/lib/common/exportRecord.ts
```diff
@@ -14,8 +14,17 @@ export const getExportRecordWithErrorHandling: (
     return await callback(args);
   } catch (error) {
     // look for the rate limit exceeded status code.
+    let retryIn = Math.floor(Math.random() * 10) + 1;
+
     if (error?.response?.status === 429) {
-      const retryIn = Math.floor(Math.random() * 10) + 1;
+      const dailyRequestsLeftStr =
```
bleonard, line 20: 👍

bleonard (approved):
Looks good. Did you test it with real data on the `roo demo` dataset?

evantahler:
Testing this, I also was getting a lot of 429 errors, and didn't think things were working right.  It turns out that this plugin didn't have the `parallelism` method exported as part of the app!  I did that in (https://github.com/grouparoo/grouparoo/pull/2686/commits/314c5b1fa7ec90842d5cf09fbedefba2c683f578), along with a rebase from main, and now things look great.  1000 people uploaded to Pipedrive.   

This pattern is a bit wasteful with all the retries the exports do, but at least we get them all done now.  This is 👍 from me for a first pass.

plugins/@grouparoo/pipedrive/src/initializers/plugin.ts
```diff
@@ -53,7 +54,7 @@ export class Plugins extends Initializer {
               description: "Pipedrive API token",
             },
           ],
-          methods: { test },
+          methods: { test, parallelism },
```
evantahler, line 57: This was the missing part!

evantahler (approved):

plugins/@grouparoo/pipedrive/src/initializers/plugin.ts
aelafifi: Oooops!

PR #2701, opened 2021-12-20
#2551 Add Airtable as a Destination

### LINKED ISSUE #2551: [Feature]: Add Airtable as a Destination

Add Airtable bases and tables as destinations to Grouparoo

An Airtable plugin maintained by Grouparoo that allows users to sync arbitrary data from sources to Airtable tables

Alternatives to a grouparoo maintained solution would be one maintained by an external contributor/ a plugin in a seperate repository

### Additional context

This is currently being used as a solution with our team since the team we are interfacing with already has all of their data in Airtable and wants to incorporate data from our data warehouse.

brandon-segal:
## Change description

Addresses issue #2551 . 
Adds Airtable as a destination for Grouparoo. The Airtable application takes in an API Key host and base Id from the user and the destination options ask for a table and foreign key that should be used to designate the exact table that data will be sent to and what is going to be used as the 

### Development

### Impact

### Code review


plugins/@grouparoo/airtable/LICENSE.txt
```diff
@@ -0,0 +1,373 @@
+Mozilla Public License Version 2.0
```
evantahler, line 1: 👍 MOZ2 licence.

plugins/@grouparoo/airtable/__tests__/lib/destination/fixtures.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  },
+  {
+    fields: successfulGetTableFields,
+    id: 'table3',
+    name: 'table 3',
+    primaryFieldId: primaryFieldID,
+  },
+];
+
+test.todo("Fixtures don't need tests y'all");
```
evantahler, line 77: 😆

plugins/@grouparoo/airtable/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "test": "jest",
+    "pretest": "npm run lint && npm run prepare",
+    "lint": "prettier --check src ",
+    "format": "prettier --write src",
+    "watch": "tsc --watch"
+  },
+  "dependencies": {
+    "@grouparoo/app-templates": "0.7.6",
+    "@grouparoo/core": "0.7.6",
+    "actionhero": "27.2.0",
```
evantahler: Please rebase this branch on main and update these versions

plugins/@grouparoo/airtable/src/lib/client/client.ts
```diff
@@ -0,0 +1,188 @@
...
+  async createRecords(tableId: string, records: CreateRecord<FieldSet>[]) {
+    return this.baseClient(tableId).create(records);
+  }
+
+  /**
+   * NOOP method for mocking out a clossing of the client
+   */
+  async close(): Promise<void> {
+    return Promise.resolve();
+  }
```
evantahler: You can skip the method if you want, and then not implement the disconnect method on the grouparoo plugin

plugins/@grouparoo/airtable/src/lib/client/client.ts
```diff
@@ -0,0 +1,188 @@
+import Airtable, { FieldSet, Record, RecordData, Records } from "airtable";
```
evantahler, line 1: This is a very well typed client!
I like how you extended the `airbase` client with the missing methods you needed

plugins/@grouparoo/airtable/src/lib/destination/destinationMappingOptions.ts
```diff
@@ -0,0 +1,107 @@
...
+    case AirtablePropertyTypes.EMAIL:
+      return "email";
+    case AirtablePropertyTypes.MULTILINE:
+      return "string";
+    case AirtablePropertyTypes.SINGLELINE:
+      return "string";
+    default:
+      return "string";
+  }
+}
```
evantahler: 👍 good mappings

plugins/@grouparoo/airtable/src/lib/destination/exportArrayProperties.ts
```diff
@@ -0,0 +1,19 @@
+import { ExportArrayPropertiesMethod } from "@grouparoo/core";
```
evantahler, line 1: This may be one of the first destinations that supports array properties in a meaningful way!

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
```diff
@@ -0,0 +1,244 @@
...
+  destinationOptions: SimpleDestinationOptions;
+  syncOperations?: DestinationSyncOperations;
+  exports: BatchExport[];
+}
+export async function exportBatch(exportBatchOptions: ExportBatchOptions) {
+  const { syncOperations, appOptions, destinationOptions, exports } =
+    exportBatchOptions;
+  const { primaryKey } = destinationOptions;
+  const batchSize = 10;
+  const findSize = 100;
```
evantahler: Where did these constant values come from?

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
```diff
@@ -0,0 +1,244 @@
...
+  };
+}
+
+const addToGroups: BatchMethodAddToGroups = async () => {
+  // This destination doesn't do groups
+};
+
+const removeFromGroups: BatchMethodRemoveFromGroups = async () => {
+  // This destination doesn't do groups
+};
```
evantahler, line 180: If we aren't going to use all the facilities of the batch helper, does it make sense to use it here?  cc @bleonard

evantahler (commented):
Really awesome work @brandon-segal! 

That's some really strong testing and typing work.  I believe you have been working closely with @bleonard, so let's wait for his review as well.

As a note, don't worry about the REDSHIFT test failing (we don't share that ENV with forks, it's a real redshift test database).  For the failing linter test, you can run `./node_modules/.bin/prettier --fix src` to have prettier fix up your files for you

plugins/@grouparoo/airtable/src/lib/appOptions.ts
```diff
@@ -0,0 +1,43 @@
...
+    displayName: "Airtable Base",
+    required: true,
+    description: "Airtable Base to read tables from",
+  },
+  {
+    key: "airtableHost",
+    type: "text",
+    displayName: "Airtable API Host",
+    required: false,
+    defaultValue: "https://api.airtable.com",
```
bleonard: maybe use the `DEFAULT_AIRTABLE_API` from above.

plugins/@grouparoo/airtable/src/lib/destination/destinationOptions.ts
```diff
@@ -0,0 +1,99 @@
...
+  private async getSchema(tableId: string): Promise<string[]> {
+    return this.getTableFields(tableId);
+  }
+
+  private async getRecordOptions(
+    destinationOptions: AirtableDestinationOptions
+  ): Promise<DestinationOptionsMethodResponse> {
+    const out: DestinationOptionsMethodResponse = {
+      tableId: { type: "list", options: [], descriptions: [] },
+      primaryKey: { type: "pending", options: [] },
```
bleonard, line 53: good use of `pending`

plugins/@grouparoo/airtable/src/lib/destination/destinationOptions.ts
```diff
@@ -0,0 +1,99 @@
...
+      .sort((a, b) => a.description.localeCompare(b.description));
+
+    out.tableId.options = sortedTables.map((value) => value.option);
+    out.tableId.descriptions = sortedTables.map((value) => value.description);
+
+    const tableId = destinationOptions.tableId;
+    if (tableId) {
+      const fields = await this.getSchema(tableId);
+      out.primaryKey.type = "typeahead";
+      out.primaryKey.options = fields;
```
bleonard, line 72: we'd want to only check ones we can write here. Aren't some calculated?
Ideally, also making sure they map to Grouparoo types, though everything seems to map for now.

plugins/@grouparoo/airtable/src/lib/destination/destinationMappingOptions.ts
```diff
@@ -0,0 +1,107 @@
...
+  const out: Array<{
+    key: string;
+    type: DestinationMappingOptionsResponseType;
+  }> = [];
+  for (const field of fields) {
+    if (!requiredFields.includes(field.name)) {
+      const type: DestinationMappingOptionsResponseType =
+        mapTypesFromAirtableToGrouparoo(field.type);
+      if (type) {
+        out.push({
```
bleonard, line 78: came here as on the destination option --- these have to be ones we can write. is it on the response?

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
```diff
@@ -0,0 +1,244 @@
...
+    if (found) {
+      found.destinationId = record.id;
+      found.result = record.fields;
+    }
+  });
+};
+
+function attachErrorsToErroneousRecords(
+  records: BatchExport[],
+  error: AirtableError
```
bleonard, line 145: is there anything on this error that says which `record id` failed? Or is it always all or nothing on the batch? Or some other scheme?

bleonard: In the ideal case, some of theme succeed and there is a record-specific error on each one.

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
```diff
@@ -0,0 +1,244 @@
...
+}) => {
+  const batchExports = buildBatchExports(recordsToExport);
+  try {
+    return await exportBatch({
+      appOptions,
+      destinationOptions,
+      syncOperations,
+      exports: batchExports,
+    });
+  } catch (error) {
```
bleonard, line 248: probably not needed here. though this is the place where we'd do `429` rate-limiting kinds of things. ever see anything like that?

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
brandon-segal: Unfortunately there is not much of a helpful error that is given back from the client. It seems that if any of the records error than the entire batch errors. I agree, ideally if I had control of the API I'd have it reply with a 207 multi response and have the client parse the response.

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
brandon-segal: I was thinking about doing that but this is currently done in the Airtable client supposedly internally and the actions not handled by their client would only be done with human involvement during initial configuration. 
[Link to exponential backoff implementation in Airtable JS repository](https://github.com/Airtable/airtable.js/blob/9d40666979af77de9546d43177cab086a03028bf/src/run_action.ts#L67)

plugins/@grouparoo/airtable/src/lib/destination/destinationOptions.ts
brandon-segal: I see. I can add a guard there when getting possible fields that only returns fields that are single line of text, dates, and emails

plugins/@grouparoo/airtable/src/lib/destination/destinationMappingOptions.ts
brandon-segal: I can have the `mapTypesFromAirtabletoGrouparoo` method return a null by default. In that case if a field doesn't match the expected writable fields they will not be included in the selection

brandon-segal:
Not sure why its failing in CI but when I run their command locally it shows that it is not failing Linting: 
Output from my side
```
pnpm run --stream -r lint --workspace-concurrency=2 --filter !'@grouparoo/grouparoo'
Scope: 48 of 49 workspace projects
plugins/@grouparoo/airtable lint$ prettier --check src 
cli lint$ prettier --check src
plugins/@grouparoo/airtable lint: Checking formatting...
plugins/@grouparoo/airtable lint: All matched files use Prettier code style!
plugins/@grouparoo/airtable lint: Done
```

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
brandon-segal: Added a constant for these values. They should both be 10 for now since that is the amount of records that the Airtable Team recommends for batching requests

evantahler:
Note to future @evantahler: When this is merged, we'll need to update the version in `package.json`

plugins/@grouparoo/airtable/src/lib/disconnect.ts
```diff
@@ -4,5 +4,5 @@ import { IClient } from "./client/interfaces/iClient";
 export const disconnect: DisconnectPluginAppMethod<IClient> = async ({
   connection,
 }) => {
-  await connection.close();
+  return;
 };
```
evantahler, line 8: I mean to say that you can just skip this file entierly if you don't really need to disconnect, and remove it from the initializer too

plugins/@grouparoo/airtable/src/lib/appTest.ts
```diff
@@ -0,0 +1,11 @@
+import { TestPluginMethod } from "@grouparoo/core";
+import { IClient } from "./client/interfaces/iClient";
+
+export const appTest: TestPluginMethod<IClient> = async ({ connection }) => {
+  try {
+    const { statusCode, body } = await connection.health();
+    return { success: statusCode === 200, body };
+  } catch (exception) {
+    return { success: false, message: exception.toString() };
```
evantahler, line 9: It's a shame that we don't get more info from the API when something goes wrong:
Screen Shot 2021-12-21 at 2 34 37 PM

plugins/@grouparoo/airtable/src/lib/appTest.ts
```diff
@@ -0,0 +1,11 @@
+import { TestPluginMethod } from "@grouparoo/core";
```
evantahler, line 1: convention is to name this file `test` (the method name should be the same, as that's the part of the plugin interface that it is implementing)

plugins/@grouparoo/airtable/src/lib/client/client.ts
```diff
@@ -0,0 +1,181 @@
...
+          statusCode: resp.status,
+        };
+      })
+      .catch((reason: AxiosError) => {
+        return {
+          body: reason.response.data,
+          statusCode: reason.response.status,
+        };
+      });
+  }
```
evantahler, line 50: This method is going to be a problem.  Right now, Airbase is not granting new API access tokens to the `meta` API.  So for me, I can't pass the health check.  It's a shame, because there's no way for me to test this right now.

The same problem exists for the `listTables()` API below.  At the very least, we need to note somewhere that this plugin requires [`meta` API access](https://airtable.com/shrWl6yu8cI8C5Dh3)

evantahler (commented):
I can't finish checking this PR until we get `meta` API access from Airtable...

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
bleonard: We could make it optional in `app-templates` later. This seems fine for now.

bleonard:
Nothing this on my base: 
> The API is limited to 5 requests per second per base. If you exceed this rate, you will receive a 429 status code and will need to wait 30 seconds before subsequent requests will succeed.

That's pretty aggressive. Seen anything like that @brandon-segal ?

plugins/@grouparoo/airtable/src/lib/destination/destinationOptions.ts
bleonard: Many APIs have a `calculated: true` kind of idea, but I don't see that. Just types.
Here's the list: https://airtable.com/api/meta
Ones that definitely look calculated:
* formula
* lastModifiedBy
* lastModifiedTime
* rollup?
* autoNumber?

A lot of the other ones more than you mentioned look settable (rating?, checkbox?, multilineText).

We've found it's less frustrating for people to include too much instead of leaving one out, so probably have a blacklist of types to leave out.

bleonard: Oh wait! They have `Permission levels` on there.
That should do it.

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
bleonard: hmm, we'd want to handle that probably. otherwise, they'd tie up the thread.
Tjhis is fine for now, but we can investigate setting `_noRetryIfRateLimited` later.

evantahler:
@brandon-segal so it looks like no one at Grouparoo has access to the `meta` API from Airtable, and it might be a while until they allow access to new users:

> Thanks for requesting access to the Airtable Metadata API! We're currently pausing generating new metadata API tokens. We'll keep this request on file however, and reach out with additional information as it becomes available.

This suggests that perhaps this plugin should be modified to ask users for more information rather than using the meta API (the custom Axios requests you are making).   Perhaps the path forward is:

App Option:
* Just the APIKey

Destination Option:
* BaseID
* TableID

Also, we'll need to come up with some other way to to implement the `test` method... I'm really surprised that there isn't a 'user' or 'ping' API endpoint - those have been common across the board.  We could host publicly available base we could hard-code as a read test maybe if all else fails

bleonard:
Moved to here: https://github.com/grouparoo/grouparoo/pull/2716

PR #2703, opened 2021-12-21
Rebuild state from props when changing from one property edit page to another

evantahler:
Fixes a bug when using the new Property Preview widget when switching from one property page to another.  We store the initial page props in local state variables (so they can be edited by the user), and we need to reset those with the new props.

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx
+++ b/ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx
@@ -66,8 +66,11 @@ export default function Page(props) {
   if (hydrationError) errorHandler.set({ message: hydrationError });
 
   useEffect(() => {
+    setProperty(props.property);
+    setPluginOptions(props.pluginOptions);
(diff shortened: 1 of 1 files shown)


ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx
```diff
@@ -66,16 +66,19 @@ export default function Page(props) {
...
+  }, [propertyId]);
 
   useEffect(() => {
     updatePluginOptions();
 
     return () => {
       clearTimeout(timer);
     };
-  }, [JSON.stringify(property.options)]);
+  }, [property.id, JSON.stringify(property.options)]);
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx:81]: What's the difference between `propertyId` above and `property.id`?

Comment (evantahler) [reply ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx:81]: I think this hook wants to be called after property is set, which the previous hook will do after the URL (propertyID) changes

edmundito (approved):

PR #2704, opened 2021-12-21
Custom flow for creating primary source

edmundito:
## Change description

Updates the way the first source is created so that it becomes the primary source and prevents the user from creating additional sources.

Changes:

* Mapping is now editable through the source Edit page and the Mapping page is gone!
* Refreshed the look of the Source overview page to be similar to the Model overview page
* When creating the first source, we assume this will be the primary source. Only the apps that have the capability to bootstrap a unique property will be enabled in the flow.
* Cannot create another source until the first source is mapped and set to the `ready` state.
* Fixes logic in the model overview page regarding when the secondary source and the destinations can be added
* `bootstrapUniqueProperty` has been updated to generate an id and guess the type if none of these are provided. This allows the UI side to only specify the source column name and let the API generate the property mapped to it.

New Source overview:
image

Edit Source:
Screen Shot 2021-12-22 at 1 48 26 PM

New Source without unique properties:

edit_source

Video:

https://user-images.githubusercontent.com/168664/147003957-d3674d6b-1df5-4215-a249-12405ab1b153.mov

### Development

### Impact

### Code review


ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -72,27 +115,95 @@ export default function Page(props) {
...
+      { ...source, state, mapping }
     );
+
     if (response?.source) {
       setSource(response.source);
       sourceHandler.set(response.source);
-      if (response.source.state !== "ready") {
+
+      // this source can have a schedule, and we have no schedules yet
+      if (scheduleCount === 0 && response.source.scheduleAvailable) {
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:207]: Creating a schedule automatically and redirecting to the schedule comes directly from the mapping page logic so I moved this over here. However, this only works for the first schedule on the page as the `scheduleCount` is pulled from the schedules in the model and not particularly for this source. Should this logic be the same?

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -16,32 +18,79 @@ import { ErrorHandler } from "../../../../../utils/errorHandler";
...
+  scheduleCount: number;
+  source: Models.SourceType;
+  types: string[];
+}
+
+interface InjectedProps extends NextPageContext {
+  errorHandler: ErrorHandler;
+  successHandler: SuccessHandler;
+  sourceHandler: SourceHandler;
+}
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:47]: This is a temp change to make the props on this page work. I created a chore in our board to make something like this available across all page components, but I did not want to apply it in this PR.

evantahler (commented):
I like the new look of the page!

I think that one of the core parts of the story was to 'guess' what the Grouparoo Property Key should be from the table's column.  In your demo video, you still need to enter that in.  I think this PR should still include 'guessing' what the Grouparoo Property Key should be

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
Comment (evantahler) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:207]: Yeah - I think the logic should be the same.  We know you want a schedule on your primary source... less likely on the secondary sources

edmundito:
I'm feeling good about the state of this. I pushed the ability to bootstrap the unique property without a key and a type, made some layout adjustments to the source editor so that the example data shows on the right on XL screens, and fixed a few issues.

core/src/classes/plugin.ts
```diff
@@ -99,6 +99,11 @@ export interface PluginConnection {
   };
 }
 
+export interface PluginConnectionApiData
+  extends Omit<PluginConnection, "methods"> {
+  methods?: string[];
+}
```
evantahler, line 105: Good update! This won't be the full plugin

core/src/modules/ops/source.ts
```diff
@@ -511,76 +514,119 @@ export namespace SourceOps {
...
 
-      // build the default options
-      const { pluginConnection } = await source.getPlugin();
-      if (!local) {
-        let ruleOptions: SimplePropertyOptions = {};
+    let didGenerateKey = false;
+    const generateKey = (index?: number) =>
+      `${ConfigWriter.generateId(model.name)}_${ConfigWriter.generateId(
+        mappedColumn
+      )}${!index ? "" : "_" + index}`;
```
evantahler, line 536: I think there should be some tests locking in the behavior of this method when there is a conflict (I don't think I saw any of those added)

core/src/modules/ops/source.ts
```diff
@@ -511,76 +514,119 @@ export namespace SourceOps {
...
+    }
 
-        if (propertyOptions) {
-          Object.assign(ruleOptions, propertyOptions);
-        }
+    if (!type) {
+      const preview = await source.sourcePreview(sourceOptions);
+      const samples = preview.map((row) => row[mappedColumn]);
+      type = TableSpeculation.columnType(mappedColumn, samples);
+    }
```
evantahler, line 547: Is there any case where we can't guess?  Default to a string then?

ui/ui-components/components/tabs/Source.tsx
```diff
@@ -5,10 +5,6 @@ import { grouparooUiEdition } from "../../utils/uiEdition";
 export default function SourceTabs({ source }: { source: Models.SourceType }) {
   const tabs = ["overview", "edit"];
 
-  if (source.previewAvailable) {
-    tabs.push("mapping");
-  }
```
evantahler, line 10: 👋 bye!

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -428,44 +590,55 @@ export default function Page(props) {
...
       </Row>
     </>
   );
-}
+};
+
+export default Page;
 
-Page.getInitialProps = async (ctx: NextPageContext) => {
+export const getServerSideProps: GetServerSideProps<Props> = async (ctx) => {
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:611]: 👍 gradually moving us from getInitialProps to getSreverSideProps

core/__tests__/models/source/source.ts
```diff
@@ -847,6 +863,57 @@ describe("models/source", () => {
       await property.destroy();
     });
 
+    test("bootstrapUniqueProperty without key and type will create a new unique property with generated key and guessed type", async () => {
```
evantahler, line 866: Found the tests!

evantahler (approved):
This is great!  I like the API change to the bootstrap methods too - much clerer

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -521,11 +521,19 @@ const Page: NextPage<Props & InjectedProps> = ({
...
+                      this Source. The Property mapped to this column will be
+                      assigned as the Model's Primary Key.
+                    </p>
+                  ) : (
+                    <p>
+                      Select a column that relates each record in this Source
+                      back to the Model's Primary Source. You can map the column
+                      to any Property in another Source in the Model.
+                    </p>
+                  )}
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:536]: I like this language!  As a note, some sources aren't columnar (eg mongo), but I think that this langauge is still clear in those cases

evantahler (approved):

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
Comment (edmundito) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:536]: Yep! that's some context we haven't resolved yet that was part of the original mapping page.

PR #2706, opened 2021-12-21
Clear `destinationId` from exports when deleting related Destination

pedroslopez:
## Change description

After #2515, because we are now keeping old exports around we could get in a case where a deleted destination comes back with the same ID and our old exports would be associated to this new destination. This PR `null`s out `destinationId` for exports when their destination is deleted to prevent this from happening.

Also addresses some issues in the UI with displaying exports with invalid destinations.

Shows `deleted` badge if has no destination on list and detail pages:

image

image

### Development

### Impact

### Code review


core/src/migrations/000094-exportDestinationAllowNull.ts
```diff
@@ -0,0 +1,20 @@
...
+  down: async (
+    queryInterface: Sequelize.QueryInterface,
+    DataTypes: typeof Sequelize
+  ) => {
+    await queryInterface.changeColumn("exports", "destinationId", {
+      type: DataTypes.STRING(191),
+      allowNull: false,
+    });
+  },
+};
```
pedroslopez: I'm not sure if this migration should also `null` out `destinationId`s for exports whose `destinationId` does not exist anymore. It would match the behavior going forward, but the sweepers should still take care of them eventually and going through and confirming that each export's destination still exists might be a long migration if someone has a ton of exports.

core/src/models/Destination.ts
```diff
@@ -740,4 +740,12 @@ export class Destination extends LoggedModel<Destination> {
...
   }
+
+  @AfterDestroy
+  static async unassociateRelatedExports(instance: Destination) {
+    return Export.update(
+      { destinationId: null },
+      { where: { destinationId: instance.id } }
+    );
+  }
 }
```
evantahler, line 751: 👍 this is great as a hook on the destination - a single query to update all the exports

core/src/migrations/000094-exportDestinationAllowNull.ts
```diff
@@ -0,0 +1,20 @@
+import Sequelize, { DataTypes } from "sequelize";
```
evantahler: I believe that https://github.com/grouparoo/grouparoo/pull/2672 has migration 94.  This should be 95.

core/src/migrations/000094-exportDestinationAllowNull.ts
evantahler: 👍 This is a "going-forward" fix.

evantahler (approved):
This is 👍 once the migration name is changed

PR #2668, opened 2021-12-09
add space after app environment var options

pedroslopez:
## Change description

Ran into a mild annoyance trying to copy/paste the avilable env options into the text boxes. Double clicking to select an option would select all options. Adding a space after each one fixes it and lets you select a single one for easy copy/pasting 😄 

**Before**

copy-env-opts-before

**After**

copy-env-opts-after

### Development

- [ ] Application changes have been tested appropriately

### Impact

- [x] Code follows company security practices and guidelines
- [x] Security impact of change has been considered
- [x] Performance impact of change has been considered
- [x] Possible migration needs considered (model migrations, config migrations, etc.)

Please explain any security, performance, migration, or other impacts if relevant:

### Code review

- [x] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.
- [x] Relevant tags have been added to the PR (bug, enhancement, internal, etc.)

PR #2708, opened 2021-12-22
Ignore empty strings when setting options

pedroslopez:
## Change description

When manually deleting option values through the UI, an empty string was being saved. This meant that things like default values were not being applied in these cases, because there was technically a value. 

This changes the behavior so when setting options to any of the supported models, empty strings are ignored.

There's also a fix for a `missing key` error on the app edit page that was introduced by #2668 (the key had to be moved up to the fragment)

### Development

### Impact

### Code review


evantahler (approved):
Love it

PR #2709, opened 2021-12-22
Update Salesforce Destination options when in enrich sync mode

pauloouriques:
## Change description

This PR contains the code responsible for when the syncMode is "Enrich", only require the primary keys in the data mapping.

### Development

### Impact

### Code review


plugins/@grouparoo/salesforce/__tests__/export-contacts/destination-mapping-options.ts
```diff
@@ -1,25 +1,46 @@
...
+    ({ model } = await helper.factories.properties());
+    app = await helper.factories.app();
+    destination = await Destination.create({
+      name: "Salesforce Test Destination",
+      type: "salesforce-export-accounts",
+      syncMode: DestinationSyncModeData.sync.key,
+      appId: app.id,
+      modelId: model.id,
+    });
+  });
```
evantahler, line 43: I see where you were making destinations now!

plugins/@grouparoo/salesforce/src/lib/export/mapping.ts
```diff
@@ -194,7 +199,10 @@ const extractFields = (
 
     if (!field.nillable && !field.defaultedOnCreate) {
       // needs to be set to create
-      required.push({ key, type });
+      const syncMode = await destination.getSyncMode();
+      if (syncMode !== DestinationSyncModeData.enrich.key) {
+        required.push({ key, type });
+      }
```
evantahler, line 205: This seems to be the main logical change

evantahler (approved):
This is good!  

I think there's a place to improve the tests however.  Removing the duplication seems desirable, so use a `jest.each` block to run them all with destinations in each sync mode, and share the tests that way.  This way we can be sure that everything is the same except the options

plugins/@grouparoo/salesforce/src/lib/export/mapping.ts
bleonard: I wonder if we should pass in the this stuff from core to be more like `exportRecords` instead of the implicit stuff hanging off the Destination. Is that a good idea @evantahler ?
```
export interface DestinationSyncOperations {
    create: boolean;
    update: boolean;
    delete: boolean;
}
```

PR #2711, opened 2021-12-22
Improve patterns on accessing web app props and event handlers

edmundito:
## Change description

This change cleans up how global props are accessed and introduces these new patterns:

1. Event Handlers have always been global to the application, instead of accessing them via the props, they can be accessed via import. This has always been true but now it makes it official.

```ts
// Before
import { ErrorHandler } from '../../utils/errorHandler'

function Page(props) {
  const { errorHandler }: { errorHandler: ErrorHandler } = props
}

// After
import { errorHandler } from '../../eventHandlers'

function Page(props) {
}
```

2. Global page props are accessed from a context

Taking advantage of React's features, instead of having to pass down the properties to all the pages, instead we can use a context to access these globals:

```tsx
// Before:
function Page(props) {
  const { navigation } = props
}

// After:
function Page(props) {
  const { navigation } = usePageContext();
}
```

3. Utility to infer props from getServerSideProps

Usage:
```tsx
import { PageWithInferredProps } from '../../types/page'

export const getServerSideProps = () => {
 return { props: { ... } }
}

const Page: PageWithInferredProps<typeof getServerSideProps> = (props) => {
}
```

### Development

### Impact

### Code review


ui/ui-components/hooks/useApi.ts
```diff
@@ -18,7 +18,7 @@ export interface ApiHook {
 }
 
 export function UseApi(
-  ctx: GetServerSidePropsContext | NextPageContext,
```
edmundito, line 21: This change may benefit from refactoring useApi first. I'm not familiar with next but it seems that the context is only needed on the server-side requests. Once the page renders and uses this "hook", does it need to also pass a context from the props? Where does it come from?

ui/ui-components/pages/_app.tsx
```diff
@@ -10,21 +10,26 @@ import PageTransition from "../components/PageTransition";
...
 import { getModelFromUrlOrCookie, setModelCookie } from "../utils/modelHelper";
 
-export default function GrouparooWebApp(props) {
+export default function GrouparooWebApp(
+  props: AppProps & GrouparooWebAppProps & { err: any }
+) {
   const { Component, pageProps, err, hydrationError } = props;
 
-  const combinedProps = Object.assign({}, pageProps || {}, {
+  const combinedProps: GrouparooNextPageProps<any> = {
```
edmundito: I have to set this as `any` here because when I try to inject the `err` prop when rendering the `<Component`, it causes a typing error. I'm still trying to figure out how these all come together.

edmundito:
Just posted this WIP with a few examples to show the direction I'm heading. Once I gather enough feedback I can apply to the rest of the pages.

evantahler (commented):
I like where this is going! 

Possibly related, possibly not - Is this when we switch from GetInitialProps to GetServersideProps?  Then we could use InferGetServerSidePropsType maybe?

edmundito:
I made an update to use InferGetServerSideProps type, forcing the type to expect a `typeof getServerSideProps` as a generic parameter.

The only thing here is that this would imply that our props are loosely defined by `getServerSideProps`, so it's a matter of preference whether we want to explicitly declare the type or not. The advantage of this current implementation is that it's less boilerplate.

ui/ui-components/hooks/useApi.ts
krishnaglick: Client side it won't have the context.

ui/ui-components/pages/account.tsx
```diff
@@ -1,25 +1,26 @@
...
-    errorHandler,
-    successHandler,
-    sessionHandler,
-  }: {
-    errorHandler: ErrorHandler;
-    successHandler: SuccessHandler;
-    sessionHandler: SessionHandler;
-  } = props;
-  const { execApi } = UseApi(props, errorHandler);
+export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
```
edmundito, line 12: One tricky part about inferring return types from getServeSideProps using `InferGetServerSidePropsType` is that If I were to declare this function with the type:

```ts
getServerSideProps: GetServerSideProps = () =>
```

It would infer the types as `any` because the GetServerSideProps definition is `GetServerSideProps<P = any>`

ui/ui-components/pages/account.tsx
```diff
@@ -1,25 +1,26 @@
...
+  const { execApi } = UseApi(ctx);
+  const { teamMember } = await execApi<Actions.AccountView>("get", `/account`);
+  return { props: { teamMember } };
+};
+
+const Page: GrouparooPage<typeof getServerSideProps> = ({
```
evantahler: This is very nice

ui/ui-components/pages/apps.tsx
```diff
@@ -12,20 +13,29 @@ import AppIcon from "../components/AppIcon";
...
+};
+
+const Page: GrouparooPage<typeof getServerSideProps> = ({
+  errorHandler,
+  successHandler,
+  ...props
+}) => {
   const router = useRouter();
-  const { execApi } = UseApi(props, errorHandler);
+  const { execApi } = UseApi(undefined, errorHandler);
```
evantahler, line 37: What if execApi was passed in from _App, already configured to use the errorHandler?  We might want to refactor that first.  Ideally, the new API would provide it to the component already initialized, and then we could override the  default  error handler as needed

ui/ui-components/types/app.ts
```diff
@@ -0,0 +1,25 @@
+import {
```
evantahler: Perhaps this shouldn't be called `app` (which might be confused with an `App`).  Perhaps "webApp" or "UI"?

edmundito:
I decided to make a major change to this and embrace React and Next's features to improve the pattern. Take a look at the updated description and let me know what you think.

ui/ui-components/eventHandlers/index.ts
```diff
@@ -0,0 +1,35 @@
...
+import { PropertiesHandler } from "../utils/propertiesHandler";
+import { RunsHandler } from "../utils/runsHandler";
+import { ScheduleHandler } from "../utils/scheduleHandler";
+import { SessionHandler } from "../utils/sessionHandler";
+import { SetupStepHandler } from "../utils/setupStepsHandler";
+import { StatusHandler } from "../utils/statusHandler";
+import { SourceHandler } from "../utils/sourceHandler";
+import { TeamHandler } from "../utils/teamHandler";
+import { TeamMemberHandler } from "../utils/teamMembersHandler";
+import { UploadHandler } from "../utils/uploadHandler";
```
edmundito, line 17: Eventually, the event handlers will be moved to this subfolder but while we're referencing these types everywhere this may cause a lot of file changes. So I will keep them in utils for now.

ui/ui-components/pages/_app.tsx
```diff
@@ -10,27 +11,51 @@ import PageTransition from "../components/PageTransition";
...
+  navigationModel: Actions.NavigationList["navigationModel"];
+}
+
+export default function GrouparooNextApp(
+  props: AppProps & GrouparooNextAppProps & { err: any }
+) {
   const { Component, pageProps, err, hydrationError } = props;
 
-  const combinedProps = Object.assign({}, pageProps || {}, {
+  const combinedProps = {
```
edmundito, line 33: This `combinedProps` is still being used for legacy pages so we need to keep it. We can eliminate this When all pages are migrated to use context and imported event handlers.

ui/ui-components/pages/apps.tsx
edmundito: Now I believe that we should lean in on the hooks pattern instead of passing these down through props. For these examples it's a bit ugly to pass undefined but I believe we can clean up useAPI to be more hook friendly

ui/ui-components/types/page.ts
```diff
@@ -0,0 +1,9 @@
+import {
```
edmundito: I could probably move this to `utils/pageHelper`. Thoughts?

ui/ui-components/hooks/useApi.ts
edmundito: Right! So It does make sense to rebuild `useAPI` as a proper hook or a server-side utility. That will be done separately from this PR.

ui/ui-components/contexts/page.tsx
```diff
@@ -0,0 +1,15 @@
...
+  currentTeamMember: Partial<Actions.NavigationList["teamMember"]>;
+  hydrationError?: string;
+  navigation: Actions.NavigationList["navigation"];
+  navigationMode: Actions.NavigationList["navigationMode"];
+  navigationModel: Actions.NavigationList["navigationModel"];
+}
+
+export const PageContext = createContext<PageContext>(undefined);
+
+export const usePageContext = () => useContext(PageContext);
```
edmundito: I'm looking for better suggestions for the name of this context. The other I had in mind was `WebAppContext`.

krishnaglick: Could call it NavigationContext?

ui/ui-components/pages/_app.tsx
```diff
@@ -10,27 +11,51 @@ import PageTransition from "../components/PageTransition";
...
+      navigationMode: props.navigationMode,
+      navigationModel: props.navigationModel,
+      clusterName: props.clusterName,
+      currentTeamMember: props.currentTeamMember,
+    };
+  }, [props]);
 
   return (
-    <Injection {...combinedProps}>
+    <Injection contextValues={[[PageContext, pageContext]]}>
```
krishnaglick: I think I see what you're doing here, though I'm partial to [this](https://github.com/grouparoo/cloud/blob/main/web/contexts/ContextProvider.component.tsx)

ui/ui-components/pages/account.tsx
krishnaglick: You can't pass it the props at that point of us?

ui/ui-components/types/page.ts
krishnaglick: Since this is a single file to export one thing, probably worth it.

ui/ui-components/utils/contextHelper.tsx
```diff
@@ -0,0 +1,25 @@
+export type ContextValue = [Context: React.Context<any>, value: any];
+
+export const renderContextProvider = <T extends object>(
+  Context: React.Context<T>,
+  value: T,
+  children: React.ReactNode
+) => <Context.Provider value={value}>{children}</Context.Provider>;
+
+export const renderNestedContextProviders = (
```
krishnaglick, line 9: Since these are components, they should be capitalized.
```suggestion
export const RenderNestedContextProviders = (
```

ui/ui-components/components/Navigation.tsx
```diff
@@ -38,31 +40,22 @@ export const iconConstrainedStyle = { width: 20 };
...
-    statusHandler: StatusHandler;
   } = props;
-  const { execApi } = UseApi(props, errorHandler);
+  const {
+    navigationModel,
+    navigationMode,
+    navigation,
+    clusterName,
+    currentTeamMember,
+  } = usePageContext();
```
evantahler: Yeah, this is much better!

evantahler (commented):
I like the stated goals and proposed API a lot.  I'll leave it to you and @krishnaglick to talk about the best way to accomplish this with the react internals (contexts, providers, etc)

ui/ui-components/pages/_app.tsx
edmundito: Yep, it's pretty much doing the same thing. I could eliminate injection entirely. I don't think it's necessary to have it and have my helper function provide all the contexts. The reason I use the function is to avoid having another nested component in the React inspector.

ui/ui-components/utils/contextHelper.tsx
edmundito: It's purposely not a component so that it doesn't render an unnecessary <NestedContextProvider> in the React inspector.

krishnaglick (approved):

PR #2716, opened 2021-12-27
Add Airtable as a Destination

bleonard:
Thanks @brandon-segal!

## Change description

Addresses issue #2551 . 
Adds Airtable as a destination for Grouparoo. The Airtable application takes in an API Key host and base Id from the user and the destination options ask for a table and foreign key that should be used to designate the exact table that data will be sent to and what is going to be used as the key.

### Development

### Impact

### Code review


plugins/@grouparoo/airtable/LICENSE.txt
```diff
@@ -0,0 +1,373 @@
+Mozilla Public License Version 2.0
```
evantahler, line 1: 👍 MOZ2

plugins/@grouparoo/airtable/__tests__/no-meta-access/client.ts
```diff
@@ -0,0 +1,182 @@
+import { Client } from "../../src/lib/client/client";
+import { helper } from "@grouparoo/spec-helper";
+import { loadAppOptions, getUpdater, loadTableData } from "../utils/nockHelper";
+import * as utils from "../utils/shared";
+import { AirtableAppOptions } from "../../src/lib/appOptions";
+import { IClient } from "../../src/lib/client/interfaces/iClient";
+
+helper.useNock(__filename, getUpdater("BASIC"));
+const appOptions = loadAppOptions("BASIC");
+const tableData = loadTableData("BASIC");
```
evantahler, line 10: This is a cool pattern to try a few variations on an API key (meta/basic)

plugins/@grouparoo/airtable/__tests__/utils/nockHelper.ts
```diff
@@ -0,0 +1,101 @@
...
+    allPrimaryKey: "Name",
+    allKnownKeys: ["Brian", "Evan", "Andy"],
+    emptyId: parsed[`AIRTABLE_${baseType}_EMPTY_TABLE_ID`],
+    emptyName: "Empty",
+    emptyPrimaryKey: "Name",
+    emptyKnownKeys: [],
+  };
+}
+
+export function getUpdater(baseType: "BASIC" | "META") {
```
evantahler: This `"BASIC" | "META"` type can probably be shared `type BaseType = "BASIC" | "META"`

plugins/@grouparoo/airtable/__tests__/utils/nockHelper.ts
```diff
@@ -0,0 +1,101 @@
...
+          throw err;
+        }
+        // 12 = start of {"bases":
+        // 3  = ", [" at end of line
+        const reply = rest.substring(12, endIndex - 3);
+        JSON.parse(reply); // just to make sure it parses
+        nockCall = nockCall.replace(
+          reply,
+          `{"bases":[{"id":"${useBaseId}","name":"[CI Test] Marketing Campaign Tracking","permissionLevel":"create"}]}`
+        );
```
evantahler, line 97: Fancy!

plugins/@grouparoo/airtable/jest.config.js
```diff
@@ -0,0 +1,18 @@
+const { helper } = require("@grouparoo/spec-helper");
+
+module.exports = {
+  testEnvironment: "node",
+  testTimeout: helper.mediumTime,
+  maxWorkers: process.env.NOCK ? "1" : "50%",
```
evantahler, line 6: Think this is something we should do fora ll nocking plugins?

plugins/@grouparoo/airtable/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "type": "git",
+    "url": "https://github.com/grouparoo/grouparoo.git",
+    "directory": "plugins/@grouparoo/airtable"
+  },
+  "scripts": {
+    "prepare": "rm -rf dist && tsc --declaration",
+    "test": "jest",
+    "pretest": "npm run lint && npm run prepare",
+    "lint": "prettier --check src ",
+    "format": "prettier --write src",
```
evantahler, line 28: Not that it's hurting anything, but we normally don't have a `format` command and rely on the IDE

plugins/@grouparoo/airtable/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "watch": "tsc --watch"
+  },
+  "dependencies": {
+    "@grouparoo/app-templates": "0.8.0-alpha.1",
+    "@grouparoo/core": "0.8.0-alpha.1",
+    "actionhero": "28.1.1",
+    "airtable": "0.11.1",
+    "axios": "0.24.0"
+  },
```
evantahler: 👍 
* no `^` or `~` dependencies
* versions match the rest of the monorepo

plugins/@grouparoo/airtable/src/lib/appOptions.ts
```diff
@@ -0,0 +1,43 @@
...
+    required: true,
+    description: "Airtable (API) key",
+  },
+  {
+    key: "baseId",
+    type: "text",
+    displayName: "Airtable Base",
+    required: true,
+    description: "Airtable Base to read tables from",
+  },
```
evantahler, line 33: Question: Should the baseId be part of the destination options?  This way, the app (just API ID) can be used for multiple destinations

plugins/@grouparoo/airtable/src/lib/client/client.ts
```diff
@@ -0,0 +1,385 @@
...
+            message: `${error}. Invalid API Key`,
+          };
+        case 404:
+          // doesn't have meta access
+          return {
+            health: true,
+            access: false,
+            message:
+              "API key valid, but account does not have meta API access. Verify base id manually.",
+          };
```
evantahler, line 71: Not the greatest test, but I agree this is all we can do without meta api access.

plugins/@grouparoo/airtable/src/lib/client/client.ts
```diff
@@ -0,0 +1,385 @@
...
+      // also support name
+      for (const table of tables) {
+        if (table.name === tableId) {
+          return Object.assign({}, table, { idOrName: table.id });
+        }
+      }
+    }
+    const records = await this.listRecords(tableId);
+    return makeTableFromRecords(tableId, records);
+  }
```
evantahler: I want to confirm that I'm reading this logic right:
* If your API key has meta access, you can reference the table id or table name
* If your API key does not have meta access, you need to just use the table id.

Is it hard to find the table IDs?  Why support both modes? In general, I think we should aim for "one way to do things" to have clearer docs and examples.  

I think this is the only usage of `.hasMeta()`

plugins/@grouparoo/airtable/src/lib/destination/connection.ts
```diff
@@ -0,0 +1,43 @@
...
+import { exportArrayProperties } from "./exportArrayProperties";
+
+export const objectsSupportedSyncModes: DestinationSyncMode[] = [
+  "sync",
+  "additive",
+  "enrich",
+];
+
+export const objectsDestinationConnection: PluginConnection = {
+  name: "airtable-export-records",
```
evantahler: Perhaps the object exported should match the name of the destination... like `AirtableRecordDestination`?

evantahler: Or, is "objects" the language Airtable would use?  If that's the case, perhaps the name and displayName should be `aitable-export-objects`.

In general, we try to use the language of the Destination vs Grouparoo's words (Contacts vs Recrods)

plugins/@grouparoo/airtable/src/lib/destination/exportRecords.ts
```diff
@@ -0,0 +1,322 @@
...
+  syncOperations?: DestinationSyncOperations;
+  exports: BatchExport[];
+}
+
+export async function exportBatch(exportBatchOptions: ExportBatchOptions) {
+  const { syncOperations, appOptions, destinationOptions, exports } =
+    exportBatchOptions;
+  const { primaryKey } = destinationOptions;
+  const batchSize = BATCH_SIZE;
+  const findSize = BATCH_SIZE;
```
evantahler, line 284: If these are constants from higher up, it's probably not nescisary to reassign them

plugins/@grouparoo/airtable/src/lib/disconnect.ts
```diff
@@ -0,0 +1,8 @@
+import { DisconnectPluginAppMethod } from "@grouparoo/core";
+import { IClient } from "./client/interfaces/iClient";
+
+export const disconnect: DisconnectPluginAppMethod<IClient> = async ({
+  connection,
+}) => {
+  return;
+};
```
evantahler, line 8: Rather than have a method that does nothing, just remove it!  Destinations don't need a `disconnect()` method

ui/ui-components/pages/app/[id]/edit.tsx
```diff
@@ -365,7 +365,8 @@ export default function Page(props) {
                   </Alert>
                 ) : ranTest ? (
                   <Alert variant="warning">
-                    <strong>Test Failed</strong> {testResult.error}
+                    <strong>Test Failed</strong>{" "}
+                    {testResult.error || testResult.message}
                   </Alert>
```
Comment (evantahler) [inline ui/ui-components/pages/app/[id]/edit.tsx:377]: This API probably could be better.  The case you added means that there is both an Error and Message set on the test... is that common?

evantahler (commented):
Excellent test updates @bleonard!

I've got mostly structural questions below, the main ones being:
* Should BaseID be a destination option?
* Naming of the Destination

... Did you get Meta API Access?  I haven't yet :(

plugins/@grouparoo/airtable/jest.config.js
bleonard: No. I needed it here because of the assumption that the table was empty in some tests, but is also used to write data into in others. This is a nice pattern if that turns out to be the case because I kept forgetting `--runInBand`

plugins/@grouparoo/airtable/src/lib/client/client.ts
bleonard: It takes either the table Id (in the url) or the table name (relatively visible in the UI), but if you have have the meta access, it seems better to upgrade the experience.
Using the meta API and inspecting the columns is the main win of using the meta API.

plugins/@grouparoo/airtable/src/lib/destination/connection.ts
bleonard: Airtable calls each row "records" so I think this fits with that philosophy.

bleonard: If by "Objects' you mean the variable name, I bet brandon copied that from Salesforce or other. I can change the variable name

bleonard: Ah, there was one spot. Fixed.
https://github.com/grouparoo/grouparoo/pull/2716/commits/e0847241eb8ee5fe89680604f0166d737449b6f2

plugins/@grouparoo/airtable/src/lib/disconnect.ts
bleonard: https://github.com/grouparoo/grouparoo/pull/2716/commits/1695abffbcc007a1816850d9d61d9556508dc1f4

ui/ui-components/pages/app/[id]/edit.tsx
Comment (bleonard) [reply ui/ui-components/pages/app/[id]/edit.tsx:377]: There is no such thing as an `error` actually. Plugins only send `message`. I don't know where `error` came from, but it wasn't in the scope of making this plugin to mess with that.

plugins/@grouparoo/airtable/__tests__/utils/nockHelper.ts
bleonard: https://github.com/grouparoo/grouparoo/pull/2716/commits/0d307875ad3030d2560f07afd164363656e5e47d

bleonard:
I think it's right that `baseId` is an app option. This parallels with all the other destinations we have that are like databases.
Made some updates to the naming.
I did get Meta access for the one base. Thanks @brandon-segal !

plugins/@grouparoo/airtable/src/lib/disconnect.ts
bleonard: Actually

> error: Error from Initializer name=@grouparoo/airtable file=/Users/brian/grouparoo/grouparoo/plugins/@grouparoo/airtable/dist/initializers/plugin.js type=initialize stack=Plugin validation failure for "@grouparoo/airtable": both app.methods.connect and app.methods.disconnect are required for a Grouparoo Plugin App

plugins/@grouparoo/airtable/__tests__/no-meta-access/export-records.ts
```diff
@@ -0,0 +1,6 @@
+import { testExportRecords } from "../utils/export-records";
+import { helper } from "@grouparoo/spec-helper";
+import { getUpdater } from "../utils/nockHelper";
+
+helper.useNock(__filename, getUpdater("BASIC"));
+testExportRecords("BASIC");
```
evantahler, line 6: What's this test doing?

plugins/@grouparoo/airtable/__tests__/with-meta-access/export-records.ts
```diff
@@ -0,0 +1,6 @@
+import { testExportRecords } from "../utils/export-records";
+import { helper } from "@grouparoo/spec-helper";
+import { getUpdater } from "../utils/nockHelper";
+
+helper.useNock(__filename, getUpdater("META"));
+testExportRecords("META");
```
evantahler, line 6: Same here - what's this test up to?

plugins/@grouparoo/airtable/src/lib/disconnect.ts
evantahler: Well that's a shame!

evantahler (approved):

plugins/@grouparoo/airtable/__tests__/with-meta-access/export-records.ts
bleonard: These both run the same set of tests (from "../utils/export-records") because it should works exactly the same in both.
I decided not to duplicate the test exactly twice because it's the most important one and I didn't want them to diverge.

PR #2749, opened 2022-01-04
Vero plugin creation

aelafifi:
## Change description

Description here

### Development

### Impact

### Code review


plugins/@grouparoo/vero/src/initializers/plugin.ts
```diff
@@ -0,0 +1,80 @@
...
+          ],
+          methods: { test },
+        },
+      ],
+      connections: [
+        {
+          name: "vero-export-persons",
+          displayName: "Vero Export Persons",
+          direction: "export",
+          description: "Export records to Vero as Person contacts",
```
bleonard: The API says they call them "Users" - should that be here or is the API different than their UI?

plugins/@grouparoo/vero/src/lib/export/destinationMappingOptions.ts
```diff
@@ -0,0 +1,30 @@
...
+      group: {
+        singular: "Vero Filter",
+        plural: "Vero Filters",
+      },
+    },
+    properties: {
+      known: [
+        {
+          key: "email",
+          type: "email",
```
bleonard: Add `important: true`

plugins/@grouparoo/vero/src/lib/export/exportRecord.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (!newRecordProperties["id"]) {
+    throw new Error(`newRecordProperties[id] is a required mapping`);
+  }
+
+  const client = await connect(appOptions);
+
+  const newID = newRecordProperties["id"];
+  const oldID = oldRecordProperties["id"];
+
+  if (oldID && oldID !== newID) {
```
bleonard: I wonder if this should be below the delete, so it doesn't change it.

bleonard: OI'm also curious how synchronous this is? What have you seen? For example, does it work as expected when it immediately gets updated or groups are updated.

plugins/@grouparoo/vero/src/lib/export/listMethods.ts
```diff
@@ -0,0 +1,18 @@
...
+  newGroups: Array<string>,
+  newRecordProperties: { [p: string]: any }
+) {
+  const groupsToRemove = oldGroups.filter(
+    (groupName) => !newGroups.includes(groupName)
+  );
+
+  await client.updateUserTags(
+    newRecordProperties["id"],
+    newGroups.map((groupName) => `In Group: ${groupName}`),
```
bleonard: We don't need to prepend`"In Group:"`. They can set it to the tag name they want in the config

bleonard (commented):
Even if we can't check as well as in other plugins, let's make a regular test suite that adds people and deletes them, etc and make it clear what it should look like at the end.

plugins/@grouparoo/vero/src/initializers/plugin.ts
aelafifi: In API they call them "Users", but in UI they call them "Customers"!
I know that both of them are not "Contacts", but which one you think it should be used?

plugins/@grouparoo/vero/src/lib/export/exportRecord.ts
bleonard: So let's move this down below the `toDelete` clause

plugins/@grouparoo/vero/src/initializers/plugin.ts
bleonard: "Export records to Vero as Customers"
Let's change the `name` as well to `vero-export-customers`

plugins/@grouparoo/vero/__tests__/utils/shared.ts
```diff
@@ -0,0 +1,8 @@
+import { helper } from "@grouparoo/spec-helper";
+
+export async function indexUsers(isNewNock: boolean) {
```
bleonard: Is this not used?

plugins/@grouparoo/vero/__tests__/export/export-record.ts
```diff
@@ -0,0 +1,572 @@
...
+      newRecordProperties,
+      oldGroups,
+      newGroups,
+      toDelete,
+    },
+  });
+}
+
+const manualCheck = async (...instructions: string[]) => {
+  if (process.env.CI) {
```
bleonard: This should probably actually be `if(newNock)` because it doesn't make sense to check unless it's really making requests. (and we don't nock on CI, so that will still work)

bleonard (approved):

PR #2750, opened 2022-01-04
Allow users to map using the Salesforce Id as primary key in enrich mode via configs

pauloouriques:
## Change description

This PR contains a treatment to allow Id to be the recordMatchField when in enrich mode.

### Development

### Impact

### Code review

### DIFF
--- a/plugins/@grouparoo/salesforce/src/lib/export/salesforceExporter.ts
+++ b/plugins/@grouparoo/salesforce/src/lib/export/salesforceExporter.ts
@@ -25,6 +25,7 @@ import {
 } from "@grouparoo/app-templates/dist/destination/batch";
 import { SalesforceModel } from "./model";
 
+const VALID_ID_LENGTHS = [15, 18];
 const idType = "Id";
 
 // return an object that you can connect with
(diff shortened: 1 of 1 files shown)


plugins/@grouparoo/salesforce/src/lib/export/salesforceExporter.ts
```diff
@@ -43,24 +44,25 @@ const findAndSetDestinationIds: BatchMethodFindAndSetDestinationIds = async ({
...
-
-  const query = { [recordMatchField]: parsedForeignKeys };
-  const fields = [idType, recordMatchField];
+  let validForeignKeys = parseForeignKeys(foreignKeys, foreignKeyType);
+  let fields = [idType, recordMatchField];
+  if (idType === recordMatchField) {
+    validForeignKeys = foreignKeys.filter((foreignKey) =>
+      VALID_ID_LENGTHS.includes(foreignKey.length)
+    );
+    fields = [idType];
```
bleonard, line 58: This looks to be the main change. What happens if we don't filter above. Does the whole query freak of if there is one that is not length 15 or 18? Where did we determine the 15 or 18?

pauloouriques: We can determine the length based on [this](https://help.salesforce.com/s/articleView?id=000324087&type=1)

pauloouriques: Yes, when there is an invalid ID on the query the entire query seems to freak out.

`warning: ExportError  Destination: accounts_destination_dsc_enrich - Salesforce Accounts DSC Enrich - Record: Unknown error=INVALID_QUERY_FILTER_OPERATOR: 
SELECT Id FROM Account WHERE Id IN ('sfabc')
                                                            ^
ERROR at Row:1:Column:30
invalid ID field: sfabc`

bleonard: So if the `id` gets left off here, they there will be an error on that export saying what? I'm thinking we shoudl add an explicit error on that bad export here.

bleonard: So instead of 
```
validForeignKeys = foreignKeys.filter((foreignKey) =>
      VALID_ID_LENGTHS.includes(foreignKey.length)
    );
  ```
  something like this:
  ```
  const passed = [];
for (const foreignKey of foreignKeys) {
  if(VALID_ID_LENGTHS.includes(foreignKey.length)) {
      passed.push(foreignKey);
  }
  else {
      const found = getByForeignKey(foreignKey);
      if(found) {
            found.error = new Error(`Invalid Salesforce id: ${foreignKey`)
      }
}
}
validForeignKeys = passed;
```

bleonard (approved):

PR #2751, opened 2022-01-04
Add states to Imports

pedroslopez:
## Change description

- Imports now have states: `associating`, `importing`, `exporting`, `complete`, `failed`
- Updated logic in multiple places to use these states instead of timestamps to determine where in the lifecycle the import was at
- Updated import list views to allow filtering by states
- `recordUpdatedAt` and `groupsUpdatedAt` timestamps were always set at the same time, so rolled those up into one `importedAt` matching state name

image

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/import.ts
+++ b/core/__tests__/models/import.ts
@@ -28,16 +28,21 @@ describe("models/import", () => {
       creatorId: "",
     });
     expect(_import.recordId).toBeFalsy();
+    expect(_import.state).toBe("associating");
 
     await _import.associateRecord();
+    expect(_import.state).toBe("pending");
     expect(_import.recordId).toBe(record.id);
   });
 
   test("an error can be set", async () => {
     const _import = await helper.factories.import();
+    expect(_import.state).toBe("associating");
+
     await _import.setError(new Error("oh no"), "associate");
     await _import.reload();
 
+    expect(_import.state).toBe("failed");
     expect(_import.errorMessage).toMatch("oh no");
     expect(_import.errorMetadata).toMatch("oh no");
   });
--- a/core/__tests__/models/import.ts
+++ b/core/__tests__/models/import.ts
@@ -31,7 +31,7 @@ describe("models/import", () => {
     expect(_import.state).toBe("associating");
 
     await _import.associateRecord();
-    expect(_import.state).toBe("pending");
+    expect(_import.state).toBe("importing");
     expect(_import.recordId).toBe(record.id);
   });
 
--- a/core/src/migrations/000097-addStateToImports.ts
+++ b/core/src/migrations/000097-addStateToImports.ts
@@ -19,23 +19,23 @@ export default {
     });
 
     await queryInterface.sequelize.query(
-      `UPDATE "imports" SET "state"='associating' WHERE "recordAssociatedAt" IS NULL`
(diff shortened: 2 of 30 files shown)


core/__tests__/models/import.ts
```diff
@@ -28,16 +28,21 @@ describe("models/import", () => {
       creatorId: "",
     });
     expect(_import.recordId).toBeFalsy();
+    expect(_import.state).toBe("associating");
 
     await _import.associateRecord();
+    expect(_import.state).toBe("importing");
```
evantahler, line 34: good test re-use!

core/src/migrations/000097-addStateToImports.ts
```diff
@@ -0,0 +1,71 @@
...
+    );
+
+    await queryInterface.sequelize.query(
+      `UPDATE "imports" SET "state"='failed' WHERE "errorMessage" IS NOT NULL`
+    );
+
+    await queryInterface.changeColumn("imports", "state", {
+      type: DataTypes.STRING(191),
+      allowNull: false,
+    });
```
evantahler, line 44: 👍 we end up with a non-null state column

evantahler: Do you think we need any addtional exclusivity checks on the updates, eg:

```sql
UPDATE "imports" SET "state"='exporting' WHERE "recordAssociatedAt" IS NOT NULL AND "importedAt" IS NOT NULL AND "exportedAt" IS NULL ((AND state IS NULL))
```

evantahler: 👍 this will fail to migrate if there are any imports that still don't have a state

core/src/modules/ops/runs.ts
```diff
@@ -111,17 +111,9 @@ export namespace RunOps {
...
-              },
-            },
-          }),
-          groupsUpdated: await run.$count("imports", {
-            where: {
-              groupsUpdatedAt: {
+              importedAt: {
                 [Op.gte]: lastBoundary,
                 [Op.lt]: nextBoundary,
               },
```
evantahler, line 119: so much better!

core/src/tasks/record/enqueueExports.ts
```diff
@@ -26,9 +26,7 @@ export class GrouparooRecordsEnqueueExports extends CLSTask {
         [Sequelize.fn("DISTINCT", Sequelize.col("recordId")), "recordId"],
       ],
       where: {
-        recordUpdatedAt: { [Op.not]: null },
-        groupsUpdatedAt: { [Op.not]: null },
-        exportedAt: null,
+        state: "exporting",
       },
```
evantahler, line 30: nice and clean!

ui/ui-components/components/import/List.tsx
```diff
@@ -26,6 +37,7 @@ export default function ImportList(props) {
   // pagination
   const limit = 100;
   const { offset, setOffset } = useOffset();
+  const [state, setState] = useState(router.query.state?.toString() || "all");
```
evantahler: this could be `useState<typeof states>` if `states as const`

evantahler (approved):
Really 💪 work!

Just some comments below, nothing blocking.

core/src/migrations/000097-addStateToImports.ts
pedroslopez: Added `state IS NULL` checks to the migration on ec560ca

PR #2752, opened 2022-01-04
Fixing Snowflake Connection Missing Error w/ Multiple Workers

krishnaglick:
## Change description

When you had more than 1 worker, there was an issue in the connection logic resulting in some workers receiving no connection.

### Development

### Impact

### Code review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -720,6 +720,61 @@ jobs:
       - run:
           name: test
           command: cd core && ./node_modules/.bin/jest __tests__/notifiers --ci --maxWorkers 2
+  test-core-ops:
+    environment:
+      GROUPAROO_TELEMETRY_ENABLED: "false"
+    docker:
+      - image: circleci/node:16.8.0
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+      - image: redis:latest
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+      - image: circleci/postgres:9
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+        environment:
+          POSTGRES_PASSWORD: password
+    steps:
+      - checkout
+      - restore_cache:
+          <<: *module-other-cache
+      - restore_cache:
+          <<: *plugin-1-cache
+      - restore_cache:
+          <<: *plugin-2-cache
+      - restore_cache:
+          <<: *plugin-3-cache
+      - restore_cache:
+          <<: *plugin-4-cache
+      - restore_cache:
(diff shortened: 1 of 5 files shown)


teallarson (approved):
👏 nice.

Do you know why we only saw the bug with the Snowflake plugin?  Does it build its connection differently?

evantahler:
Please be sure to check all the items in the checklist before merging

evantahler (approved):
Nice find!

It might be more trouble than it is worth, but can this be tested some way?  Maybe a fake app can be made and multiple requests in a Promise.all() try to connect()?

.circleci/config.yml
```diff
@@ -720,6 +720,61 @@ jobs:
       - run:
           name: test
           command: cd core && ./node_modules/.bin/jest __tests__/notifiers --ci --maxWorkers 2
+  test-core-ops:
```
krishnaglick, line 723: Not sure why this is here, rebased from main.

core/__tests__/ops/app.ts
```diff
@@ -0,0 +1,19 @@
...
+  it("should always return a connection object when given a connect method, even when run multiple times", async () => {
+    const app = new App({
+      name: "test app",
+      type: "test-plugin-app",
+    });
+    await app.save();
+
+    for (let i = 0; i < 8; i++) {
+      expect(await AppOps.connect(app)).toBe(true);
+    }
```
evantahler, line 17: I don't think this test is actually checking what we want - since the specHelper connect method returns a literal `true` each time.  You might want to mock the `connect` method on the plugin to return a specific value and make sure you get that value back.  Or maybe return a Math.random() or something and assert that they are the same each time

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -293,6 +293,8 @@ export namespace helper {
             appQuery: async () => {
               return [new Date().getTime()];
             },
+            connect: async () => true,
+            disconnect: async () => true,
```
evantahler, line 297:
```suggestion
            connect: async () => Math.random(),
            disconnect: async () => true,
```

.circleci/config.yml
evantahler, line 723: You could move this new test into a model test too if you don't want to start the whole ops section

evantahler (changes requested):

core/__tests__/ops/app.ts
krishnaglick: I half-agree. However the root cause of the issue is simply existing connections are not returned. So this test fails if that's the case.
I'm not sure we _need_ to do deeper testing since, if the connection is not returned, this is not true.

evantahler: That’s a good point! Ship it!

evantahler (approved):

PR #2760, opened 2022-01-05
add ui edition to preview record localstorage token

teallarson:
## Change description

Previously, when adding a sample record in Config UI and then switching to Community or Enterprise UI, the preview record would throw an error that it could not be found (because the record Ids are different for the config ui sample records).

This PR adds the UI edition to the localstorage token so that we are always referencing a record from the correct scope.

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/components/record/SampleRecordCard.tsx
+++ b/ui/ui-components/components/record/SampleRecordCard.tsx
@@ -46,17 +46,24 @@ export interface SampleRecordCardProps {
 const isConfigUI = grouparooUiEdition() === "config";
 
 const getCachedSampleRecordId = (modelId: string): string => {
-  return globalThis.localStorage?.getItem(`sampleRecord:${modelId}`);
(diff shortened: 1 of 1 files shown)


bleonard:
Separating the version is great, but I think we'll also just want to generically handle that record not being there more gracefully.

evantahler:
> Separating the version is great, but I think we'll also just want to generically handle that record not being there more gracefully.

I think that what we do already is sufficient! If you really did somehow get in a case where the sample record you used to be using is suddenly gone (eg: a schedule removed it, you deleted the sample record in config mode), you should see an error telling you that your sample is gone.  I think that picking a new one at random would be surprising?  

We could maybe use an "alert" toast (yellow?) rather than red in this case?  this would be the first one of those

evantahler (approved):

bleonard:
Yeah, it's the red box I was talking about. I suppose we can see how often it happens with this change and go from there.

PR #2762, opened 2022-01-06
Schedule Filters UI cleanup

teallarson:
## Change description

Makes the UI for Schedule Filters match that of Property Filters.  Now, the "value" box only appears if the operation uses a match value.  

Also fixes the default option in the UI for adding a new Property Filter.  It was previously set to "Exists", however not all plugins support `Exist`/ `notExist` yet (I'm looking at you, Mongo).

Example:
Screen Shot 2022-01-06 at 9 27 33 AM

Screen Shot 2022-01-06 at 9 27 25 AM

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx
+++ b/ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx
@@ -197,7 +197,7 @@ export default function Page(props) {
 
     _localFilters.push({
       key: filterOptions[0].key,
-      op: "exists",
+      op: filterOptions[0].ops[0],
       match: "",
     });
 
--- a/ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
+++ b/ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
@@ -393,24 +393,21 @@ export default function Page(props) {
                     <hr />
                     <strong>Filters</strong>
                     <p>
-                      Are there any criteria where you’d want to filter out rows
-                      from being included in this Schedule?
+                      Only include rows that meet the following criteria when
+                      checking this Schedule:
                     </p>
 
                     <Table bordered size="sm">
                       <thead>
                         <tr>
-                          <td />
-                          <td>
+                          <th />
+                          <th>
                             <strong>Key</strong>
-                          </td>
-                          <td>
+                          </th>
+                          <th>
                             <strong>Operation</strong>
-                          </td>
-                          <td>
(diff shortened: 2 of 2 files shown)


ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
```diff
@@ -497,50 +495,53 @@ export default function Page(props) {
...
-                              <td>
-                                {localFilter.key === "occurredAt" ? (
-                                  <DatePicker
-                                    selected={
-                                      localFilter.match &&
-                                      localFilter.match !== "null"
-                                        ? new Date(
-                                            parseInt(
-                                              localFilter.match.toString()
+                                {!localFilter.op
```
Comment (teallarson) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx:499]: The diff here makes it look like there are many changes.  There are two changes:

(1) the `match` input is no longer housed in its own `<td>`, but shares with operation (like in the Group Rule and Property Filter UIs)
(2)  The `match` input is also now conditional -- only displays if the `op` doesn't contain the string "exist"

ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx
```diff
@@ -197,7 +197,7 @@ export default function Page(props) {
 
     _localFilters.push({
       key: filterOptions[0].key,
-      op: "exists",
+      op: filterOptions[0].ops[0],
```
Comment (pedroslopez) [inline ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx:200]: There will always be at least one available here, yeah? Just making sure we don't run into issues trying to access undefined indexes

pedroslopez (approved):
Looks good! 

With these being so similar now, I wonder if there's an opportunity to make a component that handles these filters that we could reuse. What do you think?

ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx
Comment (teallarson) [reply ui/ui-components/pages/model/[modelId]/property/[propertyId]/edit.tsx:200]: All property types and plugins support, at the very least, equals, so yes!  Could also just statically set it to "equals"

teallarson:
> Looks good!
> 
> With these being so similar now, I wonder if there's an opportunity to make a component that handles these filters that we could reuse. What do you think?

Seems like it.  It's also possible that Group Rules could utilize the same component.

teallarson:
After looking into it, it was a bit more involved to try and combine into a shared, reusable component than I think this story warrants for something that is only reused on two pages (group rules are different in structure and have quite a bit of extra logic).

PR #2766, opened 2022-01-06
Mark records and recordProps as `pending` when a property type changes

rwfeather:
## Change description

When an existing Property has its type modified, we should be checking existing data to see if it conforms to the new datatype.

Ex:
```
email (string) -> email (email)
balance (string) -> balance (float)
```

To enqueue the records for validation, we mark the affected RecordProperties and parent Records as`pending`. 

### Development

### Impact

### Code review

### DIFF
--- a/core/src/models/Property.ts
+++ b/core/src/models/Property.ts
@@ -364,7 +364,7 @@ export class Property extends LoggedModel<Property> {
   @AfterSave
   static async ensureRecordsMarkedPendingOnTypeChange(instance: Property) {
     // Check to see if type changed
-    if (instance.changed("type")) {
+    if (instance.changed("type") && instance.previous("type") != undefined) {
       //mark all relevant recordproperties as "pending"
       await RecordProperty.update(
         { state: "pending" },
--- a/core/src/models/Property.ts
+++ b/core/src/models/Property.ts
@@ -379,7 +379,6 @@ export class Property extends LoggedModel<Property> {
         { state: "pending" },
         { where: { modelId: modelId } }
       );
-
     }
   }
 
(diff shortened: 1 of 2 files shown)


rwfeather:
Performed some manual testing to ensure that the `state` and `invalid` fields are being properly updated and enqueued after a change.

core/src/models/Property.ts
```diff
@@ -361,6 +361,25 @@ export class Property extends LoggedModel<Property> {
...
+        { state: "pending", startedAt: null },
+        { where: { propertyId: instance.id } }
+      );
+
+      const source = await instance.$get("source");
+
+      await GrouparooRecord.update(
+        { state: "pending" },
+        { where: { modelId: source.modelId } }
+      );
```
evantahler, line 379: Noting a realization:

We don't need to look up which Records are effected by this property - by definition, it will be all records in the Model

evantahler (approved):
👍 
Excellent work!

core/src/models/Property.ts
```diff
@@ -361,6 +361,25 @@ export class Property extends LoggedModel<Property> {
     );
   }
 
+  @AfterSave
+  static async ensureRecordsMarkedPendingOnTypeChange(instance: Property) {
+    // Check to see if type changed
+    if (instance.changed("type") && instance.state === "ready") {
```
bleonard, line 367: Does somewhere else do this same thing on changing the options or filter?

For better or worse, they seem to do 
```
await PropertyOps.enqueueRuns(this);
```
Should that be done in this case, too?
Should the "pending" work have been done in those cases, too?

PR #2770, opened 2022-01-07
Primary Source creation UI bug fixes

teallarson:
## Change description

Fixes two bugs:

(1) We had a double `router.push()` to `.../schedules` after creating a source that would leave us with a suspended loading bar at the top of the page.  One has been removed, so NextJS knows when the route change is complete now.

(2) Update to the order of a few things when bootstrapping a property for Primary Sources and the redirect that follows that.  this also involved updating how we track a primary source and whether it's currently bootstrapping a property.  Of note: no more flashing form before redirecting. (thanks @krishnaglick!)

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/components/source/FormMappingSelector.tsx
+++ b/ui/ui-components/components/source/FormMappingSelector.tsx
@@ -111,7 +111,7 @@ const FormMappingSelector: React.FC<Props> = ({
 
   return (
     <Row>
-      <Col md={hasAvailableProperties ? 6 : 12} xl={12}>
+      <Col md={!mappingDisabled && hasAvailableProperties ? 6 : 12} xl={12}>
         <FormInputContainer
           controlId="mapping_source_column"
           label="Source Column"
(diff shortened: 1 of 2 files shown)


ui/ui-components/components/source/FormMappingSelector.tsx
```diff
@@ -138,46 +140,48 @@ const FormMappingSelector: React.FC<Props> = ({
...
-            <FormInputContainer
-              controlId="mapping_property"
-              label="Mapped to Grouparoo Property"
-              required
-            >
-              <Form.Control
-                as="select"
+      {!mappingDisabled && (
+        <Col>
+          {hasAvailableProperties && (
```
edmundito: Consolidate these two checks into one. I think where mappingDisabled is located (above the Col) is the right place to check.

Note that in the first `<Col>` above you will also need to apply similar logic for the medium sizing.

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -83,10 +83,9 @@ const Page: NextPage<Props & InjectedProps> = ({
   );
   const isPrimarySource = useMemo(
     () =>
-      (totalSources === 1 && source.state !== "ready") ||
       properties.filter(
         ({ isPrimaryKey, sourceId }) => isPrimaryKey && sourceId === source.id
-      ).length > 0,
+      ).length > 0 || totalSources === 1,
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx]: The logic can be switched to `totalSources == 1 || properties.filter...` because checking if a value is 1 is a less expensive operation than filter

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -237,6 +216,21 @@ const Page: NextPage<Props & InjectedProps> = ({
       }
     }
 
+    const { properties, examples } = await execApi<Actions.PropertiesList>(
+      "get",
+      `/properties`,
+      {
+        includeExamples: true,
+        state: "ready",
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:226]: I see that `unique: true` was removed when it was moved here. I don't remember a the details but I take it it's not necessary?

edmundito (changes requested):
Looking really good. I have one change to request on the layout of the FormMappingSelector and some questions/suggestions.

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
Comment (teallarson) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:226]: Looks like a copy/paste error.  It doesn't appear to affect the bug fix to have it/not have it.  I put `unique: true` back!

krishnaglick (approved):

edmundito (approved):

PR #2771, opened 2022-01-07
Update navigation to be centered around models

edmundito:
## Change description

This change primarily makes the models and the model overview the center of the navigation in the UI apps.

image

Changes:
- Source schedule is created as "recurring"
- Once the primary source schedule is created, user is redirected to the multiple properties page
- Updates the navigation to show all the models on the side. Records is now tucked away in Platform, and Sources/Group/Destinations list pages are gone
- In nav bar, moved Apps outside of platform and under Dashboard
- Breadcrumbs within the model change from `Type / Topic Name / Tab` to `Model Name / Topic Name / Tab`
- Added breadcrumbs and tabs to model overview to have consistent navigation across all ares
- Spiffied up the navigation bar a bit (minor size and spacing changes)
- Removed navigationModel from global props

### Development

### Impact

### Code review


bleonard:
`Records` (and anything else that's model-specific. is there more?) should be removed from the left menu. That will now be accessible via `model/overview` (See All Records)

core/src/actions/navigation.ts
```diff
@@ -190,13 +188,22 @@ export class NavigationList extends OptionallyAuthenticatedAction {
...
 
     const platformItems: NavigationItem[] = [
-      { type: "link", title: "Apps", href: "/apps" },
       { type: "link", title: "Imports", href: "/imports" },
       { type: "link", title: "Exports", href: "/exports" },
       { type: "link", title: "Runs", href: "/runs" },
       { type: "link", title: "Export Processors", href: "/exportProcessors" },
     ];
 
+    if (modelId) {
```
bleonard: I think if my design works, this endpoint would no longer take `modelId` because there is no more concept of "current model" - I noticed there is still several in an `if` block below in `authenticatedConfigNav`. Might some refactoring to combine then here, too.

edmundito:
Updated to address feedback and I removed the ideas of the current `navigationModel` based on a cookie since it's no longer relevant as the user selects the model

edmundito:
Ok, I made all the changes I wanted to make and this should be ready for review. I  resolved a couple of issues in UI config in regards to where to go after completing the setup or signing in. Previously it was going to the record list of the selected model. Now it will redirect the first model.

teallarson (commented):
This is a nice looking iteration!  

In the future we may want to decide what styling lives on the component/page files and what lives in SCSS modules as there seems to be a mix of it across the UI projects, but I don't see that that needs to be solved here.

core/src/actions/properties.ts
```diff
@@ -108,7 +108,7 @@ export class PropertiesList extends AuthenticatedAction {
       }
     }
 
-    const total = await Property.scope(null).count();
+    const total = await Property.scope(null).count({ where });
```
pedroslopez: Could also use `Property.findAndCountAll` above instead, which will consider the `where` and `includes` for the returned count

ui/ui-components/__tests__/components/tabs.tsx
```diff
@@ -7,24 +7,28 @@ describe("<PropertyTab />", () => {
...
+    render(<PropertyTab property={property} source={source} model={model} />);
   });
 
-  it("renders the plural word properly", async () => {
-    const link = screen.getByRole("link", { name: "Properties" });
-    expect(link).toHaveAttribute("href", "/model/xyz321/properties");
+  it("renders the model link properly", async () => {
+    const link = screen.getByRole("link", { name: model.name });
+    expect(link).toHaveAttribute("href", `/model/${model.id}/overview`);
   });
```
pedroslopez, line 24: Why was this one removed?

ui/ui-components/components/Navigation.tsx
```diff
@@ -205,60 +203,35 @@ export default function Navigation(props) {
...
-                            </>
-                          ) : null}
-                        </>
-                      }
-                      icon={nav.icon}
-                      idx={idx}
-                    />
-                  </Fragment>
+                  <HighlightingNavLink
+                    key={idx}
```
pedroslopez: `nav.href` as a key like was being used before is probably more appropriate, unless we don't have unique links now?

ui/ui-components/pages/object/[id].tsx
```diff
@@ -123,21 +121,7 @@ export default function FindObject(props) {
...
-      "properties",
-      "schedules",
-    ];
-    if (byModel.includes(tableName) && props.navigationModel.value) {
-      return `/model/${navigationModel.value}/${tableName}`;
-    }
-
-    return `/${tableName}`;
-  }
+  const getListPage = (tableName: string) => `/${tableName}`;
```
Comment (pedroslopez) [inline ui/ui-components/pages/object/[id].tsx]: I believe this means that in community edition, where we show list pages instead of edit pages for most "tables", it would link to something like `/groups` or `/properties`, which would 404. Let's check what the behavior is for going to `/object/some_group_id` in community edition after this change and ensure it still works. I think some additional change would need to be made to instead take them to the model page or something like that.

ui/ui-components/pages/setup.tsx
```diff
@@ -37,10 +35,16 @@ export default function Page(props) {
...
+      models: [model],
+    } = await execApi<Actions.ModelsList>("get", "/models", {
+      limit: 1,
+    });
+
     const response = await execApi(
       "get",
       `/setupSteps`,
-      { modelId: navigationModel.value },
+      { modelId: model?.id },
```
pedroslopez: I don't think it's necessary to have this as part of the action input anymore since the model won't be coming from the UI. It can probably be determined somewhere in the action or closer to where it's being used.

ui/ui-components/utils/modelHelper.ts
```diff
@@ -1,45 +1,6 @@
...
-      pathname: newPath,
-    });
-  } else {
-    // model is not in url,
-    // but we should still refresh to update nav menu
-    router.reload();
-  }
-};
-
 export const getModelFromUrlOrCookie = (ctx: NextPageContext) => {
```
pedroslopez: Is the cookie part of this still needed? (and the `setModelCookie` function?)

pedroslopez (changes requested):
Nice work!

Mainly some non-blocking questions- only thing I would consider blocking is to check out the `/object/` page in community edition with the change to the function to get the relevant object's list page and ensuring that still works.

ui/ui-components/__tests__/components/tabs.tsx
edmundito: The breadcrumbs have changed from:

`Properties / {Property Name} / Edit`

To:

`{Model Name} / {Property Name} / Edit`

ui/ui-components/components/Navigation.tsx
edmundito: Make sense, I must have replaced it with the wrong thing when I removed the Fragment.

ui/ui-components/pages/setup.tsx
edmundito: Agreed. I was working on this while @evantahler was making a change in #2773 but now that it's done and merged this can be removed and moved into the action itself.

ui/ui-components/utils/modelHelper.ts
edmundito: Nope, will remove.

ui/ui-components/pages/object/[id].tsx
Comment (edmundito) [reply ui/ui-components/pages/object/[id].tsx]: Ok. I need to rethink how this is done right now because I will need to resolve the model id some other way.

core/src/actions/setupSteps.ts
```diff
@@ -4,13 +4,14 @@ import { AsyncReturnType } from "type-fest";
...
 import { ActionPermission } from "../models/Permission";
 import { ParamsFrom } from "actionhero";
+import { GrouparooModel } from "..";
 
 export class SetupStepsList extends AuthenticatedAction {
   name = "setupSteps:list";
   description = "List the SetupSteps and their status";
   permission: ActionPermission = { topic: "setupStep", mode: "read" };
   outputExample = {};
-  inputs = { modelId: { required: false } };
```
pedroslopez, line 13: let's also remove the `modelId` from the `runWithinTransaction` method param types - looks like `{ params }: { params: { modelId?: string } }` is still left over

ui/ui-components/pages/object/[id].tsx
```diff
@@ -54,10 +54,24 @@ export default function FindObject(props) {
       response.records.length === 1 &&
       grouparooUiEdition() === "community"
     ) {
-      const listPage = getListPage(response.records[0].tableName.toLowerCase());
-      router.push(listPage);
+      const tableName = response.records[0].tableName.toLowerCase();
```
Comment (pedroslopez) [inline ui/ui-components/pages/object/[id].tsx]: Note that there's an incoming change in this area here https://github.com/grouparoo/grouparoo/pull/2783

core/src/actions/properties.ts
```diff
@@ -108,7 +108,7 @@ export class PropertiesList extends AuthenticatedAction {
       }
     }
 
-    const total = await Property.scope(null).count();
```
pedroslopez, line 111: Sorry, I should have been more clear in my suggestion. The tests are failing because `findAndCountAll` returns an object with the rows and the count, so you can be sure the total count refers to whatever you're looking up. My suggestion was to replace `await Property.scope(null).findAll({` on line 71 with `await Property.scope(null).findAndCountAll({` - here's an example on the `exports` action:

https://github.com/grouparoo/grouparoo/blob/c7bce3cac78f48bfc6ace66da041c4c538a95257/core/src/actions/exports.ts#L45

pedroslopez (commented):
Looking good- I think I know why that test is failing. More info in the comments :D

ui/ui-components/pages/object/[id].tsx
Comment (edmundito) [reply ui/ui-components/pages/object/[id].tsx]: Ahh maybe I should apply the change here instead.

edmundito:
Ok, I pushed a few more changes to fix some issues with the community edition:

- All model links point to the overview page instead of the edit page
- The edit links are removed from the model overview page in the community edition
- Groups point to members in the community edition and rules everywhere else
- Hide the "New Model" link in community edition
- The `/object/{recordId}` page now redirects to the record edit page (implemented the changes from #2783)

Finally, addressed @pedroslopez's comments

core/src/actions/navigation.ts
```diff
@@ -102,16 +91,45 @@ export class NavigationList extends OptionallyAuthenticatedAction {
...
+    }
+
+    navigationItems.push({
+      type: "link",
+      title: "New Model",
+      icon: "plus",
+      href: `/model/new`,
+      mainPathSectionIdx: 2,
+      small: hasModels,
+      hideOn: ["community"],
```
pedroslopez: Note that you should have access to the ui edition here through `process.env.GROUPAROO_UI_EDITION`, so passing this along might not be necessary- not a huge deal though

core/src/actions/properties.ts
pedroslopez: This works, but now the data is being fetched twice. 

I'm referring to removing this line from here completely and making a change higher up:

https://github.com/grouparoo/grouparoo/blob/041e196e850bf83d12ff6012dfd555d9437b99c0/core/src/actions/properties.ts#L70-L80

This would instead be something like:

```ts
    const { rows, count: total } = await Property.scope(null).findAndCountAll({ 
        include, 
        limit: params.limit, 
        offset: params.offset, 
        order: params.order, 
        where, 
    });

    const properties = rows.filter((p) => 
	params.modelId ? p.source?.modelId === params.modelId : true 
   ); 
```
That way we're grabbing the total and finding in the same call, guaranteeing the same args.
Let me know if you want to have a quick chat

edmundito: Oh of course. 🤦🏼  Despite the name of the function being super clear about what it didn't I didn't think too much into it when I swapped it.

edmundito:
@pedroslopez I fixed the couple of things you mentioned in your review.

pedroslopez (approved):
Woo! As far as the changes this PR is addressing, looks good to me.

I did notice that sometimes the breadcrumbs show the model capitalized and other times it's not, but I'm not sure why (this was tested in community edition):

Destination exports:
image

Single Record view:
image

I assume this is probably because of showing the ID vs the name in some cases.

Also, this is unrelated but I'm noticing now after this change community edition has no way of running a specific schedule and can now only run all. Added a story for that here https://www.pivotaltracker.com/story/show/180894470.

core/src/actions/navigation.ts
```diff
@@ -12,29 +12,26 @@ type NavigationMode =
...
 
 export class NavigationList extends OptionallyAuthenticatedAction {
   name = "navigation:list";
   description = "returns a list of pages for the UI navigation for this user";
   permission: ActionPermission = { topic: "*", mode: "read" };
   outputExample = {};
-  inputs = {
-    modelId: { required: false },
-  };
+  inputs = {};
```
evantahler: Note: You can delete them if they aren't in use - not a required property for an Action

ui/ui-components/__tests__/utils/modelHelper.ts
```diff
@@ -1,109 +0,0 @@
-import type { NextPageContext } from "next";
```
evantahler, line 1: 👋

ui/ui-components/components/import/List.tsx
```diff
@@ -13,6 +13,7 @@ import { ErrorHandler } from "../../utils/errorHandler";
 import { ImportRecordPropertiesDiff, ImportGroupsDiff } from "./Diff";
 import StateBadge from "../badges/StateBadge";
 import { capitalize } from "../../utils/languageHelper";
+import { grouparooUiEdition } from "../../utils/uiEdition";
```
evantahler: Un-used

evantahler (approved):
This looks great.

Simply because of the number of UI files changed, we'll likely want to do a pre-v0.8 release bug bash, just to make sure that everything still works as expected. (cc @bleonard)

edmundito:
@pedroslopez I addressed the couple of issues you mentioned and we should be good to go!

I did notice that we were still redirecting to `/models` in some instances so I made a couple of updates to redirect to either `/` or `/model/new` depending on the context.

PR #2772, opened 2022-01-08
Find one airtable record at a time to prevent duplicate issues.

bleonard:
## Change description

Customer report:
I think there may be a bug in the airtable client that checks the table for records by field value. I ran our export job on a table that had a very large amount of items and some of those items it turned out had duplicates. These duplicate foreign keys resulted in the list records by field to build the conditional and return all records associated with that foreign key but didn't return any of the items for other foreign keys in the conditional statement. This then resulted in those foreign keys that weren't found being created again.

I propose we use a similar method to find the records but rather than batch the records in the list by field we have that method only return a single record if it exists in the table.

### Development

### Impact

### Code review


plugins/@grouparoo/airtable/src/lib/client/client.ts
```diff
@@ -209,19 +209,25 @@ export class Client implements IClient {
     return this.baseClient(tableIdOrName).select().all();
   }
 
-  async listRecordsByField(
```
bleonard, line 212: This method is the main change.

plugins/@grouparoo/airtable/src/lib/client/client.ts
```diff
@@ -209,19 +209,25 @@ export class Client implements IClient {
...
-    const conditionals = foreignKeys.map((key) => `{${primaryKey}}="${key}"`);
-    const filterString = `OR(${conditionals.join(",")})`;
-    return this.baseClient(tableIdOrName)
+    foreignKey: string
+  ): Promise<Record<FieldSet>> {
+    // escape separators
+    foreignKey = foreignKey.replace(/\"/gi, '\\"');
+    primaryKey = primaryKey.replace(/\{/gi, "\\{");
+    const filterByFormula = `{${primaryKey}}="${foreignKey}"`;
+    const list = await this.baseClient(tableIdOrName)
```
evantahler, line 221: Confirming that you want to send up a literal `{}` => `{user_id}="123"`

plugins/@grouparoo/airtable/__tests__/utils/shared.ts
```diff
@@ -5,16 +5,7 @@ export async function getRecordByName(
...
-    name,
-  ]);
-  if (records.length === 0) {
-    return null;
-  }
-  if (records.length > 1) {
-    throw new Error(`More than one (${records.length}) result: ${name}`);
-  }
-  return records[0];
+  return client.getRecordByKey(tableIdOrName, "Name", name);
```
evantahler, line 8: That's a nice helper

evantahler (commented):
Code-wise this looks Ok.

I don't think I understand what used to be happening - the Destination Options require you to provide a single `Primary Key`, so where did the old `foreignKeys: string[]` come from?  Isn't there ever only 1 foreignKey for each record?

bleonard:
This was from the batch helper so we sent say, 10 emails at a time to look up the 10 rows all at once (which rows match these emails). But it turns out the world is more complicated and maybe that sheet has one of those emails repeated 102 times in that sheet. So then we get back that email 100 times and not the other 9 row it should have found. The new code does one at a time to find the first row for that email.

evantahler (approved):

PR #2774, opened 2022-01-10
Rename Import `exporting` state to `processing`

pedroslopez:
## Change description

The `exporting` state introduced in #2751 didn't quite represent what was going on accurately, since exports are only being _created_ and the import does not wait for exports to be fully completed. The "exporting" name could cause users to expect to see imports in that state until they've been exported.

This also changes the timestamp to match the new name, since `exportedAt` was also not quite accurate and could cause confusion.

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/ops/runs.ts
+++ b/core/src/modules/ops/runs.ts
@@ -72,9 +72,9 @@ export namespace RunOps {
     const data = [];
     const start = run.createdAt.getTime();
 
-    const lastExportedImport = await Import.findOne({
+    const lastProcessedImport = await Import.findOne({
       where: { creatorId: run.id },
-      order: [["exportedAt", "desc"]],
+      order: [["processedAt", "desc"]],
       limit: 1,
     });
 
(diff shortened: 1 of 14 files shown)


core/src/modules/ops/runs.ts
```diff
@@ -119,9 +119,9 @@ export namespace RunOps {
               },
             },
           }),
-          exported: await run.$count("imports", {
+          processed: await run.$count("imports", {
```
evantahler, line 122: This method is used on the run details page to show this chart
Screen Shot 2022-01-10 at 8 50 26 AM

Are there any changes needed in the UI, or do we just display whatever the API returns?

evantahler (approved):
Looks great! 
Please confirm that the run details page still looks good, and then this is ready to merge!

core/src/modules/ops/runs.ts
pedroslopez: image

Looks ok even on old runs

PR #2781, opened 2022-01-11
Update `uzmug`, `ah-sequelize-plugin` and `sequelize-typescript` packages

evantahler:
### Development

### Impact

### Code review


core/src/models/Property.ts
```diff
@@ -40,7 +40,7 @@ import { Run } from "./Run";
 import { Source } from "./Source";
 
 const jsMap = {
-  boolean: config.sequelize.dialect === "sqlite" ? "text" : "boolean", // there is no boolean type in SQLite
+  boolean: config?.sequelize?.dialect === "sqlite" ? "text" : "boolean", // there is no boolean type in SQLite
```
evantahler, line 43: This is interesting - when running the `./bin/migrator` command, we don't actually have the full actionhero config loaded up.  We did refrence the FilterDictionary in a migration, which then imports Filter, then Property...

core/package.json
```diff
@@ -64,11 +64,11 @@
     "reflect-metadata": "0.1.13",
     "semver": "7.3.5",
     "sequelize": "6.11.0",
-    "sequelize-typescript": "2.1.1",
+    "sequelize-typescript": "2.1.2",
     "sqlite3": "5.0.2",
     "tar": "6.1.11",
     "ts-node": "10.4.0",
-    "umzug": "v3.0.0-beta.5",
+    "umzug": "v3.0.0",
```
pedroslopez: Why are we using the `v` prefix exclusively for umzug? 
Also, is the explicit dependency necessary? I don't see we're using it directly, so I _believe_ it should be able to be removed like in https://github.com/grouparoo/telemetry/pull/355

evantahler:
> Why are we using the `v` prefix exclusively for umzug? Also, is the explicit dependency necessary? I don't see we're using it directly, so I _believe_ it should be able to be removed like in [grouparoo/telemetry#355](https://github.com/grouparoo/telemetry/pull/355)

Great point! Trying that now...

pedroslopez (approved):

PR #2789, opened 2022-01-12
Mailchimp OAuth App

pedroslopez:
## Change description

This PR builds on the addition of mailchimp as an oauth provider to the Grouparoo Auth server.

- Adds new `oauth-token` option type to handle oauth popups and loading options for apps
    - Tweaks option obfuscation a bit to support also obfuscating this new type of option, in addition to `password`
- Fixes an issue with processing export queues when a plugin defines multiple apps
- Adds oauth support for Mailchimp plugin via new `mailchimp-oauth` app

mailchimp-oauth

### Development

### Impact

### Code review


plugins/@grouparoo/mailchimp/__tests__/test.ts
```diff
@@ -0,0 +1,37 @@
...
+      appOptions,
+    });
+
+    expect(success).toBe(true);
+  });
+
+  test("Can connect with an OAuth Token", async () => {
+    const appOptions: SimpleAppOptions = loadAppOptions(newNock, true);
+
+    const { success } = await testConnection({
```
pedroslopez, line 28: Because we're using a Mailchimp client library for everything and do not have custom methods, it seems like testing if we can connect is enough to know if oauth is doing its job

core/src/config/tasks.ts
```diff
@@ -22,7 +22,9 @@ export const DEFAULT = {
           ["system", "imports", "runs", "schedules", "groups", "exports"],
           api?.plugins?.plugins
             .filter((plugin) => plugin.apps?.length > 0)
-            .map((plugin) => plugin.apps.map((app) => `exports:${app.name}`)),
+            .flatMap((plugin) =>
+              plugin.apps.map((app) => `exports:${app.name}`)
+            ),
```
pedroslopez, line 27: This was the fix to get the tasks processing properly when a plugin defined more than one app. It used to be returning something like
```
[
   "groups",
   "exports",
   ["exports:hubspot"],
   [ "exports:mailchimp", "exports:mailchimp-oauth"],
   "records"
]
```

and is now returning something like 

```
[
   "groups",
   "exports",
   "exports:hubspot",
   "exports:mailchimp", 
   "exports:mailchimp-oauth",
   "records"
]
```

core/src/classes/plugin.ts
```diff
@@ -135,6 +135,7 @@ export type PluginOptionType =
   | "text"
   | "textarea"
   | "password"
+  | "oauth-token"
```
pedroslopez, line 139: New option type: `oauth-token`

plugins/@grouparoo/mailchimp/src/lib/connect.ts
```diff
@@ -2,6 +2,12 @@ import { SimpleAppOptions } from "@grouparoo/core";
 import Mailchimp from "mailchimp-api-v3";
 
 export async function connect(appOptions: SimpleAppOptions) {
-  const client = new Mailchimp(appOptions.apiKey?.toString());
-  return client;
+  if (appOptions.apiKey) {
+    return new Mailchimp(appOptions.apiKey?.toString());
+  } else if (appOptions.oAuthToken) {
+    const [authToken, dc] = appOptions.oAuthToken.toString().split("::");
```
pedroslopez: Doing a bit of pre-processing on the token to extract the `dc` (datacenter).

Could also take in the oAuthToken plain and determine the `dc` here by doing a `meta` request, but would require adding something like axios. See my comments here: https://github.com/grouparoo/telemetry/pull/356#discussion_r783170760

plugins/@grouparoo/mailchimp/src/lib/connect.ts
```diff
@@ -2,6 +2,12 @@ import { SimpleAppOptions } from "@grouparoo/core";
...
+
+    //@ts-ignore Mailchimp client typings have not been updated for OAuth support (with dc param)
```
pedroslopez, line 30: Made a PR for this here https://github.com/thorning/node-mailchimp/pull/78 but has not been merged in yet. Once it is, we can remove this `@ts-ignore`

ui/ui-components/components/session/SignIn.tsx
```diff
@@ -53,7 +53,7 @@ export default function SignInForm(props) {
     setLoadingOauthProviders(true);
     const response: Actions.OAuthListProviders = await execApi(
       "get",
-      `/oauth/providers`
+      `/oauth/providers?type=user`
```
pedroslopez, line 56: Note that until this is merged in, if we merge the telemetry counterpart of this PR that adds the mailchimp provider, it will start showing up on the sign in page.

ui/ui-components/pages/app/[id]/edit.tsx
```diff
@@ -132,17 +206,17 @@ export default function Page(props) {
...
                 <Form.Label>Name</Form.Label>
                 <Form.Control
                   required
                   type="text"
+                  name="name"
                   placeholder="Name"
-                  value={app.name}
                   disabled={loading}
-                  onChange={(e) => update(e)}
+                  ref={register}
```
Comment (pedroslopez) [inline ui/ui-components/pages/app/[id]/edit.tsx:207]: Moved to using `react-hook-form` on this page so that we could more easily update the fields as needed on incoming oauth responses.

core/src/modules/ops/app.ts
```diff
@@ -97,9 +97,10 @@ export namespace AppOps {
       throw new Error(`cannot find a pluginApp type of ${app.type}`);
     }
 
-    let sanitizedOptions = await OptionHelper.replaceObfuscatedPasswords(
+    const filteredOptions = OptionHelper.filterEmptyOptions(options);
+    let sanitizedOptions = await OptionHelper.replaceObfuscatedOptions(
```
krishnaglick, line 101: Minor note, you could just do:
```typescript
const sanitizedOptions = OptionHelper.sourceEnvironmentVariableOptions(
  app,
  await OptionHelper.replaceObfuscatedOptions(app, OptionHelper.filterEmptyOptions(options))
);
```

plugins/@grouparoo/mailchimp/__tests__/test.ts
krishnaglick: +1

ui/ui-components/pages/app/[id]/edit.tsx
```diff
@@ -211,54 +269,66 @@ export default function Page(props) {
...
-                                        >
-                                          <em>
-                                            Descriptions:{" "}
-                                            {opt.descriptions
-                                              ? opt.descriptions.join(", ")
-                                              : "None"}
-                                          </em>
-                                        </small>,
-                                      ];
+                                  <Controller
```
Comment (krishnaglick) [inline ui/ui-components/pages/app/[id]/edit.tsx:271]: Controller component for Typeahead integration or react-hook-form didn't work right 🙃

krishnaglick (approved):

evantahler:
Note - looks like the logo for the Mailchimp oAuth app needs to be updated (old blue logo)

core/src/actions/oAuth.ts
```diff
@@ -13,16 +13,23 @@ export type TelemetryError = {
 export class OAuthListProviders extends Action {
   name = "oAuth:listProviders";
   description = "list the available oAuth Providers";
+  inputs = {
+    type: { required: false },
+  };
```
evantahler, line 18: 👍

core/src/actions/oAuth.ts
```diff
@@ -47,6 +56,8 @@ export class OAuthClientStart extends CLSAction {
       id: params.id,
       type: params.type,
       provider: params.provider,
+      appId: params.appId,
+      appOption: params.appOption,
```
evantahler, line 60: These maybe aren't in use?

core/src/models/OAuthRequest.ts
```diff
@@ -64,6 +64,7 @@ export class OAuthRequest extends CommonModel<OAuthRequest> {
       type: this.type,
       consumed: this.consumed,
       provider: this.provider,
+      token: this.consumed ? null : this.token,
```
evantahler, line 67: This seems a bit dangerous to expose, but... I guess it's OK for now.  We could come up with another scheme to load values from `GROUPAROO_OPTION__OAUTH_{ID}` sort of like how we do with ENV, but we wait until later for that

core/src/modules/optionHelper.ts
```diff
@@ -16,7 +16,7 @@ import { LockableHelper } from "./lockableHelper";
 import { plural } from "pluralize";
 import { modelName } from "./modelName";
 
-export const ObfuscatedPasswordString = "__ObfuscatedPassword";
+export const ObfuscatedOptionString = "__ObfuscatedOption";
```
evantahler, line 19: This change is only between the local API and UI - shouldn't cause any breaking changes

plugins/@grouparoo/mailchimp/__tests__/test.ts
evantahler: I was about to say that `__tests__/test` is a silly name for a file, but it makes sense!

plugins/@grouparoo/mailchimp/src/lib/connect.ts
evantahler: Nice!

plugins/@grouparoo/mailchimp/src/lib/connect.ts
evantahler: There's something a bit inflexible that's nagging at me using `::` as a delimiter.  Since we are building a deployed service with Telemetry, making changes there require more effort & coordination.

Is there any value to returning a (serialized) object from Telemetry rather than a :: separated string? If we did that, then we could add additional properties as we need without breaking older versions:

```ts
const {authToken, dc} = JSON.parse(appOptions.oAuthToken.toString()) // oAuthToken = {"authToken":"abc123","dc":"us-east"}
```

So, if one say we also needed to add "language" (us-en), for example, to the response from Telemetry, we could, and it would be backwards compatible.

plugins/@grouparoo/mailchimp/src/initializers/plugin.ts
```diff
@@ -67,6 +67,21 @@ export class Plugins extends Initializer {
...
+              key: "oAuthToken",
+              type: "oauth-token",
+              displayName: "OAuth Access Token",
+              required: true,
+              description:
+                "Mailchimp OAuth access token and datacenter prefix.",
+            },
+          ],
+          methods: { test, parallelism },
+        },
```
evantahler: This small-ish change here makes me very happy!

ui/ui-components/pages/app/[id]/edit.tsx
```diff
@@ -47,14 +54,37 @@ export default function Page(props) {
...
+  useEffect(() => {
+    if (!oAuthPopup) return;
+
+    const onMessage = (event: MessageEvent) => {
+      if (event.source === oAuthPopup) {
+        loadOAuthRequest(event.data.requestId);
+      }
+    };
+
+    window.addEventListener("message", onMessage, false);
```
Comment (evantahler) [inline ui/ui-components/pages/app/[id]/edit.tsx]: I *think* we should be using `globalThis` rather than `window` in these modern times

ui/ui-components/pages/app/[id]/edit.tsx
```diff
@@ -100,17 +132,48 @@ export default function Page(props) {
...
+    setLoadingOAuth(true);
+
+    const response: Actions.OAuthClientStart = await execApi(
+      "post",
+      `/oauth/${providerName}/client/start`,
+      { type: "app", appId: id, appOption: optionKey }
+    );
+    if (response.location) {
+      const windowFeatures =
+        "toolbar=no, menubar=no, width=600, height=850, top=100, left=100";
```
Comment (evantahler) [inline ui/ui-components/pages/app/[id]/edit.tsx:145]: Facinating!

ui/ui-components/pages/oauth/callback.tsx
```diff
@@ -23,10 +23,11 @@ export default function OauthCallbackPage(props) {
     if (response.oAuthRequest) {
       if (response.oAuthRequest.type === "user") {
         router.replace(`/session/sign-in?requestId=${requestId}`);
+      } else if (response.oAuthRequest.type === "app" && window.opener) {
+        window.opener?.postMessage({ requestId });
+        window.close();
```
evantahler, line 28: That's cool

evantahler (approved):

evantahler (commented):
I wonder if the oAuth login flow should also be moved to a pop-up

evantahler:
Great work @pedroslopez - this is a sweeping PR with lots of edges that were well thought out

core/src/actions/oAuth.ts
pedroslopez: `appOption` is being used, but `appId` isn't really. Was going to remove it but thought it couldn't hurt to have just in case, to know which app that `appOption` is referring to

pedroslopez:
> I wonder if the oAuth login flow should also be moved to a pop-up

I think it's ok - it was more of a big deal here since it's a form and we don't want to clear things that users might have typed into other fields. In the case of user login there's nothing that would be cleared.

pedroslopez:
@evantahler re-requesting review since I made some changes to how `dc` is handled. It's now being determined in the plugin's `connect` method so that the auth server only needs to worry about auth and details of what other things the plugin may need to connect are left to the plugin itself. This has a nice benefit of keeping the auth token option in a format that people might expect when looking at it.

side effect is a new dependency, axios, to perform the request but I think it's worth the avoided complexity.

ui/ui-components/pages/app/[id]/edit.tsx
```diff
@@ -47,14 +54,37 @@ export default function Page(props) {
...
+  }, [oAuthPopup]);
+
+  async function edit(appData: Models.AppType) {
     const state = app.state === "ready" ? undefined : "ready";
     setLoading(true);
     const response: Actions.AppEdit = await execApi(
       "put",
       `/app/${id}`,
-      Object.assign({}, app, { state })
+      Object.assign({}, appData, { state })
```
Comment (krishnaglick) [inline ui/ui-components/pages/app/[id]/edit.tsx]: Should use object spread syntax

ui/ui-components/pages/app/[id]/edit.tsx
Comment (krishnaglick) [reply ui/ui-components/pages/app/[id]/edit.tsx:145]: Chrome ignores most of these lol

krishnaglick (approved):

plugins/@grouparoo/mailchimp/__tests__/test.ts
evantahler: Under the hood, this also checks that the DC can be looked up 👍

plugins/@grouparoo/mailchimp/src/lib/connect.ts
```diff
@@ -1,7 +1,26 @@
...
+      `https://login.mailchimp.com/oauth2/metadata`,
+      {
+        headers: {
+          Authorization: `OAuth ${oAuthToken}`,
+          accept: "application/json",
+        },
+      }
+    );
+
+    const datacenter = metaResponse.data?.dc;
```
evantahler: Should we throw if we can't have a dc?

evantahler (approved):
Love it

PR #2792, opened 2022-01-12
Show Pending Properties on Sample Records when a new Property is created

evantahler:
When using `grouparoo config`, after creating a new Property, Sample Records prior to this PR did not properly show that there was a pending property.... now they do!  Having this badge that something is pending will be a good motivator for the user to hit that 'import' button again.

Screen Shot 2022-01-12 at 10 54 00 AM

In other modes, adding a new Property would enqueue a Run to update all the Properties, but that doesn't happen in `grouparoo config` mode.  For this use-case, have a new `@afterSave` hook. 

---

To clarify the environment Grouparoo is running in, this PR also swaps the usage of `process.env.GROUPAROO_RUN_MODE` for `config.general.runMode`.  This provides proper typing and validation for this config setting.  As we modify this value often in tests, Actionhero needed to be updated to the latest version which now exports `rebuildConfig()` which can be used to re-create the config object after ENV changes

### Development

### Impact

### Code review

### DIFF
--- a/apps/staging-config/package.json
+++ b/apps/staging-config/package.json
@@ -22,7 +22,7 @@
     "jest": "27.4.5"
   },
   "scripts": {
-    "dev": "rm -f grouparoo_config.sqlite && cd node_modules/@grouparoo/core && DATABASE_URL=\"sqlite://grouparoo_config.sqlite\" NEXT_DEVELOPMENT_MODE=true GROUPAROO_RUN_MODE=\"cli:config\" WORKERS=0 ./bin/dev",
+    "dev": "cd node_modules/@grouparoo/core && NEXT_DEVELOPMENT_MODE=true GROUPAROO_RUN_MODE=\"cli:config\" WORKERS=0 ./bin/dev",
     "demo": "./node_modules/.bin/grouparoo demo --config",
     "debug": "cd node_modules/@grouparoo/core && ./bin/debug"
   },
--- a/core/__tests__/modules/configWriter.ts
+++ b/core/__tests__/modules/configWriter.ts
@@ -344,6 +344,11 @@ describe("modules/configWriter", () => {
         configObject = await app.getConfigObject();
       });
 
+      afterEach(() => {
+        process.env.GROUPAROO_RUN_MODE = undefined;
+        rebuildConfig();
+      });
+
       test('returns "config:writer" for JS files (LOCKED)', async () => {
         process.env.GROUPAROO_RUN_MODE = "cli:config";
         rebuildConfig();
@@ -356,9 +361,8 @@ describe("modules/configWriter", () => {
           objects: [configObject],
         });
         expect(ConfigWriter.getLockKey(configObject)).toEqual("config:writer");
-        process.env.GROUPAROO_RUN_MODE = undefined;
-        rebuildConfig();
       });
+
       test("returns null for JSON files (UNLOCKED)", async () => {
         process.env.GROUPAROO_RUN_MODE = "cli:config";
         rebuildConfig();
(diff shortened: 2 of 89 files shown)


apps/staging-config/package.json
```diff
@@ -22,7 +22,7 @@
     "jest": "27.4.5"
   },
   "scripts": {
-    "dev": "rm -f grouparoo_config.sqlite && cd node_modules/@grouparoo/core && DATABASE_URL=\"sqlite://grouparoo_config.sqlite\" NEXT_DEVELOPMENT_MODE=true GROUPAROO_RUN_MODE=\"cli:config\" WORKERS=0 ./bin/dev",
+    "dev": "cd node_modules/@grouparoo/core && NEXT_DEVELOPMENT_MODE=true GROUPAROO_RUN_MODE=\"cli:config\" WORKERS=0 ./bin/dev",
```
evantahler, line 25: I also updated the config and `.env.example` for staging-config to use in-memory SQLite, which 
1. is more like what customers will see
2. allows the application to hot-reload without issue when in development mode

core/src/actions/properties.ts
```diff
@@ -168,7 +168,7 @@ export class PropertyCreate extends AuthenticatedAction {
     key: { required: false },
     type: { required: true },
     unique: { required: false, formatter: APIData.ensureBoolean },
-    isArray: { required: false },
+    isArray: { required: false, formatter: APIData.ensureBoolean },
```
evantahler, line 171: This should have been a boolean

evantahler:
Actionhero v28.4.1 released to export the `rebuildConfig` method

https://github.com/actionhero/actionhero/releases/tag/v28.1.4

core/src/models/Property.ts
```diff
@@ -568,6 +549,37 @@ export class Property extends LoggedModel<Property> {
...
+    if (instance.state !== "ready") return;
+
+    const source = await instance.$get("source");
+    if (!source) return;
+
+    const records = await GrouparooRecord.findAll({
+      where: { modelId: source.modelId },
+    });
+    for (const record of records) await record.buildNullProperties();
+  }
```
evantahler, line 581: This is the main logical change of the PR

core/src/initializers/plugins.ts
```diff
@@ -73,9 +73,7 @@ export class Plugins extends Initializer {
     if (uiPlugins) {
       const relevantUiPlugin =
         uiPlugins[
-          process.env.GROUPAROO_RUN_MODE === "ui:config"
-            ? uiPlugins.length - 1
-            : 0
```
evantahler, line 78: Bug! `GROUPAROO_RUN_MODE` was never `ui:*` - it's `cli:*`.

core/__tests__/modules/configWriter.ts
```diff
@@ -364,11 +372,14 @@ describe("modules/configWriter", () => {
...
         expect(ConfigWriter.getLockKey(configObject)).toEqual(null);
         process.env.GROUPAROO_RUN_MODE = undefined;
+        rebuildConfig();
       });
       test('returns "code:config" when in start mode', async () => {
         process.env.GROUPAROO_RUN_MODE = "cli:start";
+        rebuildConfig();
         expect(ConfigWriter.getLockKey(configObject)).toEqual("config:code");
         process.env.GROUPAROO_RUN_MODE = undefined;
+        rebuildConfig();
```
krishnaglick: Do you need to run this since it's the end of the method? Maybe we should just call it in `beforeEach`?

krishnaglick (commented):
Glad I added `rebuildConfig` to Actionhero!

krishnaglick (approved):

core/src/models/Property.ts
```diff
@@ -568,6 +549,37 @@ export class Property extends LoggedModel<Property> {
...
+    if (config.general.runMode !== "cli:config") return;
+    if (instance.state !== "ready") return;
+
+    const source = await instance.$get("source");
+    if (!source) return;
+
+    const records = await GrouparooRecord.findAll({
+      where: { modelId: source.modelId },
+    });
+    for (const record of records) await record.buildNullProperties();
```
pedroslopez, line 580: since config mode should only have a small number of records to test with, loading all the records is probably ok

pedroslopez (approved):
👍🏼

PR #2798, opened 2022-01-13
Suggest an available key on the multiple props page

rwfeather:
## Change description

It was possible for the "add multiple properties" page to suggest property keys that are already being used, leading to an unhelpful error message. 

This change will suggest something less likely to conflict with a previous property.

The logic for making a suggested name is now in the `source:preview` action and occurs at the same time as column types and uniqueness are being speculated on.

Screen Shot 2022-01-13 at 2 42 04 PM
Screen Shot 2022-01-13 at 2 48 25 PM

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/actions/sources/sources.ts
+++ b/core/__tests__/actions/sources/sources.ts
@@ -17,7 +17,6 @@ import {
 import { PropertyDestroy } from "../../../src/actions/properties";
 import { ConfigWriter } from "../../../src/modules/configWriter";
 
-
 describe("actions/sources", () => {
   helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
   const configSpy = jest.spyOn(ConfigWriter, "run");
--- a/core/__tests__/actions/sources/sources.ts
+++ b/core/__tests__/actions/sources/sources.ts
@@ -202,9 +202,17 @@ describe("actions/sources", () => {
         { id: 2, fname: "luigi", lname: "mario" },
       ]);
       expect(columnSpeculation).toEqual({
-        id: { isUnique: true, type: "integer" },
-        fname: { isUnique: false, type: "string" },
-        lname: { isUnique: false, type: "string" },
+        id: { isUnique: true, type: "integer", suggestedPropertyKey: "id" },
+        fname: {
+          isUnique: false,
+          type: "string",
+          suggestedPropertyKey: "fname",
+        },
+        lname: {
+          isUnique: false,
+          type: "string",
+          suggestedPropertyKey: "lname",
+        },
       });
     });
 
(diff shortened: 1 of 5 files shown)


core/__tests__/actions/sources/sources.ts
```diff
@@ -309,6 +317,58 @@ describe("actions/sources", () => {
       expect(source.state).toBe("ready");
     });
 
+    describe("a source with column(s) that conflict with an existing property", () => {
```
evantahler, line 320: good testing!

core/src/modules/tableSpeculation.ts
```diff
@@ -57,4 +59,17 @@ export namespace TableSpeculation {
...
   }
+
+  export function suggestKey(
+    key: string,
+    model: GrouparooModel,
+    existingProperties: Property[]
+  ): string {
+    for (const property of existingProperties) {
+      if (property.key === key) {
+        return `${ConfigWriter.generateId(model.name)}_${key}`;
```
evantahler: Nice addition! This ensures that the name suggested is human-friendly, even before we safe the config files for something new in ui-config.

core/src/modules/tableSpeculation.ts
```diff
@@ -57,4 +59,17 @@ export namespace TableSpeculation {
...
     return databaseType;
   }
+
+  export function suggestKey(
+    key: string,
+    model: GrouparooModel,
+    existingProperties: Property[]
+  ): string {
+    for (const property of existingProperties) {
+      if (property.key === key) {
```
evantahler: Rather than doing a whole loop like this, you can pluck a matching property from the array:

```ts
export function suggestKey(
    key: string,
    model: GrouparooModel,
    existingProperties: Property[]
  ): {
  const matchingProperty = existingProperties.find(p => p.key === key);
  return matchingProperty ? `${ConfigWriter.generateId(model.name)}_${key}` : key;
}

The return type of the method will also be discovered automatically since we are always returning a string - no need to type the method's return up top
```

ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx
```diff
@@ -199,6 +197,12 @@ export default function Page(props) {
               onChange={(e) => setKey(e.target.value)}
               disabled={disabled}
             />
+            {!exactProperty && key !== column ? (
+              <Alert variant="info" className="mt-2">
+                Property with key "{column}" already exists, suggesting "{key}"
+                instead.
+              </Alert>
+            ) : null}
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx]: I wonder if the tooltip component could be useful here https://react-bootstrap.netlify.app/forms/validation/#tooltips

evantahler (approved):
Good work! There's nothing blocking here, but there may be some refactoring opportunities below.

ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx
Comment (rwfeather) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx]: I could definitely move this language into a form tooltip, but I'm not sure that fits. It looks like our choices for the tooltip are `valid | invalid` and I don't know if what we're communicating fits into that. We're saying essentially: "this value looks different because we needed to pick something that is available in the grouparoo app". Since it's just a small informational tidbit, I thought using the `info` variant would be the semantically consistent approach.

The tooltip/feedback component seems to be more focused on form validation and user input. I think it has the right look stylistically, and if it had the full `primary | info | warning | error | etc` options that alert does I think it would work. But as-is, I'm not sure how to map our key "suggestions" into the green/red options.

Comment (rwfeather) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx]: I'll play with it and take some screenshots

rwfeather:
The alert seems like it stands out a bit too much, I've moved the informational text into a `muted` form label:

Screen Shot 2022-01-19 at 1 03 53 PM

evantahler:
oh! the `muted` label is nice.

PR #2800, opened 2022-01-13
Incremental Schedules

evantahler:
This PR provides the option to have an incremental Schedule - which starts new Runs where the old Run left off, or non-incremental schedules, which import the whole table each time.  Supporting incremental Schedules is up to the plugin.

This allows users to supply the required ORDER BY column without Grouparoo assuming this column is also meant to filter previously seen data out of the Run.  Before this PR, if you didn't have a real HighWatermark column, we might have been skipping records.

Screen Shot 2022-01-25 at 3 49 11 PM

* We still require a column to `ORDER BY`, so LIMIT and OFFSET return sorted data.  The ORDER BY prefers the column supplied, then the mapping column to break ties on secondary tables.  E.g: `SELECT * FROM USERS ORDER BY first_name ASC, user_id ASC`
* We will still store the previous Highwater Mark on Runs, but if the Schedule is not incremental, we won't use it on the next run. 
* Plugins can opt-into providing incremental schedules or not (CSV, Google Sheet)

## Plugin Notes:
- [x] Mongo Order of Aggregation Operations
- [x] CSV should use Source Offset not HWM
- [x] Google Sheets should use Source Offset not HWM
- [x] `getChangedRows` renamed to `getRows`

### Development

### Impact

### Code review


core/src/migrations/000099-addIncrementalToSchedules.ts
```diff
@@ -0,0 +1,23 @@
...
+export default {
+  up: async (
+    queryInterface: Sequelize.QueryInterface,
+    DataTypes: typeof Sequelize
+  ) => {
+    await queryInterface.addColumn("schedules", "incremental", {
+      type: DataTypes.BOOLEAN,
+      allowNull: true,
+      defaultValue: true,
+    });
```
evantahler: Can we can assume that any schedules we have to-date are incremental?

bleonard: Yes, I think so.

bleonard: Though it's interesting that google sheets and CSV and query sources don't really have incremental. So it's not fully clear to me this is a column on schedules and not an option that SQL table sources use to respond to `records`

fabform:
Handy to [find google sheet id](https://fabform.io/a/how-to-find-google-sheet-id)

plugins/@grouparoo/app-templates/src/source/table/getExamples.ts
```diff
@@ -137,10 +137,7 @@ export const getSortableColumnExamples: GetColumnExamplesMethod = async ({
   // only handle columns that support >=
   const choices = [];
   Object.keys(columns).map((colName) => {
-    const { filterOperations } = columns[colName];
-    if (filterOperations.includes(FilterOperation.GreaterThanOrEqual)) {
-      choices.push(colName);
-    }
```
evantahler, line 143: You can now sort by any column. This is for ORDER BY (and potentially HWM)

plugins/@grouparoo/mailchimp/src/lib/import/connection.ts
```diff
@@ -10,6 +10,7 @@ const connection: PluginConnection = {
   displayName: "Mailchimp Import Contacts",
   direction: "import",
   description: "Import or update records with data from Mailchimp contacts.",
+  supportIncrementalSchedule: false,
```
evantahler: ... I think?

.circleci/config.yml
```diff
@@ -4158,7 +4098,6 @@ workflows:
             - test-core-models
             - test-core-modules
             - test-core-notifiers
-            - test-core-ops
```
evantahler, line 4162: There was only one small test here.  Historically, we keep these tests under `/models`.  Moved the test file to save a CI chunk

core/src/migrations/000099-addIncrementalToSchedules.ts
```diff
@@ -0,0 +1,36 @@
...
+        `SELECT * FROM "sources" WHERE type IN ('csv-import-table', 'google-sheet-import')`
+      );
+
+    if (nonIncrementalSources.length > 0) {
+      await queryInterface.sequelize.query(
+        `UPDATE schedules SET incremental=false WHERE id IN (${nonIncrementalSources
+          .map((s) => s.id)
+          .join(", ")})`
+      );
+    }
```
evantahler: Any non-incremental sources I missed?

evantahler:
There were a lot of discrepancies in our file naming conventions which made updating the tests difficult - everything is updated to camelCase to match the method names in the sources.

plugins/@grouparoo/app-templates/src/source/query/meta.ts
```diff
@@ -28,7 +28,7 @@ export interface BuildConnectionMethod {
     options?: ConnectionOptionsOption[];
     executeQuery: ExecuteQueryMethod;
     validateQuery?: ValidateQueryMethod;
-    getChangedRows?: GetChangedRowsMethod;
+    getRows?: GetRowsMethod;
```
evantahler, line 31: Many method names were changed because we `getRows` now... they might not only be the changed ones

core/src/migrations/000099-addIncrementalToSchedules.ts
bleonard: calculated property?
sql queries ones for sure aren't incremental.

plugins/@grouparoo/app-templates/src/source/table/records.ts
```diff
@@ -82,7 +84,10 @@ export const getRecords: GetRecordsMethod = ({ getChangedRows }) => {
       const currentValue = highWaterMark[highWaterMarkAndSortColumnASC];
       const newValue = lastRow[highWaterMarkKey].toString();
 
-      if (currentValue && newValue === currentValue) {
+      if (!schedule.incremental) {
+        nextSourceOffset = sourceOffset + limit;
+        nextHighWaterMark[highWaterMarkAndSortColumnASC] = newValue;
```
bleonard, line 89: is this right? This would change the `where` which out make the limit/offset different.
I would think it would only be `nextSourceOffset = sourceOffset + limit;`

plugins/@grouparoo/mailchimp/src/lib/import/connection.ts
bleonard: actually, it uses ` nextHighWaterMark["last_changed"] = newValue;` in the current `records.ts` code.

core/src/models/Schedule.ts
```diff
@@ -353,15 +376,17 @@ export class Schedule extends LoggedModel<Schedule> {
...
-  static async ensureSourceCanUseSchedule(instance: Schedule) {
-    const source = await Source.findById(instance.sourceId);
-    if (source.state !== "ready") throw new Error("source is not ready");
-
-    const scheduleAvailable = await source.scheduleAvailable();
-    if (!scheduleAvailable) {
-      throw new Error(
-        `source ${source.name} (${instance.sourceId}) cannot have a schedule`
-      );
+  @BeforeUpdate
```
pedroslopez: Should this just be `@BeforeSave`? (as opposed to both `@BeforeCreate` and `@BeforeUpdate`)

core/src/models/Schedule.ts
```diff
@@ -289,6 +298,7 @@ export class Schedule extends LoggedModel<Schedule> {
       name,
       sourceId,
       recurring,
+      incremental,
       recurringFrequency,
```
pedroslopez: nitpick: maybe keep `recurring` and `recurringFrequency` next to eachother on generated config objects?

plugins/@grouparoo/app-templates/src/source/query/meta.ts
```diff
@@ -56,22 +56,23 @@ export const buildConnection: BuildConnectionMethod = ({
...
-  const records = getChangedRows ? getRecordsMethod(getChangedRows) : null;
+  // const scheduleOptions = getRows ? getScheduleOptions() : null;
+  const importRecords = getRows ? getRecordsMethod(getRows) : null;
 
   const plugin: PluginConnection = {
     name,
     displayName,
     direction: "import",
     description,
+    supportIncrementalSchedule: true,
```
pedroslopez: Do query sources do anything "incremental" with the high watermark? Fuzzy on this one.

plugins/@grouparoo/app-templates/src/source/table/records.ts
evantahler: We decided to keep storing the highWaterMarks, even though we aren't using them.  This way, you could choose to turn on  incremental later and use the most recent highWaterMark from the last run

core/src/migrations/000099-addIncrementalToSchedules.ts
```diff
@@ -0,0 +1,36 @@
...
+        `SELECT * FROM "sources" WHERE type LIKE '%-import-table'`
+      );
+
+    if (incrementalSources.length > 0) {
+      await queryInterface.sequelize.query(
+        `UPDATE schedules SET incremental=true WHERE id IN (${incrementalSources
+          .map((s) => s.id)
+          .join(", ")})`
+      );
+    }
```
pedroslopez: The select statement above is looking for sources, whereas this update query is targeting schedules based on those IDs we just looked up. Did you mean to use WHERE sourceId here?

plugins/@grouparoo/app-templates/src/source/query/records.ts
```diff
@@ -30,6 +31,7 @@ export const getRecordsMethod = (getChangedRows: GetChangedRowsMethod) => {
     appOptions,
     properties,
     sourceOptions,
+    schedule,
   }) => {
     let offset = highWaterMark.offset
```
pedroslopez: Is this also a case where a source is using highWatermark as a `sourceOffset`?

plugins/@grouparoo/csv/src/lib/shared/parseProfiles.ts
```diff
@@ -1,30 +1,24 @@
...
-import {
-  plugin,
-  Source,
-  SourceMapping,
-  Run,
-  HighWaterMark,
-} from "@grouparoo/core";
+import { plugin, Source, SourceMapping, Run } from "@grouparoo/core";
 
 export async function parseProfiles({
```
pedroslopez: should this (and the file) be called `getRows` to match the other plugins?
not a big deal, but since we're doing so much renaming... 😛

plugins/@grouparoo/mailchimp/src/lib/import/connection.ts
```diff
@@ -10,6 +10,7 @@ const connection: PluginConnection = {
   displayName: "Mailchimp Import Contacts",
   direction: "import",
   description: "Import or update records with data from Mailchimp contacts.",
+  supportIncrementalSchedule: true,
```
pedroslopez, line 13: might be more to do to handle `incremental: false`? unless I missed the changes to `getChangedMembers` 🤔

plugins/@grouparoo/mysql/__tests__/table-import/importProperty.ts
```diff
@@ -286,7 +286,7 @@ describe("mysql/table/recordProperty", () => {
     //   relativeMatchDirection?: string;
     // }
 
-    describe.only("exists", () => {
+    describe("exists", () => {
```
pedroslopez, line 289: nice catch!

plugins/@grouparoo/mailchimp/src/lib/import/connection.ts
evantahler: It should be handled! 

https://github.com/grouparoo/grouparoo/blob/08fc49dee423593c56ba44b2f882b05a66081925/plugins/%40grouparoo/mailchimp/src/lib/import/members.ts#L147-L152

The only change would be that if incremental=false, when a run starts, it doesn't get the last run's HWM, so it starts over.  The plugin handles the HWM being null already by not adding that to the query

plugins/@grouparoo/app-templates/src/source/query/meta.ts
evantahler: Oh! not yet!

plugins/@grouparoo/app-templates/src/source/query/records.ts
evantahler: It sure is! Good catch!  Fixed up in 1308c3130f092034755fe8d7aaf5b4cf17e54173 (CC @bleonard)

core/src/migrations/000099-addIncrementalToSchedules.ts
```diff
@@ -0,0 +1,36 @@
...
+        `SELECT * FROM "sources" WHERE type LIKE '%-import-table'`
+      );
+
+    if (incrementalSources.length > 0) {
+      await queryInterface.sequelize.query(
+        `UPDATE schedules SET incremental=true WHERE "scheduleId" IN (${incrementalSources
+          .map((s) => `'${s.id}'`)
+          .join(", ")})`
+      );
+    }
```
pedroslopez:
```suggestion
    if (incrementalSources.length > 0) {
      await queryInterface.sequelize.query(
        `UPDATE schedules SET incremental=true WHERE "sourceId" IN (${incrementalSources
          .map((s) => `'${s.id}'`)
          .join(", ")})`
      );
    }
```

core/src/models/Schedule.ts
```diff
@@ -352,16 +375,17 @@ export class Schedule extends LoggedModel<Schedule> {
...
-      throw new Error(
-        `source ${source.name} (${instance.sourceId}) cannot have a schedule`
-      );
+  @BeforeSave
+  static async checkIncremental(instance: Schedule) {
+    const { pluginConnection } = await instance.getPlugin();
+    if (instance.incremental && !pluginConnection.supportIncrementalSchedule) {
+      if (instance.isNewRecord) {
+        instance.incremental = false;
+      } else {
```
pedroslopez: Would this validation be skipped when loading config files for the first time? (it's a new record so it would just be set to false and not throw?)

pedroslopez (changes requested):
Nice!

Looks like the migration is still invalid

core/src/models/Schedule.ts
evantahler: I agree this is strange!
I think this was an artifact of having a incremental=true default value, but really, the UI/action should make sure that you've set a good value.

pedroslopez (approved):
🎉

PR #2801, opened 2022-01-13
Create an InternalRun when Property Type changes

evantahler:
This PR updates the logic added in https://github.com/grouparoo/grouparoo/pull/2766 to use an InternalRun to update Records when a Property's type changes.  While the logic is similar, and execution time is slower, this approach has a few  benefits:

* Shared Code with other Property Updates
* Related models will also be marked pending and updated, e.g. Groups & Group Members
* Imports will be created to track any changes to Record Properties (e.g. data is invalid now)
* For larger datasets, doing this work in a task will spread out the load over time

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/ops/property.ts
+++ b/core/src/modules/ops/property.ts
@@ -2,6 +2,8 @@ import { Property, SimplePropertyOptions } from "../../models/Property";
 import { Group } from "../../models/Group";
 import { Option } from "../../models/Option";
 import { Mapping } from "../../models/Mapping";
+import { RecordProperty } from "../../models/RecordProperty";
+import { GrouparooRecord } from "../../models/GrouparooRecord";
 import { GroupRule } from "../../models/GroupRule";
 import { internalRun } from "../internalRun";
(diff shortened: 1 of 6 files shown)


rwfeather (approved):

evantahler:
Note - For UI Config, we sort of need to do the old behavior, and make the updates in-line because there are no running tasks

core/src/modules/ops/property.ts
```diff
@@ -14,22 +16,34 @@ export namespace PropertyOps {
...
+      await RecordProperty.update(
+        { state: "pending", startedAt: null },
+        { where: { propertyId: property.id } }
+      );
+
+      const source = await property.$get("source", { scope: null });
+      await GrouparooRecord.update(
+        { state: "pending" },
+        { where: { modelId: source.modelId } }
+      );
```
evantahler: In `cli:confg` mode, we'll do the pending updates in-line.  We don't need to worry about imports or groups in this mode.

rwfeather (approved):

PR #2805, opened 2022-01-14
GROUPAROO_RUN_MODE Bugs

evantahler:
This PR fixes a number of problems on the Grouparoo CLI introduced by https://github.com/grouparoo/grouparoo/pull/2792.  The setting of `process.env.GROPUAROO_RUN_MODE` is atypical in that it is set by the running process itself (part of the init phase of a CLI command).  As such, it is likely that some of other file/module has already loaded the config object and has a local copy in memory.... leading to incorrect values for `config.general.runMode`.  So, this value is not a good candidate for storing in the config.   

This PR reverts those changes, and moves to using a getter utility instead. 
 
### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/runMode.ts
+++ b/core/src/modules/runMode.ts
@@ -1,0 +1,7 @@
+export const getGrouparooRunMode = () =>
+  process.env.GROUPAROO_RUN_MODE as
+    | "cli:run"
+    | "cli:start"
+    | "cli:config"
+    | "cli:apply"
+    | "cli:validate";
--- a/cli/src/utils/loadLocalCommands.ts
+++ b/cli/src/utils/loadLocalCommands.ts
@@ -15,9 +15,10 @@ type ActionheroCLIStub = {
   inputs: {
     [key: string]: ActionheroCLIInputStub;
   };
-  preInitialize: () => Promise<void>;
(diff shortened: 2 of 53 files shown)


core/src/modules/runMode.ts
```diff
@@ -0,0 +1,7 @@
+export const getGrouparooRunMode = () =>
+  process.env.GROUPAROO_RUN_MODE as
+    | "cli:run"
+    | "cli:start"
+    | "cli:config"
+    | "cli:apply"
+    | "cli:validate";
```
evantahler, line 7: At least we can still get types

pedroslopez (commented):
Bummer :/ 

From #2792, I know we added a bunch of `rebuildConfig()`s to the tests... is this PR missing their removal? I don't think it would be needed anymore.

evantahler:
> From #2792, I know we added a bunch of `rebuildConfig()`s to the tests... is this PR missing their removal? I don't think it would be needed anymore.

Oh yeah!  all of those are removed now via https://github.com/grouparoo/grouparoo/pull/2805/commits/54a9e00f25045cbdf3ca4db158c55abcd93fc1b8

pedroslopez (approved):
Once tests pass, looks good!

PR #2807, opened 2022-01-14
Mailchimp OAuth works in config mode

pedroslopez:
## Change description

This PR contains a couple fixes for app oauth:

1. Add the callback page to `ui-config` so that `grouparoo config` can configure oauth apps properly.
2. Disable the "sign in with oauth" button if the app is locked due to being code-configured
3. Small fix to the mailchimp connect method so it always returns a client 

Because of the way we were checking the options after #2789, we could end up without a client if the options were both falsy, leading to ambiguous errors down the line:

image

After this change we're always returning a client again so we get better errors:

image

### Development

### Impact

### Code review

### DIFF
--- a/plugins/@grouparoo/mailchimp/src/lib/connect.ts
+++ b/plugins/@grouparoo/mailchimp/src/lib/connect.ts
@@ -3,9 +3,7 @@ import Mailchimp from "mailchimp-api-v3";
 import axios from "axios";
 
 export async function connect(appOptions: SimpleAppOptions) {
-  if (appOptions.apiKey) {
-    return new Mailchimp(appOptions.apiKey?.toString());
-  } else if (appOptions.oAuthToken) {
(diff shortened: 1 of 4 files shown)


plugins/@grouparoo/mailchimp/src/lib/connect.ts
```diff
@@ -3,9 +3,7 @@ import Mailchimp from "mailchimp-api-v3";
 import axios from "axios";
 
 export async function connect(appOptions: SimpleAppOptions) {
-  if (appOptions.apiKey) {
-    return new Mailchimp(appOptions.apiKey?.toString());
-  } else if (appOptions.oAuthToken) {
+  if (appOptions.oAuthToken) {
```
evantahler, line 6: 👍

evantahler (approved):

PR #2811, opened 2022-01-18
fix: debounce loading destination options on change

pedroslopez:
## Change description

When a destination loads options dynamically based on a text input, it would fetch the new options `onChange`, after every keystroke. This PR adds debouncing to the `loadOptions` function in the destination edit page so it only loads after the user has stopped typing. 

Moving to another event like `onBlur` instead of `onChange` was also considered, but the debounce applied on the function level proved to be more appropriate to handle the different types of options that also exhibited similar issues, like the typeahead.

*Before*:
contacts-predebounce

*After*:
contacts-postdebounce

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/hooks/useDebouncedCallback.ts
+++ b/ui/ui-components/hooks/useDebouncedCallback.ts
@@ -1,0 +1,18 @@
+import { useCallback, useEffect, useRef } from "react";
+
+export const useDebouncedCallback = (callback: Function, ms: number) => {
+  const timeout = useRef<ReturnType<typeof setTimeout>>();
+
+  useEffect(() => {
+    return () => {
+      clearTimeout(timeout.current);
+    };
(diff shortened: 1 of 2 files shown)


ui/ui-components/hooks/useDebouncedCallback.ts
```diff
@@ -0,0 +1,18 @@
...
+    return () => {
+      clearTimeout(timeout.current);
+    };
+  }, []);
+
+  return useCallback(() => {
+    clearTimeout(timeout.current);
+    timeout.current = setTimeout(() => {
+      callback();
+    }, ms);
```
krishnaglick: Nitpick
```suggestion
    timeout.current = setTimeout(callback, ms);
```

krishnaglick (approved):

PR #2799, opened 2022-01-13
Fix method signature for trackAnonymous in customer.io plugin

asdfryan:
## Change description

Customer.IO integration fails at testing connection with the error `Test Failed data.name is required`.

This happens because it's using the `TrackClient` from `customerio-node`, whose method signature for `trackAnonymous` expects 2 parameters: `anonymousId` and `data`. ([Code](https://github.com/customerio/customerio-node/blob/main/lib/track.ts#L74)) Currently, the code is only passing in `data`.

### Development

- [ ] Application changes have been tested appropriately

### Impact

- [ ] Code follows company security practices and guidelines
- [ ] Security impact of change has been considered
- [ ] Performance impact of change has been considered
- [ ] Possible migration needs considered (model migrations, config migrations, etc.)

Please explain any security, performance, migration, or other impacts if relevant:

### Code review

- [ ] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.
- [ ] Relevant tags have been added to the PR (bug, enhancement, internal, etc.)

PR #2812, opened 2022-01-18
Fix method signature for trackAnonymous in customer.io plugin

evantahler:
## Change description 

> This PR updates https://github.com/grouparoo/grouparoo/pull/2799 by @asdfryan

Customer.IO integration fails at testing connection with the error `Test Failed data.name is required`.

This happens because it's using the `TrackClient` from `customerio-node`, whose method signature for `trackAnonymous` expects 2 parameters: `anonymousId` and `data`. ([Code](https://github.com/customerio/customerio-node/blob/main/lib/track.ts#L74)) Currently, the code is only passing in `data`.

### Development

### Impact

### Code review


asdfryan:
Thanks!

plugins/@grouparoo/customerio/src/lib/client.ts
```diff
@@ -18,8 +18,8 @@ class CustomerioClient {
     return this.client.identify(customerId, payload);
   }
 
-  async trackAnonymous(data: object) {
-    return this.client.trackAnonymous(data);
+  async trackAnonymous(anonymousId: string | number, data: object) {
+    return this.client.trackAnonymous(anonymousId, data);
```
rwfeather, line 22: Should we update the type of client to be more specific? I think if we changed line 5 `client: any;` to specify the type then we would have caught this earlier.

rwfeather (approved):

plugins/@grouparoo/customerio/src/lib/client.ts
evantahler: That's a great suggestion!  Better types added in e1cffd4b0c0ad641c3ecab758904907a02b3ae6d

Screen Shot 2022-01-18 at 10 02 27 AM

PR #2814, opened 2022-01-18
fix: disconnect app after options changed

pedroslopez:
## Change description

Apps were not properly being disconnected after options had changed, which meant old options were still being used after connecting.

The issue comes down to using `!this.isSoftDeleted` in the hook that would disconnect the app. `isSoftDeleted` is a function that's always defined on the model, so this would always resolve to false. I did not call the function because I don't think it's appropriate in this case - `isSoftDeleted` only applies to "paranoid" models, which we're not using and would cause an error to be thrown.

See `isSoftDeleted` reference here https://sequelize.org/master/class/lib/model.js~Model.html#instance-method-isSoftDeleted

### Development

### Impact

### Code review

### DIFF
--- a/core/src/models/App.ts
+++ b/core/src/models/App.ts
@@ -114,7 +114,7 @@ export class App extends LoggedModel<App> {
   }
 
   async afterSetOptions(hasChanges: boolean) {
-    if (hasChanges) {
+    if (hasChanges && this.state === "ready") {
       await redis.doCluster(
         "api.rpc.app.disconnect",
         [this.id],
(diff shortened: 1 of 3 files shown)


core/src/models/App.ts
```diff
@@ -114,7 +114,7 @@ export class App extends LoggedModel<App> {
   }
 
   async afterSetOptions(hasChanges: boolean) {
-    if (hasChanges && !this.isSoftDeleted) {
+    if (hasChanges && this.state !== "draft" && !this.isNewRecord) {
```
evantahler, line 117: Yay! 🟢 tests

evantahler (approved):

PR #2817, opened 2022-01-19
Use `GROUPAROO_INTERNAL_BATCH_SIZE` to configure `config.batchSize.recordProperties`

evantahler:
Prior to this PR, the environment variable `GROUPAROO_INTERNAL_BATCH_SIZE` was doing double-duty, controlling how many Records/Properties were being read from Sources at once, and how many Records/Properties we were processing at a time internally.  This was bad because you may want a large batch of records from your source, but internally, the Grouparoo Postgres database should still continue to work on ~500 Records at a time. This PR separates the environment variables:

- `GROUPAROO_IMPORTS_BATCH_SIZE` - How many to records to import at once? (default 1000)
-  (new!)`GROUPAROO_INTERNAL_BATCH_SIZE` - How many to records/recordProperties to process at once? (default 500)
- `GROUPAROO_EXPORTS_BATCH_SIZE` - How many to records to export at once? (default 100) 

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/tasks/recordProperty/enqueue/smallerBatchSize.ts
+++ b/core/__tests__/tasks/recordProperty/enqueue/smallerBatchSize.ts
@@ -1,4 +1,4 @@
-process.env.GROUPAROO_IMPORTS_BATCH_SIZE = "1";
+process.env.GROUPAROO_INTERNAL_BATCH_SIZE = "1";
 
 import { helper } from "@grouparoo/spec-helper";
 import { api, specHelper } from "actionhero";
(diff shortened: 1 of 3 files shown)


bleonard (approved):
This can't hurt, for sure. But I think it's still interesting that it doesn't cap those spots where it's applied to working with RecordProperties which is `GROUPAROO_INTERNAL_BATCH_SIZE * number of exact properties its batching`. We can handle that later if that's the right approach.

evantahler:
> This can't hurt, for sure. But I think it's still interesting that it doesn't cap those spots where it's applied to working with RecordProperties which is `GROUPAROO_INTERNAL_BATCH_SIZE * number of exact properties its batching`. We can handle that later if that's the right approach.

👍 Yep - I want to get this out first to enable folks to up `GROUPAROO_IMPORTS_BATCH_SIZE` without crashing things.

PR #2818, opened 2022-01-19
export record locks

teallarson:
## Change description

Previously, it was possible for two exports to be created for the same Record to a single Destination, resulting in multiple entries in the Destination. 

This PR adds a Redis lock during Export on a Record/Destination pairing.  When a batch is being sent, we now have the following logic:
1. For every export given: 
    i. Ensure it is the newest export for a record/destination pair.  If not, cancel it and look at the next given export.
    ii. Try to get a lock for the record/destination pair for the given export.  If no lock could be obtained, there is already an export in progress for that pair.  Do nothing, it will get tried again later.  If a lock was obtained, push it to an array of all current locks, push the export to an array of exports to be sent.
2. No change: If there are exports to be sent still, gather the needed methods and options and send them.
3. Finally, unlock all locks.

Also introduces a new type of lock `getLock`, which either returns an unlocking function if a lock was created or `null` if a lock was already in place for the given key.

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/destination/plugins/exportRecord.ts
+++ b/core/__tests__/models/destination/plugins/exportRecord.ts
@@ -961,12 +961,14 @@ describe("models/destination - with custom exportRecord plugin", () => {
     });
 
     test("the app can be rate-limited with an error and the export will have a sendAt in the future", async () => {
+      //set a response ahead of time
       exportProfileResponse = {
         success: false,
         error: new Error("oh no!") as ErrorWithRecordId,
         retryDelay: 1000,
       };
 
+      //make a group and some destination group memberships
       const group = await helper.factories.group();
       const destinationGroupMemberships: Record<string, any> = {};
       destinationGroupMemberships[group.id] = group.name;
@@ -974,6 +976,7 @@ describe("models/destination - with custom exportRecord plugin", () => {
         destinationGroupMemberships
       );
 
+      //make a record and export it
       const record = await helper.factories.record();
       await destination.exportRecord(record);
       const _export = await Export.findOne({
@@ -982,11 +985,13 @@ describe("models/destination - with custom exportRecord plugin", () => {
 
       await specHelper.runTask("export:enqueue", {});
 
+      //it was sent!
       let foundSendTasks = await specHelper.findEnqueuedTasks("export:send");
       expect(foundSendTasks.length).toBe(1);
 
       await specHelper.runTask("export:send", foundSendTasks[0].args[0]);
 
+      //now reload it after it was sent to check its columns... should have a "sendAt"
       await _export.reload();
       expect(_export.startedAt).toBeFalsy();
       expect(_export.sendAt.getTime()).toBeGreaterThan(new Date().getTime());
@@ -996,13 +1001,15 @@ describe("models/destination - with custom exportRecord plugin", () => {
       expect(_export.state).toBe("pending");
       expect(_export.retryCount).toBe(0);
 
-      // when the response is back to success
+      //then set the response back to success
       exportProfileResponse = {
         success: true,
         error: undefined,
         retryDelay: undefined,
       };
 
+      //and export it again, it should process now
+      foundSendTasks = await specHelper.findEnqueuedTasks("export:send");
       await specHelper.runTask("export:send", foundSendTasks[0].args[0]);
       await _export.reload();
       expect(_export.completedAt).toBeTruthy();
--- a/core/__tests__/models/destination/plugins/exportRecord.ts
+++ b/core/__tests__/models/destination/plugins/exportRecord.ts
@@ -961,14 +961,12 @@ describe("models/destination - with custom exportRecord plugin", () => {
     });
 
     test("the app can be rate-limited with an error and the export will have a sendAt in the future", async () => {
-      //set a response ahead of time
       exportProfileResponse = {
         success: false,
         error: new Error("oh no!") as ErrorWithRecordId,
         retryDelay: 1000,
       };
 
-      //make a group and some destination group memberships
       const group = await helper.factories.group();
       const destinationGroupMemberships: Record<string, any> = {};
       destinationGroupMemberships[group.id] = group.name;
@@ -976,7 +974,6 @@ describe("models/destination - with custom exportRecord plugin", () => {
         destinationGroupMemberships
       );
 
-      //make a record and export it
       const record = await helper.factories.record();
       await destination.exportRecord(record);
       const _export = await Export.findOne({
@@ -985,13 +982,11 @@ describe("models/destination - with custom exportRecord plugin", () => {
 
       await specHelper.runTask("export:enqueue", {});
 
-      //it was sent!
       let foundSendTasks = await specHelper.findEnqueuedTasks("export:send");
       expect(foundSendTasks.length).toBe(1);
 
       await specHelper.runTask("export:send", foundSendTasks[0].args[0]);
 
-      //now reload it after it was sent to check its columns... should have a "sendAt"
       await _export.reload();
       expect(_export.startedAt).toBeFalsy();
       expect(_export.sendAt.getTime()).toBeGreaterThan(new Date().getTime());
@@ -1001,14 +996,12 @@ describe("models/destination - with custom exportRecord plugin", () => {
       expect(_export.state).toBe("pending");
       expect(_export.retryCount).toBe(0);
 
-      //then set the response back to success
       exportProfileResponse = {
         success: true,
         error: undefined,
         retryDelay: undefined,
       };
 
-      //and export it again, it should process now
       foundSendTasks = await specHelper.findEnqueuedTasks("export:send");
       await specHelper.runTask("export:send", foundSendTasks[0].args[0]);
       await _export.reload();
--- a/core/__tests__/models/destination/plugins/exportRecord.ts
+++ b/core/__tests__/models/destination/plugins/exportRecord.ts
@@ -1002,7 +1002,6 @@ describe("models/destination - with custom exportRecord plugin", () => {
         retryDelay: undefined,
       };
 
-      foundSendTasks = await specHelper.findEnqueuedTasks("export:send");
       await specHelper.runTask("export:send", foundSendTasks[0].args[0]);
       await _export.reload();
       expect(_export.completedAt).toBeTruthy();
--- a/core/src/modules/locks.ts
+++ b/core/src/modules/locks.ts
@@ -25,7 +25,7 @@ export async function waitForLock(
     );
   }
 
-  const set = await client.setnx(lockKey, requestId);
+  const set = await client.setnx(lockKey, requestId); // lock:grouparoo:export-locker:rec_abc:des_efg
   const checkValue = await client.get(lockKey);
 
   if (!set || checkValue !== requestId) {
@@ -41,3 +41,32 @@ export async function waitForLock(
 
   return { releaseLock, attempts };
 }
+
+
+// option 1: sleep 
+ 
+try {
+  const {releaseLock} = await waitForLock('mystring')
+  // do stuff like send exports
+} finally {
+  releaseLock()
+}
+
+// option 2: re-enuque job for later
+
+  const {myturn, releaseLock} = await getLock(_export) // use the record and destination id to make a lock string // lock:grouparoo:export-locker:rec_abc:des_efg
+  if (myturn) {
+    try {
+    // do stuff like send exports
+    finally {
+      releaseLock()
+    }
+  } else (
+    // A) re-enqueue this job
+    // B) do nothing, and the export plumbing already knows to try again later << --- WINNER
+   //   * If I'm a newer Export, just wit an I'll be retried
+         // How do I know what new / old means?  
+   //   * If I'm an older Export, cancel me 
+
+  )
+} 
--- a/core/src/modules/locks.ts
+++ b/core/src/modules/locks.ts
@@ -53,10 +53,11 @@ export type getLockResponse = {
 export async function getLock(
   key: string,
   requestId: string = uuid.v4(), // for exports, this will be the export id, we may have use cases that don't pass a request id though, so we'll assign a default
-  ttl = LOCK_DURATION_MS
+  ttl = LOCK_DURATION_MS,
+  lockType?: string
 ): Promise<getLockResponse> {
   const client = api.redis.clients.client;
-  const lockKey = `grouparoo:lock:${key}`;
+  const lockKey = `grouparoo:lock:${lockType && `${lockType}:`}${key}`;
 
   await client.expire(lockKey, Math.ceil(ttl / 1000));
 
@@ -64,19 +65,12 @@ export async function getLock(
 
   const isLocked = !set;
 
-  if (isLocked) {
-    // if it couldn't set
-    // return no unlock function, isLocked = true, and who locked it,
-    const lockValue = await client.get(lockKey);
-
-    return { releaseLock: null, isLocked, lockedBy: lockValue }; //
-  }
-
-  //if it could set it
   async function releaseLock() {
     await client.del(lockKey);
   }
 
+  const lockValue = await client.get(lockKey);
+
   //return how to release it, the fact that it was set, and who it was set by
-  return { releaseLock, isLocked, lockedBy: requestId };
+  return { releaseLock, isLocked, lockedBy: lockValue };
 }
--- a/core/src/modules/locks.ts
+++ b/core/src/modules/locks.ts
@@ -42,25 +42,17 @@ export async function waitForLock(
   return { releaseLock, attempts };
 }
 
-export async function getLock({
-  key,
-  requestId = uuid.v4(),
-  ttl = LOCK_DURATION_MS,
-}: {
-  key: string;
-  requestId?: string;
-  ttl?: number;
-}): Promise<() => void> {
+export async function getLock(key: string, ttl = LOCK_DURATION_MS) {
   const client = api.redis.clients.client;
   const lockKey = `grouparoo:lock:${key}`; // grouparoo:lock:export:record123:dest456
-  let releaseLock: () => void = null;
+  let releaseLock: typeof client.del = null;
 
-  const set = await client.setnx(lockKey, requestId);
+  const set = await client.setnx(lockKey, new Date().getTime());
 
   if (set) {
     await client.expire(lockKey, Math.ceil(ttl / 1000));
     releaseLock = async () => {
-      await client.del(lockKey);
+      return client.del(lockKey);
     };
   }
 
--- a/core/src/modules/locks.ts
+++ b/core/src/modules/locks.ts
@@ -70,14 +70,13 @@ export async function getLock(
     const lockValue = await client.get(lockKey);
 
     return { releaseLock: null, isLocked, lockedBy: lockValue }; //
-  } else {
-    //if it could set it
-
-    async function releaseLock() {
-      await client.del(lockKey);
-    }
-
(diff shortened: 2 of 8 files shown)


teallarson:
@evantahler @pedroslopez this is still a WIP (needs testing)... before I dive into that, would like any feedback on the implementation now that I've got it working and existing tests passing!

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -124,7 +124,7 @@ export namespace helper {
   export const longTime = 60 * 1000; // 60 seconds
   export const mediumTime = 20 * 1000; // 20 seconds
   export const shortTime = 10 * 1000; // 10 seconds
-  export const defaultTime = process.env.CI ? mediumTime : shortTime;
+  export const defaultTime = process.env.CI ? mediumTime : longTime;
```
evantahler: Why the change?  You can tell a specific test of describe to have a longer timeout if needed

core/src/modules/locks.ts
```diff
@@ -41,3 +41,35 @@ export async function waitForLock(
 
   return { releaseLock, attempts };
 }
+
+export type getLockResponse = {
+  releaseLock: Function;
+  isLocked: boolean;
+  lockedBy: string;
+};
```
evantahler: Since you aren't re-using this type anywhere, it's usually better to just have the getLock method determine it's own type from the implementation.   This way, there's nothing to change downsteam

core/src/modules/locks.ts
```diff
@@ -41,3 +41,35 @@ export async function waitForLock(
...
+
+export async function getLock(
+  key: string,
+  requestId: string = uuid.v4(),
+  ttl = LOCK_DURATION_MS,
+  lockType?: string
```
evantahler: for pattern matching, it might be best to give this a default, like `lockType = 'lock'`, this way all the lock keys will have the same number of "parts", eg: `grouparoo:lock:{type}:key` if we need to find or clear them in the future

core/src/modules/locks.ts
```diff
@@ -41,3 +41,35 @@ export async function waitForLock(
...
+): Promise<getLockResponse> {
+  const client = api.redis.clients.client;
+  const lockKey = `grouparoo:lock:${
+    lockType.length > 1 && `${lockType}:`
+  }${key}`;
+
+  const set = await client.setnx(lockKey, requestId);
+
+  await client.expire(lockKey, Math.ceil(ttl / 1000));
```
evantahler: This changes the expire TTL on the key regardless of this caller created it.  I would guess you don't want to mess with someone else's lock unless set===true

core/src/modules/locks.ts
```diff
@@ -41,3 +41,35 @@ export async function waitForLock(
...
+
+  const isLocked = !set;
+
+  async function releaseLock() {
+    await client.del(lockKey);
+  }
```
evantahler: Similar to the above, perhaps if you didn't create the lock, you shouldn't have access to the releaseLock method

core/src/modules/locks.ts
```diff
@@ -41,3 +41,35 @@ export async function waitForLock(
...
+
+  const lockValue = await client.get(lockKey);
```
evantahler: Is the lock value meaningful for the caller?  Perhaps I only care if I was able to get the lock or not

core/src/modules/ops/destination.ts
```diff
@@ -912,13 +913,41 @@ export namespace DestinationOps {
...
   }> {
-    const _exports: Export[] = []; // only ones we are sending
+    const _exports: Export[] = [];
+
     for (const _export of givenExports) {
-      // TODO: maybe should recalcuate hasChanges and from current mostRecent
-      if (!_export.hasChanges) {
-        await _export.complete(); // do not include exports with hasChanges=false
+      const { isLocked, lockedBy } = await getLock(
+        `${_export.recordId}:${_export.destinationId}`,
```
evantahler: 👍 the lock key is a tuple comprised of the record.id and destination.id

core/src/modules/ops/destination.ts
```diff
@@ -912,13 +913,41 @@ export namespace DestinationOps {
...
+            state: "canceled",
+            sendAt: null,
+            errorMessage: `Outdated duplicate of export ${lockingExport.id}`,
+            errorLevel: null,
+            completedAt: new Date(),
+          });
+          return;
+        } else if (lockingExport.id !== _export.id) {
+          //if the export we're looking at didn't create this lock, keep it pending for now
+          return;
```
evantahler: This return would exit the whole function. perhaps `continue` instead to go to the next _export in the loop

core/src/modules/ops/destination.ts
```diff
@@ -981,6 +1010,18 @@ export namespace DestinationOps {
...
     }
 
+    for (const _export of _exports) {
+      // release locks for all of the exports that were processed regardless of outcome - re-enqueues are handled elsewhere
+      const { releaseLock } = await getLock(
+        `${_export.recordId}:${_export.destinationId}`,
+        _export.id,
+        null,
+        "export"
+      );
```
evantahler: Ah, I see while you are already returning `releaseLock` now.  Perhaps above in the first instance of getLock, you could store all the `releaseLock`s you get in an array and `Promise.all(locks.map(releaseLock => releaseLock())`.  This way, you don't need to talk to redis again

evantahler: It's also probably a good idea to wrap most of this method in a try/finally block so you can still release the locks if something goes wrong

core/src/modules/ops/destination.ts
evantahler: I think `lockedBy` is a random UUID, so won't the `await Export.findById(lockedBy);` below always fail to find an Export?  I'll add an idea below to not need store any information in the lock

core/src/modules/ops/destination.ts
```diff
@@ -912,13 +913,41 @@ export namespace DestinationOps {
...
-        await _export.complete(); // do not include exports with hasChanges=false
+      const { isLocked, lockedBy } = await getLock(
+        `${_export.recordId}:${_export.destinationId}`,
+        _export.id,
+        null,
+        "export"
+      );
+
+      if (isLocked) {
+        const lockingExport = await Export.findById(lockedBy);
```
evantahler: Perhaps this should be 

```ts
const mostRecentExport = await Export.findOne({where: {recordId: export.recordId, destinationId: export.destinationId}, order: ['createdAt', 'ASC']})
```

Then, you can be sure you have the most recent export and you can compare if `mostRecentExport.id === _export.id`.  Either "I" am the most recent export or not

core/src/modules/ops/destination.ts
```diff
@@ -912,13 +913,41 @@ export namespace DestinationOps {
...
+        //if the export we're looking at is older than the one that locked this pair, cancel it
+        if (lockingExport.createdAt > _export.createdAt) {
+          await _export.update({
+            state: "canceled",
+            sendAt: null,
+            errorMessage: `Outdated duplicate of export ${lockingExport.id}`,
+            errorLevel: null,
+            completedAt: new Date(),
+          });
+          return;
```
evantahler: Same note about swapping `return` for `continue`

evantahler (commented):
Comments commented!

core/src/modules/locks.ts
bleonard: Yeah, maybe we call this `getLockIfAvailable` or something. It would return `null` if it's already locked and the `releaseLock` if we were able to get it.

core/src/modules/ops/destination.ts
```diff
@@ -912,13 +913,45 @@ export namespace DestinationOps {
...
+        const mostRecentExport = await Export.findOne({
+          where: {
+            recordId: givenExport.recordId,
+            destinationId: givenExport.destinationId,
+            state: "pending",
+          },
+          order: ["createdAt", "ASC"],
+        });
+        //if the export we're looking at is older than the one that locked this pair, cancel it
+        if (mostRecentExport.createdAt > givenExport.createdAt) {
```
bleonard: So if I read this right, we couldn't get the lock (because an export is being processsed in another thread or earlier in this batch) _and_ now we check against other pending exports. That right?

If the `mostRecentExport` is newer, I understand cancelling this one and `continue`.
But if `givenExport` is newer, we will still want to try it later. I think it should still `continue` though because letting it fall through like it is now means we still end up with them in parallel. right?

core/src/modules/ops/destination.ts
```diff
@@ -912,13 +913,45 @@ export namespace DestinationOps {
...
+
+    for (const givenExport of givenExports) {
+      const lock = await getOrSetLock({
+        key: `${givenExport.recordId}:${givenExport.destinationId}`,
+      });
+      locks.push(lock);
+
+      if (lock.isLocked) {
+        // is _export the newest export for this pair?  if not, cancel it.
+        const mostRecentExport = await Export.findOne({
```
bleonard, line 935: Note this could, and mostly commonly would, return `givenExport`.
Maybe it should say `not` that one and then check for null.

core/src/modules/ops/destination.ts
```diff
@@ -981,6 +1014,10 @@ export namespace DestinationOps {
       await app.checkAndUpdateParallelism("decr");
     }
 
+    for (const lock of locks) {
```
bleonard: put this in a `finally` so that if it crashes, we will still release them.

core/src/modules/ops/destination.ts
teallarson: I did what you suggested below, checking for a `moreRecentExport` (the same query as before, but excluding the current export by id) and if one exists, cancelling the one we're looking at.  I think that should address the case you're talking about here, yes?

core/src/modules/ops/destination.ts
tealjulia: I've done that with locks... implementing the try/finally block sounds right but the scoping of it is escaping me.  There are a handful of loops happening here, any of which could fail.

core/src/modules/ops/destination.ts
bleonard: But if `givenExport` is the most recent and `lock.isLocked`, doesn't that mean that we are currently processing an older export that got the lock? In that case, we should wait to process `givenExport` until later. As it is now, it still gets processed in parallel with the one that got the lock.

It's all very tricky, so let me know if I'm wrong.

core/src/modules/ops/destination.ts
```diff
@@ -912,13 +913,45 @@ export namespace DestinationOps {
...
+            errorMessage: `Replaced by more recent export ${mostRecentExport.id}`,
+            errorLevel: null,
+            completedAt: new Date(),
+          });
+          continue;
+        }
+      }
+
+      //things are locked, make sure we're using the absolute most recent data.
+      const updatedExport = await Export.findById(givenExport.id);
```
bleonard: how can this be updated? What changes do you expect?

core/src/modules/locks.ts
tealjulia: I think that makes sense.  During the unlocking in this particular instance, it doesn't much matter who created the lock.  The locks are just stored in an array and processed one-by-one after export.  But in terms of overall design, I think it makes sense to only give the `releaseLock()` out once.

core/src/modules/ops/destination.ts
teallarson: This makes sure we're using the most recent properties.  `oldProperties` and `newProperties` in the export are populated with getters/setters on column read.

core/src/modules/ops/destination.ts
teallarson: Nope that sounds right.  I had a `continue` in there earlier for that case...  It should have stayed.

core/src/modules/ops/destination.ts
tealjulia: I'm wondering if we want `startedAt` or `createdAt` here... `startedAt` has to do with the batch it's in I believe.  In testing, I'm seeing two exports may have the same `startedAt` while they have differing `createdAt`s.

core/src/modules/ops/destination.ts
```diff
@@ -912,82 +912,124 @@ export namespace DestinationOps {
...
+            moreRecentExport.id !== givenExport.id
+          ) {
+            await givenExport.update({
+              state: "canceled",
+              sendAt: null,
+              errorMessage: `Replaced by more recent export ${moreRecentExport.id}`,
+              errorLevel: null,
+              completedAt: new Date(),
+            });
+            continue;
```
bleonard: this one is now extra

teallarson: oh yep!

core/src/modules/ops/destination.ts
bleonard: really? that's fascinating. You're saying by reading `newProperties`, it actually does a `SELECT FROM recordProperties` and not just reading it from what is set in the column? -- that seems surprising to me

I see this:
```
  get newRecordProperties(): ExportRecordProperties {
    return ExportOps.deserializeExportRecordProperties(
      //@ts-ignore
      this.getDataValue("newRecordProperties")
    );
  }
```
But that doesn't refresh the Record/Properties as far as I can tell.

teallarson: You're right.  It calls that function on each read, but that function isn't where they're actually gathered from.  On it.

bleonard: I'm thinking for now we don't actually have to update them. The "newest" checks you do above means it should work out.

bleonard: So I think you can just use `givenExport` below here.

bleonard: Well.... it's tricky. we are only doing that when you _don't_ get the lock now.
I'd say pair with @evantahler today and see where you end up.

teallarson: I think we ended up in a good direction, though I'm still seeing some test failures remotely, so it may need adjusting.

core/src/modules/ops/destination.ts
```diff
@@ -912,82 +923,109 @@ export namespace DestinationOps {
...
+          where: {
+            recordId: givenExport.recordId,
+            destinationId: destination.id,
+          },
+          order: [["createdAt", "DESC"]],
+        });
+
+        const isNewest = mostRecentExport.id === givenExport.id;
+        if (!isNewest) console.info(mostRecentExport);
+        const gotLock = typeof lock === "function";
```
bleonard: what else can it be?

core/src/modules/ops/destination.ts
```diff
@@ -912,82 +923,109 @@ export namespace DestinationOps {
...
+
+        if (gotLock) locks.push(lock);
+        if (!isNewest) {
+          await cancelOldExport(givenExport, mostRecentExport);
```
bleonard, line 945: I like canceling either way. Maybe do this before getting the lock if we are going to do it either way and bail. One less thing to unlock.

core/src/modules/ops/destination.ts
```diff
@@ -912,82 +923,109 @@ export namespace DestinationOps {
...
+          continue;
+        }
+        if (gotLock && isNewest && givenExport.hasChanges)
+          _exports.push(givenExport);
+      }
```
bleonard, line 960: maybe return early here if `_exports.length === 0` - it's now possible.

core/src/modules/ops/destination.ts
```diff
@@ -912,82 +923,109 @@ export namespace DestinationOps {
...
-    }
+    const _exports: Export[] = [];
+    const locks: Awaited<ReturnType<typeof getLock>>[] = [];
 
-    const exportRecords: ExportRecordsPluginMethod = await getBatchFunction(
-      destination
-    );
+    try {
+      for (const givenExport of givenExports) {
+        if (!givenExport.hasChanges) await givenExport.complete();
```
bleonard: probably just `continue` here. Weird to also cancel it lower or something.
Then can remove the `hasChanges` check below as well.

core/src/modules/ops/destination.ts
teallarson: `null`

bleonard: I meant `typeof` - I was just thinking we would use `!lock`

bleonard (approved):

core/src/modules/ops/destination.ts
teallarson: For sure.  @evantahler had recommended this, I believe as a slightly tighter guarantee.

bleonard: As you like. I think it's even weirder if it somehow returns something else and we don't unlock it. I'd rather see the crash in `Promise.all` below if it's not a function somehow. But not strong feelings here.

core/src/modules/ops/destination.ts
evantahler: 👍 agree - might as well always cancel old exports.

core/src/modules/ops/destination.ts
```diff
@@ -912,82 +922,115 @@ export namespace DestinationOps {
...
+        await app.checkAndUpdateParallelism("decr");
+      }
+      return updateExports(
+        destination,
+        _exports,
+        exportResult,
+        combinedError,
+        synchronous
+      );
     } finally {
```
evantahler, line 1031: Theses nested try/catch/finally areas are a little worrysome to me.  Is there a away that we can only have one try/catch here?  The finally's both are doing cleanup work so it's possible to both release all the locks and fixup the parallelism in one go.  Another variable will be needed to keep track of if we should `app.checkAndUpdateParallelism("decr");`, but I think that's worth if to avoid the nesting

core/__tests__/models/destination/plugins/exportRecord.ts
```diff
@@ -996,13 +996,13 @@ describe("models/destination - with custom exportRecord plugin", () => {
...
       expect(_export.retryCount).toBe(0);
 
-      // when the response is back to success
       exportProfileResponse = {
         success: true,
         error: undefined,
         retryDelay: undefined,
       };
 
+      foundSendTasks = await specHelper.findEnqueuedTasks("export:send");
```
evantahler: Did you intend to assert something about foundSendTasks?

core/src/modules/ops/destination.ts
```diff
@@ -912,82 +922,115 @@ export namespace DestinationOps {
...
+        if (!givenExport.hasChanges) {
+          await givenExport.complete();
+          continue;
+        }
 
-    const syncMode = await destination.getSyncMode();
-    const syncOperations: DestinationSyncOperations = syncMode
-      ? DestinationSyncModeData[syncMode].operations
-      : DestinationSyncModeData.sync.operations; // if destination does not support sync modes, allow all
+        const mostRecentExport = await Export.findOne({
```
evantahler, line 935: Calling out that this will, in the short term, slow down this method because we need to SELECT from exports 100 times now (once for each export in the batch).  Once we do https://www.pivotaltracker.com/story/show/181115564, we can skip this lookup

evantahler (approved):
Well done! There was some tricky logic here.

I think there's one more option to clarify what's going on with try/catches, but not required.

core/src/modules/ops/destination.ts
teallarson: The `finally`s can be cleaned up fairly simply.  It seems like we're depending on things throwing in that "inside" try/catch though ([example](https://github.com/grouparoo/grouparoo/blob/04570d22a1772065d3ca17f288cc8b1904bef7b1/core/__tests__/models/destination/plugins/exportRecords.ts#L866))  I'm sure there's a way to get them unnested.  I have not found it yet though.

PR #2821, opened 2022-01-19
Replace UI eventHandlers props with global handlers

edmundito:
## Change description

Since the beginning the event handlers have been global to the app. This changes makes it so that this is only way to use them and it makes it easier to pull an event handler that is already typed instead of having to pull it out of the props.

Additionally, all the event handlers are moved to the `eventHandlers` folder.

Usage:

```ts
// Before:
import { ErrorHandler } from '../../utils/errorHandler'
import { SuccessHandler } from '../../utils/successHandler'

function Page({ errorHandler, successHandler }: { errorHandler: ErrorHandler, successHandler: SuccessHandler }) {
   // Use your handler props here

  <MyComponent successHandler={successHandler} />
}

// Now:
import { errorHandler, successHandler } from '../../eventHandlers'

function Page() {
   // Use your global handlers here

  // No need to pass the successHandler, MyComponent can import it itself
  <MyComponent />
}
```

### Development

### Impact

### Code review


ui/ui-components/pages/_app.tsx
```diff
@@ -12,7 +12,7 @@ import "../components/Icons";
 
 import { WebAppContext } from "../contexts/webApp";
 
-import * as eventHandlers from "../eventHandlers";
+import "../eventHandlers";
```
edmundito, line 15: This just warm boots all those objects before so that they're instanced once the page or component uses them.

evantahler:
Rebase from main to fix the version-checker test

ui/ui-components/client/client.ts
```diff
@@ -6,9 +6,9 @@ import Axios, {
 } from "axios";
 import type { IncomingMessage, ServerResponse } from "http";
 
+import { UploadHandler } from "../eventHandlers/uploadHandler";
```
evantahler, line 9: We may not need this any more as we don't every upload files any more

ui/ui-config/pages/model/[modelId]/records.tsx
```diff
@@ -1,28 +1,24 @@
+import { useRouter } from "next/router";
 import { useState } from "react";
 
+import {
+  errorHandler,
+  recordsHandler,
+  successHandler,
+} from "@grouparoo/ui-components/eventHandlers";
```
Comment (evantahler) [inline ui/ui-config/pages/model/[modelId]/records.tsx:8]: 👍 works form imports.

ui/ui-config/pages/app/new/index.tsx
```diff
@@ -4,7 +4,8 @@ import { useState, useEffect } from "react";
 import { Form, Modal, Spinner, Alert } from "react-bootstrap";
 import { useRouter } from "next/router";
 import AppSelectorList from "@grouparoo/ui-components/components/AppSelectorList";
-import { ErrorHandler } from "@grouparoo/ui-components/utils/errorHandler";
+import { errorHandler } from "@grouparoo/ui-components/eventHandlers";
+import { ErrorHandler } from "@grouparoo/ui-components/eventHandlers/errorHandler";
```
evantahler: I don't think you need the type any more

evantahler (approved):
Yeah! more deletions than additions!  All these *handers were always singletons... we should treat them that way.

Screen Shot 2022-01-19 at 1 30 02 PM

ui/ui-config/pages/app/new/index.tsx
edmundito: Removed!

ui/ui-components/client/client.ts
edmundito: Good note. We may be able to get rid of it when we rebuild useAPI

PR #2825, opened 2022-01-19
Retry failed exports within a time range

pedroslopez:
## Change description

Adds the ability to retry failed exports for a destination within a time range via a new "Retry" tab on Destinations.

image

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/export.ts
+++ b/core/__tests__/models/export.ts
@@ -555,6 +555,269 @@ describe("models/export", () => {
     });
   });
 
+  describe("retryFailed", () => {
+    const startDate = new Date("2022-01-01T10:00:00Z");
+    const endDate = new Date("2022-01-01T12:00:00Z");
+
+    let oldFailedExport: Export;
+    let inRangeFailedExport: Export;
+    let foreignFailedExport: Export;
+    let completeExport: Export;
+    let retryingExport: Export;
+
+    beforeAll(async () => {
+      await Export.truncate();
+      const destination2 = await helper.factories.destination();
+
+      oldFailedExport = await Export.create({
+        recordId: record.id,
+        destinationId: destination.id,
+        oldRecordProperties: {},
+        newRecordProperties: {},
+        newGroups: [],
+        oldGroups: [],
+        sendAt: null,
+        startedAt: new Date(1),
+        errorMessage: "Oh No!",
+        errorLevel: "error",
+        state: "failed",
+        retryCount: 1,
+        createdAt: new Date(0),
+      });
+
+      inRangeFailedExport = await Export.create({
+        recordId: record.id,
+        destinationId: destination.id,
+        oldRecordProperties: {},
+        newRecordProperties: {},
+        newGroups: [],
+        oldGroups: [],
+        sendAt: null,
+        startedAt: new Date("2022-01-01T11:05:00Z"),
+        errorMessage: "Oh No!",
+        errorLevel: "error",
+        state: "failed",
+        retryCount: 1,
+        createdAt: new Date("2022-01-01T11:00:00Z"),
+      });
+
+      foreignFailedExport = await Export.create({
+        recordId: record.id,
+        destinationId: destination2.id,
+        oldRecordProperties: {},
+        newRecordProperties: {},
+        newGroups: [],
+        oldGroups: [],
+        sendAt: null,
+        startedAt: new Date("2022-01-01T11:05:00Z"),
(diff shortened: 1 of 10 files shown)


pedroslopez:
Note: this page uses `react-hook-form`, and its usage syntax changes after #2717

core/__tests__/models/export.ts
```diff
@@ -555,6 +555,269 @@ describe("models/export", () => {
     });
   });
 
+  describe("retryFailed", () => {
```
evantahler, line 558: This is a very thourough test!

core/src/actions/exports.ts
```diff
@@ -108,3 +108,32 @@ export class ExportView extends AuthenticatedAction {
...
 }
+
+export class ExportsRetryFailed extends AuthenticatedAction {
+  name = "exports:retryFailed";
+  description = "retry failed exports within a timeframe";
+  permission: ActionPermission = { topic: "destination", mode: "write" };
+  inputs = {
+    destinationId: { required: false },
+    startTimestamp: { required: true, formatter: APIData.ensureNumber },
+    endTimestamp: { required: true, formatter: APIData.ensureNumber },
```
evantahler, line 119: I was curious if we would have ended up with an `APIData.ensureDate` - but passing timestamps is better! That's what we do on the way out with model.ApiData()

core/src/actions/exports.ts
```diff
@@ -108,3 +108,32 @@ export class ExportView extends AuthenticatedAction {
...
+    preview: { required: false, formatter: APIData.ensureBoolean },
```
evantahler: [nit] maybe a default value of `false` would be more clear

core/src/modules/ops/export.ts
```diff
@@ -137,4 +137,42 @@ export namespace ExportOps {
...
+        {
+          state: "pending",
+          sendAt: new Date(),
+          startedAt: null,
+          errorMessage: null,
+          errorLevel: null,
+          retryCount: 0,
+        },
+        { where }
+      );
```
evantahler, line 171: 👍 one bulk update

ui/ui-enterprise/pages/model/[modelId]/destination/[destinationId]/retry.tsx
```diff
@@ -0,0 +1 @@
+export { default } from "@grouparoo/ui-components/pages/model/[modelId]/destination/[destinationId]/retry";
```
Comment (evantahler) [inline ui/ui-enterprise/pages/model/[modelId]/destination/[destinationId]/retry.tsx:1]: 👍 added to ui-enterprise

ui/ui-community/pages/model/[modelId]/destination/[destinationId]/retry.tsx
```diff
@@ -0,0 +1 @@
+export { default } from "@grouparoo/ui-components/pages/model/[modelId]/destination/[destinationId]/retry";
```
Comment (evantahler) [inline ui/ui-community/pages/model/[modelId]/destination/[destinationId]/retry.tsx:1]: 👍 added to ui-community

evantahler (approved):
This looks great!

Thanks for keeping the update logic in a single Op so it can be reused by the CLI too!

core/src/modules/ops/export.ts
evantahler: We may want to add a LIMIT here - Check out the work done here https://github.com/grouparoo/grouparoo/pull/2840 , with `GROUPAROO_INTERNAL_BATCH_SIZE` being the limit we might want to use

pedroslopez: @evantahler after trying to test this it looks like limits on update queries won't have an effect for pg or sqlite

image

Since this isn't a recurring thing and will only run when someone wants to retry things explicitly, maybe it's ok?
If we really wanted to do the limit here we could find the rows first and do specific IDs to the `update` queries

core/src/actions/exports.ts
pedroslopez: done in https://github.com/grouparoo/grouparoo/pull/2825/commits/6d61192a3b80a08f8b0510cfc1e3dfc2860a00b4

core/src/modules/ops/export.ts
evantahler: Whoa! https://blog.crunchydata.com/blog/simulating-update-or-delete-with-limit-in-postgres-ctes-to-the-rescue

That's an important fact to know!  I guess it doesn't matter then... feel free to skip it for now (as you say, it's an action not part of the tasks)

PR #2826, opened 2022-01-20
Memory and Query Use updates

evantahler:
The staging server is having OOM issues.  Here are some fixes:

1. The `/api/v1/status/private` endpoint will now throw an error if there is no active leader + `cli:start` is the active mode.  This will make it very easy to tell that something is wrong with the cluster
2. Fixes to the `Property.cached` lookup methods - this will remove a /lot/ of unnecessary SQL statements
3. Ensure that the batch sizes processed in the `recordProperty:importRecordProperties` are small and obey the provided limit

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/property/localCache.ts
+++ b/core/__tests__/models/property/localCache.ts
@@ -9,6 +9,7 @@ describe("models/property", () => {
   let originalPropertyCount: number;
 
   beforeEach(() => (CachedProperties.expires = new Date().getTime() + 5000));
+  beforeEach(() => (CachedProperties.TTL = 0));
   afterAll(() => (CachedProperties.expires = 0));
 
   beforeAll(async () => {
--- a/core/__tests__/models/property/localCache.ts
+++ b/core/__tests__/models/property/localCache.ts
@@ -8,6 +8,9 @@ describe("models/property", () => {
   let source: Source;
   let originalPropertyCount: number;
 
+  beforeEach(() => (CachedProperties.expires = new Date().getTime() + 5000));
+  afterAll(() => (CachedProperties.expires = 0));
+
   beforeAll(async () => {
     await helper.factories.properties();
     source = await Source.findOne({ where: { state: "ready" } });
@@ -103,20 +106,47 @@ describe("models/property", () => {
   });
 
   describe("#findOneWithCache", () => {
-    let emailProperty: Property;
+    let firstNameProperty: Property;
 
     beforeAll(async () => {
-      emailProperty = await Property.findOne({ where: { key: "email" } });
+      firstNameProperty = await Property.findOne({
+        where: { key: "firstName" },
+      });
+      await firstNameProperty.update({ key: "FIRST NAME" });
+    });
+
+    test("after a Property is updated, the local cache should be invalid", async () => {
+      expect(CachedProperties.expires).toBeGreaterThan(0);
+      const lastNameProperty = await Property.findOne({
+        where: { key: "lastName" },
+      });
+      await lastNameProperty.update({ key: "LAST NAME" });
+      expect(CachedProperties.expires).toEqual(0);
     });
 
     test("it will find by id by default", async () => {
-      const found = await Property.findOneWithCache(emailProperty.id);
-      expect(found.key).toBe("email");
+      const found = await Property.findOneWithCache(firstNameProperty.id);
+      expect(found.id).toBe("firstName");
+      expect(found.key).toBe("FIRST NAME");
+    });
+
+    test("it will avoid using SQL when a cached property exists", async () => {
+      const cachedEmail = CachedProperties.properties.find(
+        (p) => p.id === firstNameProperty.id
+      );
+      // @ts-ignore
+      cachedEmail.__isCached = true;
+      const found = await Property.findOneWithCache(firstNameProperty.id);
+      expect(found.id).toBe("firstName");
+      expect(found.key).toBe("FIRST NAME");
+      // @ts-ignore
+      expect(found.__isCached).toBe(true);
     });
 
     test("it can find by other keys", async () => {
-      const found = await Property.findOneWithCache("email", "key");
-      expect(found.key).toBe("email");
+      const found = await Property.findOneWithCache("FIRST NAME", null, "key");
+      expect(found.id).toBe("firstName");
+      expect(found.key).toBe("FIRST NAME");
     });
 
     test("a cache miss will invalidate the cache", async () => {
--- a/core/__tests__/models/property/localCache.ts
+++ b/core/__tests__/models/property/localCache.ts
@@ -113,7 +113,8 @@ describe("models/property", () => {
       firstNameProperty = await Property.findOne({
         where: { key: "firstName" },
       });
-      await firstNameProperty.update({ key: "FIRST NAME" }); // should reset the cache
+      await firstNameProperty.update({ key: "FIRST NAME" });
+      CachedProperties.expires = 0;
     });
 
     test("after a Property is updated, the local cache should be invalid", async () => {
--- a/core/__tests__/models/property/localCache.ts
+++ b/core/__tests__/models/property/localCache.ts
@@ -117,6 +117,7 @@ describe("models/property", () => {
     });
 
     test("after a Property is updated, the local cache should be invalid", async () => {
+      CachedProperties.expires = new Date().getTime() + 1000 * 30;
       const lastNameProperty = await Property.findOne({
         where: { key: "lastName" },
       });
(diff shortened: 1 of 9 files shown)


core/src/models/Property.ts
```diff
@@ -331,13 +331,17 @@ export class Property extends LoggedModel<Property> {
...
+    lookupKey: keyof Property = "id"
+  ) {
     const properties = await Property.findAllWithCache(modelId);
-    let property = properties.find((p) => p.key === value);
+    let property = properties.find((p) => p[lookupKey] === value);
 
     if (!property) {
       property = await Property.findOne({
-        where: { [key]: value },
+        where: { [lookupKey]: value },
```
evantahler, line 343: Note the `[] vs .` lookup

This bug was leading to a /ton/ of SQL selects on every usage of `record.getProprties()` -> `recordProperty.getValue()` -> `recordProperty.ensureProperty()` -> `Property.findOneWithCache()`.

Prior to this fix, we were never finding a match because p.key (literal) was never matching an id, which is what "key" should have been.  This wasn't caught in tests, because we disable the cache in tests :/

andyjih: Any idea how long this bug has been out in the wild?

core/src/modules/ops/recordProperty.ts
```diff
@@ -228,10 +233,12 @@ async function preparePropertyImports(
...
-    await CLS.enqueueTask(`recordProperty:importRecordProperties`, {
-      propertyIds: properties.map((p) => p.id),
-      recordIds: uniqueRecordIds,
-    });
+    while (uniqueRecordIds.length > 0) {
+      await CLS.enqueueTask(`recordProperty:importRecordProperties`, {
+        propertyIds: properties.map((p) => p.id),
+        recordIds: uniqueRecordIds.splice(0, limit),
+      });
+    }
```
evantahler, line 241: Note the `.splice()` now present.

On staging, the task that was running out of memory looked like this:

```
debug: [ worker ] working job recordProperties workerId=1 class=recordProperty:importRecordProperties queue=recordProperties args={"propertyIds":["rul_fa41c80c-7a31-449a-9be6-fd5e0ebd58eb","rul_29cb27cb-c696-4527-929f-e302b969a423","rul_b7e3c82a-5ae2-4fc9-a0b4-f9e0eb319e29","rul_d5dbb7af-56f0-47c5-a4c8-128eb724b4c5","gender","rul_2a0e0039-9360-4b95-a1d0-75582841e23f","rul_85a8d395-080f-4974-b3db-36e59d2001b9","ip_address","vip"],"recordIds":"2997 items"}
```

That's a reasonable number of properties for an unreasonable number of records.  This change enqueues a few jobs (rather than one huge one) never to exceed the `limit` number of records at a time.

core/src/models/Property.ts
evantahler: @pedroslopez any suggestions on how we could test this one?

evantahler: @andyjih probably for at least a few months

core/src/modules/ops/recordProperty.ts
evantahler: This is a rather timely addition to https://github.com/grouparoo/grouparoo/pull/2817 (cc @bleonard)

core/__tests__/tasks/recordProperty/enqueue/grouppedAggregations.ts
```diff
@@ -0,0 +1,134 @@
...
+      await mario.markPending();
+      await luigi.markPending();
+      await toad.markPending();
+      await peach.markPending();
+
+      // shuffle the order of things so all the updated properties will be found in the first group
+      for (const person of [mario, luigi, toad, peach]) {
+        await api.sequelize.query(
+          `UPDATE "recordProperties" SET "updatedAt"=NOW() WHERE "recordId"='${person.id}' and "propertyId"='${email.id}'`
+        );
```
pedroslopez: Curious why not use the ORM for this

core/src/models/Property.ts
pedroslopez: I think we should definitely check that we're returning the cached properties in the tests. One idea is setting `CachedProperties.expires` explicitly to something in the future, and setting something like `_isCached` on the cached property within the test that we can check on the returned property from `findOneWithCache`.

Here's a quick example as an update to https://github.com/grouparoo/grouparoo/blob/463d74aaf43e6d9635efea9c9dd333cf396f11a4/core/__tests__/models/property/localCache.ts#L112

```ts
CachedProperties.expires = new Date().getTime() + 5000;
const cachedEmail = CachedProperties.properties.find(
  (p) => p.id === emailProperty.id
);
// @ts-ignore
cachedEmail.__isCached = true;
const found = await Property.findOneWithCache(emailProperty.id);
expect(found.key).toBe("email");
// @ts-ignore
expect(found.__isCached).toBe(true);
```

Typescript yells at you because `__isCached` isn't in the type, but I think it's _ok_ for the test.

Now, one thing I did notice is that in the tests the `id` and `key` are both the same (`email`), so the test would still pass in this case. They should be different to actually test this properly.

core/__tests__/tasks/recordProperty/enqueue/grouppedAggregations.ts
evantahler: I tried a few ways to bump the updatedAt and Seqeulize *really* doesn't let you.  We do have a utility for this kind of thing (also using raw SQL) that I'll move over to using - https://github.com/grouparoo/grouparoo/blob/118f95913d6bab113a91d1d10f214e8629127af7/plugins/%40grouparoo/spec-helper/src/lib/specHelper.ts#L661-L673

evantahler: 74e28410f98e9ad49f3e250a57059d5e4ef0914e makes things a little nicer

core/src/models/Property.ts
evantahler: Nice suggestion! messing with the cached object to ensure we get it back is a good path forward

evantahler: better cache tests via 09d6e77dc9e18b0c5638ee9d22c00a8640643ddf

core/src/actions/status.ts
```diff
@@ -32,6 +33,14 @@ export class PrivateStatus extends AuthenticatedAction {
...
 
     const samples = await Status.get();
+    const { leader } = await task.details();
+    const runMode = getGrouparooRunMode();
+
+    if (!leader && runMode === "cli:start") {
+      throw new Error(
+        `no leader for this Grouparoo cluster "${clusterName}" (${runMode})`
+      );
+    }
```
evantahler, line 43: This change to the private status action will ensure that we are notified if there are no workers/leaders for a given cluster

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -659,16 +659,26 @@ export namespace helper {
...
   ) {
-    const { api } = await import("actionhero");
-    const tableName = instance.constructor["tableName"];
+    const { api, config } = await import("actionhero");
+    const tableName = instances[0].constructor["tableName"];
+    const sqlDate =
+      timestamp === true
+        ? config.sequelize.dialect === "sqlite"
+          ? "strftime('%Y-%m-%d %H-%M-%S','now')"
+          : "NOW()"
```
evantahler, line 672: Looks like "NOW()" doesn't really exist in SQLite

core/__tests__/models/property/localCache.ts
```diff
@@ -103,20 +106,48 @@ describe("models/property", () => {
...
+      await firstNameProperty.update({ key: "FIRST NAME" });
+      expect(CachedProperties.expires).toEqual(0);
+    });
+
+    test("after a Property is updated, the local cache should be invalid", async () => {
+      const lastNameProperty = await Property.findOne({
+        where: { key: "lastName" },
+      });
+      await lastNameProperty.update({ key: "LAST NAME" });
+      expect(CachedProperties.expires).toEqual(0);
```
pedroslopez, line 125: Should we set or check that `CachedProperties.expires` is something else before this? I believe it would be `0` anyway at this point.

core/__tests__/models/property/localCache.ts
```diff
@@ -103,20 +106,48 @@ describe("models/property", () => {
...
     test("it will find by id by default", async () => {
-      const found = await Property.findOneWithCache(emailProperty.id);
-      expect(found.key).toBe("email");
+      const found = await Property.findOneWithCache(firstNameProperty.id);
+      expect(found.id).toBe("firstName");
+      expect(found.key).toBe("FIRST NAME");
+    });
+
+    test("it will avoid using SQL when a cached property exists", async () => {
+      const cachedEmail = CachedProperties.properties.find(
```
pedroslopez: This should be `cachedFirstName` or something, `email` isn't quite right

pedroslopez (commented):
Looks good! Just some minor testing notes

evantahler:
@pedroslopez test updated! one more review please :D

pedroslopez (approved):

PR #2827, opened 2022-01-20
Google Sheets destination plugin

pauloouriques:
## Change description

This PR introduces the Google Sheets destination plugin.

### Development

### Impact

### Code review


plugins/@grouparoo/google-sheets/src/lib/sheet-export/exportRecords.ts
```diff
@@ -0,0 +1,251 @@
...
+}) => {
+  for (const foreignKey of foreignKeys) {
+    const user = getByForeignKey(foreignKey);
+    try {
+      const row = await client.getRowByPrimaryKey(
+        config.foreignKey,
+        foreignKey
+      );
+      if (row) {
+        user.destinationId = row._rowNumber;
```
bleonard: I think this is what I'm most worried about here, though I'm not sure of a better approach. If this row number changes because someone (or grouparoo) inserts/deletes, things are going to get crazy.

plugins/@grouparoo/google-sheets/src/lib/shared/Spreadsheet.ts
```diff
@@ -0,0 +1,252 @@
...
+    const cacheDurationMs = 1000 * 60 * 3; // 3 minutes
+    const contentId = `${this.docId}-${this.sheetId}`;
+    const cacheKey = ["sheetContent", contentId];
+    return objectCache(
+      { objectId: contentId, cacheKey, cacheDurationMs },
+      async () => {
+        return await this.getAllRows();
+      }
+    );
+  }
```
evantahler: I see what you mean about using memory to hold the whole sheet!

I'm also worried about this filling up all of redis and causing problems system-wide (eg: the background tasks stop working).  I don't have a great idea for a better solution other than emulating what we do on the source side: downloading the whole sheet locally as a CSV and then scanning the file until we find the row we want.

Pros:
* doesn't use Redis' memory

Cons:
* probably slower
* increases the drift between local and remove versions 

Maybe using Redis is OK, but I'd want some sort of upper limit on how big the cached object is... perhaps something like `if (rows * columns > 1M) throw new Error('too big') `

evantahler: Wait... this /doesn't/ store everything in redis! that's just the method to reload the table once the key expires!  @pauloouriques were you instead worried about  `this.getAllRows()` in general possibly filling up Node's memory (rather than Redis')?

pauloouriques: It still has the `getAllRows` but now with no cache.

plugins/@grouparoo/google-sheets/src/lib/shared/Spreadsheet.ts
```diff
@@ -0,0 +1,315 @@
+import {
+  objectCache,
+  objectCacheInvalidate,
+  SimpleAppOptions,
+} from "@grouparoo/core";
+
+const { GoogleSpreadsheet } = require("google-spreadsheet");
+
+const GOOGLE_SHEETS_ROWS_LIMIT = 10000;
+const MINUTE_IN_MILLIS = 60 * 1000;
```
bleonard: maybe wait less if the rate limit is per minute. 20 seconds? it's fine to try twice and not get it.

plugins/@grouparoo/google-sheets/src/lib/shared/Spreadsheet.ts
```diff
@@ -0,0 +1,315 @@
...
+      return rows.filter((row) => row[column] === value)[0] || null;
+    }
+    return null;
+  }
+
+  async getRowObjectByPrimaryKey(column, value) {
+    await this.load();
+    const row = await this.getRowByPrimaryKey(column, value);
+    if (row?._rowNumber) {
+      return await this.getRowObjectByRowNumber(row?._rowNumber);
```
bleonard: is it necessary to get the index _and then_ look up the object. This could be quite a bit of extra calls if it i further down.

bleonard: And this function is the only one calling `getRowByPrimaryKey`

plugins/@grouparoo/google-sheets/src/lib/sheet-export/exportRecords.ts
```diff
@@ -0,0 +1,261 @@
...
+  client,
+  users,
+  config,
+}) => {
+  for (const user of users) {
+    try {
+      const { newRecordProperties, oldRecordProperties } = user;
+      const id = newRecordProperties[config.foreignKey];
+      const oldId = oldRecordProperties[config.foreignKey];
+      if (oldId && oldId !== id && config.syncOperations) {
```
bleonard: missing a `syncOperations.Something` here?

plugins/@grouparoo/google-sheets/src/lib/sheet-export/exportRecords.ts
```diff
@@ -0,0 +1,261 @@
...
+      newGroups
+    );
+  }
+  return formattedDataFields;
+}
+
+function getGroupsListToExport(record, groupsColumn, oldGroups, newGroups) {
+  const groups = newGroups;
+  if (record.destinationId && record.result && record.result[groupsColumn]) {
+    const removedGroups = [];
```
bleonard: I think maybe you don't have to go through `oldGroups` here. You just want the new group names

bleonard: In other words, it doesn't seem useful to keep groups that people set by hand in this case.

bleonard (commented):
A few things. We still have to talk about parallelism or something here and/or deleting rows.
Is it right to say that if we only `-` rows then we can actually run two of thse im parallel? something about the rows being added at the end being fine in parallel?
But if we do really delete rows, all kinds of things are going to get messed up.

plugins/@grouparoo/google-sheets/__tests__/export/destination-mapping-options.ts
```diff
@@ -0,0 +1,50 @@
+import "@grouparoo/spec-helper";
+import { helper } from "@grouparoo/spec-helper";
+import { loadAppOptions, updater } from "../utils/nockHelper";
+import { destinationMappingOptions } from "../../src/lib/sheet-export/destinationMappingOptions";
+
+// NOCK=true on cmd line to record new nock file: have to change "assertion" afterwards for google auth
+helper.useNock(__filename, updater);
+
+const url =
```
bleonard, line 9: This should probably be in an ENV Variable too and read like `loadAppOptions`

plugins/@grouparoo/google-sheets/src/lib/sheet-export/exportRecords.ts
```diff
@@ -0,0 +1,238 @@
...
+    .filter((k) => !newPropertyKeys.includes(k))
+    .forEach((k) => (deletePropertiesPayload[k] = null));
+  let payload = Object.assign(newRecordProperties, deletePropertiesPayload);
+  const formattedDataFields = {};
+  for (const key of Object.keys(payload)) {
+    formattedDataFields[key] = formatVar(payload[key]);
+  }
+  const groupsColumn = config?.destinationOptions?.groupsColumn?.toString();
+  if (groupsColumn) {
+    formattedDataFields[groupsColumn] = newGroups.join(",");
```
bleonard: might be nice to `sort()` here so it's predictable.

bleonard (approved):
Looking great. Very interested to see on testing if there are issues when in parallel.

PR #2828, opened 2022-01-20
Fix issues with updating recurring schedule

edmundito:
## Change description

This fixes some issues related to updating a recurring schedule to not updating. It would cause some validation failures because setting a recurring schedule to false will not necessarily update if the recurring frequency was also reset to 0. Now the `recurring` flag is always updated and the `recurring frequency` will be reset to 0 when updating recurring to `false` to avoid confusion.

### Development

### Impact

### Code review

### DIFF
--- a/core/src/actions/schedules.ts
+++ b/core/src/actions/schedules.ts
@@ -210,13 +210,17 @@ export class ScheduleEdit extends AuthenticatedAction {
 
   async runWithinTransaction({ params }: { params: ParamsFrom<ScheduleEdit> }) {
     const schedule = await Schedule.findById(params.id);
+
+    const recurringFrequency =
+      params.recurring && params.recurringFrequency
+        ? params.recurringFrequency
+        : 0;
+
     // these timing options are validated separately, and should be set first
-    if (params.recurringFrequency || params.recurring) {
-      await schedule.update({
-        recurringFrequency: params.recurringFrequency,
-        recurring: params.recurring,
-      });
-    }
+    await schedule.update({
+      recurring: params.recurring,
+      recurringFrequency,
+    });
 
     if (params.options) await schedule.setOptions(params.options);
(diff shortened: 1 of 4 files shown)


core/src/actions/schedules.ts
```diff
@@ -210,13 +210,17 @@ export class ScheduleEdit extends AuthenticatedAction {
...
   async runWithinTransaction({ params }: { params: ParamsFrom<ScheduleEdit> }) {
     const schedule = await Schedule.findById(params.id);
+
+    const recurringFrequency =
+      params.recurring && params.recurringFrequency
+        ? params.recurringFrequency
+        : 0;
+
     // these timing options are validated separately, and should be set first
-    if (params.recurringFrequency || params.recurring) {
```
edmundito, line 214: Here's the problem: Only updates recurring if it's truthy.

evantahler (approved):
Nice.

ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
```diff
@@ -46,37 +46,44 @@ export default function Page(props) {
...
+    const scheduleToSave: Models.ScheduleType = {
+      ...schedule,
+      state: "ready",
+      filters: localFilters,
+      recurringFrequency: schedule.recurring
+        ? recurringFrequencyMinutes * (60 * 1000)
+        : 0,
+    };
+
+    console.log(scheduleToSave);
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx]: Will remove

core/src/actions/schedules.ts
edmundito: @evantahler Another question that came to mind: If we call this action and we don't set a recurring/recurring frequency, should it update the current recurring frequency or should I really check if these values are in the params before making updates?

PR #2830, opened 2022-01-20
fix: generate unique ids when creating properties

pedroslopez:
## Change description

When creating properties via the multiple properties page and `bootstrapUniqueProperty` after #2086 , we were explicitly setting the ID of the new property to be the same as the key, instead of letting grouparoo generate an ID for it as it used to and still does for the regular "Create a property" page. 

This was causing conflicts with the auto key generation logic within `bootstrapUniqueProperty`, because while it would keep iterating to find a valid key if a property with the same _key_ already existed, it would fail if a property with the same _id_ already existed. 

To make things more consistent and also prevent this issue, the changes in this PR lets Grouparoo generate UUIDs like it usually does for properties created through this way. Creating properties in config mode still generates the pretty IDs as expected, since it still switches the IDs when generating the config objects.

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/tasks/system/destroy.ts
+++ b/core/__tests__/tasks/system/destroy.ts
@@ -226,7 +226,7 @@ describe("tasks/destroy", () => {
                 type: "integer",
                 mappedColumn: "id",
               });
-              await userIdProperty.update({ isPrimaryKey: true });
+              await source.setMapping({ id: "userId" });
             });
 
             test("it will enqueue destroy task by model for records when there is no primary key property", async () => {
--- a/core/src/modules/ops/source.ts
+++ b/core/src/modules/ops/source.ts
@@ -554,7 +554,7 @@ export namespace SourceOps {
       retry = false;
 
       const property = Property.build({
-        id: id ?? ConfigWriter.generateId(key),
+        id,
         key,
         type,
         state: "ready",
--- a/ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx
+++ b/ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx
@@ -130,7 +130,6 @@ export default function Page(props) {
         `/property`,
         {
           sourceId: source.id,
-          id: key, // reuse the key as ID
           key,
           type,
           unique,
--- a/core/__tests__/models/source/source.ts
+++ b/core/__tests__/models/source/source.ts
@@ -880,6 +880,37 @@ describe("models/source", () => {
       await property.destroy();
     });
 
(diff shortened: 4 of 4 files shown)


ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx
```diff
@@ -130,7 +130,6 @@ export default function Page(props) {
         `/property`,
         {
           sourceId: source.id,
-          id: key, // reuse the key as ID
```
Comment (pedroslopez) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx:133]: @evantahler do you know what the reasoning for this was initially? (reusing they key as id when creating properties through the multiple properties page, instead of letting the ID be generated randomly)

core/__tests__/tasks/system/destroy.ts
```diff
@@ -226,7 +226,7 @@ describe("tasks/destroy", () => {
                 type: "integer",
                 mappedColumn: "id",
               });
-              await userIdProperty.update({ isPrimaryKey: true });
+              await source.setMapping({ id: "userId" });
```
pedroslopez, line 229: @rwfeather This was why those tests started failing - the `.update()` call triggered some other hooks that tried to get the mapping, but the old mapped property no longer existed so errors were being thrown and `isPrimaryKey` was never being set. Properly setting the mapping on the source to the newly created property and letting `isPrimaryKey` be determined as normal lines everything up so the mapping on the source itself isn't referring to something that no longer exists.

ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx
Comment (evantahler) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/multipleProperties.tsx:133]: Presumably we wanted to try to make pretty IDs (eg: user_id vs prop_abc123).  BUT, the config writer should take care of this for us... so I'm not that worried about removing this.

core/src/modules/ops/source.ts
```diff
@@ -554,7 +554,7 @@ export namespace SourceOps {
       retry = false;
 
       const property = Property.build({
-        id: id ?? ConfigWriter.generateId(key),
+        id,
```
evantahler, line 557: So there's never a time id would be null?

evantahler (approved):

core/src/modules/ops/source.ts
pedroslopez: If an `id` is not passed into `bootstrapUniqueProperty`, then Grouparoo will generate a UUID for the property as usual. Added explicit tests for this here cbed94dd5b8703c65fe661205cbcbdc0130f3581

PR #2838, opened 2022-01-20
Ensure that property keys are unique regardless of case

edmundito:
## Change description

This fixes an issue where a user can create two properties with the same key but with different cases. For example `LTV` and `ltv`. Because ids in config mode are generated out of the key, it would cause the creation of properties with the same ID. This makes it so that the check is case-insensitive.

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/property/property.ts
+++ b/core/__tests__/models/property/property.ts
@@ -90,6 +90,31 @@ describe("models/property", () => {
       await property.destroy();
     });
 
+    test("ready properties cannot share the same key regardless of key case", async () => {
+      const ruleOne = await Property.create({
+        sourceId: source.id,
+        key: "CASE",
+        type: "string",
+      });
+      const ruleTwo = await Property.create({
+        sourceId: source.id,
+        key: "case",
+        type: "string",
+      });
+
+      await ruleOne.setOptions({ column: "abc123" });
+      await ruleOne.update({ state: "ready" });
+
+      await ruleTwo.setOptions({ column: "abc123" });
+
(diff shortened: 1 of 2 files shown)


core/__tests__/models/property/property.ts
```diff
@@ -90,6 +90,31 @@ describe("models/property", () => {
       await property.destroy();
     });
 
+    test("ready properties cannot share the same key regardless of key case", async () => {
```
evantahler, line 93: 👍 good test

core/src/models/Property.ts
```diff
@@ -372,10 +372,14 @@ export class Property extends LoggedModel<Property> {
     const count = await Property.count({
       where: {
         id: { [Op.ne]: instance.id },
-        key: instance.key,
+        key: Sequelize.where(
+          Sequelize.fn("LOWER", Sequelize.col("key")),
+          instance.key.toLowerCase()
+        ),
```
evantahler, line 378: Yep, that works!

Are there any other `ensureUnique*` methods we should update in a similar way?
A quick glance shows:
* `App.ensureUniqueName`
* `Destination.ensureUniqueName`
* `Group.ensureUniqueName`
* `Schedule.ensureUniqueName`
* `Source.ensureUniqueName`

evantahler (approved):

core/src/models/Property.ts
edmundito: Good question. We should expand these into a story in our backlog.

PR #2839, opened 2022-01-20
fix: only do one redirect after schedule creation

pedroslopez:
## Change description

There were multiple `router.push` firing on the case where you were setting up your first source/schedule, which would cause an infinite spinner after taking you to one of the two. This makes sure we're only doing the right one.

The redirect to add multiple properties was also being performed even if you were just editing the schedule. This changes the behavior so it only happens upon first creation.

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
+++ b/ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
@@ -73,10 +73,17 @@ export default function Page(props) {
 
     if (response?.schedule) {
       if (response.schedule.state === "ready" && schedule.state === "draft") {
-        router.push(
-          "/model/[modelId]/source/[sourceId]/overview",
-          `/model/${source.modelId}/source/${source.id}/overview`
-        );
+        if (totalSources === 1 && totalProperties === 1) {
+          router.push(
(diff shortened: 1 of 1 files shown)


krishnaglick (approved):
I feel like it'd be helpful to have a test for this!

ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx
```diff
@@ -73,10 +73,17 @@ export default function Page(props) {
...
+          router.push(
+            "/model/[modelId]/source/[sourceId]/multipleProperties",
+            `/model/${source.modelId}/source/${source.id}/multipleProperties`
+          );
+        } else {
+          router.push(
+            "/model/[modelId]/source/[sourceId]/overview",
+            `/model/${source.modelId}/source/${source.id}/overview`
+          );
+        }
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/schedule.tsx]: A better implementation is to not duplicate the calls but instead duplicate the final page value because we know in either case we'll redirect.

pedroslopez:
Since there's a future story to do integration tests and set up the mocking/seed data bits necessary for testing these flows, added a new story to add tests after that's done. pivotal-180997677

krishnaglick (approved):

PR #2840, opened 2022-01-20
Add `GROUPAROO_RECORDS_BATCH_SIZE` separate from `GROUPAROO_INTERNAL_BATCH_SIZE`

evantahler:
Adds even more tuning ability when importing lots of data.

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/ops/recordProperty.ts
+++ b/core/src/modules/ops/recordProperty.ts
@@ -10,7 +10,7 @@ import { Op } from "sequelize";
 import { Source } from "../../models/Source";
 import { AggregationMethod, PluginConnection } from "../../classes/plugin";
 import { Filter } from "../../models/Filter";
-import { log } from "actionhero";
+import { config, log } from "actionhero";
 
 export namespace RecordPropertyOps {
   const defaultRecordPropertyProcessingDelay = 1000 * 60 * 5;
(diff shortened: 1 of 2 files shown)


core/src/modules/ops/recordProperty.ts
```diff
@@ -199,6 +199,17 @@ export namespace RecordPropertyOps {
 }
 
 // utilities (private)
+async function updateIds(ids: string[], values: { [key: string]: any }) {
+  const max = config.batchSize.internalWrite;
+  const queue = ids; // this prevents it from changing the passed in one, right?
```
evantahler, line 204: No... you would need to do `const queue = [...ids]` to duplicate the array

core/src/modules/ops/recordProperty.ts
```diff
@@ -199,6 +199,17 @@ export namespace RecordPropertyOps {
 }
 
 // utilities (private)
+async function updateIds(ids: string[], values: { [key: string]: any }) {
+  const max = config.batchSize.internalWrite;
```
evantahler, line 203: I think this is the only place that uses the new `GROUPAROO_INTERNAL_BATCH_SIZE` ENV.  I don't **think** this would be a limiting problem necessarily because it's just one small-ish statement (UPDATE recordProperties where ID in ...).

That said, this will allow us to play with this setting and try out new values, so it's a 👍 from me!

evantahler: Also, this is a nice utility method

core/src/modules/ops/recordProperty.ts
evantahler: ... probably why the tests are failing

evantahler:
Also, please update https://github.com/grouparoo/grouparoo/blob/main/documents/environment-variables.md

core/src/modules/ops/recordProperty.ts
bleonard: The problem the customer was seeing is that the size of this is variable on the number of properties that can be batched together, so it got to be quite large (100k) on a wide table and froze things up.

bleonard: I assume we'll find other places to use it, roughly as a way to think about how much simultaenous updating the database can do.

bleonard:
Replaced by https://github.com/grouparoo/grouparoo/pull/2843

PR #2841, opened 2022-01-20
Fix typescript errors in dbt plugin

asdfryan:
## Change description

Without this, I ran into errors during installation:
Screen Shot 2022-01-20 at 3 44 13 PM

After this, installation succeeded:
Screen Shot 2022-01-20 at 3 45 03 PM

### Development

- [ ] Application changes have been tested appropriately

### Impact

- [ ] Code follows company security practices and guidelines
- [ ] Security impact of change has been considered
- [ ] Performance impact of change has been considered
- [ ] Possible migration needs considered (model migrations, config migrations, etc.)

Please explain any security, performance, migration, or other impacts if relevant:

### Code review

- [ ] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.
- [ ] Relevant tags have been added to the PR (bug, enhancement, internal, etc.)

PR #2844, opened 2022-01-21
Enable `noImplicitAny` for `@grouparoo/dbt`

evantahler:
... and fixup the types.  Follows https://github.com/grouparoo/grouparoo/pull/2841

### Development

### Impact

### Code review

### DIFF
--- a/plugins/@grouparoo/dbt/src/settings/parse.ts
+++ b/plugins/@grouparoo/dbt/src/settings/parse.ts
@@ -122,8 +122,8 @@ const parseProject: ParseProjectMethod = async (args) => {
     );
   }
 
-  const contents = fs.readFileSync(projectPath);
-  const document = yaml.load(contents);
+  const contents = String(fs.readFileSync(projectPath));
+  const document: Record<string, any> = yaml.load(contents);
   // console.log({ project: document });
 
   profile = document.profile;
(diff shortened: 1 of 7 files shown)


plugins/@grouparoo/dbt/src/settings/parse.ts
```diff
@@ -122,8 +122,8 @@ const parseProject: ParseProjectMethod = async (args) => {
     );
   }
 
-  const contents = fs.readFileSync(projectPath);
-  const document = yaml.load(contents);
+  const contents = String(fs.readFileSync(projectPath));
```
krishnaglick, line 125: I feel like we've discussed this, but I believe this should be:
```suggestion
  const contents = fs.readFileSync(projectPath, {encoding: "utf8"});
```

evantahler: I don't think this is a great idea - you can't assume a user's hard drive is in a certain encoding

krishnaglick: I thought this set the expectation for what _we_ want to work with. Docs don't explicitly say no, but imply that's what we're reading. Okay!

krishnaglick (approved):

PR #2846, opened 2022-01-21
execApi -> useApi

krishnaglick:
## Change description

This should be functionally transparent, but all references to execApi have been replaced with either a new Client for server-side calls, or the client object stored in a react context.

There's some minor cleanup work I did while going through everything, but I tried to keep those to a minimum.

### Development

### Impact

- [X] Code follows company security practices and guidelines
- [X] Security impact of change has been considered
- [X] Performance impact of change has been considered
- [X] Possible migration needs considered (model migrations, config migrations, etc.)

### Code review

- [X] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.


ui/ui-components/client/client.ts
```diff
@@ -107,33 +112,39 @@ export class Client {
...
       return window.localStorage.getItem("session:csrfToken");
     }
-  }
+  };
+
+  private static readonly optionDefaults = {
+    useCache: true,
+    errorHandler,
+    uploadHandler,
+  };
```
edmundito: Should probably go at the top of the class?

ui/ui-components/client/client.ts
```diff
@@ -107,33 +112,39 @@ export class Client {
...
+  public request = async <Response = any>(
     verb: Method = "get",
     path: string,
     data: AxiosRequestConfig["data"] = {},
-    useCache = true,
-    uploadHandler?: UploadHandler,
-    req?: IncomingMessage,
-    res?: ServerResponse
-  ): Promise<Response> {
+    options: Partial<typeof Client.optionDefaults> = {}
```
edmundito: Maybe the Options should be explicit instead of derived from `optionDefaults`. There is no optionDefaults can be broken at any time without having the right Options contract

ui/ui-components/client/client.ts
```diff
@@ -147,7 +158,7 @@ export class Client {
         const uploadPercentage = Math.round(
           (progressEvent.loaded / progressEvent.total) * 100
         );
-        return uploadHandler ? uploadHandler.set({ uploadPercentage }) : null;
+        return options.uploadHandler.set({ uploadPercentage });
```
edmundito: I put a note on the story that we can get rid of the uploadHandler entirely because we no longer upload files. It would also be fine to save it for another change.

ui/ui-components/client/client.ts
```diff
@@ -210,16 +221,32 @@ export class Client {
...
     }
-  }
+  };
+}
+
+export const generateClient = (
+  ctx: AppContext | NextPageContext | GetServerSidePropsContext
+) => new Client(getRequestContext(ctx));
+
+function isAppContext(ctx: unknown): ctx is AppContext {
```
edmundito: This function has nothing to do with the client, should probably be a util

ui/ui-components/contexts/api.tsx
```diff
@@ -0,0 +1,5 @@
+import { createContext, useContext } from "react";
+import type { Client } from "../client/client";
+
+export const ApiContext = createContext<{ client: Client }>(undefined);
+export const useApi = () => useContext(ApiContext);
```
edmundito, line 7: After looking at the code. I'm torn if this particular function should be in `/hooks/useApi` along the rest of the other context functions. The implementation of it as a context seems irrelevant when it's being used.

ui/ui-components/pages/_app.tsx
```diff
@@ -47,8 +44,16 @@ export default function GrouparooNextApp(
...
   }, [props]);
 
+  const client = typeof window === "undefined" ? props.client : new Client();
   return renderNestedContextProviders(
-    [[WebAppContext, pageContext]],
+    [
+      [WebAppContext, pageContext],
+      [
+        ApiContext,
+        // The serialized object passed from the server is not an isntance of the Client class
```
edmundito: This comment could probably be moved above the client declaration

ui/ui-components/pages/_app.tsx
```diff
@@ -108,10 +114,12 @@ GrouparooNextApp.getInitialProps = async (appContext: AppContext) => {
       navigation: navigationResponse.navigation,
       clusterName: navigationResponse.clusterName,
       hydrationError,
+      client,
     };
   } catch (_error) {
     const { formattedErrorMessage, formattedErrorObject } = renderError(_error);
     return {
+      client,
```
edmundito, line 119: To confirm: This client will only be used the first time the page renders on the server side?

ui/ui-components/client/client.ts
```diff
@@ -210,16 +221,32 @@ export class Client {
...
       } else {
+        options.errorHandler.set({ message: error });
         throw error;
       }
     }
-  }
+  };
+}
+
+export const generateClient = (
```
edmundito, line 241: Could also be `getClient`

ui/ui-components/client/client.ts
krishnaglick: I'm not sure what you're getting at here. optionDefaults are defaulted, and you can't pass anything not in it. They're override only.

ui/ui-components/contexts/api.tsx
krishnaglick: This could potentially just go into `WebAppContext` but I'd have to find-replace both the useApi and the import and that would be a _lot_ of gruntwork for not much gain. This is a context, we should keep it here.

ui/ui-components/pages/_app.tsx
krishnaglick: I'll be pushing up a slight logic change. As long as this client is a client and not some serialized object, it will be used.

ui/ui-components/client/client.ts
edmundito: My point is that it's strange that this is deriving which options are valid from a constant that assigns default values Using explicit typing is a lot stronger way to establish what's valid.

ui/ui-components/pages/_app.tsx
edmundito: It will always get a serialized object when the browser side runs. I'm wondering if we ever need to share this instance with the page server side.

ui/ui-components/client/client.ts
krishnaglick: Fair enough, added an interface.

ui/ui-components/pages/_app.tsx
krishnaglick: This is run on the client-side as well.
We should perhaps migrate to [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props) instead.

evantahler:
@krishnaglick can you call out in a self-review what some of the new improvements are?  I see how's much easier to use in components now (eg not top-level pages), but I'm looking for new/better error handling and missing it :/

edmundito (approved):

krishnaglick:
> @krishnaglick can you call out in a self-review what some of the new improvements are? I see how's much easier to use in components now (eg not top-level pages), but I'm looking for new/better error handling and missing it :/

The biggest change to the error handling was partially from Edmundo's work to centralize the errorHandler. In addition, the client.request call allows you to override the errorHandler in the options if you want or need to.
The goal of this PR was to remove execApi since it _looked_ like a React hook but actually wasn't. This turns it into a real hook and should make its functionality far more expected for engineers, new or otherwise.

evantahler (approved):
let's give this a try!

PR #2850, opened 2022-01-22
Authentication via `Authorization` header

evantahler:
This PR allows the passing of a Grouparoo API Key via header.   Depending on your infrastructure and logging, it is often more secure to pass the API Key via header than via query param. 

Docs @ https://github.com/grouparoo/www.grouparoo.com/pull/881

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/middleware/authentication.ts
+++ b/core/src/modules/middleware/authentication.ts
@@ -226,9 +226,16 @@ function parseHeaders(data: ActionProcessor<any>) {
   if (data.connection.type !== "web") return;
 
   if (data.connection.rawConnection.req.headers["authorization"]) {
-    data.params.apiKey =
-      data.connection.rawConnection.req.headers["authorization"];
+    const [scheme, token]: string[] =
+      data.connection.rawConnection.req.headers["authorization"].split(" "); // should resemble `Bearer: abc123`
+    if (scheme.toLowerCase() !== "bearer") {
+      throw new Error(
+        `APIKeys should be sent with the \`Authorization: Bearer <token>\` scheme`
+      );
+    }
+    data.params.apiKey = token;
   }
+
   if (data.connection.rawConnection.req.headers["x-grouparoo-server-token"]) {
     data.params.serverToken =
       data.connection.rawConnection.req.headers["x-grouparoo-server-token"];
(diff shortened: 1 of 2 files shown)


pedroslopez (commented):
Question: why not use the standard `Authorization` header for this?

core/src/modules/middleware/authentication.ts
```diff
@@ -218,3 +220,17 @@ async function authenticateTeamMemberInRoom(
...
   });
 }
+
+// check if the apiKey was sent via the header
+function parseHeaders(data: ActionProcessor<any>) {
+  if (data.connection.type !== "web") return;
+
+  if (data.connection.rawConnection.req.headers["x-grouparoo-api-key"]) {
+    data.params.apiKey =
+      data.connection.rawConnection.req.headers["x-grouparoo-api-key"];
```
edmundito: My only concern about the implementation here is that you're the data object's parsed `?apiKey=` to assign the value the header, therefore you'd have to understand the implementation of this function to know where the value is really coming from.

The ideal case would be to have the apiKey come from either params or the header itself, but that would take a lot more code changes. Same for the server token below.

evantahler:
> Question: why not use the standard `Authorization` header for this?

Love it.

core/src/modules/middleware/authentication.ts
evantahler: Well, the goal was to allow for both - the user can choose to send the key and server token either via header or query param - with the header winning if they are both set

edmundito:
Are we using the Authorization spec correctly here?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

evantahler:
Talking it over with @pedroslopez, We are going with type=Bearer

https://swagger.io/docs/specification/authentication/bearer-authentication/

> Bearer authentication (also called token authentication) is an HTTP authentication scheme that involves security tokens called bearer tokens. The name “Bearer authentication” can be understood as “give access to the bearer of this token.” The bearer token is a cryptic string, usually generated by the server in response to a login request. The client must send this token in the Authorization header when making requests to protected resources:

evantahler:
Now with proper use of `Authentication: Bearer <token>` headers!

pedroslopez (approved):
Nice!

PR #2852, opened 2022-01-24
Fix recurring and refreshEnabled schedule update

edmundito:
## Change description

This ensures that `recurring` and `refreshEnabled` can be updated in the schedule regardless of true or false value. Like before, it allows the user to skip those values (and they will not get updated instead.)

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/actions/schedules.ts
+++ b/core/__tests__/actions/schedules.ts
@@ -245,6 +245,53 @@ describe("actions/schedules", () => {
         expect(schedule.recurringFrequency).toBe(0);
       });
 
+      test("an administrator can edit refresh enabled", async () => {
+        connection.params = {
+          csrfToken,
+          id,
+          refreshEnabled: true,
+        };
+
+        const { schedule: firstSchedule, error: firstError } =
+          await specHelper.runAction<ScheduleEdit>("schedule:edit", connection);
+        expect(firstError).toBeUndefined();
+        expect(firstSchedule.refreshEnabled).toBe(true);
+
+        connection.params.refreshEnabled = false;
+
+        const { error, schedule } = await specHelper.runAction<ScheduleEdit>(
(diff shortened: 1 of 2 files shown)


core/src/actions/schedules.ts
```diff
@@ -211,21 +211,25 @@ export class ScheduleEdit extends AuthenticatedAction {
...
+      const recurringFrequency =
+        params.recurring && params.recurringFrequency
+          ? params.recurringFrequency
+          : 0;
+
+      // these timing options are validated separately, and should be set first
+      await schedule.update({
+        recurring: params.recurring,
+        recurringFrequency,
+      });
```
pedroslopez, line 224: From what I'm reading, we would only be updating `recurringFrequency` if `recurring` is passed at the same time. Could there be an issue here if you don't touch the `recurring` field on the form but do want to change the frequency?

core/__tests__/actions/schedules.ts
```diff
@@ -245,6 +245,53 @@ describe("actions/schedules", () => {
         expect(schedule.recurringFrequency).toBe(0);
       });
 
+      test("an administrator can edit refresh enabled", async () => {
```
pedroslopez, line 248: Do we need similar tests for `recurring`?

core/src/actions/schedules.ts
edmundito: Yep, this is how it was previously set before #2828, and wanted to see if this is what was intended. However, one change I did in that PR is that you can only set `recurringFrequency` > 0 if `recurring` is `true`.

core/__tests__/actions/schedules.ts
edmundito: These were setup in #2828

pedroslopez (approved):

PR #2858, opened 2022-01-24
Improve how queries get built in SQLite plugin

edmundito:
## Change description

This change fixes an issue with the SQLite plugin where clauses don't often format strings correctly. It improves the way that SQLite queries are prepared in the plugin so that the values are formatted correctly by taking advantage of the formatter available in the SQLite module:

https://github.com/mapbox/node-sqlite3/wiki/API#databaserunsql-param--callback

It introduces a `SQLLiteQueryBuilder` that assists with putting parts of a query together. Its really simple usage allows pushing different parts of a query and parameters to eventually build a full list.

TODO:
- [x] Migrate table-import

### Development

### Impact

### Code review

### DIFF
--- a/plugins/@grouparoo/sqlite/__tests__/data/records.csv
+++ b/plugins/@grouparoo/sqlite/__tests__/data/records.csv
@@ -3,8 +3,8 @@ id,first_name,last_name,email,gender,ip_address,ios_app,android_app,vip,ltv,date
 2,Cacilie,Eate,ceate1@example.com,Female,253.146.41.201,false,true,false,94.36,2020/02/02,2020/02/02 12:13:14
 3,Gretel,Groucock,ggroucock2@example.com,Female,211.89.110.134,true,false,true,668.30,2020/02/03,2020/02/03 12:13:14
 4,Amata,Cotesford,acotesford3@example.com,Female,193.9.240.246,true,true,true,489.00,2020/02/04,2020/02/04 12:13:14
-5,Lira,Johnston,ljohnston4@example.com,Female,53.161.58.99,false,true,false,823.16,2020/02/05,2020/02/05 12:13:14
-6,Deni,Scalia,dscalia5@example.com,Female,138.131.148.74,true,false,true,273.98,2020/02/06,2020/02/06 12:13:14
+5,Lira,"""Airquotes"" Johnston",ljohnston4@example.com,Female,53.161.58.99,false,true,false,823.16,2020/02/05,2020/02/05 12:13:14
+6,Deni,O‘Hara,dscalia5@example.com,Female,138.131.148.74,true,false,true,273.98,2020/02/06,2020/02/06 12:13:14
 7,Helga,Kleis,hkleis6@example.com,Female,85.248.8.158,true,true,true,407.05,2020/02/07,2020/02/07 12:13:14
 8,Gertie,Cubitt,gcubitt7@example.com,Female,123.255.130.24,false,false,true,841.45,2020/02/08,2020/02/08 12:13:14
 9,Lurleen,Browell,lbrowell8@example.com,Female,231.159.55.43,true,false,false,623.34,2020/02/09,2020/02/09 12:13:14
--- a/plugins/@grouparoo/sqlite/__tests__/data/records.csv
+++ b/plugins/@grouparoo/sqlite/__tests__/data/records.csv
@@ -1,10 +1,10 @@
 id,first_name,last_name,email,gender,ip_address,ios_app,android_app,vip,ltv,date,stamp
-1,Erie,Jervois,ejervois0@example.com,Male,15.247.38.72,true,false,true,259.12,2020/02/01,2020/02/01 12:13:14
-2,Cacilie,Eate,ceate1@example.com,Female,253.146.41.201,false,true,false,94.36,2020/02/02,2020/02/02 12:13:14
+1,Erie,O‘Hara,eohara0@example.com,Male,15.247.38.72,true,false,true,259.12,2020/02/01,2020/02/01 12:13:14
+2,"Cacilie ""Cici""",Eate,ceate1@example.com,Female,253.146.41.201,false,true,false,94.36,2020/02/02,2020/02/02 12:13:14
 3,Gretel,Groucock,ggroucock2@example.com,Female,211.89.110.134,true,false,true,668.30,2020/02/03,2020/02/03 12:13:14
 4,Amata,Cotesford,acotesford3@example.com,Female,193.9.240.246,true,true,true,489.00,2020/02/04,2020/02/04 12:13:14
-5,Lira,"""Airquotes"" Johnston",ljohnston4@example.com,Female,53.161.58.99,false,true,false,823.16,2020/02/05,2020/02/05 12:13:14
-6,Deni,O‘Hara,dscalia5@example.com,Female,138.131.148.74,true,false,true,273.98,2020/02/06,2020/02/06 12:13:14
+5,Lira,Johnston,ljohnston4@example.com,Female,53.161.58.99,false,true,false,823.16,2020/02/05,2020/02/05 12:13:14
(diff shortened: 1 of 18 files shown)


teallarson:
I notice this is currently only on the export side -- is there a reason to not use it for imports too?

edmundito:
> I notice this is currently only on the export side -- is there a reason to not use it for imports too?

Working on it now, hence why this is a Draft PR...

plugins/@grouparoo/sqlite/src/lib/sqlite.ts
```diff
@@ -28,15 +29,22 @@ export class SQLite extends Database {
     });
   }
 
-  async asyncQuery(query: string): Promise<any[]> {
+  async asyncQuery(
+    query: string,
+    params?: SQLiteQueryParamValue[] | Record<string, SQLiteQueryParamValue>
+  ): Promise<Record<string, any>[]> {
```
evantahler, line 35: Maybe asyncQuery should take an instance of QueryBuilder, rather than the string query?

plugins/@grouparoo/sqlite/src/lib/export/exportRecord.ts
```diff
@@ -98,9 +107,12 @@ export const exportRecord: ExportRecordPluginMethod = async ({
     // --- Groups --- //
 
     // delete existing groups
-    const deleteGroupsQuery = `DELETE FROM "${groupsTable}" WHERE ${groupForeignKey} = ${newRecordProperties[primaryKey]}`;
-    validateQuery(deleteGroupsQuery);
-    await connection.asyncQuery(deleteGroupsQuery);
+    await connection.asyncQuery(
+      ...SQLiteQueryBuilder.build(
+        `DELETE FROM "${groupsTable}" WHERE ${groupForeignKey} = ?`,
```
evantahler, line 114: might as well make all interpolations a `?`, so the args would be `[groupsTable, groupForeignKey , newRecordProperties[primaryKey]` and the query `DELETE FROM ? WHERE ? = ? `

evantahler (commented):
I think the interaction between the new SQLiteQueryBuilder and asyncQuery can be tightened up... there won't be one without the other.

I would propose that all interpolated strings be moved to using `?`, even for table names etc.

plugins/@grouparoo/sqlite/src/lib/export/exportRecord.ts
edmundito: I tried, but it's not how the way this version of the module's parser works. I got an error when I tried. It seems that it only parses values but not the table names for some reason. I'll dig a little deeper to see if this was a mistake.

plugins/@grouparoo/sqlite/src/lib/sqlite.ts
edmundito: I looked into this and there's a couple of issues:

1. It's also being used in the test utils. I could switch everything to use the query builder but then I'm coupling the builder with the connection
2. It's unusual to pass a builder as a parameter. The convention is generally to build the output and return the result.

One thought I had was to change the signature so that it's `asyncQuery({ query, params})` but then all simple calls would need to be mad as `asyncQuery({ query: 'select...' })`

plugins/@grouparoo/sqlite/src/lib/export/exportRecord.ts
edmundito: I did another try and it failed. It looks like the implementation of the library goes all the way down the sqlite C implementation and looks like not all values can be substituted:

https://stackoverflow.com/a/597198/20494

evantahler (commented):
Can you please add a test that would have failed in the past?  Perhaps a recordProperty with a "`" in someone's last name.  Otherwise, given the limitations of how many "?" we can add, this is looking good!

plugins/@grouparoo/sqlite/src/lib/util.ts
```diff
@@ -0,0 +1,20 @@
...
+};
+
+export const toValuesArray = (data: DataParam): SQLiteQueryParamValue[] =>
+  Array.isArray(data) ? data : Object.values(data);
+
+export const toValuePlaceholders = (
+  values: SQLiteQueryParamValue[]
+): string => {
+  const placeholders = values.map(() => "?");
+  return placeholders.length > 1 ? `(${placeholders})` : placeholders.join(",");
```
pedroslopez: nit: the .join at the end isn't doing much, might be clearer to write 

```suggestion
  return placeholders.length > 0 ? `(${placeholders})` : "";
```

pedroslopez (commented):
Looks good overall, only comment I left was a small nitpick.
 
I'll defer to evan for the final approval once that test is added.

edmundito:
I resolved some conflicts with recent changes from `main` and added a test to exportRecord to confirm that we can export to a table where a string is the primary key which was the main problem we had. I also sprinkled the data with some quotes and special characters

plugins/@grouparoo/sqlite/__tests__/data/records.csv
```diff
@@ -3,8 +3,8 @@ id,first_name,last_name,email,gender,ip_address,ios_app,android_app,vip,ltv,date
 2,Cacilie,Eate,ceate1@example.com,Female,253.146.41.201,false,true,false,94.36,2020/02/02,2020/02/02 12:13:14
 3,Gretel,Groucock,ggroucock2@example.com,Female,211.89.110.134,true,false,true,668.30,2020/02/03,2020/02/03 12:13:14
 4,Amata,Cotesford,acotesford3@example.com,Female,193.9.240.246,true,true,true,489.00,2020/02/04,2020/02/04 12:13:14
-5,Lira,Johnston,ljohnston4@example.com,Female,53.161.58.99,false,true,false,823.16,2020/02/05,2020/02/05 12:13:14
-6,Deni,Scalia,dscalia5@example.com,Female,138.131.148.74,true,false,true,273.98,2020/02/06,2020/02/06 12:13:14
+5,Lira,"""Airquotes"" Johnston",ljohnston4@example.com,Female,53.161.58.99,false,true,false,823.16,2020/02/05,2020/02/05 12:13:14
+6,Deni,O‘Hara,dscalia5@example.com,Female,138.131.148.74,true,false,true,273.98,2020/02/06,2020/02/06 12:13:14
```
evantahler: Nice.

plugins/@grouparoo/sqlite/src/lib/table-import/getRows.ts
```diff
@@ -17,33 +14,29 @@ export const getRows: GetRowsMethod = async ({
...
   highWaterMarkKey,
   incremental,
-}: {
-  highWaterMarkCondition: MatchCondition;
-  [key: string]: any;
 }) => {
   // Begin with SELECT statement.
-  let query = `SELECT *, ${highWaterMarkAndSortColumnASC} AS ${highWaterMarkKey} FROM "${tableName}"`;
+  const queryBuilder = new SQLiteQueryBuilder(
+    `SELECT *, ${highWaterMarkAndSortColumnASC} AS ${highWaterMarkKey} FROM "${tableName}"`
```
evantahler: Do these column names need to be quoted?

evantahler (approved):

plugins/@grouparoo/sqlite/__tests__/data/records.csv
evantahler: Although... I don't see those values tested for in a test?  I would have expected `expect(users[6].lastName).toBe()` somewhere

plugins/@grouparoo/sqlite/src/lib/table-import/getRows.ts
edmundito: Good catch. I did another pass to ensure that column names are always quoted

plugins/@grouparoo/sqlite/__tests__/data/records.csv
edmundito: True. Will fix

PR #2862, opened 2022-01-25
`scheduleIds` param validation for `grouparoo run`

pedroslopez:
## Change description

Checks if the passed schedules are valid and adds some more output.

The `grouparoo run` command now also stops currently running schedule runs at the start. We are assuming that if you are using `grouparoo run` you only expect things to run when you tell them to, and the previous behavior of continuing old schedules that may have been in a weird state was unexpected. Initial runs enqueued by the creation of schedules through config files will be stopped, but new runs will be enqueued for them anyway as part of the manually triggered `schedules:enqueueRuns` so the behavior remains as expected.

### Development

### Impact

### Code review

### DIFF
--- a/core/src/bin/run.ts
+++ b/core/src/bin/run.ts
@@ -3,6 +3,7 @@ import { CLI, Task, api, config, ParamsFrom } from "actionhero";
 import { Reset } from "../modules/reset";
 import { Worker } from "node-resque";
 import { getGrouparooRunMode } from "../modules/runMode";
+import { Schedule } from "../models/Schedule";
 
 export class RunCLI extends CLI {
   name = "run";
@@ -65,7 +66,9 @@ export class RunCLI extends CLI {
     >;
   }) {
     GrouparooCLI.logCLI(this.name, false);
+
     this.checkWorkers();
+    await this.checkSchedules(params.scheduleIds);
 
     if (!params.web) GrouparooCLI.disableWebServer();
     if (params.reset) await Reset.data(getGrouparooRunMode());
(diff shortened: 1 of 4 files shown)


evantahler:
Noted that this might be a `breaking-change` as the run behavior has changed

core/src/bin/run.ts
```diff
@@ -93,6 +99,44 @@ export class RunCLI extends CLI {
...
+
+    const schedules = await Schedule.findAll({ where: { id: scheduleIds } });
+    const foundScheduleIds = schedules.map((s) => s.id);
+    scheduleIds.forEach((id) => {
+      if (!foundScheduleIds.includes(id))
+        return GrouparooCLI.logger.fatal(
+          `Schedule with id "${id}" was not found`
+        );
+    });
+  }
```
evantahler, line 118: 👍 batch validation

core/src/bin/run.ts
```diff
@@ -93,6 +99,44 @@ export class RunCLI extends CLI {
...
+
+  async stopScheduleRuns() {
```
evantahler, line 120: I wonder if we should stop other types of runs too: Group, Property, Internal Run...

core/src/tasks/schedule/enqueueRuns.ts
```diff
@@ -38,12 +38,26 @@ export class ScheduleEnqueueRuns extends CLSTask {
...
       });
-      if (shouldRun) await schedule.enqueueRun();
+
+      if (shouldRun) {
+        await schedule.enqueueRun();
+        enqueuedSchedules.push(schedule);
+      }
+    }
+
+    if (getGrouparooRunMode() === "cli:run" && enqueuedSchedules.length > 0) {
```
evantahler: You could log this all the time, regardless of runMode

evantahler (approved):
This implementation looks great.

I'd want to chat a little more (perhaps with the whole group?) about whether or not we should cancel the other types of runs too...

core/src/tasks/schedule/enqueueRuns.ts
pedroslopez: Done in d3052ef, though I still changed the log level depending on runMode since it's a normal part of what happens in other run modes, but wanted to make it stand out more on `cli:run`

core/src/bin/run.ts
pedroslopez: From standup discussion: we'll only worry about Schedules for now

PR #2866, opened 2022-01-25
Write and load refreshEnabled

bleonard:
## Change description

This change ensures that `refreshEnabled` is both loaded and written into the config file.

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/modules/configLoaders/schedule.ts
+++ b/core/__tests__/modules/configLoaders/schedule.ts
@@ -1,31 +1,8 @@
 import { helper } from "@grouparoo/spec-helper";
-import { AsyncReturnType } from "type-fest";
 import { ScheduleConfigurationObject } from "../../../src/classes/codeConfig";
 import { Schedule } from "../../../src/models/Schedule";
 import { loadSchedule } from "../../../src/modules/configLoaders/schedule";
 import { ConfigWriter } from "../../../src/modules/configWriter";
-
-type ScheduleApiData = AsyncReturnType<Schedule["apiData"]>;
-
-const expectApiDataToMatchConfig = (
-  apiData: ScheduleApiData,
-  scheduleConfig: ScheduleConfigurationObject
-) => {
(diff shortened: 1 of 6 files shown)


core/src/modules/configLoaders/schedule.ts
```diff
@@ -54,6 +53,7 @@ export async function loadSchedule(
     incremental: configObject.incremental,
     recurringFrequency: configObject.recurringFrequency,
     confirmRecords: configObject.confirmRecords,
+    refreshEnabled: configObject.refreshEnabled,
```
evantahler, line 56: So for all existing codeConfig files which are missing refreshEnabled, they will get the default value from the model, which is `true`... which I guess is the behavior today

core/__tests__/modules/configLoaders/schedule.ts
```diff
@@ -0,0 +1,125 @@
+import { helper } from "@grouparoo/spec-helper";
```
evantahler, line 1: We don't have tests like this yet... I like it!

evantahler (approved):

pedroslopez (commented):
Nice! Will just note that the config loaders had been tested implicitly via `__tests__/modules/codeConfig/codeConfig.ts`, which loads real files and checks if the right things were created/updated for all the models. Could probably also add checks for `refreshEnabled` there.

edmundito:
@pedroslopez I just made the changes there as well and I noticed that this is really similar to the config loader test. We could keep both the new one is more of a unit test.

This at least ensures it fixes the issues for now but I feel we have a bad test design in that the tests pass when we introduce new properties to these grouparoo objects. I'm writing out a chore to revisit these tests so that they fail if we miss a step.

edmundito:
Pushed a fix to the configLoader test to make it ready to fail if we add more propeties to getConfigData. I forgot about that function yesterday and realized that it's much better choice than apiData.

PR #2747, opened 2022-01-04
`noImplicitAny` enabled for `@grouparoo/core`

evantahler:
A strongly-typed `@grouparoo/core` - `noImplicitAny=true`.

This PR should have no functional changes.

### Development

- [x] Application changes have been tested appropriately

### Impact

- [x] Code follows company security practices and guidelines
- [x] Security impact of change has been considered
- [x] Performance impact of change has been considered
- [x] Possible migration needs considered (model migrations, config migrations, etc.)

Please explain any security, performance, migration, or other impacts if relevant:

### Code review

- [x] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.
- [x] Relevant tags have been added to the PR (bug, enhancement, internal, etc.)

PR #2868, opened 2022-01-26
Add `retryExports` CLI command

pedroslopez:
## Change description

Adds a new cli command to retry failed exports, following the functionality introduced in #2825 

Example `grouparoo retryExports --startAgo 5000`

```
🦘 Grouparoo: retryExports

Searching for failed Exports:

Start: 1/27/2022, 10:11:21 AM (5000 seconds ago)
End: 1/27/2022, 11:34:41 AM

┌-- Summary @ 2022-01-27T15:34:41.836Z ---
| My Destination (my_destination)
| * Failed Exports: 6
|
| Accounts Destination (accounts_destination)
| * Failed Exports: 0
└-----------------------------------------

✅ Success! 6 failed Exports marked to be retried.
```

Date/times for `--start` and `--end` can be passed in as anything that JS considers valid ("Jan 23, 2022 10:53", "1642949580000", "2022-01-23T14:53:00.000Z")

Also in this PR:
- Brought in `formatter` and `validator` support into CLI following [actionhero/actionhero#2062](https://github.com/actionhero/actionhero/pull/2062)
- Fixed an issue where final status reporter would show `undefined` in the Records section
- Final status reporter now shows destinations that have just failed/completed old exports 

### Development

### Impact

### Code review

### DIFF
--- a/cli/src/utils/loadLocalCommands.ts
+++ b/cli/src/utils/loadLocalCommands.ts
@@ -29,6 +29,8 @@ type ActionheroCLIInputStub = {
   flag?: string;
   placeholder?: string;
   variadic?: boolean;
+  formatter?: Function;
+  validator?: Function;
 };
 
 export async function loadLocalCommands(program: Command): Promise<boolean> {
@@ -228,6 +230,16 @@ async function runCommand(
   clearRequireCache();
 
   params["_arguments"] = _arguments;
+
+  for (const [key, inputOpts] of Object.entries(instance.inputs)) {
+    if (typeof inputOpts.formatter === "function") {
+      params[key] = await inputOpts.formatter(params[key]);
+    }
+
+    if (typeof inputOpts.validator === "function") {
+      await inputOpts.validator(params[key]);
+    }
+  }
 
   if (instance.initialize === false && instance.start === false) {
     toStop = await instance.run({ params });
--- a/cli/src/utils/loadLocalCommands.ts
+++ b/cli/src/utils/loadLocalCommands.ts
@@ -25,6 +25,7 @@ type ActionheroCLIInputStub = {
   description?: string;
   default?: boolean;
   required?: boolean;
+  requiredValue?: boolean;
   letter?: string;
   flag?: string;
   placeholder?: string;
@@ -191,7 +192,8 @@ async function convertCLIToCommanderAction(
 
   for (const key in instance.inputs) {
     const input = instance.inputs[key];
-    const separators = input.required ? ["<", ">"] : ["[", "]"];
+    const separators =
+      input.required || input.requiredValue ? ["<", ">"] : ["[", "]"];
     const methodName = input.required ? "requiredOption" : "option";
     const argString = `${input.letter ? `-${input.letter}, ` : ""}--${key} ${
       input.flag
--- a/cli/src/utils/loadLocalCommands.ts
+++ b/cli/src/utils/loadLocalCommands.ts
@@ -4,7 +4,7 @@ import Ora from "ora";
 import * as glob from "glob";
 import { readPackageJSON } from "./readPackageJSON";
 import { ensureNoTsHeaderFiles } from "./ensureNoTsHeaderFiles";
-import { Command } from "commander";
+import { Command, InvalidArgumentError } from "commander";
 
 // We don't want to require actionhero in the project (so we can rely on core's version), so we need to stub this type
 type ActionheroCLIStub = {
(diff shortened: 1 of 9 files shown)


cli/src/utils/loadLocalCommands.ts
```diff
@@ -229,6 +231,16 @@ async function runCommand(
...
 
+  for (const [key, inputOpts] of Object.entries(instance.inputs)) {
+    if (typeof inputOpts.formatter === "function") {
+      params[key] = await inputOpts.formatter(params[key]);
+    }
+
+    if (typeof inputOpts.validator === "function") {
+      await inputOpts.validator(params[key]);
+    }
+  }
```
pedroslopez: From https://github.com/actionhero/actionhero/pull/2062

core/src/modules/statusReporters.ts
```diff
@@ -601,7 +601,8 @@ export namespace FinalSummaryReporters {
       });
       const allRecords = await GrouparooRecord.count();
 
-      const recordData = {
+      const recordData: RecordData = {
+        name: null,
```
pedroslopez, line 605: Unrelated, but noticed that having this `undefined` made the string "undefined" show up in the final status reporter as the name for the section. This was previously explicitly set to null, but removed (likely unintentionally) as part of #2747

evantahler: Whoops!

cli/src/utils/loadLocalCommands.ts
evantahler: I think we still need this special file to load plugins because Grouparoo is running from within node_modules, not on top of it... but perhaps we can load some some of these methods from within actionhero rather than re-implementing them?

core/src/modules/apiData.ts
```diff
@@ -76,7 +76,10 @@ export namespace APIData {
   export function formatDate(date: Date | string) {
     if (!date || date === "") return null;
     else if (date instanceof Date) return date.getTime();
-    else if (typeof date === "string") return new Date(date).getTime();
-    else throw new Error(`${date} is not a date`);
+    else if (typeof date === "string") {
+      const asNumber = Number(date);
+      if (!isNaN(asNumber)) return asNumber;
+      return new Date(date).getTime();
+    } else throw new Error(`${date} is not a date`);
```
evantahler, line 97: Nice

core/src/bin/retryExports.ts
```diff
@@ -0,0 +1,157 @@
...
+      params.start !== undefined && params.start !== null;
+
+    if (
+      (hasRelativeStart && hasAbsoluteStart) ||
+      (!hasRelativeStart && !hasAbsoluteStart)
+    ) {
+      return GrouparooCLI.logger.fatal(
+        "One of --start or --startAgoSeconds must be specified"
+      );
+    }
```
evantahler: Dealing with "at least one of these options" is always kind of custom

core/src/bin/retryExports.ts
```diff
@@ -0,0 +1,157 @@
...
+      required: false,
+      formatter: APIData.formatDate,
+      description:
+        "Search for failed Exports created on or after this timestamp.",
+    },
+    startAgoSeconds: {
+      required: false,
+      description:
+        "Search for failed Exports created on or after a certain number of seconds ago.",
+    },
```
evantahler: I think this is OK, but I'm curious to see if folks use this and prefer minutes/hours/ some other unit

core/src/bin/retryExports.ts
```diff
@@ -0,0 +1,157 @@
...
+    GrouparooCLI.logger.status("Summary", summaryItems);
+
+    if (params.preview) {
+      GrouparooCLI.logger.log(
+        `ℹ️  (Preview) Found ${totalCount} failed Exports to retry.`
+      );
+    } else {
+      if (totalCount) {
+        GrouparooCLI.logger.log(
+          `✅ Success! ${totalCount} failed Exports marked to be retried.`
```
evantahler: Perhaps we should add something like "Now, run `grouparoo run` or `grouparoo start` to retry them"

core/src/bin/retryExports.ts
evantahler: Coming soon - an actionhero update to help with this

cli/src/utils/loadLocalCommands.ts
pedroslopez: Gave it a (very quick) shot, but there's a comment in our `loadLocalCommands` file that says we don't want to require `actionhero` in the `cli` package, which makes this tricky...

evantahler: That's in the `getActionhero` method itself... I wonder if it is OK to use that method to load a method like

```
function doStuff (...args) {
  const originalDoStuff = getActionhero()('doStuff')
  return originalDoStuff(args...)
}
```

core/src/bin/retryExports.ts
pedroslopez: This one still needs a bit of custom processing since it's one or the other, but pretty much all other validations have been radically simplified!

core/src/bin/retryExports.ts
pedroslopez: Added support for anything Moment supports in 7559dc6 !

core/src/bin/retryExports.ts
```diff
@@ -0,0 +1,169 @@
...
+      formatter: APIData.ensureBoolean,
+      description:
+        "When set, will not make any changes and only outputs the number of Exports that would be retried.",
+      flag: true,
+      letter: "p",
+    },
+    destinationIds: {
+      required: false,
+      requiredValue: true,
+      formatter: (val: any) => val as string[],
```
pedroslopez: We can remove this formatter once actionhero is updated (swap out for a `variadic: true as const`)

cli/src/utils/loadLocalCommands.ts
pedroslopez: Ah I see! That doesn't seem _terrible_ (aside from not getting typings)

Looks like some work needs to be done on the actionhero side though - these cli methods are being defined from within the actionhero cli (and not exported to actionhero). Grouparoo also passes around the commander `program`. This probably means refactoring a bit to move these methods into some utility module that can be exported by actionhero, taking in which `program` to use.

evantahler (approved):
👍 once the new version of actionhero is in!

PR #2872, opened 2022-01-27
Fix: UI Error Handling and Redirects

evantahler:
This PR updates Grouparoo's UI Error handing when the user should be redirected to the Sign In page.

Recently, we've started transitioning from Next.js's `getInitialProps()` to `getServerSideProps()`.  `getInitialProps` allowed for modification of the `res`ponse object, including setting headers and ending the connection, where as `getServerSideProps` does not - as it tries to be smarter about error handling and building up the user response payload.  This PR cppies the `withServerErrorHandler` utility from cloud to help with redirects from `getServerSideProps`

### Development

### Impact

### Code review


core/src/modules/errors.ts
```diff
@@ -17,6 +17,7 @@ export namespace Errors {
     constructor(message: string, code = "AUTHENTICATION_ERROR") {
       super(message);
       this.code = code;
+      delete this.stack;
```
evantahler, line 20: Also in this PR: I'm cleaning up our logs when we hit an `AUTHENTICATION_ERROR` - the stack trace is not meaningful and the action log tells you enough:

Before:

```
2022-01-27T02:25:56.460Z - error: [ action @ web ] to=127.0.0.1 action=destinations:list params={"action":"destinations:list","apiVersion":"1","modelId":"people","serverToken":"change_me"} duration=20 method=GET pathname=/api/v1/destinations error=Error: Please log in to continue code=AUTHENTICATION_ERROR stack=Error: Please log in to continue
    at cls_1.CLS.wrap.catchError (/Users/evan/workspace/grouparoo/grouparoo/core/src/modules/middleware/authentication.ts:75:17)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async /Users/evan/workspace/grouparoo/grouparoo/core/src/modules/cls.ts:87:21
    at async /Users/evan/workspace/grouparoo/grouparoo/node_modules/.pnpm/sequelize@6.11.0_98f032adade92d92d5295670c0365597/node_modules/sequelize/lib/sequelize.js:1097:24
    at async wrapInternal (/Users/evan/workspace/grouparoo/grouparoo/core/src/modules/cls.ts:86:5)
    at async Object.CLS.wrap (/Users/evan/workspace/grouparoo/grouparoo/core/src/modules/cls.ts:48:27)
    at async authenticateTeamMemberFromSession (/Users/evan/workspace/grouparoo/grouparoo/core/src/modules/middleware/authentication.ts:63:10)
    at async authenticateTeamMember (/Users/evan/workspace/grouparoo/grouparoo/core/src/modules/middleware/authentication.ts:144:13)
    at async ActionProcessor.preProcessAction (/Users/evan/workspace/grouparoo/grouparoo/node_modules/.pnpm/actionhero@28.1.4/node_modules/actionhero/dist/classes/actionProcessor.js:162:17)
    at async ActionProcessor.runAction (/Users/evan/workspace/grouparoo/grouparoo/node_modules/.pnpm/actionhero@28.1.4/node_modules/actionhero/dist/classes/actionProcessor.js:331:40)
    at async WebServer.processAction (/Users/evan/workspace/grouparoo/grouparoo/node_modules/.pnpm/actionhero@28.1.4/node_modules/actionhero/dist/classes/server.js:113:22)
```

After:

```
2022-01-27T02:25:20.916Z - error: [ action @ web ] to=127.0.0.1 action=destinations:list params={"action":"destinations:list","apiVersion":"1","modelId":"people","serverToken":"change_me"} duration=17 method=GET pathname=/api/v1/destinations error=Error: Please log in to continue code=AUTHENTICATION_ERROR
```

ui/ui-components/eventHandlers/setupStepsHandler.ts
```diff
@@ -1,3 +1,6 @@
+import { Models } from "../utils/apiData";
 import { EventDispatcher } from "../utils/eventDispatcher";
 
-export class SetupStepHandler extends EventDispatcher<string[]> {}
+export class SetupStepsHandler extends EventDispatcher<
+  Models.SetupStepType[]
+> {}
```
evantahler, line 6: Also in this PR: The Type and name of this hander wasn't correct

pedroslopez:
> An alternative approach would be to try/catch around every usage of getServerSideProps() and return a {redirect:{}} payload... but that didn't seem worth it. What do you think, PR reviewers?

@evantahler FWIW, in cloud we took a similar approach and introduced a `withServerErrorHandler` wrapper that handles this and returns the `redirect`. https://github.com/grouparoo/cloud/blob/main/web/util/withServerErrorHandler.ts

ui/ui-components/pages/dashboard.tsx
```diff
@@ -52,3 +54,10 @@ export default function Page(props) {
     </>
   );
 }
+
+// this getServerSideProps doesn't actually return anything, but we to want to make a request to see if the user is signed in
+export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
+  const { execApi } = UseApi(ctx);
```
krishnaglick: This should be using the new `generateClient` call.

ui/ui-components/pages/dashboard.tsx
```diff
@@ -52,3 +54,10 @@ export default function Page(props) {
     </>
   );
 }
+
+// this getServerSideProps doesn't actually return anything, but we to want to make a request to see if the user is signed in
```
krishnaglick, line 59: I feel like we should make this call in `_app`

krishnaglick (commented):
Please test this with `roo start`, I was seeing different functionality in `npm run dev` vs `roo start`.

ui/ui-components/pages/dashboard.tsx
evantahler: We technically do (to load the nav), but that always works... we need to load *something* that requires authentication :/

evantahler:
> Please test this with `roo start`, I was seeing different functionality in `npm run dev` vs `roo start`.

Yep! The `dev` command produces the red next errors now, and the `start` command does what we want - redirects the user with our custom error page

evantahler:
> @evantahler FWIW, in cloud we took a similar approach and introduced a withServerErrorHandler wrapper that handles this and returns the redirect. https://github.com/grouparoo/cloud/blob/main/web/util/withServerErrorHandler.ts 

Oh, I like that pattern better!  Since we don't have that many `getServerSideProps` yet, it's probably better to get this in place now.  This makes the dev and run experiences the same!

ui/ui-components/pages/dashboard.tsx
krishnaglick: Making this call in _app would trigger a redirect, no?

evantahler: A few problems I see:
- not every page is authenticated, so then _app would need to be aware of which pages should have this call or not, or at least how to handle the error differently on each page
- this is an extra request on every page, since most (all but 2) pages have some hydration step of their own which then in turn checks for auth

krishnaglick: Maybe we should try to do what Cloud does, and create a `useAuthenticatedPage` hook?
That way redirects are triggered at the React level rather than the request level, and we can just authenticate in `_app` to know if we're logged in or not.

ui/ui-components/pages/model/[modelId]/overview.tsx
```diff
@@ -36,6 +37,8 @@ const Page: NextPage<Props> = ({
   schedules,
   destinations,
 }) => {
+  if (!model) return <Alert variant="warning">Model not found</Alert>;
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/overview.tsx:40]: Note - if you went to a model overview page for a model that doesn't exist, you would have previously gotten a 'not iterable error'

ui/ui-components/pages/dashboard.tsx
evantahler: Perhpas! I think that's a new story though.

krishnaglick: Are there other pages that need something like this so the fail properly?

evantahler: AFAIK this is the only authenticated page that didn't make a `getInitialProps` or `getServerSideProps`

krishnaglick (approved):

ui/ui-components/pages/model/[modelId]/overview.tsx
Comment (edmundito) [reply ui/ui-components/pages/model/[modelId]/overview.tsx:40]: I just noticed this change in main and one thing to note is that it's bad practice to return before all the hooks are set. React has a lot of problems with it.

PR #2874, opened 2022-01-27
Select from multiple apps after installing plugin from config UI

edmundito:
## Change description

This change ensures when adding a new app that needs to be installed first, if there are multiple apps in the plugin, it goes to the app selection page instead of creating the first app it finds.

Additionally, it cleans up the page code and fixes some issues with checking on restarting the app.

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-config/pages/app/new/index.tsx
+++ b/ui/ui-config/pages/app/new/index.tsx
@@ -154,10 +154,10 @@ const Page: NextPageWithInferredProps<typeof getServerSideProps> = ({
       setShowSpinner(false);
       setInstallingMessage(
         <Alert variant="warning">
-          There was a problem restarting the Grouparoo config app.
+          There was a problem restarting.
           <br />
           <br />
-          Please restart config app via the command line.
+          Please restart the Grouparoo config app via the CLI.
         </Alert>
       );
       return;
--- a/ui/ui-config/pages/app/new/index.tsx
+++ b/ui/ui-config/pages/app/new/index.tsx
@@ -1,14 +1,18 @@
-import { useApi } from "../../../../ui-components/contexts/api";
+import { GetServerSidePropsContext } from "next";
 import Head from "next/head";
 import { useState, useEffect } from "react";
 import { Form, Modal, Spinner, Alert } from "react-bootstrap";
 import { useRouter } from "next/router";
 import AppSelectorList from "@grouparoo/ui-components/components/AppSelectorList";
+import { useApi } from "@grouparoo/ui-components/contexts/api";
 import { errorHandler } from "@grouparoo/ui-components/eventHandlers";
 import { EventDispatcher } from "@grouparoo/ui-components/utils/eventDispatcher";
 import { Actions } from "@grouparoo/ui-components/utils/apiData";
 import { generateClient } from "@grouparoo/ui-components/client/client";
-import { NextPageContext } from "next";
+import { NextPageWithInferredProps } from "@grouparoo/ui-components/utils/pageHelper";
+import React from "react";
+
+type PluginWithVersion = Actions.PluginsList["plugins"][number];
 
 class CustomErrorHandler extends EventDispatcher<{ message: string }> {
   message: Error | string | any = null;
@@ -20,7 +24,7 @@ class CustomErrorHandler extends EventDispatcher<{ message: string }> {
       this.message = message;
 
       if (this.message?.config?.url === "/api/v1/status/public") {
-        console.info(`could not reach API: ${message}`);
+        console.info(`Could not reach API: ${message}`);
       } else {
         errorHandler.set({ message: this.message });
       }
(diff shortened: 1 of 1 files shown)


ui/ui-config/pages/app/new/index.tsx
```diff
@@ -20,87 +24,96 @@ class CustomErrorHandler extends EventDispatcher<{ message: string }> {
...
-      const response: Actions.AppCreate = await client.request("post", `/app`, {
-        type: plugin.apps[0].name,
-      });
-      if (response?.app) {
-        return router.push("/app/[id]/edit", `/app/${response.app.id}/edit`);
-      } else {
-        setLoading(false);
-      }
-    } else {
-      router.push(`/app/new/${plugin.name}`);
```
evantahler, line 93: it looks like this page was /trying/ to do the proper logic before.  Can you call out what the problem was?

ui/ui-config/pages/app/new/index.tsx
```diff
@@ -20,87 +24,96 @@ class CustomErrorHandler extends EventDispatcher<{ message: string }> {
...
-  }
 
-  function sleep(ms) {
-    return new Promise((resolve) => setTimeout(resolve, ms));
+    setLoading(true);
+    const response = await client.request<Actions.AppCreate>("post", `/app`, {
+      type: plugin.apps[0].name,
+    });
+    if (response?.app) {
+      return router.push("/app/[id]/edit", `/app/${response.app.id}/edit`);
```
evantahler, line 102: I think you can just do `return router.push(`/app/${response.app.id}/edit`);` now

evantahler (approved):

ui/ui-config/pages/app/new/index.tsx
```diff
@@ -20,87 +24,96 @@ class CustomErrorHandler extends EventDispatcher<{ message: string }> {
...
+}) => {
   const router = useRouter();
   const { client } = useApi();
-  const [plugin, setPlugin] = useState<
-    Partial<Actions.PluginsList["plugins"][number]>
-  >({ name: "" });
-  const [cards, setCards] = useState([]);
+  const [plugin, setPlugin] = useState<Partial<PluginWithVersion>>(() => ({
+    name: "",
+  }));
```
krishnaglick, line 58: I don't precisely recall how the computed state works, but since you're setting this to a static object I don't think you need to do so.

ui/ui-config/pages/app/new/index.tsx
```diff
@@ -20,87 +24,96 @@ class CustomErrorHandler extends EventDispatcher<{ message: string }> {
...
-        .sort((a, b) => (a.name > b.name ? 1 : -1))
-        .sort((a, b) => {
-          if (a.installed === b.installed) return 0;
-          if (a.installed) return -1;
-          return 1;
-        })
-    );
+    const nextCards = _plugins
+      .filter((p) => p.source || p.destination)
+      .sort((a, b) => (a.name > b.name ? 1 : -1))
```
krishnaglick, line 75:
```suggestion
      .sort((a, b) => (a.name - b.name))
```
I _think_ this will function more consistently.

ui/ui-config/pages/app/new/index.tsx
```diff
@@ -139,28 +154,40 @@ export default function Page(props) {
...
-    let response: Actions.PublicStatus = await client.request(
-      "get",
-      "/status/public",
-      undefined,
-      { useCache: false }
-    );
+    const retry = async () => {
+      console.log("Server down. Trying again in 1s...");
+      await sleep(1000);
+      return await waitForServer(failTime);
```
krishnaglick, line 169: Why not just add 1000 to failTime?

ui/ui-config/pages/app/new/index.tsx
```diff
@@ -139,28 +154,40 @@ export default function Page(props) {
...
+        undefined,
+        { useCache: false, errorHandler: new CustomErrorHandler() }
+      );
 
-    if (response["status"] !== "ok") {
-      console.log("Server down. Trying again in 0.25 seconds ...");
-      await sleep(250);
-      response = await waitForServer(failTime);
-    }
+      if (response["status"] !== "ok") {
```
krishnaglick, line 180:
```suggestion
      if (response.status !== "ok") {
```
Nitpick

ui/ui-config/pages/app/new/index.tsx
edmundito: It was only doing the proper logic when the plugin was already installed. But if it needed to install the plugin, it would not do the same thing.

ui/ui-config/pages/app/new/index.tsx
edmundito: I didn't know! I've been keeping these as they were. I will keep that in mind for future updates.

ui/ui-config/pages/app/new/index.tsx
edmundito: I'm just cleaning the syntax here. This file needed more work but wanted to focus on fixing the bug.

ui/ui-config/pages/app/new/index.tsx
edmundito: It's because while this is happening the app is restarting so if we wait for the server it's going to fail immediately due to network errors

PR #2875, opened 2022-01-27
Add Grouparoo model global context

edmundito:
## Change description

This moves the grouparooModel context to be global so that any page can get the model without having to fetch it individually. All pages have been migrated to get the model from the global context.

Usage:

```ts
import { useGrouparooModel } from '../contexts/grouparooModel'

const PageOrComponent = (props) => {
   const { model } = useGrouparooModel()
}
```

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/contexts/grouparooModel.tsx
+++ b/ui/ui-components/contexts/grouparooModel.tsx
@@ -1,4 +1,4 @@
-import { createContext, Dispatch, SetStateAction, useContext } from "react";
+import { createContext, useContext } from "react";
 import { Models } from "../utils/apiData";
 
 interface GrouparooModelContext {
--- a/ui/ui-components/contexts/grouparooModel.tsx
+++ b/ui/ui-components/contexts/grouparooModel.tsx
@@ -21,7 +21,7 @@ export const useGrouparooModel = () => useContext(GrouparooModelContext);
 export const GrouparooModelContextProvider: React.FC<{
   value: Models.GrouparooModelType;
 }> = ({ value: modelProp, children }) => {
-  const [updatedModel, setModel] = useState(modelProp);
+  const [model, setModel] = useState(modelProp);
 
   useEffect(() => {
     setModel(modelProp);
@@ -29,7 +29,7 @@ export const GrouparooModelContextProvider: React.FC<{
 
   return (
     <GrouparooModelContext.Provider
-      value={{ model: updatedModel || modelProp, setModel }}
+      value={{ model: model ?? modelProp, setModel }}
     >
       {children}
     </GrouparooModelContext.Provider>
--- a/ui/ui-components/contexts/grouparooModel.tsx
+++ b/ui/ui-components/contexts/grouparooModel.tsx
@@ -21,14 +21,16 @@ export const useGrouparooModel = () => useContext(GrouparooModelContext);
 export const GrouparooModelContextProvider: React.FC<{
   value: Models.GrouparooModelType;
 }> = ({ value: modelProp, children }) => {
-  const [model, setModel] = useState(modelProp);
+  const [updatedModel, setModel] = useState(modelProp);
 
   useEffect(() => {
     setModel(modelProp);
   }, [modelProp]);
 
   return (
-    <GrouparooModelContext.Provider value={{ model, setModel }}>
+    <GrouparooModelContext.Provider
+      value={{ model: updatedModel || modelProp, setModel }}
+    >
       {children}
     </GrouparooModelContext.Provider>
   );
--- a/ui/ui-components/contexts/grouparooModel.tsx
+++ b/ui/ui-components/contexts/grouparooModel.tsx
@@ -1,37 +1,11 @@
-import {
-  createContext,
-  Dispatch,
-  SetStateAction,
-  useContext,
-  useEffect,
-  useState,
-} from "react";
+import { createContext, Dispatch, SetStateAction, useContext } from "react";
 import { Models } from "../utils/apiData";
 
 interface GrouparooModelContext {
   model: Models.GrouparooModelType;
-  setModel: Dispatch<SetStateAction<Models.GrouparooModelType>>;
 }
 
 export const GrouparooModelContext =
   createContext<GrouparooModelContext>(undefined);
 
 export const useGrouparooModel = () => useContext(GrouparooModelContext);
-
-export const GrouparooModelContextProvider: React.FC<{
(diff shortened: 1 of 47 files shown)


ui/ui-components/contexts/grouparooModel.tsx
```diff
@@ -1,15 +1,37 @@
...
+}> = ({ value: modelProp, children }) => {
+  const [updatedModel, setModel] = useState(modelProp);
+
+  useEffect(() => {
+    setModel(modelProp);
+  }, [modelProp]);
+
+  return (
+    <GrouparooModelContext.Provider
+      value={{ model: updatedModel || modelProp, setModel }}
```
krishnaglick: Does this update automatically if you just pass in modelProp to the value?

ui/ui-components/pages/_app.tsx
```diff
@@ -74,15 +79,30 @@ GrouparooNextApp.getInitialProps = async (appContext: AppContext) => {
...
+      client.request<Actions.NavigationList>("get", `/navigation`),
+    ];
+
+    if (modelId) {
+      promises.push(
+        client.request<Actions.ModelView>("get", `/model/${modelId}`)
+      );
+    }
+
+    [navigationResponse, modelResponse] = await Promise.all(promises);
```
krishnaglick:
```suggestion
    const [navigationResponse, modelResponse] = await Promise.all([
        client.request<Actions.NavigationList>("get", `/navigation`),
        modelId ? client.request<Actions.ModelView>("get", `/model/${modelId}`) : Promise.resolve()
    ]);
```
Something like this should work, and then you don't have a `Promise<any>[]`

ui/ui-components/pages/_app.tsx
```diff
@@ -106,6 +126,7 @@ GrouparooNextApp.getInitialProps = async (appContext: AppContext) => {
 
     return {
       ...appProps,
+      model: model || null,
```
krishnaglick: Nitpick
```suggestion
      model: model ?? null,
```

ui/ui-components/contexts/grouparooModel.tsx
edmundito: This change was made for setModel to work  and the timing of Next js. Here's what happens:

1. I navigate to a page without the model: model prop is `undefined`, model state is also `undefined`.
2. I navigate from the page without the model to a page that does have a modelId: model prop is now defined, model state is still `undefined`
3. Page renders
4. useEffect kicks in: model state is now the same as model prop.

krishnaglick: If value was `{ model: modelProp }` wouldn't that just update when the prop changed instead of trigger a `useEffect` to do the same?

edmundito: unless you update the prop from `setModel()` when you do an async update from the client-side rendering

edmundito:
After chatting with @krishnaglick it made sense to simplify the context provider. There is only one place where the model object is currently updated and the solution can be remediated by reloading the route which is the way it works now.

ui/ui-components/pages/_app.tsx
```diff
@@ -74,15 +79,33 @@ GrouparooNextApp.getInitialProps = async (appContext: AppContext) => {
...
+
+    const promises: Promise<any>[] = [
+      client.request<Actions.NavigationList>("get", `/navigation`),
+    ];
+
+    if (modelId) {
+      promises.push(
+        client.request<Actions.ModelView>("get", `/model/${modelId}`)
+      );
+    }
```
krishnaglick:
```suggestion
```

evantahler (commented):
As a pattern, I love it.  I'll let others comment on the implementation specifics

ui/ui-components/contexts/grouparooModel.tsx
```diff
@@ -1,15 +1,11 @@
-import { createContext, useContext } from "react";
+import { createContext, Dispatch, SetStateAction, useContext } from "react";
```
pedroslopez: Are `Dispatch` and `SetStateAction` being used here?

ui/ui-components/pages/model/[modelId]/edit.tsx
```diff
@@ -66,7 +61,7 @@ export default function Page(props) {
   const update = async (event) => {
     const _model = Object.assign({}, model);
     _model[event.target.id] = event.target.value;
-    setModel(_model);
+    setModelToUpdate(_model);
   };
 
   return (
```
Comment (pedroslopez) [inline ui/ui-components/pages/model/[modelId]/edit.tsx]: Looks like we need to move to use `modelToUpdate` in a couple places here - the auto-updating breadcrumb and title broke after this change (use in `ModelTabs`, `PageHeader`, `Head`)

pedroslopez (changes requested):
Nice! There are still some duplicate requests being made in `_app.tsx`'s `getInitialProps` as @krishnaglick pointed out.

We should also look at the model edit page to keep its behavior consistent with our other edit pages (titles/headers/breadcrumbs auto update as name is changed)

ui/ui-components/pages/model/[modelId]/edit.tsx
Comment (edmundito) [reply ui/ui-components/pages/model/[modelId]/edit.tsx]: I considered this and it's intentional. To me, it didn't make sense to auto-update until the user presses "save" because the changes haven't been applied yet.

Comment (pedroslopez) [reply ui/ui-components/pages/model/[modelId]/edit.tsx]: Yeah, I agree - though I bring it up because recently the same happened with the App edit page and a story was made to revert the behavior and make it match with all of our other edit pages. We should either do this everywhere or keep it consistent.

Comment (edmundito) [reply ui/ui-components/pages/model/[modelId]/edit.tsx]: @pedroslopez not a big fan but I updated it so that it's using the state on this page.

ui/ui-components/pages/_app.tsx
pedroslopez: @edmundito This was kc's change I was referring to in my earlier review - We're still calling `client.request` twice for both the nav list and model, presumably from an earlier implementation. This `promises` array doesn't seem to be used.

pedroslopez (changes requested):

ui/ui-components/pages/_app.tsx
edmundito: @pedroslopez Whoops, I thought I deleted those! I just pushed a change that removes them.

krishnaglick (approved):

pedroslopez (approved):

PR #2876, opened 2022-01-28
Display sign in link for older `getInitialProps` pages on auth error

evantahler:
Following the work in https://github.com/grouparoo/grouparoo/pull/2872, for our remaining pages that still hydrate via `getInitialProps`, we no longer auto-redirect to the sign in page when there's an auth error.  As we intend to swap `getInitialProps` for `getServerSideProps` on all pages, this PR is a temporary stop-gap. This PR makes the error message show as a link so folks can sign in easily. 

Screen Shot 2022-01-28 at 8 38 55 AM

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/components/alerts/HydrationError.tsx
+++ b/ui/ui-components/components/alerts/HydrationError.tsx
@@ -1,15 +1,17 @@
 import { Fragment } from "react";
 import { Card } from "react-bootstrap";
 import Link from "next/link";
-import { isBrowser } from "../../utils/isBrowser";
+import { useRouter } from "next/router";
 
 export default function HydrationError({
   hydrationError,
 }: {
   hydrationError: string;
 }) {
+  const router = useRouter();
   let error: Error;
   let errorData: { [key: string]: any };
+
   try {
     const parsed = JSON.parse(hydrationError);
     error = new Error(parsed.message);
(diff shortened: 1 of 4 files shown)


ui/ui-components/components/alerts/HydrationError.tsx
```diff
@@ -23,7 +25,19 @@ export default function HydrationError({
         <Card.Header>There was an error loading this Page</Card.Header>
         <Card.Body>
           <blockquote className="blockquote mb-0">
-            <p>{error.message}</p>
+            {error?.message.match(/sign in to continue/) && isBrowser() ? (
```
krishnaglick: Why only show this on the client? If we're SSRing it should be fine to render on the server.

krishnaglick (commented):
Fun fact (that I haven't fact-checked): Sign in translates better than log in.

evantahler:
> Fun fact (that I haven't fact-checked): Sign in translates better than log in.

Yep! I had that repeated to me many times before... and then I forgot the first time we made this error :D

ui/ui-components/components/alerts/HydrationError.tsx
evantahler: It actually still works on an initial page load!  It just changes the body once it gets to the client and re-renders.  I was having trouble getting the nextPage on the server, since we are already in an error "catch", the response object was odd

krishnaglick: Looking at this, the `pathname` is the only hangup?
Iirc we have a `useRouter` or something from Next, no?
Unless there's something else I've missed ^^

evantahler: Boom - b5be1b3ddac8e13e6ab449d5963a25e236d90b18

krishnaglick (approved):

PR #2886, opened 2022-02-01
Remove unused Import metadata and start `record:export` sooner

### LINKED ISSUE #2870: [Feature]: Performance Improvements

_No response_

_No response_

### Additional context

_No response_

evantahler:
This PR removes a number of unused data columns on the Import table.  This is done to make the Import table smaller and speed up writes.  This PR also removes a timed wait step (the old `records:enqueueExports` task) in favor of chaining tasks together to speed things up.

We were storing a ton of data on every Import - old and new record properties, old and new groups, and the full data of the source's row that produced this Import.  We aren't using this data in the UI, and it has not proved to be useful in debugging.  Having large write like this slows things down, so we are going to stop doing it!

This PR also stops using Imports as a trigger to know when a Record is ready to Export, further decoupling Imports from Exports.
* We no longer need the `record:enqueueExports` task, as we can enqueue the `record:export` task directly from the `record:makeExports` task.  
  * A new backstop task, `records:reEnqueueStuckExportingRecords` is used when those tasks are lost to recover - if a record has been in the "exporting" state for over an hour, we re-enqueue the `record:export` task

### Development

### Impact

### Code review


core/src/models/GrouparooRecord.ts
```diff
@@ -29,7 +29,7 @@ import { Source } from "./Source";
 import { GrouparooModel } from "./GrouparooModel";
 import { ModelGuard } from "../modules/modelGuard";
 
-const STATES = ["draft", "pending", "ready", "deleted"] as const;
+const STATES = ["draft", "pending", "exporting", "ready", "deleted"] as const;
```
evantahler: Alternatively to adding a new state, we could make a new column on the record `readyForExport (bool)`

bleonard (commented):
I'm thinking a `state` might be overloading things. How about a boolean `needsExporting`. Be sure to checked for `ready` and that boolean in the export task.

core/src/models/Import.ts
```diff
@@ -32,22 +32,13 @@ export interface ImportRecordProperties {
...
+const STATES = ["associating", "importing", "failed", "complete"] as const;
 
 const STATE_TRANSITIONS: StateMachine.StateTransition[] = [
   { from: "associating", to: "failed", checks: [] },
   { from: "associating", to: "importing", checks: [] },
   { from: "importing", to: "failed", checks: [] },
   { from: "importing", to: "complete", checks: [] },
-  { from: "importing", to: "processing", checks: [] },
-  { from: "processing", to: "failed", checks: [] },
-  { from: "processing", to: "complete", checks: [] },
```
evantahler: There is no more `processing` state for imports.  They are `complete` once the record they are associated to goes ready.

core/src/modules/ops/record.ts
```diff
@@ -1105,54 +1087,20 @@ export namespace RecordOps {
...
+    await Import.update(
+      { state: "complete", importedAt: now },
+      {
+        where: {
+          recordId: { [Op.in]: records.map((p) => p.id) },
+          state: "importing",
+          recordAssociatedAt: { [Op.lt]: now },
+        },
+      }
     );
```
evantahler: There's a possible race condition here where a new import may have appeared in the middle of this method, the record should have been moved to 'pending' again.  BUT, we looked up only those records who had no pending properties a few lines up - the transactional safety of this method should make this OK

core/src/tasks/record/export.ts
```diff
@@ -88,10 +74,12 @@ export class RecordExport extends RetryableTask {
       }
     } catch (error) {
       if (env !== "test") log(`[EXPORT ERROR] ${error}`, "alert");
+    }
 
-      for (const i of imports) {
-        await i.setError(error, this.name);
-      }
+    try {
+      await record.update({ state: "ready" });
```
evantahler, line 80: After this task, the record goes back to the ready state

core/src/tasks/record/export.ts
```diff
@@ -88,10 +74,12 @@ export class RecordExport extends RetryableTask {
       }
```
pedroslopez, line 72: Looks like we're still setting `processedAt` and updating import state here, which would not be necessary

core/src/tasks/record/export.ts
```diff
@@ -19,59 +20,44 @@ export class RecordExport extends RetryableTask {
...
     if (record.state !== "ready") return;
 
     const imports = await Import.findAll({
       where: {
-        state: "processing",
+        state: "complete",
         recordId: record.id,
       },
       order: [["createdAt", "asc"]],
     });
```
pedroslopez: Previously we had only been looking up imports that had not created exports, whereas now we are looking up all complete imports. I _believe_ this is only being used to check which specific destinations we need to export to, but it would be including all destinations we've ever wanted to export to instead of the ones from the latest imports.

core/src/tasks/record/makeExports.ts
```diff
@@ -16,6 +17,12 @@ export class GrouparooRecordsCheckReady extends CLSTask {
...
       : true;
 
-    await RecordOps.makeReadyAndCompleteImports(limit, toExport);
+    const records = await RecordOps.makeExports(limit);
+
+    if (toExport) {
+      for (const record of records) {
+        await CLS.enqueueTask("record:export", { recordId: record.id });
+      }
+    }
```
pedroslopez, line 26: Previously, when `toExport=true` we would not create the exports in the db since the imports were going directly to `complete` instead of passing through `processing`. 

In this case we are now creating the exports, but just not enqueuing the task. I think this change is ok since we are still checking for `GROUPAROO_DISABLED_EXPORTS` within `records:reEnqueueStuckExportingRecords`, but wanted to note the change.

ui/ui-components/components/import/List.tsx
```diff
@@ -171,16 +168,6 @@ export default function ImportList(props) {
...
-                    <br /> Processed:{" "}
-                    {_import.processedAt
-                      ? formatTimestamp(_import.processedAt)
-                      : "pending"}
-                  </td>
-                  <td>
-                    <ImportRecordPropertiesDiff _import={_import} />
-                  </td>
-                  <td>
-                    <ImportGroupsDiff _import={_import} groups={groups} />
```
pedroslopez, line 183: Hmm... I wonder if writing less data is worth losing the visibility/traceability of changes 🤔

core/src/tasks/record/export.ts
evantahler: Missed that one, thanks! e9ade5adcf213ca02206264c909bc77fdcde1168

ui/ui-components/components/import/List.tsx
evantahler: That's the crux of the whole PR - @bleonard and I came to the conclusion that this data is mostly redundant.  We keep change timestamps on each RecordProperty now so you can see when values were checked & changed.  GroupMembers have similar timestamps.  What's lost is data for the negative case (eg: I used to be in this group, and now I'm not).  If we do want to bring that data back at some point we can, but I believe it belongs in some other format, like a new "RecordSnapshots" table - something that doesn't live in the critical import/export path whose writes would slow things down.

We do want to continue to keep this information on the export, because that's actually used to compute the diffs about what we send on the way out, and is what users care more about.

core/src/migrations/000100-removeImportData.ts
```diff
@@ -0,0 +1,60 @@
...
+    await queryInterface.addColumn("records", "readyToExport", {
+      type: DataTypes.BOOLEAN,
+      allowNull: false,
+      defaultValue: false,
+    });
+
+    await queryInterface.changeColumn("records", "readyToExport", {
+      type: DataTypes.BOOLEAN,
+      allowNull: false,
+    });
```
evantahler: Should probably make the new readyToExport column first, and populate the initial values from the old import columns - readyToExport=true if processedAt IS NULL

core/src/tasks/record/export.ts
evantahler: That's a really good point - maybe the `processing` state needs to remain for the imports for this reason

evantahler:
@pedroslopez please give this another look! I've rolled back a number of the previous changes:
* Imports still have a processing state - this fixes your comment about filtering down the collection of imports we check for meta destinations
* We don't need the record.readyToExport column based on the above

That said, I still think we'll get a good speedup here as we are still writing less, and we don't need to read imports at all in the  `records:makeExports` task (just a bulk write to update the sate & timestamps)

core/__tests__/tasks/record/makeExports/makeExports.ts
```diff
@@ -0,0 +1,210 @@
...
+    enableTestPlugin: true,
+    disableTestPluginImport: true,
+  });
+  beforeEach(async () => await api.resque.queue.connection.redis.flushdb());
+  beforeAll(async () => await helper.factories.properties());
+
+  beforeAll(async () => {
+    source = await Source.findOne();
+    schedule = await helper.factories.schedule(source);
+  });
```
krishnaglick: Nitpick: You can combine `beforeAll`. I'm not sure what logic jest uses to order them.

core/src/tasks/record/export.ts
```diff
@@ -29,49 +33,30 @@ export class RecordExport extends RetryableTask {
...
-            .map((d) => d.id)
-            .includes(_import.data?._meta?.destinationId)
-        ) {
-          const destination = await Destination.scope(null).findOne({
-            where: { id: _import.data._meta.destinationId },
-          });
-          if (destination) destinations.push(destination);
-        }
+    // check for explicit destinations to export to from each import
+    for (const i in imports) {
```
krishnaglick: Nitpick: I think you can do `for import of imports` and reference it directly instead of `_import`

evantahler:
Thanks @krishnaglick! The nits have been picked.

krishnaglick (approved):

pedroslopez (approved):
👍🏼

PR #2930, opened 2022-02-01
Update `sequelize` and `ioredis-mock` to latest versions

evantahler:
Minor code changes were required for each of these plugins.  
1) We need to ignore a modified return type from a `SELECT COUNT` call
2) Simplify the redis connection options now that `ioredis` and `ioredis-mock` have the same constructor signatures
3) Removed a no-longer-used 'totals' action

### Development

### Impact

### Code review


core/src/actions/totals.ts
```diff
@@ -79,15 +79,14 @@ export class TotalsAction extends AuthenticatedAction {
...
-        const dateKey = Object.keys(row).find((r) => r !== "count");
-        return {
-          date: dateString(row[dateKey]),
-          count: row.count,
-        };
-      });
+    ).map((row) => {
+      const dateKey = Object.keys(row).find((r) => r !== "count");
+      return {
+        // @ts-ignore
```
edmundito: How come this needs to be ignored?

edmundito (approved):

core/src/actions/totals.ts
evantahler: because we added a `GROUP BY` above.  The new Seqeulize type fixed one thing (now `.count()` calls can return more than one row), but they broke another by assuming that the only 'column' returned is `"count"` .  In our case, we are grouping by by the date of the createdAt column, so the response actually looks something like:

```
[
{agg: 2020-01-01, count : 50},
{agg: 2020-01-02, count : 51},
]
```

evantahler:
@edmundito I took another look at that totals action, and I don't think we are using it anymore!  I've removed it entirely :D 

Tests re-running now

PR #2931, opened 2022-02-02
Remove `Log` model and table

### LINKED ISSUE #2870: [Feature]: Performance Improvements

evantahler:
This PR removes the `Log` model and table within Grouparoo.  This table is rarely used and we've moved any interesting information to timestamps on other models.  Not writing to this (large) table will speed things up!

### Development

### Impact

### Code review


core/src/models/DestinationGroupMembership.ts
```diff
@@ -8,13 +8,13 @@ import {
...
 import { Op } from "sequelize";
-import { LoggedModel } from "../classes/loggedModel";
 import { Group } from "./Group";
 import { Destination } from "./Destination";
 import { APIData } from "../modules/apiData";
+import { CommonModel } from "../classes/commonModel";
 
 @Table({ tableName: "destinationGroupMemberships", paranoid: false })
-export class DestinationGroupMembership extends LoggedModel<DestinationGroupMembership> {
+export class DestinationGroupMembership extends CommonModel<DestinationGroupMembership> {
```
teallarson, line 17: I see a handful of previously logged models that were not given the broadcasting hook, like this one.  Was the thought that those logs aren't as necessary?  That the information can be inferred or accessed somewhere else?

core/__tests__/models/run/run.ts
```diff
@@ -252,7 +243,7 @@ describe("models/run", () => {
         await GroupMember.create({ groupId: group.id, recordId: recordC.id });
 
         await run.determinePercentComplete();
-        expect(run.percentComplete).toBe(70);
+        expect(run.percentComplete).toBe(99);
```
krishnaglick, line 246: What caused this change?

core/src/classes/codeConfig.ts
```diff
@@ -217,7 +217,7 @@ export function validateConfigObjectKeys(
  * Log a Sequelize Model
  */
 export function logModel(
-  instance: LoggedModel<unknown> & {
+  instance: CommonModel<unknown> & {
```
krishnaglick: Just a thought:
`logModel<T = unknown>(instance: CommonModel<T>`

core/src/migrations/000101-removeLogsTable.ts
```diff
@@ -0,0 +1,11 @@
+import Sequelize from "sequelize";
+
+export default {
+  up: async (queryInterface: Sequelize.QueryInterface) => {
+    await queryInterface.dropTable("logs");
+  },
+
+  down: async () => {
+    throw new Error("irreversible migration");
```
krishnaglick, line 9: Should this throw, or just log? Not sure if we have a `run all down` type command.

core/src/modules/modelName.ts
```diff
@@ -1,7 +1,7 @@
 import { Model } from "sequelize-typescript";
-import { LoggedModel } from "../classes/loggedModel";
+import { CommonModel } from "../classes/commonModel";
 
-export function modelName(instance: Model | typeof LoggedModel) {
+export function modelName(instance: Model | typeof CommonModel) {
```
krishnaglick: Not sure this adds any value but:
```suggestion
export function modelName<T>(instance: Model | CommonModel<T>) {
```

core/src/modules/ops/runs.ts
```diff
@@ -196,37 +195,22 @@ export namespace RunOps {
       });
 
       if (totalGroupMembers === 0 && membersAlreadyUpdated === 0) return 0;
+      if (group.state === "deleted") return 99;
```
krishnaglick, line 198: What's the thought process behind this?

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -203,24 +199,29 @@ export namespace helper {
...
+        try {
+          require("@grouparoo/core/src").plugin.mountModels();
+        } catch (error) {}
+
+        if (options.resetSettings) await helper.resetSettings();
+        if (options.enableTestPlugin) helper.enableTestPlugin();
+        if (options.disableTestPluginImport) helper.disableTestPluginImport();
+      } catch (e) {
+        console.error(e); // this produces better logs from sequelize
+        throw e;
```
krishnaglick, line 223: Pretty sure none of this needs to go into fork-process function.

core/__tests__/models/run/run.ts
evantahler: This was the only logical place in the code that used data in the Log table - when deleting a group, we used the previous size from the log to know what % we were done.  Now we can't do that, but I didn't think it was a valuable feature.  So... we are always 99% done :D

core/src/migrations/000101-removeLogsTable.ts
evantahler: We've got a couple of these already.  They mainly appear when you can't recover the data you lost.  I'm on the fence about this one because the log table was always meant to be ephemeral, so we could possibly just re-create an empty table and that could be OK

core/src/modules/ops/runs.ts
evantahler: This was the only logical place in the code that used data in the Log table - when deleting a group, we used the previous size from the log to know what % we were done.  Now we can't do that, but I didn't think it was a valuable feature.  So... we are always 99% done :D

core/src/classes/codeConfig.ts
evantahler: Tighter types in 5ba6c3fb90b1258169c3244a25fd35e20ad10711

core/src/modules/modelName.ts
evantahler: Since T will be optional, might as well!

core/src/models/DestinationGroupMembership.ts
evantahler: Actually, I don't think we need *any* of them any more! 

These broadcast methods send changes about the model over the websocket to the browser client.  In the olden days we used these to update the badges on the left nav - "there are 3 runs running".  But, now we do all of this with the status samples... I don't think we need them at all!

`useRealtimeStream` is how the frontend chooses what to subscribe to, and we currently only listen to CLI messages and system:status

evantahler: removed via 62accc4ea

core/src/modules/ops/runs.ts
evantahler: I think to get this working again we would need to add a column to the groups table to store `lastRunGroupMemberCount` or something like that

krishnaglick (approved):

teallarson (approved):

PR #2933, opened 2022-02-02
Fix error handling issues in UI Client

edmundito:
## Change description

This change fixes issues that came up after the changes in #2846.

- Ensures that requests land on the error handler nicely
- Fixes issues with error handler fighting with hydration errors. Hydration error for server-side and error handler for browser side
- Ensure that user gets redirected on access failures from browser or pages that use `getIntialProps`
- Removes custom implementations of hydration errors. Tested that these are handled by the default hydration error.

### Development

### Impact

For security reasons I tested that a few cases when the user is logged out:

- Loading a page implemented with getServerSideProps
- Loading a page implemented with getInitialProps
- Making an async request (e.g. clicking on Run all schedules button)

### Code review

### DIFF
--- a/ui/ui-components/client/client.ts
+++ b/ui/ui-components/client/client.ts
@@ -171,6 +171,7 @@ export class Client {
 
   private async sendRequest<Response>(
     config: AxiosRequestConfig,
+    options: Partial<ClientRequestOptions>,
     unlock?: ClientCacheGetObject["unlock"]
   ) {
     try {
@@ -195,6 +196,11 @@ export class Client {
 
       if (error.response && error.response.data && error.response.data.error) {
         if (this.redirectOnAccessError(error.response.data.error)) {
+          if (options.errorHandler && error.response.data.error.message) {
+            options.errorHandler.set({
+              message: error.response.data.error.message,
+            });
+          }
           return new Promise<Response>(() => {});
         }
 
@@ -205,6 +211,7 @@ export class Client {
 
         const newError = new Error(err);
         newError["code"] = error.response?.data?.error?.code;
+
         throw newError;
       } else {
         throw error;
@@ -274,7 +281,7 @@ export class Client {
         this.cache.clear();
       }
 
-      return await this.sendRequest(config, unlock);
+      return await this.sendRequest(config, options, unlock);
     } catch (error) {
       if (options.errorHandler) {
         options.errorHandler.set({ message: error });
--- a/ui/ui-components/client/client.ts
+++ b/ui/ui-components/client/client.ts
@@ -6,7 +6,6 @@ import Axios, {
 } from "axios";
 import type { IncomingMessage, ServerResponse } from "http";
 
-import { isBrowser } from "../utils/isBrowser";
 import PackageJSON from "../package.json";
 import { errorHandler } from "../eventHandlers";
 import type { AppContext } from "next/app";
@@ -14,11 +13,17 @@ import type { GetServerSidePropsContext, NextPageContext } from "next";
 import { getRequestContext } from "../utils/appContext";
 import type { ErrorHandler } from "../eventHandlers/errorHandler";
 
-interface ClientCacheObject {
+interface ClientCacheObject<T = any> {
   locked: boolean;
-  data: any;
+  data: T;
   createdAt: number;
 }
+
+interface ClientCacheGetObject<T = any> {
+  cacheData: T;
+  unlock: (data?: T) => void;
+}
+
 export class ClientCache {
   cache: { [key: string]: ClientCacheObject };
   ttl: number;
@@ -34,7 +39,7 @@ export class ClientCache {
     });
   }
 
-  async get(key: string) {
+  async get<T>(key: string): Promise<ClientCacheGetObject<T>> {
     if (!this.cache[key]) {
       this.cache[key] = { locked: false, data: undefined, createdAt: 0 };
     }
@@ -55,7 +60,7 @@ export class ClientCache {
       this.cache[key] = { locked: true, data: undefined, createdAt: 0 };
     }
 
-    const unlock = (data) => {
+    const unlock = (data?: T) => {
       this.cache[key] = {
         data,
         createdAt: new Date().getTime(),
@@ -83,9 +88,15 @@ export class Client {
   private serverToken = process.env.SERVER_TOKEN;
   private cache = new ClientCache();
 
-  private static readonly optionDefaults: ClientRequestOptions = {
+  private static readonly defaultOptions: ClientRequestOptions = {
     useCache: true,
     errorHandler,
+  };
+
+  private static readonly defaultHeaders: AxiosRequestHeaders = {
+    Accept: "application/json",
+    "Content-Type": "application/json",
+    "X-Grouparoo-Client": `${PackageJSON.name}-v${PackageJSON.version}`,
   };
 
   constructor(
@@ -101,74 +112,10 @@ export class Client {
     }
   };
 
-  public request = async <Response = any>(
-    verb: Method = "get",
-    path: string,
-    data: AxiosRequestConfig["data"] = {},
-    options: Partial<ClientRequestOptions> = {}
-  ): Promise<Response> => {
-    options = { ...Client.optionDefaults, ...options };
-
-    const headers: AxiosRequestHeaders = {
-      Accept: "application/json",
-      "Content-Type": "application/json",
-      "X-Grouparoo-Client": `${PackageJSON.name}-v${PackageJSON.version}`,
-    };
-
-    const { req } = this.getRequestContext();
-    if (req?.headers?.cookie) {
-      headers["X-GROUPAROO-SERVER-TOKEN"] = this.serverToken;
-      headers["cookie"] = req?.headers?.cookie;
-      options.useCache = false; // do not ever cache responses on the server
-    }
-
-    const config: AxiosRequestConfig = {
-      params: null,
-      data: null,
-      url: `${this.webUrl}/api/${this.apiVersion}${path}`, // path comes with a leading "/"
-      withCredentials: true,
-      method: verb.toLowerCase() as Method,
-      headers,
-    };
-
-    data.csrfToken = this.csrfToken();
-
-    if (data) {
-      for (const i in data) {
-        if (data[i] === null || data[i] === undefined) {
-          delete data[i];
-        }
-      }
-
-      if (data.file || data._file) {
-        delete config.headers;
-        let dataForm = new FormData();
-        for (const i in data) {
-          dataForm.append(i, data[i]);
-        }
-        data = dataForm;
-      }
-
-      if (config.method === "get") {
-        config.params = data;
-      } else {
-        config.data = data;
-      }
-    }
-
-    let unlock: Function;
-    if (config.method === "get" && options.useCache) {
-      const { cacheData, unlock: _unlock } = await this.cache.get(
-        config.url + JSON.stringify(data)
-      );
-      unlock = _unlock;
-      if (cacheData) {
-        return cacheData;
-      }
-    } else {
-      this.cache.clear();
-    }
-
+  private async sendRequest<Response>(
+    config: AxiosRequestConfig,
+    unlock?: ClientCacheGetObject["unlock"]
+  ) {
     try {
       const response: AxiosResponse<Response & { error?: Error }> = await Axios(
         config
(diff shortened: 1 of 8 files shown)


edmundito:
Two open questions:

Does `checkForLoggedIn` also need to be restored if there is an error?
https://github.com/grouparoo/grouparoo/blob/9addf3514ddff985e38eed1f7a5739e5767fcf56/ui/ui-components/client/client.ts#L83-L110

Do we need an error handler on all pages or some pages?

Currently, all pages will have an assigned error handler because it's default right in the `Client` object.

In the previous iteration, not all pages passed the error handler to the previous `useAPI` hook. One example is in the property edit page where instead of having `useApi(ctx, errorHandler)` we just have `useApi(ctx)` so there is no custom error handler and it will throw an error if something is off.

https://github.com/grouparoo/grouparoo/blob/9addf3514ddff985e38eed1f7a5739e5767fcf56/ui/ui-components/pages/model/%5BmodelId%5D/source/%5BsourceId%5D/property/%5BpropertyId%5D/edit.tsx#L788-L790

ui/ui-components/client/client.ts
```diff
@@ -195,15 +142,85 @@ export class Client {
...
+        unlock = _unlock;
+      } else {
+        this.cache.clear();
+      }
+
+      return await this.sendRequest(config, unlock);
+    } catch (error) {
+      if (options.errorHandler) {
+        options.errorHandler.set({ message: error });
+        return {} as Response;
```
evantahler, line 260: Confirming that returning an empty object was what the old useApi hook used to do:

https://github.com/grouparoo/grouparoo/blob/161da0ce3daeade174361526551f2a40c6cc490b/ui/ui-components/hooks/useApi.ts#L58-L59

evantahler (commented):
Knowing what to do for `checkedForLogged` depends on what the ErrorHandler is for server-side requests.  On the server, we want any request in either getServerProps or getInitialProps to throw so we can catch it with out new [withServerErrorHandler](https://github.com/grouparoo/grouparoo/blob/main/ui/ui-components/utils/withServerErrorHandler.ts) utility.  I can't sort out what the error handler is on the server-side... I **think** `options.errorHandler` will be undefined. Can you confirm?

I've got to imagine the page you called out (https://github.com/grouparoo/grouparoo/blob/9addf3514ddff985e38eed1f7a5739e5767fcf56/ui/ui-components/pages/model/%5BmodelId%5D/source/%5BsourceId%5D/property/%5BpropertyId%5D/edit.tsx#L788-L790) that doesn't use a custom error handler is an oversight.  I can imagine a use case on a browser-side page that wants to handle errors in a custom way, but this is `getInitialProps` - there's no action the user can take here to correct an error

ui/ui-components/client/client.ts
```diff
@@ -83,92 +89,90 @@ export class Client {
   private serverToken = process.env.SERVER_TOKEN;
   private cache = new ClientCache();
 
-  private static readonly optionDefaults: ClientRequestOptions = {
+  private static readonly defaultOptions: ClientRequestOptions = {
     useCache: true,
-    errorHandler,
+    errorHandler: isBrowser() ? errorHandler : undefined,
```
edmundito, line 93: The browser side seems better fit to use the `errorHandler`. The rest of the pages can use the `hydrationError` component.

ui/ui-components/client/client.ts
edmundito: Yep, this is the same as it was before

ui/ui-components/components/Navigation.tsx
```diff
@@ -177,7 +177,7 @@ export default function Navigation(props) {
           </span>
         </div>
 
-        {!navigationMode.includes("unauthenticated") && (
+        {!navigationMode?.includes("unauthenticated") && (
```
edmundito, line 180: These were causing the app to crash when there was an error. However, with the hydration error issues resolved this may no longer be a problem.

ui/ui-components/client/client.ts
```diff
@@ -190,20 +194,94 @@ export class Client {
       }
 
       if (error.response && error.response.data && error.response.data.error) {
+        if (this.redirectOnAccessError(error.response.data.error)) {
+          return new Promise<Response>(() => {});
```
edmundito, line 176: This stops the page from resolving anything else while the page just redirects to where it needs to go.

ui/ui-components/pages/model/[modelId]/destination/[destinationId]/data.tsx
```diff
@@ -858,41 +850,38 @@ export default function Page(props) {
...
 
-    const exportArrayPropertiesResponse = await client.request(
+  const { exportArrayProperties } =
+    await client.request<Actions.DestinationExportArrayProperties>(
       "get",
       `/destination/${destinationId}/exportArrayProperties`
     );
-    exportArrayProperties = exportArrayPropertiesResponse.exportArrayProperties;
-  } catch (error) {
-    hydrationError = error.toString();
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/destination/[destinationId]/data.tsx:894]: This custom hydrationError was causing a lot of trouble so now it just falls back to the hydration error set by the next app.

ui/ui-components/pages/model/[modelId]/overview.tsx
```diff
@@ -50,8 +50,6 @@ const Page: NextPage<Props> = ({
     [properties]
   );
 
-  if (!model) return <Alert variant="warning">Model not found</Alert>;
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/overview.tsx:53]: This is no longer needed because we now fall into the hydration error when requests in getServerSideProps fail

ui/ui-components/client/client.ts
```diff
@@ -83,92 +89,91 @@ export class Client {
...
       res?: ServerResponse;
     } = () => ({})
   ) {}
 
-  private csrfToken = () => {
-    if (globalThis?.localStorage) {
-      return window.localStorage.getItem("session:csrfToken");
-    }
-  };
+  private redirectOnAccessError({ code }: { code: string }): boolean {
```
pedroslopez, line 110: I feel like there should be some central definition of where these errors redirect to - we now have three different ways of redirecting (browser, `getInitialProps`, `getServerSideProps`)... we should make sure that the behavior is consistent

ui/ui-components/client/client.ts
```diff
@@ -83,92 +89,91 @@ export class Client {
...
-        delete config.headers;
-        let dataForm = new FormData();
-        for (const i in data) {
-          dataForm.append(i, data[i]);
-        }
-        data = dataForm;
+      case "NO_TEAMS_ERROR": {
+        const requestPath = req.url.match("^[^?]*")[0];
+        res.writeHead(302, {
+          Location: `/session/sign-in?nextPage=${requestPath}`,
```
pedroslopez: This one doesn't match what the other two are doing (others go to `/` on `NO_TEAMS_ERROR`)

ui/ui-components/client/client.ts
```diff
@@ -5,20 +5,26 @@ import Axios, {
...
 import { errorHandler } from "../eventHandlers";
 import type { AppContext } from "next/app";
 import type { GetServerSidePropsContext, NextPageContext } from "next";
 import { getRequestContext } from "../utils/appContext";
+import type { NextContextType } from "../utils/appContext";
 import type { ErrorHandler } from "../eventHandlers/errorHandler";
+import { isBrowser } from "../utils/isBrowser";
 
-interface ClientCacheObject {
+interface ClientCacheObject<T = any> {
```
krishnaglick:
```suggestion
interface ClientCacheObject<T = unknown> {
```

ui/ui-components/client/client.ts
krishnaglick: I'm unsure if this is a memory leak since this promise is never resolved.

pedroslopez: should this just `throw` and let the `catch` on the `request` method handle it?
edit: ah i see, it doesn't resolve, nvm

ui/ui-components/client/client.ts
```diff
@@ -5,20 +5,26 @@ import Axios, {
...
   locked: boolean;
-  data: any;
+  data: T;
   createdAt: number;
 }
+
+interface ClientCacheGetObject<T = any> {
```
krishnaglick:
```suggestion
interface ClientCacheGetObject<T = unknown> {
```

ui/ui-components/client/client.ts
```diff
@@ -34,7 +40,7 @@ export class ClientCache {
     });
   }
 
-  async get(key: string) {
+  async get<T>(key: string): Promise<ClientCacheGetObject<T>> {
```
krishnaglick:
```suggestion
  async get<T = unknown>(key: string): Promise<ClientCacheGetObject<T>> {
```

krishnaglick: Requires we change the return to `return { cacheData: this.cache[key]?.data as T, unlock };`

Not sure these changes get us a _ton_ since this is all behind the scenes though.

ui/ui-components/client/client.ts
```diff
@@ -83,92 +89,91 @@ export class Client {
...
-      data: null,
-      url: `${this.webUrl}/api/${this.apiVersion}${path}`, // path comes with a leading "/"
-      withCredentials: true,
-      method: verb.toLowerCase() as Method,
-      headers,
-    };
-
-    data.csrfToken = this.csrfToken();
+    const { type, req, res } = this.getRequestContext();
+    if (!req || !res || (type !== "AppContext" && type !== "NextPageContext")) {
```
krishnaglick: I don't anticipate this value changing, why not just
`type === "ServerSidePropsContext"`

ui/ui-components/client/client.ts
```diff
@@ -83,92 +89,91 @@ export class Client {
...
+    }
 
-    if (data) {
-      for (const i in data) {
-        if (data[i] === null || data[i] === undefined) {
-          delete data[i];
-        }
+    // Only check requests from `Page.getInitialProps` or `NextApp.getInitialProps`
+    // getServerSideProps requests will be handled by `withServerErrorHandler`
+    switch (code) {
```
krishnaglick: `switch`!

ui/ui-components/client/client.ts
```diff
@@ -190,20 +195,100 @@ export class Client {
...
+
         throw newError;
       } else {
-        options.errorHandler.set({ message: error });
         throw error;
       }
     }
+  }
+
+  public request = async <Response = any>(
```
krishnaglick, line 194:
```
  public request = async <Response = unknown>(
```
If we did this it would force typed responses everywhere. Probably outside the scope of this PR, but worth doing.

ui/ui-components/utils/appContext.ts
```diff
@@ -1,11 +1,34 @@
...
+export type NextContextType =
+  | "AppContext"
+  | "NextPageContext"
+  | "ServerSidePropsContext";
+
 function isAppContext(ctx: unknown): ctx is AppContext {
   return !!ctx.hasOwnProperty("ctx");
 }
+
+function getNextContextType(ctx: unknown): NextContextType {
```
krishnaglick: Nitpick: `unknown` here could be `AppContext | NextPageContext | GetServerSidePropsContext`

krishnaglick (commented):
This was super cleanly done! Nice job!!

ui/ui-components/client/client.ts
edmundito: Yes, I was also wondering if it would cause a leak. Basically, it stops the request from resolving so that it can redirect.

For example:

```ts
const onButtonClick = () => {
   setLoading(true)
    // This below never resolve, so the page will stop being updated
   const response = await client.request('get', '/will-get-logged-out')
   setResponse(response)
   setLoading(false)
}
```

ui/ui-components/client/client.ts
edmundito: I was thinking about the same thing. I can massage this a bit more so that they're closer together

ui/ui-components/client/client.ts
evantahler: I would suggest throwing as well to stop processing.  Throwing also stops subsequent requests, especially during page hydration

ui/ui-components/client/client.ts
edmundito: Yep, I gave it a try but definitely outside the scope of the PR. a lot of function are still not specifying the type.

ui/ui-components/client/client.ts
evantahler: I think I'm still confused to why this is needed within the client vs the users of the client.  My understanding is that there are only 2 callers of the client: in-browser react callers or server-side hydration callers

* In all cases of the in-browser react callers, an errorHandler will be present and take care of problems. The errorHandler may want to have some redirect logic, but the client doesn't need it
* In the cases of the server-side hydrators, either
  1. `getInitialProps` - throwing will redirect the user to the `_error` page where we provide a "sign in" button for AUTH* error codes.  We are deprecating `getInitialProps`, so this can go away soon
  2. `getServerSideProps` - we wrap all uses of `getServerSideProps` with `withServerErrorHandler` which handles the redirect

krishnaglick: I think there's a case where the react renderer on the server may make requests, but I'm not 100% on that.

evantahler (approved):
Approving as-is to fix the bugs.   We've got another story coming soon (https://www.pivotaltracker.com/story/show/181142452) to refactor and simplify redirecting and error handing in general.

PR #2938, opened 2022-02-02
Package.json validation tests

teallarson:
## Change description

- Adds schema validation tests for our Package.json files
- Checks for version matches across the repo 
- Checks file contents for items in our "bad words" list

Went with invoking the version checker and spelling checker within child process inside the jest test as the scripts in place did a more efficient job of the tasks then a JavaScript solution would in my opinion.

Also removed the CI jobs that used to execute those scripts, and added one to run the new suite.

### Development

### Impact

### Code review


teallarson:
the flaky test here in models is fixed by #2943 (missing awaits)

jest.setup.js
```diff
@@ -0,0 +1,4 @@
+require("@grouparoo/spec-helper/dist/lib/environment");
+
+// tell `@grouparoo/spec-helper` to load everything from `src` in core, so we can test without re-compiling
+process.env.GROUPAROO_SPEC_HELPER_LOAD_PATH = "src";
```
evantahler: This probably isn't required at the top-level of the monorepo - this is for running the grouparoo app, which we don't need here

package.json
```diff
@@ -13,19 +13,23 @@
     "url": "git+https://github.com/grouparoo/grouparoo.git"
   },
   "devDependencies": {
+    "@grouparoo/core": "0.8.1-alpha.0",
+    "@grouparoo/spec-helper": "0.8.1-alpha.0",
```
evantahler: Having the top of the monorepo depend on something within it might get wonky.  I don't think this is needed if `jest.setup` is removed

tools/merger/data/ci/root/job.yml.template
```diff
@@ -0,0 +1,29 @@
...
+      - image: redis:latest
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+      - image: circleci/postgres:9
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+        environment:
+          POSTGRES_PASSWORD: password
```
evantahler: I don't think the root tests need redis or postgres to run

tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  //todo: this is copied from core, but likely not completely correct....
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./dist",
+    "allowJs": true,
+    "module": "commonjs",
+    "target": "es2019",
+    "esModuleInterop": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true
```
evantahler: These last 2 probably aren't needed, the rest looks good!

tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  },
+  "include": ["./src/**/*"],
```
evantahler: There is no `src` in the root project, so maybe this isn't needed?

jest.config.js
```diff
@@ -0,0 +1,18 @@
+const { helper } = require("@grouparoo/spec-helper");
```
evantahler: This can probably be skipped, and just set the testTimeout manually - when we don't need this dependency

jest.config.js
```diff
@@ -0,0 +1,18 @@
...
+
+module.exports = {
+  testEnvironment: "node",
+  testTimeout: helper.defaultTime,
+  maxWorkers: "50%",
+  setupFiles: ["<rootDir>/jest.setup.js"],
+  transform: {
+    "^.+\\.ts?$": "ts-jest",
+  },
+  testPathIgnorePatterns: [
```
evantahler: I don't think any of these directories exist in the root project

__tests__/packages.ts
```diff
@@ -0,0 +1,119 @@
...
+  test("package versions match across the monorepo", async () => {
+    const checkVersions = await new Promise<void>((res, reject) => {
+      childProcess.exec(
+        "./node_modules/.bin/syncpack list-mismatches",
+        { cwd: path.resolve(__dirname + "/../") },
+        (error, _, stderr) => {
+          if (error || stderr) {
+            reject(error || stderr);
+          }
+          res();
```
evantahler: without a return statement, this will both resolve and reject this promise.  either `return reject(...)` or do an if/else

evantahler: What does it look like when this promise throws?  The cli usage of syncpack did a good job of showing us which packages are mismatched... does that come through in the jest error?

__tests__/schemaGeneric.json
```diff
@@ -0,0 +1,244 @@
...
+          "access": "public"
+        }
+      ],
+      "required": ["access"],
+      "properties": {
+        "access": {
+          "$id": "#/properties/publishConfig/properties/access",
+          "type": "string",
+          "examples": ["public"]
+        }
```
evantahler: This is confirming that `publishConfig.access` is a string in the package.json, but it might not be `"public"` - can we confirm that it is literally "public" (except for ui-enterprise)?

__tests__/schemaGeneric.json
```diff
@@ -0,0 +1,244 @@
...
+      "examples": [
+        {
+          "node": ">=12.0.0 <17.0.0"
+        }
+      ],
+      "required": ["node"],
+      "properties": {
+        "node": {
+          "$id": "#/properties/engines/properties/node",
+          "type": "string",
```
evantahler, line 59: Same note as above - can we check the literal value of of this property vs that it is any string?

__tests__/schemaGeneric.json
```diff
@@ -0,0 +1,244 @@
...
+          "prepare": "./bin/build",
+          "dev": "./bin/dev",
+          "pretest": "npm run lint && npm run prepare",
+          "test": "jest",
+          "lint": "prettier --check src __tests__",
+          "docs": "typedoc --out docs --theme default src/index.ts",
+          "import": "./api/bin/import",
+          "watch": "tsc --watch"
+        }
+      ],
```
evantahler: this big list of examples doesn't seem to gain us anything

__tests__/schemaGeneric.json
```diff
@@ -0,0 +1,244 @@
...
+      "properties": {},
+      "additionalProperties": true
+    },
+    "dependencies": {
+      "$id": "#/properties/dependencies",
+      "type": "object",
+      "examples": [
```
evantahler: this big list of examples doesn't seem to gain us anything

__tests__/wordCheck
```diff
@@ -30,13 +31,9 @@ for word in "${BAD_WORDS[@]}"; do
...
     printf " ✅\n"
+  
   else
-    printf "🚫\n"
-    echo "> instances of $word found:"
-    echo "---------------"
-    echo "$MATCHES"
-    echo "---------------"
-    exit 1
+    echo "$word: $MATCHES" 1>&2
```
evantahler, line 36: Nice! Redirecting the echo from `stdout` to `stderr`.  This will get the promise in the jest test to reject, but it still common practice to explicitly `exit 1` to show the program has "failed" - please add that back.  This will also end the loop we are in early

__tests__/packages.ts
```diff
@@ -0,0 +1,119 @@
...
+  test("the monorepo does not contain banned words", async () => {
+    const checkBannedWords = await new Promise<void>((res, reject) => {
+      childProcess.exec(
+        "./wordCheck",
+        { cwd: path.resolve(__dirname) },
+        (error, _stuff, stderr) => {
+          if (error || stderr) {
+            reject(error || stderr);
+          }
+          res();
```
evantahler: Same as below - an early reject without a return will still also resolve later

evantahler (changes requested):
Good setup! I like the consistency of having a top-level test suite, rather then some random bash scripts we run.

I think there are a few topics to look into:
1. Don't import/require anything from grouparoo/core or grouparoo/spec-helper.  They shouldn't be needed and I'm worried that will have an adverse effect on the publishing pipeline
2. Can JSON-schema assert the contents of the JSON rather than the types?  For example, we don't really care that publish config is /any/ string in the packages, we really do want to assert that it is "public".  I think that's a better reflection of the goal.

__tests__/schemaGeneric.json
teallarson: Yes.  I did set that on the plugin-specific schema as I assumed we only guaranteed that the plugins would have `publishConfig`.  Are you saying we want that on every package.json except `ui-enterprise`?

evantahler: Sorry about that! I'm more interested in confirming that most packages have a literal `publishConig.access = 'public'` rather than any string, which I think is what this assertion does.  I would prefer a validation/test that checks for that value directly vs that any string is present.  

The nuance is that this is true for all packages except ui-enterprise.

__tests__/packages.ts
krishnaglick: I think doing `done(error || stderr)` and not wrapping all this in a promise may be the best way to handle it.

__tests__/packages.ts
krishnaglick: `done(error || stderr)` pattern worth considering like I mentioned above

teallarson: I think what we actually what here is `return reject(stdout)` since that's where the useful info is sent.

__tests__/schemaGeneric.json
teallarson: Yep!

__tests__/packages.ts
teallarson: Went with the `done()` pattern for both the version and license checkers and piped out the most meaningful messaging for each checker (license checker shows it in `stderr`, but the version checker shows it in `stdout`)

teallarson:
Changes since last review:
- Tighter assertions for the `publishConfig` (Implemented by adding a third schema for `restricted` access configs and switching some logic for which schema to use.  The only packages using the `schemaGeneric` now are ones that do not require a `publishConfig`.)
- Tighter assertion for `engine`
- TSConfig and CI Job cleanup
- Tests for version and license checking fail correctly now, and display helpful errors (ie: what package it failed on and why)

__tests__/packages.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  packageFiles.forEach((pkgFile) => {
+    const pkgJson = JSON.parse(fs.readFileSync(pkgFile, "utf-8"));
+    const nameAndVersion = `${pkgJson.name}@${pkgJson.version}`;
+    excludedPackages.push(nameAndVersion);
+  });
+
+  const pkgFile =
+    "/Users/teal/workspace/grouparoo/grouparoo/plugins/@grouparoo/customerio/package.json";
+  describe.each(packageFiles)(`%s`, (pkgFile) => {
```
evantahler: I would not have expected this to pass on CI - there's no `/teal` folder there!  This should be fixed, but also... why did it pass?

tools/merger/data/ci/root/job.yml.template
evantahler: Bump! I don't think you need the docker images other than {{{docker_image}}}

evantahler (changes requested):

__tests__/packages.ts
teallarson: Ah.  This was left over from debugging, oops.... it's still passing because the `pkgFile` referenced below is from the loop, not the one declared up here.

evantahler: Ah! Then 👍 once this is removed!

evantahler (approved):

PR #2939, opened 2022-02-03
table sources provide table options as sorted typeahead

pedroslopez:
## Change description

Provides better experience for configuring the source when users have lots of tables in their database.
Also fixes typeaheads on the source edit page to play nicely alongside react-hook-form.

image

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
+++ b/ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
@@ -338,7 +338,6 @@ const Page: NextPage<Props> = ({
   };
 
   const updateOption = async (optKey, optValue) => {
-    console.log("update option", optKey, optValue);
     const _source = { ...source };
     _source.options[optKey] = optValue;
     _source.mapping = {};
@@ -424,6 +423,7 @@ const Page: NextPage<Props> = ({
                                     id="typeahead"
                                     labelKey="key"
                                     disabled={loading || loadingOptions}
+                                    inputProps={{ required: opt.required }}
                                     options={connectionOptions[
                                       opt.key
                                     ]?.options.map((k, idx) => {
@@ -551,6 +551,7 @@ const Page: NextPage<Props> = ({
                               placeholder={opt.placeholder}
                               name={`source.options.${opt.key}`}
                               {...register(`source.options.${opt.key}`, {
+                                shouldUnregister: true,
                                 onChange: (e) =>
                                   updateOption(
                                     e.target.id.replace("_opt~", ""),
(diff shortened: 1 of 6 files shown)


ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -414,51 +414,65 @@ const Page: NextPage<Props> = ({
...
+                                      ];
+                                    }}
+                                    defaultSelected={
+                                      source.options[opt.key]
+                                        ? [source.options[opt.key]]
+                                        : undefined
+                                    }
+                                    onChange={(selected) => {
+                                      onChange(selected[0]?.key);
+                                      updateOption(opt.key, selected[0]?.key);
```
Comment (pedroslopez) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx]: calling both `react-hook-form`'s `onChange` and our `updateOption`

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -414,51 +414,65 @@ const Page: NextPage<Props> = ({
...
-                                        Descriptions:{" "}
-                                        {opt.descriptions.join(", ")}
-                                      </em>
-                                    ) : null}
-                                  </small>,
-                                ];
+                            <Controller
+                              control={control}
+                              name={`source.options.${opt.key}`}
+                              render={({ field: { onChange } }) => {
```
Comment (pedroslopez) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx]: Because `Typeahead`'s props aren't compatible with the props `react-hook-form`'s `register` function returns, we need to wrap it in a `Controller` with a custom render function

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -537,6 +551,7 @@ const Page: NextPage<Props> = ({
                               placeholder={opt.placeholder}
                               name={`source.options.${opt.key}`}
                               {...register(`source.options.${opt.key}`, {
+                                shouldUnregister: true,
```
Comment (pedroslopez) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:511]: On the first few renders until we get back the connectionOptions response, it falls back to a disabled text input. This was causing problems because register by default does not unregister on unmount, so sometimes the custom `onChange` function passed here stuck around and was still getting called whenever we changed the value on the typeahead. Passing `shouldUnregister: true` here correctly cleans up the old onChange on unmount as expected.

krishnaglick (commented):
Do you think you could write a test to test this? I feel like this is a touchy part of the codebase and could use the support.
Maybe it's worth putting this logic into its own component so it's more easily testable?

pedroslopez:
@krishnaglick separated out this react-hook-form + react-bootstrap-typeahead integration into its own component and added some tests for it. I was able to reuse it in both the Source Edit and App Edit pages- the destination edit page also has a typeahead in it, but the page doesn't use react-hook-form. We can probably move to using this component later when we go all-in react-hook-form.

krishnaglick (approved):
Really love what you did here, nice job!

PR #2941, opened 2022-02-03
Render sample record card on server side

rwfeather:
## Change description

In order to have the correct scroll position when a user navigates to `#destinations`, we're moving the sample record into server side rendering (when possible).

This includes moving some small amount of user state from localstorage to a cookie based solution (accessible in both client and server code)

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/components/record/SampleRecordCard.tsx
+++ b/ui/ui-components/components/record/SampleRecordCard.tsx
@@ -61,6 +61,8 @@ const cookies = new Cookies();
 
 const getCachedSampleRecordId = (modelId: string): string => {
   if (isBrowser()) {
+    // Migrate existing localstorage data into cookie
+    // TODO: remove the migration in a later release
     const localStorageRecordId = globalThis.localStorage?.getItem(
       `sampleRecord:${grouparooUiEdition()}:${modelId}`
     );
--- a/ui/ui-components/components/record/SampleRecordCard.tsx
+++ b/ui/ui-components/components/record/SampleRecordCard.tsx
@@ -60,23 +60,9 @@ export const isConfigUI = grouparooUiEdition() === "config";
 const cookies = new Cookies();
 
 const getCachedSampleRecordId = (modelId: string): string => {
-  if (isBrowser()) {
-    // Migrate existing localstorage data into cookie
-    // TODO: remove the migration in a later release
-    const localStorageRecordId = globalThis.localStorage?.getItem(
-      `sampleRecord:${grouparooUiEdition()}:${modelId}`
-    );
-    if (localStorageRecordId) {
-      setCachedSampleRecordId(modelId, localStorageRecordId);
-      globalThis.localStorage?.removeItem(
-        `sampleRecord:${grouparooUiEdition()}:${modelId}`
-      );
-      return localStorageRecordId;
-    } else {
-      return cookies.get(`sampleRecord:${grouparooUiEdition()}`)?.[modelId];
-    }
-  }
-  return undefined;
+  return isBrowser()
+    ? cookies.get(`sampleRecord:${grouparooUiEdition()}`)?.[modelId]
+    : undefined;
 };
 
 const setCachedSampleRecordId = (modelId: string, recordId: string): void => {
@@ -90,7 +76,7 @@ const setCachedSampleRecordId = (modelId: string, recordId: string): void => {
       {
         path: "/",
         sameSite: "strict",
-        maxAge: 60 * 60 * 24 * 14, // two weeks
+        maxAge: 60 * 60 * 24 * 30, // 1 month
       }
     );
   }
@@ -106,6 +92,7 @@ const clearCachedSampleRecordId = (modelId: string): void => {
       cookies.set(`sampleRecord:${grouparooUiEdition()}`, existing, {
         path: "/",
         sameSite: "strict",
+        maxAge: 60 * 60 * 24 * 30, // 1 month
       });
     }
   }
--- a/ui/ui-components/components/record/SampleRecordCard.tsx
+++ b/ui/ui-components/components/record/SampleRecordCard.tsx
@@ -90,6 +90,7 @@ const setCachedSampleRecordId = (modelId: string, recordId: string): void => {
       {
         path: "/",
         sameSite: "strict",
+        maxAge: 60 * 60 * 24 * 14, // two weeks
       }
     );
   }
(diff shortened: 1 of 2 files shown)


ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -46,30 +48,64 @@ export interface SampleRecordCardProps {
...
 const setCachedSampleRecordId = (modelId: string, recordId: string): void => {
-  if (recordId) {
-    globalThis.localStorage?.setItem(
-      `sampleRecord:${grouparooUiEdition()}:${modelId}`,
-      recordId
+  if (isBrowser() && recordId) {
+    const existing = cookies.get<Record<string, string>>(
+      `sampleRecord:${grouparooUiEdition()}`
+    );
+    cookies.set(
```
bleonard, line 73: Does this mean we should switch to only using cookies and stop using local storage?

rwfeather: In this case we've switched to cookies so that both client-side and server-side rendering has access to the user's cached sample record. 

The logic here is to "migrate" existing client side state into the cookie when the user first navigates to the page.

rwfeather: The only other explicit usage of localstorage in our codebase is the `session:csrfToken`, which seems fine.

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -46,30 +48,64 @@ export interface SampleRecordCardProps {
...
+  return undefined;
 };
 
 const setCachedSampleRecordId = (modelId: string, recordId: string): void => {
-  if (recordId) {
-    globalThis.localStorage?.setItem(
-      `sampleRecord:${grouparooUiEdition()}:${modelId}`,
-      recordId
+  if (isBrowser() && recordId) {
+    const existing = cookies.get<Record<string, string>>(
```
evantahler, line 70: What is `existing` already in the cookie that might want to be preserved?  It's it only this component that writes to that cookie?

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -46,30 +48,64 @@ export interface SampleRecordCardProps {
...
 };
 
 const clearCachedSampleRecordId = (modelId: string): void => {
-  globalThis.localStorage?.removeItem(
-    `sampleRecord:${grouparooUiEdition()}:${modelId}`
-  );
+  if (isBrowser()) {
+    const existing = cookies.get<Record<string, string>>(
+      `sampleRecord:${grouparooUiEdition()}`
+    );
```
evantahler, line 89: Similarly here - I think we want to delete the cookie maybe?  That's the ol' set the expiry to something in the past trick?

evantahler: Related - without setting it explicitly, do these cookies never expire?

bleonard: My recollection is that not setting it might delete it when the browser closed, so we should probably set a time.

ui/ui-components/components/record/SampleRecordCard.tsx
rwfeather: The solution we went with (paired on this with @edmundito) is to store cached record ids from multiple models in a single cookie via json, keyed by model id.

So a parsed cookie could look like:

```json
{
    "accounts": "abcdf",
    "users": "defgh",
}
```

The library we have for this (universal-cookie), handles reading and writing the whole object but doesn't seem to have a method for updating it in place. Thus we have to `get -> merge -> set` here.

I could rearrange a bit or rename `existing` to make that clearer.

ui/ui-components/components/record/SampleRecordCard.tsx
rwfeather: And same story here. We aren't clearing the entire cookie, just the data associated with a single model.

I will look into how we should manage the cookie exipiration. Agree that having this potentially be cleaned up on browser close is not ideal.

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -46,30 +48,64 @@ export interface SampleRecordCardProps {
...
+  if (isBrowser()) {
+    const localStorageRecordId = globalThis.localStorage?.getItem(
+      `sampleRecord:${grouparooUiEdition()}:${modelId}`
+    );
+    if (localStorageRecordId) {
+      setCachedSampleRecordId(modelId, localStorageRecordId);
+      globalThis.localStorage?.removeItem(
+        `sampleRecord:${grouparooUiEdition()}:${modelId}`
+      );
+      return localStorageRecordId;
```
evantahler: Is this codepath just for migrating an existing sample record from localStorage to cookie-storage?  If so, I'm not too worried about folks needing to re-select the same record again.  If you do only need this for a migration, please add a TODO so we remember to remove it in a later release

rwfeather: Added a comment and TODO.

I agree the migration may not be 100% necessary. Would you prefer I just remove it while I'm still here?

evantahler: Not that I feel strongly, but but I would vote to remove it - the user experience won't suffer that much, and at the moment, the theme of simplifying branching codepaths on our front-end is something we are focusing on

ui/ui-components/components/record/SampleRecordCard.tsx
```diff
@@ -46,30 +48,67 @@ export interface SampleRecordCardProps {
...
+    const existing = cookies.get<Record<string, string>>(
+      `sampleRecord:${grouparooUiEdition()}`
+    );
+    cookies.set(
+      `sampleRecord:${grouparooUiEdition()}`,
+      { ...existing, [modelId]: recordId },
+      {
+        path: "/",
+        sameSite: "strict",
+        maxAge: 60 * 60 * 24 * 14, // two weeks
```
evantahler: I would suggest this be far longer - at least the 30 days that our sessions last for

evantahler (approved):
Approved, as my notes are suggestions - nothing blocking!

ui/ui-components/components/record/SampleRecordCard.tsx
rwfeather: I was on the fence about including it. Removed!

ui/ui-components/components/record/SampleRecordCard.tsx
rwfeather: I've upped it to 30 days to match the sessions

ui/ui-components/components/record/SampleRecordCard.tsx
rwfeather: Cookie expiration is now being handled with a `max-age`

PR #2943, opened 2022-02-04
Newest Export Lookup is Batched

krishnaglick:
## Change description

The magic of sql

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/destination/plugins/exportRecords.ts
+++ b/core/__tests__/models/destination/plugins/exportRecords.ts
@@ -738,8 +738,8 @@ describe("models/destination - with custom exportRecords plugin", () => {
         errorLevel: null,
       });
 
-      helper.changeTimestamps([exportOne], true, new Date());
-      helper.changeTimestamps(
+      await helper.changeTimestamps([exportOne], true, new Date());
+      await helper.changeTimestamps(
         [exportTwo],
         true,
         new Date(new Date().valueOf() - 100000)
--- a/core/__tests__/models/destination/plugins/exportRecords.ts
+++ b/core/__tests__/models/destination/plugins/exportRecords.ts
@@ -770,6 +770,116 @@ describe("models/destination - with custom exportRecords plugin", () => {
       expect(exportTwo.state).toEqual("canceled");
     });
 
+    test.only("If multiple exports exist for a single record/destination pair, the newest is sent and all older ones are canceled", async () => {
+      const record = await helper.factories.record();
+      const group = await helper.factories.group();
+      await GroupMember.create({ recordId: record.id, groupId: group.id });
+      await destination.updateTracking("group", group.id);
+
+      const exportOne = await Export.create({
+        retryCount: 0,
+        destinationId: destination.id,
+        recordId: record.id,
+        startedAt: null,
+        oldRecordProperties: "{}",
+        newRecordProperties: "{}",
+        oldGroups: "[]",
+        newGroups: "[]",
+        state: "pending",
+        sendAt: new Date(),
+        hasChanges: true,
+        toDelete: false,
+        force: true,
+        exportProcessorId: null,
+        completedAt: null,
+        errorMessage: null,
+        errorLevel: null,
+      });
+
+      const exportTwo = await Export.create({
+        retryCount: 0,
+        destinationId: destination.id,
+        recordId: record.id,
+        startedAt: null,
+        oldRecordProperties: "{}",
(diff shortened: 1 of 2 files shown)


core/src/modules/ops/destination.ts
```diff
@@ -903,15 +902,16 @@ export namespace DestinationOps {
     return { success, error, retryexportIds, retryDelay };
   }
 
-  async function cancelOldExport(oldExport: Export, newExport: Export) {
```
krishnaglick, line 906: We don't have the exact new export now.

tealjulia: Noting that this means the cancel messaging is different now too

core/__tests__/models/destination/plugins/exportRecords.ts
```diff
@@ -738,8 +738,8 @@ describe("models/destination - with custom exportRecords plugin", () => {
         errorLevel: null,
       });
 
-      helper.changeTimestamps([exportOne], true, new Date());
-      helper.changeTimestamps(
+      await helper.changeTimestamps([exportOne], true, new Date());
+      await helper.changeTimestamps(
```
evantahler, line 742: good catch!

core/src/modules/ops/destination.ts
```diff
@@ -926,23 +926,37 @@ export namespace DestinationOps {
...
+          "id"
+        FROM (
+          SELECT
+            "id",
+            ROW_NUMBER() OVER (PARTITION BY "exports"."recordId" ORDER BY "exports"."createdAt" DESC) AS __rownum
+          FROM
+            "exports"
+          WHERE
+            "exports"."destinationId" = ?
+            AND "exports"."id" IN (?)) AS __ranked
```
evantahler: I had no idea you could use Sequelize's string helpers for custom queries! That's powerful.... and there are probably some other places to update to do this.

evantahler: 👍 having the where clause on the inner query vs the outer query

evantahler: I don't think this part is quire right
```sql
AND "exports"."id" IN (?)) AS __ranked
```

We want to find the newest export that isn't necessarily in  `givenExports`
```sql
AND "exports"."recordId" IN (?)) AS __ranked
```

So the values would be 

```
values: [destination.id, givenExports.map(({ recordId }) => recordId)],
```

evantahler: The tests should be updated to catch something like the above.  Have a later export that isn't in the batch being considered.

evantahler (changes requested):

core/src/modules/ops/destination.ts
krishnaglick: Done!

teallarson (approved):

evantahler (commented):
I think the tests are showing that something isn't quite right yet

krishnaglick:
> I think the tests are showing that something isn't quite right yet

Seemed to be an async issue, sorted out now.

core/src/modules/ops/destination.ts
```diff
@@ -926,23 +926,40 @@ export namespace DestinationOps {
...
+        WHERE
+          "__ranked"."__rownum" = 1;`,
+            values: [
+              destination.id,
+              givenExports.map(({ recordId }) => recordId),
+            ],
+          },
+          undefined
+        )
+        .then((exports) => exports[0].map((e) => e.id));
```
evantahler: I think you can skip this `[0]` if you add this kind of thing -> https://github.com/grouparoo/grouparoo/blob/main/core/src/modules/ops/record.ts#L1062-L1065 to get an object per row.

I think what you are getting back now is something like  rows -> exports -> id

core/src/modules/ops/destination.ts
```diff
@@ -926,23 +926,40 @@ export namespace DestinationOps {
     const locks: Awaited<ReturnType<typeof getLock>>[] = [];
 
     try {
+      const exportIds = await Export.sequelize
```
evantahler: It's pedantic, but this SQL is kind of complex - a better variable name might be something like `newestExportIds`

core/src/modules/ops/destination.ts
krishnaglick: Oh, I like this!

evantahler (approved):
If the test are 🟢 I'm 🟢 !

PR #2945, opened 2022-02-05
Cached Models and Shared Model Classes

### LINKED ISSUE #2870: [Feature]: Performance Improvements

evantahler:
## Change description

This PR introduces caching for Apps and Sources, and cleans up the Property cache.  This allows the Grouparoo application to skip a whole bunch of SQL queries.

[BEFORE](https://sentry.io/organizations/grouparoo/performance/grouparoo-development:fce6b0e0b09b4727be9fc31facf6190c/?end=2022-02-03T16%3A44%3A14&project=5673667&query=transaction.duration%3A%3C15m&start=2022-02-02T04%3A22%3A33&transaction=recordProperty%3AimportRecordProperties&unselectedSeries=p100%28%29)

Screen Shot 2022-02-10 at 11 36 10 AM

[AFTER](https://sentry.io/organizations/grouparoo/performance/grouparoo-development:927b64832a9a4b8091e3cac389af54eb/?project=5673667&query=transaction.duration%3A%3C15m&statsPeriod=1h&transaction=recordProperty%3AimportRecordProperties&unselectedSeries=p100%28%29)

Screen Shot 2022-02-10 at 11 36 08 AM

This PR also moves a few methods that were common to all models that extended CommonModel into the CommonModel itself, like `findById`, now that the can extend classes property (thanks @krishnaglick and @bleonard!)

### Things that didn't work:

1. I did not enable the tighter typecheck for sequelize models: 
 ```ts

export abstract class CommonModel<T> extends Model<T, NonAbstract<Columns<T>>> {
// This tighter type is not yet in use because it would require all places where we make a new instance to typecheck.
//  For example, an action that takes state:string would need to coerce that to state: typeof App['state'].
//  That is out of scope for now.
```

2. I had to rollback work to make a new StateTransitioningModel that extended Common Model.  Being unable to define the default value for the column, as a dynamic type, was no good.  Click to Expand

<details>

```ts
import {
  AllowNull,
  BeforeSave,
  DataType,
  Column,
  BeforeValidate,
  BeforeBulkCreate,
import { NonAbstract } from "sequelize-typescript/dist/shared/types";
import { modelName } from "../modules/modelName";
import { CommonModel } from "./commonModel";

export interface StateTransition<
  S extends readonly string[] = readonly string[]
> {
  from: S[number];
  to: S[number];
  checks: ((instance: any) => Promise<any>)[];
}

export type StateMachineModelStatic<M> = (new () => M) &
  NonAbstract<typeof StateMachineModel>;

export abstract class StateMachineModel<
  T extends StateMachineModel<T, S>,
  S extends readonly string[] = readonly string[]
> extends CommonModel<T> {
  static defaultState: string;
  static STATE_TRANSITIONS: StateTransition[];
  static STATES: readonly string[];

  @AllowNull(false)
  @Column(DataType.STRING)
  state: S[number];

  getDefaultSate() {
    const defaultState: T["state"][number] =
      // @ts-expect-error TODO: this is messy
      this.constructor.defaultState ?? "draft";
    return defaultState;
  }

  getStateTransitions() {
    const transitions: StateTransition[] =
      // @ts-expect-error TODO: this is messy
      this.constructor.STATE_TRANSITIONS;
    return transitions;
  }

  @BeforeValidate
  static async applyDefaultState<T extends StateMachineModel<T>>(
    instance: StateMachineModel<T>
  ) {
    // this can't be done via an @Default decorator because we don't have access to the instance, just the Type
    const defaultState = instance.getDefaultSate();
    if (!instance.state) instance.state = defaultState;
  }

  @BeforeBulkCreate
  static async appleDefaultStates<T extends StateMachineModel<T>>(
    instances: StateMachineModel<T>[]
  ) {
    for (const instance of instances) {
      StateMachineModel.applyDefaultState(instance);
    }
  }

  @BeforeSave
  static async confirmStateTransition<T extends StateMachineModel<T>>(
    instance: StateMachineModel<T>
  ) {
    await validateStateTransition<T>(instance);
  }
}

async function validateStateTransition<T extends StateMachineModel<T>>(
  instance: StateMachineModel<T> & {
    _previousDataValues?: { state?: T["state"][number] };
  }
) {
  const klass = modelName(instance);
  const oldState = instance["_previousDataValues"]["state"];
  if (!oldState || instance.state === oldState) return;

  const transitions = instance.getStateTransitions();
  const transition = findTransition(oldState, instance.state, transitions);
  if (!transition) {
    throw new Error(
      `cannot transition ${klass} state from ${oldState} to ${instance.state}`
    );
  }

  for (const i in transition.checks) {
    const check = transition.checks[i];
    await check(instance);
  }
}

function findTransition(
  from: string,
  to: string,
  transitions: StateTransition[]
) {
  return transitions.find((t) => t.from === from && t.to === to);
}
```

</details>

### Development

### Impact

### Code review


evantahler:
It doesn't make that much sense to share the cache methods either, as the setters and getters are going to be different for each model (eg Apps can't filter by modelId)

evantahler:
@krishnaglick and @pedroslopez There are still more places where the cache can be used, but here's the direction I'm going. I timeboxed the TS work, got about 1/2 of what I wanted in, and moved on.  We can add stronger types later

core/__tests__/models/app/modelCache.ts
```diff
@@ -0,0 +1,141 @@
...
+
+      async function makeApp() {
+        app = await helper.factories.app();
+        await helper.sleep(100); // wait for delayed RPC calls
+      }
+
+      afterEach(async () => {
+        try {
+          await app.destroy();
+        } catch {}
```
krishnaglick:
```suggestion
      await app.destroy().catch(() => null);
```
You can do this too!

core/__tests__/models/app/modelCache.ts
```diff
@@ -0,0 +1,141 @@
...
+    test("it will find by id by default", async () => {
+      const found = await AppsCache.findOneWithCache(app.id);
+      expect(found.id).toBe(app.id);
+      expect(found.name).toBe("NEW NAME");
+    });
+
+    test("it will avoid using SQL when a cached app exists", async () => {
+      const cachedApp = AppsCache.instances.find((a) => a.id === app.id);
+      // @ts-ignore
+      cachedApp.__isCached = true;
```
krishnaglick: Nitpick
```suggestion
      (cachedApp as any).__isCached = true;
```

core/__tests__/models/app/modelCache.ts
```diff
@@ -0,0 +1,141 @@
...
+      const found = await AppsCache.findOneWithCache(app.id);
+      // @ts-ignore
+      expect(found.__isCached).toBe(true);
```
krishnaglick: Nitpick
```suggestion
      expect((found as any).__isCached).toBe(true);
```

core/__tests__/models/source/modelCache.ts
```diff
@@ -0,0 +1,152 @@
...
+      let source: Source;
+
+      async function makeSource() {
+        source = await helper.factories.source();
+        await helper.sleep(100); // wait for delayed RPC calls
+      }
+
+      afterEach(async () => {
+        try {
+          await source.destroy();
```
krishnaglick: Could `.catch(() => null);` here as well

core/__tests__/modules/internalRun.ts
```diff
@@ -13,7 +13,7 @@ describe("modules/internalRun", () => {
   });
 
   test("an internal run is created and enqueued", async () => {
-    await internalRun("creator", "abc123");
+    await internalRun("teamMember", "abc123");
```
krishnaglick, line 16: Curious about this change.

core/src/classes/codeConfig.ts
```diff
@@ -217,7 +217,7 @@ export function validateConfigObjectKeys(
  * Log a Sequelize Model
  */
 export function logModel<T>(
-  instance: CommonModel<T> & {
+  instance: CommonModel<unknown> & {
```
krishnaglick: Why `unknown` instead of `T`? Typescript issues?

core/src/classes/commonModel.ts
```diff
@@ -7,9 +7,21 @@ import {
...
+import { NonAbstract } from "sequelize-typescript/dist/shared/types";
 import validator from "validator";
 import * as uuid from "uuid";
-
+import { Op, Attributes } from "sequelize";
+import { config } from "actionhero";
+
+type Columns<T> = Exclude<T, Function>;
+export type CommonModelStatic<M> = (new () => M) &
+  NonAbstract<typeof CommonModel>;
```
krishnaglick, line 18: Worth commenting out these too?

core/src/classes/commonModel.ts
```diff
@@ -70,22 +82,34 @@ export abstract class CommonModel<T> extends Model {
...
-    throw new Error("not implemented");
+  public static async findById<T extends Model>(
+    this: CommonModelStatic<T>,
+    id: string
+  ): Promise<T> {
+    const instance = await this.scope(null).findOne({ where: { id } });
+    const modelName = String(this).match(/class\s(.+)\sextends/)?.[1] ?? "item"; // TODO: It looks like at this point `this` is the stringified representation of the constructor function, literally "class Group extends CommonModel..."
+    if (!instance) throw new Error(`cannot find ${modelName} ${id}`);
+    return instance;
+  }
```
krishnaglick, line 96: 👍

core/src/initializers/rpc.ts
```diff
@@ -1,6 +1,18 @@
...
+import { AppsCache } from "../modules/caches/appsCache";
+import { SourcesCache } from "../modules/caches/sourcesCache";
+import { PropertiesCache } from "../modules/caches/propertiesCache";
+
+declare module "actionhero" {
+  export interface Api {
+    rpc: {
+      app: Record<string, (arg: any) => void | Promise<void>>;
+      source: Record<string, (arg: any) => void | Promise<void>>;
+      property: Record<string, (arg: any) => void | Promise<void>>;
```
krishnaglick, line 12: Do these need the arg? Doesn't look like it's used below.

core/src/models/App.ts
```diff
@@ -410,4 +408,13 @@ export class App extends CommonModel<App> {
     const redis = api.redis.clients.client;
     return redis.del(key);
   }
+
+  @AfterSave
+  @AfterDestroy
+  static async invalidateCache() {
+    AppsCache.invalidate();
+    await CLS.afterCommit(
+      async () => await redis.doCluster("api.rpc.app.invalidateCache")
```
krishnaglick, line 417:
```suggestion
      () => redis.doCluster("api.rpc.app.invalidateCache")
```
This might work the same?

core/src/models/GrouparooModel.ts
```diff
@@ -122,12 +122,6 @@ export class GrouparooModel extends CommonModel<GrouparooModel> {
 
   // --- Class Methods --- //
 
-  static async findById(id: string) {
-    const instance = await this.scope(null).findOne({ where: { id } });
-    if (!instance) throw new Error(`cannot find ${this.name} ${id}`);
-    return instance;
-  }
-
```
krishnaglick, line 130: Be Gonne!

core/src/models/Property.ts
```diff
@@ -240,7 +235,7 @@ export class Property extends CommonModel<Property> {
 
   async afterSetFilters(hasChanges: boolean) {
     if (hasChanges) {
-      await Property.invalidateCache();
+      PropertiesCache.invalidate();
```
krishnaglick, line 239: This to speed things up?

core/src/models/Property.ts
```diff
@@ -623,4 +554,13 @@ export class Property extends CommonModel<Property> {
...
   }
+
+  @AfterSave
+  @AfterDestroy
+  static async invalidateCache() {
+    PropertiesCache.invalidate();
+    await CLS.afterCommit(
+      async () => await redis.doCluster("api.rpc.property.invalidateCache")
+    );
+  }
```
krishnaglick, line 566: Could this be put on the CachedModel? Or was that removed?

core/src/modules/broadcastHelper.ts
```diff
@@ -3,11 +3,11 @@ import { CommonModel } from "../classes/commonModel";
 import { modelName } from "../modules/modelName";
 
 export async function broadcastModel<T>(
-  instance: CommonModel<T>,
+  instance: CommonModel<unknown>,
```
krishnaglick: What was the `T` issue here?

core/src/modules/caches/appsCache.ts
```diff
@@ -0,0 +1,52 @@
+import { ModelCache } from "../modelCache";
+import { App } from "../../models/App";
+import { Option } from "../../models/Option";
+import { Includeable, WhereAttributeHash } from "sequelize";
+
+function getInclude() {
+  const include: Includeable[] = [{ model: Option, required: false }];
+  return include;
+}
```
krishnaglick: Seems pretty simple to be its own function.

core/src/modules/caches/appsCache.ts
```diff
@@ -0,0 +1,52 @@
...
+  modelId?: string,
+  state?: App["state"],
+  lookupKey: keyof App = "id"
+) {
+  const instances = await this.findAllWithCache();
+  let instance = instances.find((i) => i[lookupKey] === value);
+
+  if (!instance) {
+    const where: WhereAttributeHash = { [lookupKey]: value };
+    if (state) where.state = state;
```
krishnaglick, line 33: I think this should work fine (if typescript doesn't complain) since it'll either be what we want, or undefined and thus ignored.
```suggestion
    where.state = state;
```

core/src/modules/caches/appsCache.ts
```diff
@@ -0,0 +1,52 @@
...
+    instance = await App.unscoped().findOne({
```
krishnaglick: Maybe I missed it, or haven't gotten to it yet but is unscoped new?

core/src/modules/caches/propertiesCache.ts
```diff
@@ -0,0 +1,68 @@
...
+import { Filter } from "../../models/Filter";
+import { Includeable, WhereAttributeHash } from "sequelize";
+
+function getInclude() {
+  const include: Includeable[] = [
+    { model: Source.unscoped(), required: false },
+    { model: Option, required: false },
+    { model: Filter, required: false },
+  ];
+  return include;
```
krishnaglick: Seems like this could just be a const? Unless `Source.unscoped()` has some call-time context?

core/src/modules/caches/propertiesCache.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (instance) this.invalidate();
+  }
+
+  return instance;
+}
+
+export const PropertiesCache = new ModelCache<Property>(
+  findAllWithCache,
+  findOneWithCache
+);
```
krishnaglick, line 64: I do like me some singletons.

core/src/modules/caches/sourcesCache.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (state) where.state = state;
+    instance = await Source.unscoped().findOne({
+      where,
+      include: getInclude(),
+    });
+    if (instance) this.invalidate();
+  }
+
+  return instance;
+}
```
krishnaglick, line 53: I'm seeing this repeated a lot. I think we might be able to improve this pattern.
An idea:
```
const generateFindOneWithCache = <T>(include: Includable[]) => {
return async function findOneWithCache(
  this: ModelCache<T>,
  value: string,
  modelId?: string,
  state?: T["state"],
  lookupKey: keyof T = "id"
) {
  const instances = await this.findAllWithCache(modelId);
  let instance = instances.find((i) => i[lookupKey] === value);

  if (!instance) {
    const where: WhereAttributeHash = { [lookupKey]: value };
    if (state) where.state = state;
    instance = await Property.unscoped().findOne({
      where,
      include,
    });
    if (instance) this.invalidate();
  }

  return instance;
}
}
```

I haven't pulled it down and tried this but I think it's worth a try.

krishnaglick: Maybe I need to pull this down and dig into it since I see a class generic below.

core/src/modules/plugin.ts
```diff
@@ -189,7 +191,7 @@ export namespace plugin {
       data: mappedRecordProperties,
       creatorType: "run",
       creatorId: run.id,
-    });
+    } as CreationAttributes<Import>);
```
krishnaglick: I'm glad this needs types, but shouldn't it not need the `as` if it's passing what's necessary?

core/src/modules/stateMachine.ts
```diff
@@ -9,7 +10,7 @@ export namespace StateMachine {
   }
 
   export async function transition(
-    instance: Model & {
+    instance: CommonModel<any> & {
```
krishnaglick:
```suggestion
  export async function transition<T = any>(
    instance: CommonModel<T> & {
```
Maybe?

core/__tests__/modules/internalRun.ts
evantahler: `internalRun ()` is now strongly typed! So... a random string ('creator') isn't valid
Screen Shot 2022-02-10 at 10 55 16 AM

core/src/models/Property.ts
evantahler: It's not async any more!

core/src/models/Property.ts
evantahler: CachedModel didn't work out, so the models themselves need to each implement this

core/src/modules/caches/appsCache.ts
evantahler: it's been around for a while!  Some of our models have a default scope (usually state=ready), and here in the cache we want all the rows, so we can filter them by state later if needed

core/src/modules/caches/appsCache.ts
evantahler: the problem is then that you pass in either an undefined (literal) or null - both of which sequelize will try to stringily and use, because the key on the where object exists.

core/src/modules/caches/appsCache.ts
evantahler: I moved this into the constructor!

core/src/modules/plugin.ts
evantahler: 👍 looks like a leftover from back when I had the stronger ts-types on models

core/src/classes/codeConfig.ts
evantahler: 🤷  it's OK now!

core/src/classes/commonModel.ts
evantahler: These are used below in `findById`

core/src/initializers/rpc.ts
evantahler: app.disconnect uses it below

core/src/modules/caches/propertiesCache.ts
evantahler: I added a note - it needs to be looked-up at runtime to prevent circular deps.

core/src/modules/caches/sourcesCache.ts
evantahler: The where classes for each are pretty different, so creating a builder didn't really work out

core/src/models/Property.ts
krishnaglick: I'd love to try to revisit it at a later point.

core/src/modules/modelCache.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.findOneWithCache = findOneWithCache.bind(this);
+    this.include = include;
+  }
+
+  invalidate() {
+    this.expires = 0;
+  }
+}
+
+export const TTL = env === "test" ? -1 : 1000 * 30;
```
krishnaglick: I think you can use this value in the class.

core/src/modules/plugin.ts
```diff
@@ -220,7 +222,7 @@ export namespace plugin {
         data: mappedRecordProperties,
         creatorType: "run",
         creatorId: run.id,
-      });
+      } as CreationAttributes<Import>);
```
krishnaglick:
```suggestion
      };
```

krishnaglick (approved):
Looks good!

PR #2927, opened 2022-02-01
Source creation works smoother, added save success messaging

krishnaglick:
## Change description

Source creation works smoother, added save success messaging

### Development

- [x] Application changes have been tested appropriately

### Impact

- [x] Code follows company security practices and guidelines
- [x] Security impact of change has been considered
- [x] Performance impact of change has been considered
- [x] Possible migration needs considered (model migrations, config migrations, etc.)

Please explain any security, performance, migration, or other impacts if relevant:

### Code review

- [x] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached where applicable.
- [x] Relevant tags have been added to the PR (bug, enhancement, internal, etc.)

PR #2946, opened 2022-02-07
Fix edit page to use load preview only if available and serverprops hydration error

edmundito:
## Change description

This change fixes an issue with the source edit page caused by #2927 where it was loading and using the preview on a source that didn't have a preview available, for example, a postgres query source.

While figuring this out, I fixed an issue where the `hydrationError` from `withServerErrorHandler` was not being used because it gets returned as part of the `pageProps` instead of the nextApp's props (since it doesn't support getServerSideProps).

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
+++ b/ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
@@ -86,8 +86,8 @@ const Page: NextPage<Props> = ({
   }, [preview]);
 
   const mappingColumn = useMemo(
-    () => Object.keys(source.mapping)[0] ?? previewColumns?.[0],
-    [previewColumns, source.mapping]
+    () => Object.keys(source.mapping)[0],
+    [source.mapping]
   );
   const mappingPropertyKey = useMemo(
     () => Object.values(source.mapping)[0],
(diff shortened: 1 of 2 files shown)


ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -86,8 +86,8 @@ const Page: NextPage<Props> = ({
   }, [preview]);
 
   const mappingColumn = useMemo(
-    () => Object.keys(source.mapping)[0] ?? previewColumns?.[0],
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx]: I don't recall why this was added in the previews change, so I restored it to the previous version. @krishnaglick do you recall the reason?

Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:89]: I don't recall why this was added in the previews change, so I restored it to the previous version. @krishnaglick do you recall the reason?

Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx]: I don't recall why this was added in the previews change, so I restored it to the previous version. @krishnaglick do you recall the reason?

Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:89]: I don't recall why this was added in the previews change, so I restored it to the previous version. @krishnaglick do you recall the reason?

Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx]: I don't recall why this was added in the previews change, so I restored it to the previous version. @krishnaglick do you recall the reason?

Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:89]: I don't recall why this was added in the previews change, so I restored it to the previous version. @krishnaglick do you recall the reason?

Comment (krishnaglick) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx]: I believe it needed to be up here so the data was correct in all the spots. Once we do a bigger refactor this will likely go away.

Comment (edmundito) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx]: Ok I restored it and tested it and still works

edmundito:
Fixed a couple of more issues and this should be ready to go.

krishnaglick (approved):
Unsure how to test the hydration error change, but it seems additive so I don't anticipate an issue.

PR #2948, opened 2022-02-07
Cleanup client error handling

edmundito:
## Change description

- Moves `getIntialProps` redirect handling to `_app getInitialProps`
- Moves all browser-side error handling to one function
- Fixes a few browser-side async requests that were not checking if the return type was made

### Development

### Impact

### Code review


ui/ui-components/client/client.ts
```diff
@@ -1,104 +1,85 @@
...
+);
 
-    const unlock = (data?: T) => {
-      this.cache[key] = {
-        data,
-        createdAt: new Date().getTime(),
-        locked: false,
-      };
-    };
+if (isBrowser()) {
```
edmundito: The biggest change is here: All the client-side logic is consolidated here instead of the `Client` class. I tried a few different approaches and this is the cleanest I could come up with. The other is to apply it to the `useApi` hook somehow but I found it too messy.

ui/ui-components/client/clientCache.ts
```diff
@@ -0,0 +1,62 @@
+export interface ClientCacheObject<T = unknown> {
```
edmundito, line 1: New file for the cache

ui/ui-components/pages/_app.tsx
```diff
@@ -102,6 +103,17 @@ GrouparooNextApp.getInitialProps = async (appContext: AppContext) => {
...
+      const requestPath = req.url.match("^[^?]*")[0];
+      const redirect = getRedirectFromErrorCode(_error.code, requestPath);
+      if (redirect) {
+        res.writeHead(302, {
+          Location: redirect.destination,
+        });
+        res.end();
+        return {};
+      }
+
```
edmundito, line 117: This needs to exist as long as we have uses `getInitialProps`. There are 58 files that currently use `getInitialProps`. I could create a separate change to migrate them all to `getServerSideProps`

ui/ui-config/pages/app/new/index.tsx
```diff
@@ -14,6 +14,7 @@ import React from "react";
 
 type PluginWithVersion = Actions.PluginsList["plugins"][number];
 
+// TODO: Can we find an alternative to not customizing the error handler?
```
edmundito: This is the only file that customizes the error handler just to display a custom message. I was wondering if there are any ideas on how to rewrite this without having to customize the error handler.

edmundito:
I had to backtrack a bit since my first change. It turns out that axios instances are not supported with the mocking library so I rolled that back while browser check separate.

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -299,8 +299,10 @@ const Page: NextPage<Props> = ({
           modelId: source?.modelId,
         });
 
-      setProperties(properties);
-      setPropertyExamples(examples);
+      if (properties) {
+        setProperties(properties);
+        setPropertyExamples(examples);
+      }
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:305]: This is probably important to socialize to the team - the pattern we are continuing to go with for the UI is:
1. The client will never throw, and the errorHandler will handle errors.  It will always return a promise with /something/
2. Because of the above, you need to check your response payload to be sure that it has the object you want

Comment (evantahler) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:305]: I personally like the above, but an alternative pattern could have been that client would return the error string as well if something went wrong, so you could do 

```ts
      if (!response.error) {
        setProperties(properties);
        setPropertyExamples(examples);
      }
```

ui/ui-config/pages/app/new/index.tsx
evantahler: Idea: maybe rather than creating custom error handlers, we create custom "formatters".

So this page would do something like:
```ts
      const response: Actions.PublicStatus = await client.request(
        "get",
        "/status/public",
        undefined,
        { useCache: false, errorFormater: (message) => {
             // either throw to pass to the error handler or not (eg log it)
        } }
      );
```

evantahler: I think that changing this behavior is in-scope for this story

ui/ui-components/pages/_app.tsx
evantahler: I think that removing `getInitialProps` is in-scope for this story.

ui/ui-components/client/client.ts
```diff
@@ -103,161 +43,92 @@ export class Client {
...
+      config[config.method === "get" ? "params" : "data"] = data;
+    }
 
-  private csrfToken = () => {
-    if (globalThis?.localStorage) {
-      return window.localStorage.getItem("session:csrfToken");
+    let unlock: ClientCacheGetObject["unlock"];
+    if (config.method === "get" && options.useCache) {
+      const { cacheData, unlock: _unlock } = await this.cache.get<Response>(
+        config.url + JSON.stringify(data)
```
krishnaglick: I think this could potentially cause problems without a stable stringify method.

ui/ui-components/client/client.ts
```diff
@@ -103,161 +43,92 @@ export class Client {
...
-      }
+      unlock?.(response.data);
 
       if (response.data && response.data.error) {
-        throw response.data.error.message
+        const error = response.data.error.message
           ? response.data.error.message
           : response.data.error;
+
+        throw error;
```
krishnaglick: Seems silly to assign and immediately throw.

ui/ui-components/client/handleBrowserClientError.ts
```diff
@@ -0,0 +1,20 @@
...
+  errorHandler.set({ message: error });
+
+  const redirect = getRedirectFromErrorCode(
+    error.code,
+    window.location.pathname
+  );
+
+  if (redirect) {
+    window.location.href = redirect.destination;
+    return new Promise(() => {});
```
krishnaglick: Still concerned about this as a memory leak.

ui/ui-components/client/handleBrowserClientError.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  return {} as Response;
```
krishnaglick: Almost want to say don't `as Response` this so consumers have to deal with `Response | {}`

ui/ui-config/pages/app/new/index.tsx
krishnaglick: Since it's only done in one spot, seems simple enough.

ui/ui-config/pages/index.tsx
```diff
@@ -40,13 +40,16 @@ export default function Page(props) {
...
+        const { models } = await client.request<Actions.ModelsList>(
+          "get",
+          "/models",
+          {
+            limit: 1,
+            order: [["name", "asc"]],
+          }
+        );
+        if (models?.length > 0) {
+          const [model] = models;
```
krishnaglick: Was this throwing if `models.length` was  0?

ui/ui-config/pages/session/sign-in.tsx
```diff
@@ -34,13 +34,15 @@ export default function SignInPage(props) {
...
-          router.push(model ? `/model/${model.id}/overview` : "/");
+          const { models } = await client.request<Actions.ModelsList>(
+            "get",
+            "/models",
+            {
+              limit: 1,
+              order: [["name", "asc"]],
+            }
+          );
+          router.push(models.length ? `/model/${models[0].id}/overview` : "/");
```
krishnaglick:
```suggestion
          router.push(models[0]?.id ? `/model/${models[0].id}/overview` : "/");
```
Maybe?

ui/ui-components/client/client.ts
edmundito: @krishnaglick Can you explain the reasoning? This has also been the same way before I moved the file.

ui/ui-components/pages/_app.tsx
edmundito: Ok, But I will keep the change as an isolated PR. For now I'll move the support here and remove this stuff once in the migration PR

ui/ui-components/client/client.ts
krishnaglick: When you stringify an object, the resulting string is dependent on how the object was built and what props were added when.
Something like [fast-json-stable-stringify](https://github.com/epoberezkin/fast-json-stable-stringify) produces consistent results across differently build objects assuming same key/value pairings.

ui/ui-config/pages/index.tsx
edmundito: It will probably throw based on the current implementation because if there is a fetch error the `errorHandler` swallows it and we get a `{}` as a response.

ui/ui-config/pages/index.tsx
```diff
@@ -40,14 +40,18 @@ export default function Page(props) {
...
+        const { models } = await client.request<Actions.ModelsList>(
+          "get",
+          "/models",
+          {
+            limit: 1,
+            order: [["name", "asc"]],
+          }
+        );
+
+        const modelId = models?.[0]?.id;
```
krishnaglick, line 52: 👍

ui/ui-config/pages/session/sign-in.tsx
```diff
@@ -34,13 +34,16 @@ export default function SignInPage(props) {
...
+          const { models } = await client.request<Actions.ModelsList>(
+            "get",
+            "/models",
+            {
+              limit: 1,
+              order: [["name", "asc"]],
+            }
+          );
+          const modelId = models?.[0]?.id;
+          router.push(modelId ? `/model/${modelId}/overview` : "/");
```
krishnaglick, line 46: 👍

ui/ui-components/client/client.ts
```diff
@@ -103,161 +44,90 @@ export class Client {
...
+      config[config.method === "get" ? "params" : "data"] = data;
+    }
 
-  private csrfToken = () => {
-    if (globalThis?.localStorage) {
-      return window.localStorage.getItem("session:csrfToken");
+    let unlock: ClientCacheGetObject["unlock"];
+    if (config.method === "get" && options.useCache) {
+      const { cacheData, unlock: _unlock } = await this.cache.get<Response>(
+        config.url + stringify(data)
```
krishnaglick, line 105: Awesome. I'm not sure we need to do this anywhere else 👍

krishnaglick (commented):
Looks good, though holding off on approval until it's out of a draft state :)

krishnaglick (approved):

PR #2949, opened 2022-02-08
Complete Exports in Batches

### LINKED ISSUE #2870: [Feature]: Performance Improvements

evantahler:
This PR replaces the single `export.update` for a class method `ExportOps.complete()` accepting an array of Exports.  This will speed up writing updates to successful exports.

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/ops/destination.ts
+++ b/core/src/modules/ops/destination.ts
@@ -920,9 +920,9 @@ export namespace DestinationOps {
     success: boolean;
     error: Error;
   }> {
-    const _exports: Export[] = [];
-    const exportsWithChanges = givenExports.filter((e) => e.hasChanges);
-    const exportsWithoutChanges = givenExports.filter((e) => !e.hasChanges);
+    const _exports: Export[] = []; // those exports with changes + locks acquired
+    const exportsWithChanges: Export[] = [];
+    const exportsWithoutChanges: Export[] = [];
     const locks: Awaited<ReturnType<typeof getLock>>[] = [];
 
     for (const givenExport of givenExports) {
--- a/core/src/modules/ops/destination.ts
+++ b/core/src/modules/ops/destination.ts
@@ -935,10 +935,12 @@ export namespace DestinationOps {
 
     await ExportOps.completeBatch(exportsWithoutChanges);
 
-    const mostRecentExportIds = await Export.sequelize
-      .query(
-        {
-          query: `
+    const mostRecentExportIds =
+      exportsWithChanges.length > 0
+        ? await Export.sequelize
+            .query(
+              {
+                query: `
     SELECT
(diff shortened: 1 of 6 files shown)


core/src/modules/ops/destination.ts
```diff
@@ -924,37 +922,36 @@ export namespace DestinationOps {
   }> {
     const _exports: Export[] = [];
     const locks: Awaited<ReturnType<typeof getLock>>[] = [];
+    const exportsWithoutChanges = givenExports.filter((e) => !e.hasChanges);
+    const exportsWithChanges = givenExports.filter((e) => e.hasChanges);
```
teallarson: Can you help me understand this change?  `.filter` is marginally slower than a `for` loop, and now we have multiple `.filter`s and then loops through the results of those instead of the single `for` below.  It looks cleaner, but I would think performance would be slower.

teallarson (commented):
Looks like there's some work ongoing with tests.

I did have one question RE: the implementation in the `sendExports()` method.

core/src/modules/ops/destination.ts
evantahler: You are right! I should be making both of theses new arrays in one loop, rather than 2.  I was mostly concerned with the order-of-magnitude change to reduce the number of SQL statements +network requests made, but this should be better.

evantahler: cleaned up in 87f2ca0278975cf16f6f1d6e899c50eca965d176

teallarson: The SQL query redundancy should be further reduced by #2943 so even faster ⏩ ⏩ ⏩

evantahler: That explains the merge conflicts :D

core/src/modules/ops/export.ts
```diff
@@ -138,6 +138,33 @@ export namespace ExportOps {
...
+      {
+        where: { id: { [Op.in]: _exports.map((e) => e.id) } },
+        returning: true,
+      }
+    );
+
+    for (const _export of _exports) {
+      _export.state = "complete";
+      _export.completedAt = now;
+      await Export.logExport(_export);
```
teallarson, line 164: Is this doing the same thing twice? `logExport` is part of an `@AfterUpdate` on the Exports model.

evantahler: `@AfterUpdate` only fires after a single-item update.  There's `@AfterBatchUpdate` which happens after these kind of bulk updates, but they don't actually include the rows modified (because they would often be too big to fit in RAM).  So, this kind of hack is needed

teallarson (approved):

PR #2953, opened 2022-02-08
Add subnav items under each model

rwfeather:
## Change description

Adds new navigation items underneath the active model

### Development

### Impact

### Code review

### DIFF
--- a/core/src/actions/navigation.ts
+++ b/core/src/actions/navigation.ts
@@ -20,6 +20,7 @@ export interface NavigationItem {
   href?: string;
   mainPathSectionIdx?: number;
   small?: boolean;
+  children?: NavigationItem[];
 }
 
 export class NavigationList extends OptionallyAuthenticatedAction {
@@ -107,6 +108,40 @@ export class NavigationList extends OptionallyAuthenticatedAction {
           icon: model.getIcon(),
           href: `/model/${model.id}/overview`,
           mainPathSectionIdx: 2,
+          children: [
+            {
+              type: "link",
+              title: "Model Data",
+              icon: "file-import",
+              href: `/model/${model.id}/overview#model-data`,
+              mainPathSectionIdx: 3,
+              small: true,
+            },
+            {
+              type: "link",
+              title: "Sample Record",
+              icon: "flask",
+              href: `/model/${model.id}/overview#sample-record`,
+              mainPathSectionIdx: 3,
+              small: true,
+            },
+            {
+              type: "link",
+              title: "Records",
+              icon: "file-alt",
+              href: `/model/${model.id}/records`,
+              mainPathSectionIdx: 3,
+              small: true,
+            },
+            {
+              type: "link",
+              title: "Destinations",
+              icon: "file-export",
+              href: `/model/${model.id}/overview#destinations`,
+              mainPathSectionIdx: 3,
+              small: true,
+            },
+          ],
         });
       });
     }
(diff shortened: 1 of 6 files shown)


rwfeather:
https://user-images.githubusercontent.com/6589528/153256832-2fbd61fc-fcc1-40ac-8fea-294b053ad8f7.mov

core/src/actions/navigation.ts
```diff
@@ -107,6 +108,40 @@ export class NavigationList extends OptionallyAuthenticatedAction {
...
+              small: true,
+            },
+            {
+              type: "link",
+              title: "Records",
+              icon: "file-alt",
+              href: `/model/${model.id}/records`,
+              mainPathSectionIdx: 3,
+              small: true,
+            },
```
evantahler, line 135: 👍 I like that this goes to the list page

ui/ui-components/components/navigation/HighlightingNavLink.tsx
```diff
@@ -67,6 +81,28 @@ export default function HighlightingNavLink({
...
+            <HighlightingNavLink
+              href={childNav.href}
+              mainPathSectionIdx={childNav.mainPathSectionIdx ?? 1}
+              text={
+                <>
+                  {childNav.title}
+                  {childNav.title === "Runs" ? (
+                    <>
+                      {" "}
+                      <RunningRunsBadge />
```
evantahler: Oh! I guess we had lost this badge

evantahler (approved):
Looks great!

ui/ui-components/components/navigation/HighlightingNavLink.tsx
```diff
@@ -1,16 +1,29 @@
...
+import { NavigationItem } from "@grouparoo/core/src/actions/navigation";
+import RunningRunsBadge from "./RunningRunsBadge";
+
+interface Properties {
+  href: string;
+  text: React.ReactNode;
+  icon: string;
+  mainPathSectionIdx: number;
+  small: boolean;
+  navChildItems?: NavigationItem[];
```
edmundito: Nit: Maybe `subNavItems` ?

edmundito:
I noticed in the video and confirmed in my local checkout as well that when navigating from the model's records to destinations it doesn't land in the right spot. It looks like the sample record is dynamically loading there instead of hydrating from getServerSideProps.

Also, I wonder if there is a way to enable smooth scrolling on the page after the first load, but nothing too complicated. maybe something simple based on this suggestion: https://stackoverflow.com/a/61784812/20494

ui/ui-components/components/navigation/HighlightingNavLink.tsx
rwfeather: Actually, this logic is still in the parent component: https://github.com/grouparoo/grouparoo/pull/2953/files#diff-f356a14f3edf616238baef014100f4a869eaa21d8bbf00df34305b2b420d3066L194

It got added here due to copy-paste, but after taking a second look it seems unnecessary to include it here. This code path handles the subNav items only, and since the "Runs" nav item is handled as a top level item behind the accordion, it doesn't really need to be here.

ui/ui-components/components/navigation/HighlightingNavLink.tsx
rwfeather: Agree, renamed.

rwfeather:
> I noticed in the video and confirmed in my local checkout as well that when navigating from the model's records to destinations it doesn't land in the right spot. It looks like the sample record is dynamically loading there instead of hydrating from getServerSideProps.

Looks like I created this branch before #2941 landed in `main`. It should work after getting merged, but let me just do a quick rebase to find out

rwfeather:
Loading behavior looks better.

I'm currently working on adding the smooth scrolling.

PR #2962, opened 2022-02-10
fix: `incremental` default on sources depends on plugin support

pedroslopez:
## Change description

Trying to create a schedule for a source that does not support incremental schedules was erroring on the UI:

image

This was because it was being set to `true` by default, but this should depend on plugin support.

Added this as part of the model since it affected the UI actions as well as loading older config files that didn't have `incremental` on them.

### Development

### Impact

### Code review

### DIFF
--- a/core/src/models/Schedule.ts
+++ b/core/src/models/Schedule.ts
@@ -105,7 +105,6 @@ export class Schedule extends CommonModel<Schedule> {
   locked: string;
 
   @AllowNull(false)
-  @Default(true)
   @Column
   incremental: boolean;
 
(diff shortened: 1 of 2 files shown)


core/src/models/Schedule.ts
```diff
@@ -356,6 +355,16 @@ export class Schedule extends CommonModel<Schedule> {
     }
   }
 
+  @BeforeValidate
+  static async ensureSetIncremental(instance: Schedule) {
+    if (instance.isNewRecord && typeof instance.incremental === "undefined") {
```
evantahler, line 360: Could it ever be null?  I guess it depends on how the not-yet-saved instance was created

evantahler (approved):
Nice!  That's a good use of a beforeValidate

core/src/models/Schedule.ts
pedroslopez: Yeah, if `null` was set on an action or config, sequelize will catch it and say it can't be null. I wanted to keep this value-setting specifically to when it hasn't been provided.

PR #2963, opened 2022-02-10
Remove `grouparoo generate` command

teallarson:
## Change description

Gone:
- the `generate` command and all references/implementations of it
- templates used by the generate command (Config UI generates files based on the config object, not the template)
- all plugin references to templates
- all tests related to templates or the `generate` command, including the CLISpechelper (which was only used to `generate` or validate the empty generated files)
- methods used to generate/fill in the templates generated
- `roo init` no longer generates a config dir full of dirs and READMEs.  (`roo config` generates the directories needed on the fly).`roo init` directs folks to the Config UI docs.  

Updated the CLI docs a bit as well.

### Development

### Impact

Users will not be able to use `grouparoo generate` anymore.

### Code review


evantahler:
Screen Shot 2022-02-10 at 2 39 54 PM
This is the best kind of PR

cli/src/lib/initialize.ts
```diff
@@ -62,22 +62,6 @@ export default async function Initialize(
 
   safelyCreateFile("gitignore", ".gitignore"); // we need to call the source file something other than `.gitignore` so it's not ignored by NPM
 
-  /**
```
teallarson, line 65: Since Config UI generates the directories it needs, we don't need this skeleton structure anymore.  The `init` command directs folks to the Config docs, so the READMEs that were housed in the empty folders aren't needed either.

cli/README.md
```diff
@@ -23,21 +23,31 @@ Options:
...
 
-Example: `npx grouparoo generate`
+Example: `npx grouparoo config`
 
-- To generate a grouparoo project in the current directory (`process.cwd()`): `npx grouparoo generate`
-- To generate a grouparoo project in another directory: `npx grouparoo generate /path/to/project`
+- This will start the Grouparoo Config UI and launch a browser for you to configure your instance.
+- For more information on Config UI, see http://www.grouparoo.com/docs/config
+
+### Init
```
evantahler, line 37: Maybe init should be first

cli/src/utils/templates.ts
```diff
@@ -2,26 +2,15 @@ import fs from "fs-extra";
 import path from "path";
```
evantahler, line 2: It's possible this whole file can be removed?

evantahler (approved):
Woo!

cli/README.md
```diff
@@ -23,21 +23,31 @@ Options:
   -h, --help       display help for command
 
 Commands:
-  generate [path]  generate a new Grouparoo project
-  upgrade [path]   upgrade an existing Grouparoo project
-  help [command]   display help for command
+  config [options]                 Interactively configure Grouparoo.
```
pedroslopez: Hmm, this command isn't part of the initial set of commands available, and only shows up when within a Grouparoo project (as the next "Commands" section indicates". Could having `config` here cause confusion?

cli/README.md
pedroslopez: I agree- it looks like the old docs were potentially referring to an older version of the `generate` command that could set up the project itself like `init` does now

cli/README.md
tealjulia: Good point.  On that same note, `install`/`uninstall` probably don't belong here as those would only be in projects as well.

cli/src/utils/templates.ts
teallarson: The remaining pieces in this file are for setting up the structure for and replacing the placeholders in the `package.json`, `.env`, etc.  when running `grouparoo init`.  Open to moving or renaming these, but I do think they're still necessary!

cli/README.md
teallarson: Agreed, I'll take that out.  I'm somewhat surprised that `install` and `uninstall` are in there before `init`.  Axing `config` here.

cli/README.md
teallarson: Yep!  Pedro pointed out that `config` isn't there until someone runs `init` anyways, so I've pulled that out entirely.

PR #2966, opened 2022-02-13
Cache Models and Destinations

evantahler:
Adding on to the work of https://github.com/grouparoo/grouparoo/pull/2945, we can also cache Models and Destinations internally to speed up various tasks.  This PR also fixes a bug in which our`@AfterSetOptions` (and similar) hooks for cached models were invalidating the local cache, but not broadcasting the invalidate message to other members of the cluster.

This work will then be used in a later PR to batch the `imports:associate` and `record:export` tasks

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/destination/modelCache.ts
+++ b/core/__tests__/models/destination/modelCache.ts
@@ -1,0 +1,152 @@
+import { helper } from "@grouparoo/spec-helper";
+import { Destination } from "../../../src";
+import { DestinationsCache } from "../../../src/modules/caches/destinationsCache";
+
+describe("models/destinationsCache", () => {
+  helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
+
+  beforeEach(() => (DestinationsCache.TTL = 30 * 1000));
+  afterAll(() => (DestinationsCache.expires = 0));
+
+  beforeAll(async () => {
+    await helper.factories.properties();
+    const deletedDestination = await helper.factories.destination();
+    await deletedDestination.update({ state: "deleted" });
+  });
+
+  describe("#findAllWithCache", () => {
+    test("it returns all the instances", async () => {
+      const instances = await DestinationsCache.findAllWithCache();
+      expect(instances.length).toBe(1);
+    });
+
+    test("it can filter to state", async () => {
+      const destination = await helper.factories.destination();
+      expect(destination.state).toBe("ready");
+
+      const instances = await DestinationsCache.findAllWithCache(
+        undefined,
+        "ready"
+      );
+
+      expect(instances.length).toBe(1);
+      expect(instances[0].id).toBe(destination.id);
+
+      await destination.destroy();
+    });
+
+    describe("rpc", () => {
+      let destination: Destination;
+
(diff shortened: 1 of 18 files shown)


core/src/models/App.ts
```diff
@@ -117,7 +117,7 @@ export class App extends CommonModel<App> {
   }
 
   async afterSetOptions(hasChanges: boolean) {
-    if (hasChanges) AppsCache.invalidate();
+    if (hasChanges) await App.invalidateCache();
```
evantahler, line 120: This is the way to invalidate the cache locally and for other members in the cluster

core/__tests__/models/destination/modelCache.ts
```diff
@@ -0,0 +1,152 @@
...
+        "ready"
+      );
+
+      expect(instances.length).toBe(1);
+      expect(instances[0].id).toBe(destination.id);
+
+      await destination.destroy();
+    });
+
+    describe("rpc", () => {
```
pedroslopez: nit: this `describe` block for rpc is probably outside the `#findAllWithCache` describe block.

core/__tests__/models/model/modelCache.ts
```diff
@@ -0,0 +1,129 @@
...
+      afterEach(async () => {
+        await model.destroy().catch(() => null);
+      });
+
+      test("creating a model signals RPC", async () => {
+        ModelsCache.expires = new Date().getTime();
+        await makeModel();
+        await helper.sleep(10);
+        expect(ModelsCache.expires).toBe(0);
+      });
```
pedroslopez: Are these actually testing the RPC signaling? Seems like the `XYZ.invalidateCache()` method invalidates the cache locally and then triggers the RPC `afterCommit`. (so `ModelsCache.expires` would be cleared anyway whether rpc works or not)

For app disconnection we were _only_ signaling the RPC and letting the local disconnection happen through that as well, but in this case since you want to delay invalidating the rest `afterCommit`, implementation seems ok. This is probably a rename of these tests to something like "creating a model invalidates the cache" and add a test for the rpc call itself?

core/src/modules/caches/destinationsCache.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Option } from "../../models/Option";
+import { Mapping } from "../../models/Mapping";
+import { WhereAttributeHash } from "sequelize";
+
+async function findAllWithCache(
+  this: ModelCache<Destination>,
+  modelId?: string,
+  state?: Destination["state"]
+) {
+  const now = new Date().getTime();
```
krishnaglick, line 12: Can also do `Date.now()`, no need to change however.

krishnaglick (approved):
Any further comments I have are minor or nitpicky things that aren't worth the trouble. More stylistic.
Looks good to me.

core/__tests__/models/model/modelCache.ts
evantahler: Good feedback! I've added better RPC tests via 702841e55.

pedroslopez (approved):

PR #2968, opened 2022-02-14
Associate Import and Create Exports in Batches

### LINKED ISSUE #2870: [Feature]: Performance Improvements

evantahler:
This PR takes two of our slowest tasks, `import:associateRecord` and `record:export` and replaces them with bulk tasks that operate on many records at once.  We trade many slow tasks for a few big tasks, with big SQL reads and writes.   You'll notice a lot more in-memory finding/mapping/filtering of a big batch of records and properties - this is faster than lots of small SQL statements to find records and properties one-by-one for reach record being considered.

This PR also uses caching and bulk writes a little better.

### Development

### Impact

### Code review


core/src/modules/ops/import.ts
```diff
@@ -32,67 +33,90 @@ export namespace ImportOps {
...
+      state: Import["state"];
+      creatorType: Import["creatorType"];
+      creatorId: Import["creatorId"];
+      createdRecord: Import["createdRecord"];
+      recordId: Import["recordId"];
+      recordAssociatedAt: Import["recordAssociatedAt"];
+      startedAt: Import["startedAt"];
+      createdAt: Import["createdAt"];
+      updatedAt: Import["updatedAt"];
+    }[] = [];
```
evantahler: Now that this method will create many records in bulk, and will be run by a unique, timed task, we can not worry about the locks and retries.  Postgres' (and Sqlite's) `updateOnDuplicate` does what we want

core/src/modules/ops/record.ts
```diff
@@ -786,75 +787,63 @@ export namespace RecordOps {
...
-        const value = uniquePropertiesHash[id];
-        const property = uniqueProperties.find((r) => r.id === id);
+    for (const i in uniquePropertyIds) {
+      const id = uniquePropertyIds[i];
+      const value = uniquePropertiesHash[id];
+      const property = uniqueProperties.find((r) => r.id === id);
 
-        const { releaseLock } = await waitForLock(
-          `recordProperty:${id}:${value}`
-        );
```
evantahler, line 799: Similarly to the above, we don't need the lock anymore

core/src/models/Import.ts
```diff
@@ -166,10 +166,6 @@ export class Import extends CommonModel<Import> {
     }
   }
 
-  async associateRecord() {
-    return ImportOps.associateRecord(this);
-  }
```
evantahler, line 171: Association can now only be done by the Op or task

core/src/modules/ops/record.ts
```diff
@@ -293,159 +293,162 @@ export namespace RecordOps {
...
       }
-      if (bulkDeletes.where.id.length > 0) {
-        await RecordProperty.destroy(bulkDeletes);
+
+      recordOffset++;
+    }
+
+    // TODO: This select may be slower than allowing the upsert to throw like we used to
+    // However, the previous upsert method was really hard to deal with because it would throw and ruin/end the transaction
+    const duplicateRecordAnds = bulkCreates
```
evantahler: This section of the method adds an extra SQL select to find all the record properties that are unique and have the same rawValues.  Previously, we were relying on `updateOnDuplicate` to handle this automatically, but that had the side-effect of ending the SQL transaction, which makes doing anything after this method not possible in a CLSAction or CLSTask.  

This new query might be slow, so we will need to test it

evantahler:
Running `timed-run` action after latest commit [](https://github.com/evantahler)@evantahler
[consider batch sizes more often](https://github.com/grouparoo/grouparoo/pull/2968/commits/bc93b32ba1565b6943454af8d92c77be870c2a16)


https://github.com/grouparoo/grouparoo/runs/5297765816

evantahler:
Well, this is promising!  Looking like an overall 30% improvement!

Screen Shot 2022-02-23 at 8 21 51 AM

Things do appear to be getting better!   The downside is that now we've lost the ability to trace our bigger tasks in Sentry

core/src/modules/ops/group.ts
```diff
@@ -96,18 +104,25 @@ export namespace GroupOps {
...
-          bulkCreateRecordIds.map((recordId) => {
-            return { recordId, groupId: group.id };
-          })
+          bulkCreateRecordIds
+            .splice(0, config.batchSize.internalWrite)
+            .map((recordId) => {
+              return { recordId, groupId: group.id };
+            })
         );
       }
```
evantahler, line 115: In a few places we are now trying to respect `config.batchSize.internalWrite` more by only updating 1000 records/rows at a time

core/src/tasks/record/makeReady.ts
```diff
@@ -17,12 +17,16 @@ export class GrouparooRecordsMakeExports extends CLSTask {
...
-        await CLS.enqueueTask("record:export", { recordId: record.id });
-      }
-    }
+    await RecordOps.makeExports(
+      partialRecords.map((r) => r.id),
+      toExport
+    );
+
+    // re-enqueue if there is more to do
+    if (partialRecords.length > 0) await CLS.enqueueTask(this.name, {});
```
evantahler, line 28: To skip waiting for the next task in ~10s, we can immediately enqueue a new instance of the task if we got records

plugins/@grouparoo/sentry/src/initializers/sentry.ts
```diff
@@ -109,3 +113,15 @@ export class SentryInitializer extends Initializer {
...
+  transaction: ReturnType<typeof Sentry.startTransaction>
+) {
+  for (const span of transaction["spanRecorder"].spans) {
+    // trim the description (where the SQL transaction would go) down to sentrySpanLengthLimit chars max
+    if (span.description && span.description.length > sentrySpanLengthLimit) {
+      span.description =
+        span.description.slice(0, sentrySpanLengthLimit) + "...";
+    }
+  }
+}
```
evantahler: Now that we have these giant bulk update queries, sentry was dropping larger event payloads.  This truncates our SQL statement lengths so we get *something* at least

core/src/modules/ops/record.ts
```diff
@@ -747,114 +757,186 @@ export namespace RecordOps {
...
-    hash: {
+    hashes: {
       [key: string]: (string | number | boolean | Date)[];
-    },
-    source?: boolean | Source
+    }[],
+    referenceIds: string[],
+    source?: boolean | Source,
+    includeAllProperties = false
   ) {
```
evantahler, line 774: This method is drastically updated to do *everything* in batches now.  It never throws, but returns a response payload with per-record errors if needed

core/src/modules/ops/record.ts
```diff
@@ -293,159 +293,159 @@ export namespace RecordOps {
...
-            )
-            .map((p) => bulkDeletes.where.id.push(p.id));
-        }
+      recordOffset++;
+    }
 
-        recordOffset++;
+    // TODO: This select may be slower than allowing the upsert to throw like we used to
+    // However, the previous upsert method was really hard to deal with because it would throw and ruin/end the transaction
+    const duplicateRecordRawValues = bulkCreates
```
evantahler, line 403: As the comment says, we can't rely on `updateOnDuplicate` throwing like we used to as we need this method to succeed in-line

evantahler:
Ok, this is finally ready for review! 🥳 

Final timed runs 
- https://github.com/grouparoo/grouparoo/runs/5326416971?check_suite_focus=true -> [Sentry](https://sentry.io/organizations/grouparoo/performance/?breakdown=db&display=latency&end=2022-02-25T20%3A06%3A33&project=5807413&start=2022-02-25T15%3A09%3A40&unselectedSeries=p100%28%29)
- https://github.com/grouparoo/grouparoo/runs/5339344316?check_suite_focus=true -> [Sentry](https://sentry.io/organizations/grouparoo/performance/?breakdown=none&end=2022-02-26T05%3A35%3A16&project=5807413&query=transaction.duration%3A%3C15m&showTransactions=p95&sort=-p50&start=2022-02-25T20%3A04%3A30)

---

* [[TEST] revert batch for makeExports](https://github.com/grouparoo/grouparoo/pull/2968/commits/c2c869bcec626504dab1c46ddfd42f29b44878f6) | https://github.com/grouparoo/grouparoo/runs/5380958642?check_suite_focus=true -> [Sentry](). --- Stopped and got stuck
* [consistenet order for recordProperties needed](https://github.com/grouparoo/grouparoo/pull/2968/commits/987ef747570f33f5864a3aaaf2b5f47235f881b5) | https://github.com/grouparoo/grouparoo/runs/5383157733?check_suite_focus=true -> Sentry

core/src/modules/ops/destination.ts
```diff
@@ -305,199 +303,210 @@ export namespace DestinationOps {
...
+    });
 
-    let newGroups = await Group.findAll({
-      include: [{ model: GroupMember, where: { recordId: record.id } }],
+    // New and old properties and groups are in the context of this destination and what it has currently been sent
+    // If there is not a mostRecentExport, both old groups and old record properties are an empty collection
+    const mostRecentExportIds = await getMostRecentExportIds(
+      destination,
+      records.map((r) => r.id)
+    );
```
pedroslopez: should this be looking for `complete` exports? This was the previous behavior and I believe the correct one to use for pulling old properties and groups

core/src/modules/ops/import.ts
```diff
@@ -32,67 +33,106 @@ export namespace ImportOps {
...
+          });
+        }
+      }
+    }
 
-    _import.state = "importing";
-    _import.createdRecord = isNew;
-    _import.recordId = record.id;
-    _import.recordAssociatedAt = new Date();
+    await RecordOps.markPendingByIds(bulkImportUpdates.map((i) => i.recordId));
```
pedroslopez, line 119: Note: The previous `associateRecord` task used to also update the record with the properties that came in the import data. This is now happening within the `findOrCreateByUniqueRecordProperties` method.

core/src/modules/ops/record.ts
```diff
@@ -747,114 +765,186 @@ export namespace RecordOps {
...
+    const properties = await PropertiesCache.findAllWithCache(
+      undefined,
+      "ready"
+    );
 
-        const { releaseLock } = await waitForLock(
-          `recordProperty:${id}:${value}`
-        );
-        lockReleases.push(releaseLock);
+    i = 0;
```
pedroslopez: is this `i` being used anywhere?

core/src/modules/ops/record.ts
```diff
@@ -747,114 +765,186 @@ export namespace RecordOps {
...
+    );
+
+    await addOrUpdateProperties(
+      response.filter((d) => !d.error).map((d) => d.record),
+      response
+        .filter((d) => !d.error)
+        .map((d) => (includeAllProperties ? d.hash : d.uniquePropertiesHash)),
+      false,
+      true
+    );
```
pedroslopez, line 944: 👍🏼 `addOrUpdateProperties` is now happening for records that were found

core/src/modules/ops/record.ts
```diff
@@ -290,162 +290,170 @@ export namespace RecordOps {
...
+        // delete old properties we didn't update
+        existingRecordProperties
+          .filter(
+            (p) =>
+              p.recordId === record.id &&
+              p.propertyId === property.id &&
+              p.position >= position
+          )
+          .map((p) => bulkDeletes.where.id.push(p.id));
+      }
```
pedroslopez: Slightly unrelated since the previous implementation seems to be doing the same thing: does this mean that when calling `addOrUpdateProperties` with a partial hash, we lose the value of the other properties that were not included in the hash?

I'm seeing the same happen in the old code but I feel like I'm missing something since I don't think the values get cleared when associating imports to existing records🤔

pedroslopez (commented):
Nice, this is a big one! 

Looks like tests are failing after the option validation merge due to a missing method on the test plugin.

evantahler:
[987ef74](https://github.com/grouparoo/grouparoo/pull/2968/commits/987ef747570f33f5864a3aaaf2b5f47235f881b5) fixes the problem outlined in https://github.com/grouparoo/grouparoo/pull/3007#issuecomment-1055925450

evantahler:
@pedroslopez I've reverted our test [[TEST] revert batch for makeExports](https://github.com/grouparoo/grouparoo/pull/2968/commits/c2c869bcec626504dab1c46ddfd42f29b44878f6).  I did notice that the work done in https://github.com/grouparoo/grouparoo/pull/3007#issuecomment-1055925450 introduced a bug, and potentially was leading to lots of slow behavior, and sometimes getting 'stuck', which has been fixed.

While there's certainly more tuning that can be done, I think that it's time to merge this in and unblock the new work.

core/src/modules/ops/recordProperty.ts
```diff
@@ -156,6 +156,7 @@ export namespace RecordPropertyOps {
             [Op.or]: [null, { [Op.lt]: new Date().getTime() - delayMs }],
           },
         },
+        order: [["recordId", "asc"]],
```
pedroslopez, line 159: Noticed that the old one ordered by `createdAt` - this should be ok because only a consistent order is needed, and `recordId` is faster because of the index, right?

pedroslopez (approved):

core/src/modules/ops/recordProperty.ts
evantahler: You got it!

PR #2969, opened 2022-02-14
No default values for source mapping 

teallarson:
## Change description

In a previous PR, we removed the defaults for what Grouparoo Property to map to (its `mappingPropertyKey`).  We also don't want to infer or default what the source column is for a mapping (its `sourceColumn`).  

This PR removes part of a hook that updated the `sourceColumn`/`mappingColumn` based on the `sourcePreview` meaning no defaults even when a table is selected:

Screen Shot 2022-02-14 at 3 10 34 PM

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/components/source/FormMappingSelector.tsx
+++ b/ui/ui-components/components/source/FormMappingSelector.tsx
@@ -115,7 +115,7 @@ const FormMappingSelector: React.FC<Props> = ({
             as="select"
             required
             disabled={disabled || !previewColumns.length}
-            defaultValue={columnName || previewColumns?.[0] || ""}
+            defaultValue={""}
             {...register("mapping.sourceColumn")}
           >
             <option value={""} disabled>
--- a/ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
+++ b/ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
@@ -201,8 +201,8 @@ const Page: NextPage<Props> = ({
   }, [preview]);
 
   const mappingColumn = useMemo(
-    () => Object.keys(source.mapping)[0] ?? previewColumns?.[0],
-    [previewColumns, source.mapping]
+    () => Object.keys(source.mapping)[0],
+    [source.mapping]
   );
   const mappingPropertyKey = useMemo(
     () => Object.values(source.mapping)[0],


edmundito (commented):
Code change looks fine, but I'll loop @krishnaglick for the `defaultValue` because I'm not sure where we landed with this and the new version of react-hook-form

ui/ui-components/components/source/FormMappingSelector.tsx
```diff
@@ -113,7 +113,7 @@ const FormMappingSelector: React.FC<Props> = ({
             as="select"
             required
             disabled={disabled || !previewColumns.length}
-            defaultValue={columnName || previewColumns?.[0] || ""}
+            defaultValue={""}
```
krishnaglick, line 118: I think you can remove this entirely?

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -201,8 +201,8 @@ const Page: NextPage<Props> = ({
   }, [preview]);
 
   const mappingColumn = useMemo(
-    () => Object.keys(source.mapping)[0] ?? previewColumns?.[0],
-    [previewColumns, source.mapping]
+    () => Object.keys(source.mapping)[0],
+    [source.mapping]
```
Comment (krishnaglick) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:205]: I _think_ this gets used downline to set a default value with `columnName`, so we might be able to get rid of it. `mappingPropertyKey` as well perhaps.

ui/ui-components/components/source/FormMappingSelector.tsx
teallarson: Hmm.  Removing this default makes the form (a) not show the intended default (it shows blank instead of "Select an Option") and (b) results in the form defaulting to the first option rather than "Select an Option" when the table is selected.

krishnaglick: 🤷 Works for me then

ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
Comment (teallarson) [reply ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:205]: Most of the time we _do_ want this set though.  The only time we don't want it is in the default scenario IE: new source and nothing has been selected for the mapping yet.  But if a user has already created a mapping, we want it to be populated when they navigate to that page.

krishnaglick (approved):

PR #2970, opened 2022-02-14
Add Hubspot oAuth based app

edmundito:
## Change description

This adds the Hubspot app with OAuth support and updates core to support refreshing the access token using the auth API.

- Adds support to manage access tokens with refresh tokens from core
- Rebuilds the Hubspot plugin API client to manage apiKey or refresh token
- Improves the error handling of the Hubspot plugin

### Development

### Impact

### Code review


core/src/index.ts
```diff
@@ -55,6 +55,8 @@ export { AppsCache } from "./modules/caches/appsCache";
 export { SourcesCache } from "./modules/caches/sourcesCache";
 export { PropertiesCache } from "./modules/caches/propertiesCache";
 
+export { oAuthAccessTokenGetter } from "./modules/oAuth";
```
evantahler: Exporting this means that a plugin wants to use it

core/src/modules/oAuth.ts
```diff
@@ -13,3 +16,63 @@ export interface oAuthIdentity {
   name: string;
   description: string;
 }
+
+export class oAuthAccessTokenGetter {
```
evantahler, line 22: Nice utility

core/src/index.ts
evantahler: An alternative approach would be to move this into `plugin.refreshToken('hubspot', oldtoken)` or something like that

core/__tests__/modules/oAuth.ts
```diff
@@ -0,0 +1,107 @@
...
+          token: accessTokens[0],
+          expirationSeconds: 60 * 12,
+        })
+      );
+
+      const accessToken = await getter.getAccessToken();
+
+      expect(fetch).toHaveBeenCalledTimes(1);
+      expect(fetch).toHaveBeenCalledWith(
+        `https://auth.grouparoo.com/api/v1/oauth/${providerName}/client/refresh`,
```
pedroslopez: We can use the configured host instead of hardcoding it here

```suggestion
        `${config.oAuth.host}/api/v1/oauth/${providerName}/client/refresh`,
```

core/src/modules/oAuth.ts
```diff
@@ -13,3 +16,73 @@ export interface oAuthIdentity {
...
+
+        const {
+          token,
+          expirationSeconds,
+        }: {
+          token: string;
+          expirationSeconds: number;
+        } = await response.json();
+
+        const expirationMs = Math.max(0, (expirationSeconds - 60) * 1000);
```
pedroslopez: Why the `- 60`? Is this just to refresh now if it's going to expire soon? If so, we could probably do this in `isAccessTokenExpired` to be more clear

plugins/@grouparoo/hubspot/src/lib/export-contacts/listMethods.ts
```diff
@@ -69,30 +77,28 @@ async function ensureList(
...
   // need to create it
-  const url = `https://api.hubapi.com/contacts/v1/lists?hapikey=${appOptions.hapikey}`;
-  const { data }: { data: Record<string, any> } = await Axios({
-    method: "POST",
-    url,
-    headers: { "Content-Type": "application/json" },
-    data: {
-      name: groupName,
-    },
-  });
```
pedroslopez, line 80: 👍🏼  moving it into the client

pedroslopez (commented):
Nice work!

Since we do have a custom client in here, do you think we could run the `export-records` tests with both `hapikey` and `refreshToken` auth? It might be as simple as wrapping it in a 

```js
describe.each([true, false])("with oAuth=%s", (useOAuth) => {
// ...

const appOptions = loadAppOptions(newNock, { useOAuth } );

// ...
})
```

but I'm not 100% sure without running it. May have to pass the client around to some of the helper functions.

plugins/@grouparoo/hubspot/__tests__/integration/hubspot-export.ts
```diff
@@ -47,7 +48,7 @@ async function deleteUsers() {
         await client.deleteContact(contact.vid);
       }
     } catch (error) {
-      if (!error.toString().match(/Request failed with status code 404/)) {
+      if (!error.toString().match(/Request failed with status code 400/)) {
```
pedroslopez: This is in a test so it's probably fine, but why did this change? I believe the intention here was to ignore any issues with deleting non-existent users.

plugins/@grouparoo/hubspot/__tests__/utils/nockHelper.ts
```diff
@@ -10,31 +10,45 @@ const realPath = path.join(dirPath, ".env");
...
+        new RegExp(realEnv.HUBSPOT_API_KEY, "gi"),
+        nockEnv.HUBSPOT_API_KEY
+      )
+      .replace(
+        new RegExp(realEnv.HUBSPOT_OAUTH_REFRESH_TOKEN, "gi"),
+        nockEnv.HUBSPOT_OAUTH_REFRESH_TOKEN
+      )
+      .replace(
+        new RegExp(process.env.GROUPAROO_AUTH_URL, "gi"),
+        "https://auth.grouparoo.com"
```
pedroslopez: We could follow the existing pattern and make it clear to people in the future that you'll likely have to set this by putting `GROUPAROO_AUTH_URL` in the hubspot `.env.example`. 

This would then become

```suggestion
        new RegExp(realEnv.GROUPAROO_AUTH_URL, "gi"),
        nockEnv.GROUPAROO_AUTH_URL
```

pedroslopez (approved):
...and the tests are green! Looks good. Comments are non-blocking.

plugins/@grouparoo/hubspot/__tests__/integration/hubspot-export.ts
edmundito: Good catch. I think it's because I was not running nock correctly and I got some bad errors and tried to fix it this way but needs to go back.

PR #2971, opened 2022-02-15
Improve email and phone number formatting

krishnaglick:
## Change description

Emails are no longer always lower-cased.
Phone Numbers are no longer formatted.
Some Typescript improvements.

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/recordProperty/recordProperty.ts
+++ b/core/__tests__/models/recordProperty/recordProperty.ts
@@ -197,7 +197,7 @@ describe("models/recordProperty", () => {
         });
         await recordProperty.setValue("MARIO@example.com");
         const response = await recordProperty.getValue();
-        expect(response).toBe("mario@example.com");
+        expect(response).toBe("MARIO@example.com");
       });
 
       test("invalid emails set invalidValue", async () => {
@@ -232,7 +232,7 @@ describe("models/recordProperty", () => {
           "Deleted-user-id-19430-Team-5051deleted-user-id-19430-team-5051XXXXXX@example.com";
         await recordProperty.setValue(value);
         const response = await recordProperty.getValue();
-        expect(response).toBe(value.toLowerCase());
+        expect(response).toBe(value);
       });
     });
 
--- a/core/__tests__/models/recordProperty/recordProperty.ts
+++ b/core/__tests__/models/recordProperty/recordProperty.ts
@@ -212,6 +212,8 @@ describe("models/recordProperty", () => {
           "someone.com@",
           "someone@site",
           "someone with spaces@site.com",
+          "someone@site.com ",
+          " someone@site.com",
         ];
 
         for (const e of badEmails) {
--- a/core/__tests__/models/recordProperty/recordProperty.ts
+++ b/core/__tests__/models/recordProperty/recordProperty.ts
@@ -190,7 +190,7 @@ describe("models/recordProperty", () => {
         expect(response).toBe("mario@example.com");
       });
 
-      test("emails are lower cased", async () => {
+      test("emails are not lower cased", async () => {
         const recordProperty = new RecordProperty({
           recordId: record.id,
           propertyId: emailProperty.id,
@@ -211,9 +211,6 @@ describe("models/recordProperty", () => {
           "someone.com",
           "someone.com@",
           "someone@site",
-          "someone with spaces@site.com",
-          "someone@site.com ",
-          " someone@site.com",
         ];
 
         for (const e of badEmails) {
@@ -222,6 +219,41 @@ describe("models/recordProperty", () => {
           expect(response).toBe(null);
           expect(recordProperty.invalidValue).toBe(e);
           expect(recordProperty.invalidReason).toBe("Invalid email value");
+        }
+      });
+
+      test("weird emails validate", async () => {
+        const recordProperty = new RecordProperty({
+          recordId: record.id,
+          propertyId: emailProperty.id,
+        });
+
+        const weirdEmails = [
+          "email@example.com",
+          "firstname.lastname@example.com",
+          "email@subdomain.example.com",
+          "firstname+lastname@example.com",
+          "email@123.123.123.123",
+          "email@[123.123.123.123]",
+          '"email"@example.com',
+          "1234567890@example.com",
+          "email@example-one.com",
+          "_______@example.com",
+          "email@example.name",
+          "email@example.museum",
+          "email@example.co.jp",
+          "firstname-lastname@example.com",
+          "much.”more unusual”@example.com",
+          "very.unusual.”@”.unusual.com@example.com",
+          'very.”(),:;<>[]”.VERY.”very@\\ "very”.unusual@strange.example.com',
+        ];
+
+        for (const e of weirdEmails) {
+          await recordProperty.setValue(e);
+          const response = await recordProperty.getValue();
+          expect(response).toBe(e);
+          expect(recordProperty.invalidValue).toBe(null);
+          expect(recordProperty.invalidReason).toBe(null);
         }
       });
 
--- a/core/__tests__/models/recordProperty/recordProperty.ts
+++ b/core/__tests__/models/recordProperty/recordProperty.ts
@@ -269,19 +269,19 @@ describe("models/recordProperty", () => {
         });
         await recordProperty.setValue("4128889999");
         const response = await recordProperty.getValue();
-        expect(response).toBe("+1 412 888 9999");
+        expect(response).toBe("4128889999");
       });
 
-      test("invalid numbers are stored", async () => {
+      test("invalid phone numbers are stored without change", async () => {
         const recordProperty = new RecordProperty({
           recordId: record.id,
           propertyId: phoneNumberProperty.id,
         });
         await recordProperty.setValue("5");
         const response = await recordProperty.getValue();
-        expect(response).toBe(null);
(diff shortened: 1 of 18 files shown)


core/__tests__/models/recordProperty/recordProperty.ts
```diff
@@ -269,19 +269,19 @@ describe("models/recordProperty", () => {
...
           propertyId: phoneNumberProperty.id,
         });
         await recordProperty.setValue("5");
         const response = await recordProperty.getValue();
-        expect(response).toBe(null);
-        expect(recordProperty.invalidValue).toBe("5");
-        expect(recordProperty.invalidReason).toBe("Invalid phoneNumber value");
+        expect(response).toBe("5");
+        expect(recordProperty.invalidValue).toBe(null);
+        expect(recordProperty.invalidReason).toBe(null);
```
evantahler: I think the spirit of the test has changed here - the phone number "5" is no longer invalid, because we've stopped validating it

core/package.json
```diff
@@ -51,7 +51,6 @@
     "ioredis-mock": "7.0.0",
     "isomorphic-fetch": "3.0.0",
     "json5": "2.2.0",
-    "libphonenumber-js": "1.9.46",
```
evantahler, line 54: 👋

core/src/modules/ops/recordProperty.ts
```diff
@@ -47,7 +45,7 @@ export namespace RecordPropertyOps {
           rawValue = await formatEmail(stringifiedValue);
           break;
         case "phoneNumber":
-          rawValue = await formatPhoneNumber(stringifiedValue);
+          rawValue = await formatString(stringifiedValue);
```
evantahler: Above this line we are still doing `    const stringifiedValue = \`${value}\`.trim();`. I though we also wanted to stop trimming strings in this PR?

plugins/@grouparoo/demo/src/connections/snowflake.ts
```diff
@@ -28,15 +28,9 @@ function findConfig(): { [key: string]: string } {
...
-  config: { [key: string]: string };
-  lines: string[];
-  data: { [tableName: string]: { [key: string]: string | number | Date }[] };
-  constructor() {
-    super();
-    this.config = Object.assign({}, findConfig());
-    this.lines = [];
-    this.data = {};
-  }
+  config = { ...findConfig() };
```
evantahler, line 31: who needs a constructor!

plugins/@grouparoo/snowflake/__tests__/table-import/importProperty.ts
```diff
@@ -67,7 +77,7 @@ describe("snowflake/table/recordProperty", () => {
     record = await helper.factories.record();
     await record.addOrUpdateProperties({
       userId: [1],
-      email: ["ejervois0@example.com"],
+      email: ["ejervoiS0@example.com"],
```
evantahler, line 80: In the CSV sample data above, the email is `ejervois0@example.com` (lowercase) - how did it get uppercased?

evantahler (changes requested):
Looking good, but there are a few things to update, and the `trim()` should be removed I think.  It's probably good to add a test to core to show that case-sensitivity has returned and that leading/trailing " " will be left in place.

core/__tests__/models/recordProperty/recordProperty.ts
krishnaglick: I can remove this test, but felt it was perhaps worth keeping in case we decide to add validation back at some later point.

plugins/@grouparoo/snowflake/__tests__/table-import/importProperty.ts
krishnaglick: I wasn't sure how to get the csv to sync, I edited the demo data directly in snowflake. Could use a hand!

core/src/modules/ops/recordProperty.ts
krishnaglick: Oh, missed that. It didn't seem to affect the tests so I'm going to add one.

krishnaglick: email validation fails if it ends in a space. Is that desired behavior?

plugins/@grouparoo/snowflake/__tests__/table-import/importProperty.ts
evantahler: I think it was rather manual by uploading the CSV.  The CSV data and what's loaded into snowflake should match.

core/src/modules/ops/recordProperty.ts
evantahler: I don't think so... I think we said that /whatever/ the user has we try to use so ` someone@fake .com ` - spaces all over the place - we would store as-is.

plugins/@grouparoo/snowflake/__tests__/table-import/importProperty.ts
krishnaglick: Then this should have parity now I think.

core/src/modules/ops/recordProperty.ts
krishnaglick: I just want to be very intentful with any change here. Right now we do a check called `isEmail`
```
export default function isEmail(string: string) {
  if (string.match(/^\S+@\S+\.\S+$/)) return true;
  return false;
}
```
This fails if the email has leading or trailing spaces. We could scope this down to `@\S+\.` which just looks for an `@` and a `.`.
Double checking the ticket it looks like we want no validation on phone numbers, but less massaging of the emails.
I feel like if someone chooses `type: email` they'd expect at least some sort of validation compared with string, however I'm not wedded to this view.

I personally feel like this current version, where leading/trailing spaces is an error seems okay. But happy to go either way.

evantahler: I think that's a call for @bleonard!

bleonard:
This is pretty challenging. 

These are clear:
*`string` should not do any `trim` -- it's _exactly_ what it is in the database.

If `email` and `phoneNumber` should exist it only makes sense if it has some differences from `string`.
But if we mess with them, then they have issues being a primary key.
So I think those differences are in validation.

So here's the proposal:
*`string` should not do any `trim` -- it's _exactly_ what it is in the database.
* `email` and `phoneNumber` do validations and set it to null and why it's failed. Case and leading/trailing spaces are ok on emails, but would fail with missing `@` or spaces in the middle.
* But still store the original value (including spaces and cases) if it's valid.

core/__tests__/models/recordProperty/recordProperty.ts
```diff
@@ -223,6 +222,41 @@ describe("models/recordProperty", () => {
...
+          "email@example-one.com",
+          "_______@example.com",
+          "email@example.name",
+          "email@example.museum",
+          "email@example.co.jp",
+          "firstname-lastname@example.com",
+          "much.”more unusual”@example.com",
+          "very.unusual.”@”.unusual.com@example.com",
+          'very.”(),:;<>[]”.VERY.”very@\\ "very”.unusual@strange.example.com',
+        ];
```
evantahler, line 249: Love it

core/package.json
evantahler: The `^` should be removed

core/src/modules/ops/recordProperty.ts
```diff
@@ -10,6 +9,7 @@ import { Source } from "../../models/Source";
 import { AggregationMethod, PluginConnection } from "../../classes/plugin";
 import { log } from "actionhero";
 import { PropertiesCache } from "../../modules/caches/propertiesCache";
+import { plugin } from "../plugin";
```
evantahler, line 12: Not used?

core/src/modules/validators/isEmail.ts
```diff
@@ -5,6 +5,6 @@
  * We cannot use import isEmail from "validator/lib/isEmail" until https://github.com/validatorjs/validator.js/pull/1435 is merged in.
  */
 export default function isEmail(string: string) {
-  if (string.match(/^\S+@\S+\.\S+$/)) return true;
+  if (string.match(/@\S+\./)) return true;
```
evantahler, line 8: 👍 the test for email is now that is has an "@" and following "."... that's it.

plugins/@grouparoo/snowflake/__tests__/data/profiles.csv
```diff
@@ -1,4 +1,4 @@
-1,1,Erie,Jervois,ejervois0@example.com,Male,15.247.38.72,true,false,true,259.12,2020-02-01,2020-02-01 12:13:14.0
+1,1,Erie,Jervois,ejervoiS0@example.com,Male,15.247.38.72,true,false,true,259.12,2020-02-01,2020-02-01 12:13:14.0
```
evantahler, line 1: 👍 thanks for changing the source test data

evantahler (approved):
Looking good accept for the pacakge.json - please change that and this is ready to go!

core/src/modules/ops/recordProperty.ts
krishnaglick: The import was moved because I removed the phone number formatting, and then manually brought it back. It's in use [here](https://github.com/grouparoo/grouparoo/pull/2971/files#diff-f26d235951e3516c278cdfcefb9e8cbc54e164a23aec8563430624cdf5d14546R295)

PR #2975, opened 2022-02-15
Failed or Canceled Export Retry Button

krishnaglick:
## Change description

Description here

### Development

### Impact

### Code review

### DIFF
--- a/core/src/actions/exports.ts
+++ b/core/src/actions/exports.ts
@@ -155,7 +155,6 @@ export class ExportsRetryFailedById extends AuthenticatedAction {
   }: {
     params: ParamsFrom<ExportsRetryFailedById>;
   }) {
-    console.log(params);
     const count = await Export.retryFailedById(params.exportId);
 
     return { count };
--- a/core/src/actions/exports.ts
+++ b/core/src/actions/exports.ts
@@ -110,7 +110,7 @@ export class ExportView extends AuthenticatedAction {
 }
 
 export class ExportsRetryFailed extends AuthenticatedAction {
-  name = "exports:retryFailed";
+  name = "exports:retryFailures";
   description = "retry failed exports within a timeframe";
   permission: ActionPermission = { topic: "destination", mode: "write" };
   inputs = {
--- a/core/src/actions/exports.ts
+++ b/core/src/actions/exports.ts
@@ -141,3 +141,23 @@ export class ExportsRetryFailed extends AuthenticatedAction {
     return { count };
   }
 }
+
+export class ExportsRetryFailedById extends AuthenticatedAction {
+  name = "exports:retryFailedById";
+  description = "retry failed export by id";
+  permission: ActionPermission = { topic: "destination", mode: "write" };
+  inputs = {
+    exportId: { required: true },
+  };
+
+  async runWithinTransaction({
+    params,
+  }: {
+    params: ParamsFrom<ExportsRetryFailedById>;
+  }) {
+    console.log(params);
+    const count = await Export.retryFailedById(params.exportId);
+
+    return { count };
+  }
+}
--- a/core/src/actions/exports.ts
+++ b/core/src/actions/exports.ts
@@ -142,9 +142,9 @@ export class ExportsRetryFailed extends AuthenticatedAction {
   }
 }
 
-export class ExportsRetryFailedById extends AuthenticatedAction {
-  name = "exports:retryFailedById";
-  description = "retry failed export by id";
+export class ExportRetryById extends AuthenticatedAction {
+  name = "export:retry";
+  description = "Retry failed or canceled export by id";
   permission: ActionPermission = { topic: "destination", mode: "write" };
   inputs = {
     exportId: { required: true },
@@ -153,9 +153,9 @@ export class ExportsRetryFailedById extends AuthenticatedAction {
   async runWithinTransaction({
     params,
   }: {
-    params: ParamsFrom<ExportsRetryFailedById>;
+    params: ParamsFrom<ExportRetryById>;
   }) {
-    const count = await Export.retryFailedById(params.exportId);
+    const count = await Export.retryById(params.exportId);
 
     return { count };
   }
--- a/core/src/actions/exports.ts
+++ b/core/src/actions/exports.ts
@@ -25,7 +25,7 @@ export class ExportsList extends AuthenticatedAction {
       formatter: APIData.ensureArray,
       default: [["createdAt", "desc"]],
     },
-  };
+  } as const;
 
   async runWithinTransaction({ params }: { params: ParamsFrom<ExportsList> }) {
     const where: WhereAttributeHash = {};
@@ -75,7 +75,7 @@ export class ExportsTotals extends AuthenticatedAction {
   inputs = {
     recordId: { required: false },
     destinationId: { required: false },
-  };
+  } as const;
 
   async runWithinTransaction({
     params,
@@ -101,7 +101,7 @@ export class ExportView extends AuthenticatedAction {
   permission: ActionPermission = { topic: "export", mode: "read" };
   inputs = {
     id: { required: true },
-  };
+  } as const;
 
   async runWithinTransaction({ params }: { params: ParamsFrom<ExportView> }) {
     const _export = await Export.findById(params.id);
@@ -122,7 +122,7 @@ export class ExportsRetryFailed extends AuthenticatedAction {
       default: false,
       formatter: APIData.ensureBoolean,
     },
-  };
+  } as const;
 
   async runWithinTransaction({
     params,
@@ -148,7 +148,7 @@ export class ExportsRetryFailedById extends AuthenticatedAction {
   permission: ActionPermission = { topic: "destination", mode: "write" };
   inputs = {
     exportId: { required: true },
-  };
+  } as const;
 
   async runWithinTransaction({
     params,
--- a/core/src/config/routes.ts
+++ b/core/src/config/routes.ts
@@ -37,6 +37,7 @@ export const DEFAULT: { [namespace]: () => RoutesConfig } = {
         { path: "/v:apiVersion/exports", action: "exports:list" },
         { path: "/v:apiVersion/exports/totals", action: "exports:totals" },
         { path: "/v:apiVersion/exports/retryFailed", action: "exports:retryFailed" },
+        { path: "/v:apiVersion/exports/retryFailed/:exportId", action: "exports:retryFailedById" },
         { path: "/v:apiVersion/group/:groupId/records", action: "records:list" },
         { path: "/v:apiVersion/group/:id", action: "group:view" },
         { path: "/v:apiVersion/group/:id/countComponentMembers", action: "group:countComponentMembers" },
--- a/core/src/config/routes.ts
+++ b/core/src/config/routes.ts
@@ -36,8 +36,6 @@ export const DEFAULT: { [namespace]: () => RoutesConfig } = {
         { path: "/v:apiVersion/exportProcessors", action: "exportProcessors:list" },
         { path: "/v:apiVersion/exports", action: "exports:list" },
         { path: "/v:apiVersion/exports/totals", action: "exports:totals" },
-        { path: "/v:apiVersion/exports/retryFailed", action: "exports:retryFailed" },
-        { path: "/v:apiVersion/exports/retryFailed/:exportId", action: "exports:retryFailedById" },
         { path: "/v:apiVersion/group/:groupId/records", action: "records:list" },
         { path: "/v:apiVersion/group/:id", action: "group:view" },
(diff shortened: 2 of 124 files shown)


core/src/modules/ops/export.ts
```diff
@@ -175,7 +175,7 @@ export namespace ExportOps {
     saveExports = true
   ) {
     const where: WhereOptions<Export> = {
-      state: "failed",
+      state: { [Op.in]: ["failed", "canceled"] },
```
krishnaglick: @pedroslopez This is a change in existing functionality, I'm unsure if it's a problem.
I can move this logic out so retried canceled exports tell it explicitly to run them.

core/src/modules/cls.ts
```diff
@@ -31,15 +31,15 @@ export namespace CLS {
    * Wrap an Async function f in such a way that all enqueued afterCommit / enqueueTasks during invocation will be run afterwords
    * Returns the return value of function f
    */
-  export interface CLSWrapMethod {
+  export interface CLSWrapMethod<T = any> {
```
krishnaglick, line 34: Generics!

ui/ui-components/client/client.ts
```diff
@@ -64,6 +66,16 @@ export class Client {
     } = () => ({})
   ) {}
 
+  // This will replace the request method below in the future
+  public async requestAction<A extends Action>(
+    verb: Method = "get",
+    path: string,
+    data: Partial<ParamsFrom<A>>, // TODO: Make ParamsFrom handle required
```
krishnaglick: Once this handles required we can use this to type input params in the Client!

ui/ui-components/components/export/List.tsx
```diff
@@ -1,21 +1,24 @@
...
+import { capitalize } from "../../utils/languageHelper";
 import StateBadge from "../badges/StateBadge";
 import { DurationTime } from "../DurationTime";
-import { useApi } from "../../contexts/api";
-import { generateClient } from "../../client/client";
-import { GetServerSidePropsContext } from "next";
+import EnterpriseLink from "../GrouparooLink";
+import LoadingTable from "../LoadingTable";
+import Pagination from "../Pagination";
+import { ExportGroupsDiff, ExportRecordPropertiesDiff } from "./Diff";
```
krishnaglick, line 21: Sorted imports, no negative effects seen.

ui/ui-components/components/export/List.tsx
```diff
@@ -174,6 +220,15 @@ export default function ExportsList(props: Props) {
...
                     <StateBadge state={_export.state} marginBottom={0} />
+                    {["failed", "canceled"].includes(_export.state) ? (
+                      <FontAwesomeIcon
+                        className="ml-1"
+                        icon="sync-alt"
+                        size="xs"
+                        style={{ cursor: "pointer" }}
+                        onClick={() => retryExport(_export, index)}
+                      />
+                    ) : null}
```
krishnaglick, line 227: @bleonard If you get a chance to look over things, this is what this looks like.
Screen Shot 2022-02-15 at 4 00 18 PM

ui/ui-components/pages/export/[id]/edit.tsx
```diff
@@ -40,9 +86,23 @@ const Page: NextPageWithInferredProps<typeof getServerSideProps> = ({
...
+            loading={loading}
+            disabled={!["failed", "canceled"].includes(_export.state)}
+            className="ml-auto"
+            onClick={retryExport}
+            key={_export.state}
+          >
+            Retry Export
+          </LoadingButton>,
+        ]}
+      >
```
Comment (krishnaglick) [inline ui/ui-components/pages/export/[id]/edit.tsx:77]: Moved to Edmundo's managed card 👌

krishnaglick (commented):
Screen Shot 2022-02-15 at 4 06 39 PM
The list and the detail view have different ways to trigger the same thing. Could replicate the icon method on the detail view, or try to fit the button on the list.

ui/ui-components/components/export/List.tsx
```diff
@@ -52,14 +55,55 @@ export default function ExportsList(props: Props) {
...
+          `/exports/retryFailed`,
+          {
+            destinationId: _export.destination.id,
+            startTimestamp: _export.createdAt,
+            endTimestamp: _export.createdAt,
+          },
+          { useCache: false }
+        );
+        if (count) {
+          successHandler.publish({ message: "Export Retried" });
```
edmundito: Not sure why, but we use `(successHandler | errorHandler).set` in core. What's the difference between set and publish?

ui/ui-components/utils/apiData.ts
```diff
@@ -443,8 +443,9 @@ export namespace Actions {
     typeof ExportsList.prototype.runWithinTransaction
   >;
   export type ExportsRetryFailed = AsyncReturnType<
-    typeof ExportsRetryFailed.prototype.runWithinTransaction
+    typeof ExportsRetryFailedExport.prototype.runWithinTransaction
   >;
+  export type ExportsRetryFailedAction = ExportsRetryFailedExport;
```
edmundito: Do we need this or should we import from `core` directly?

ui/ui-components/pages/export/[id]/edit.tsx
```diff
@@ -27,9 +33,47 @@ export const getServerSideProps = withServerErrorHandler(async (ctx) => {
...
 const Page: NextPageWithInferredProps<typeof getServerSideProps> = ({
-  _export,
   groups,
+  ...props
 }) => {
+  const { client } = useApi();
+  const [_export, setExport] = useState(props._export);
+  const [loading, setLoading] = useState(false);
+
+  const retryExport = useCallback(async () => {
```
Comment (edmundito) [inline ui/ui-components/pages/export/[id]/edit.tsx]: A lot of this logic seems to be repeated in two files. I wonder if we could encapsulate part, if not all of this logic, maybe in a hook?

ui/ui-components/components/export/List.tsx
```diff
@@ -52,14 +55,55 @@ export default function ExportsList(props: Props) {
...
+    async (_export: Models.ExportType, index: number) => {
+      setLoading(true);
+      try {
+        const { count } = await client.requestAction<ExportsRetryFailed>(
+          "get",
+          `/exports/retryFailed`,
+          {
+            destinationId: _export.destination.id,
+            startTimestamp: _export.createdAt,
+            endTimestamp: _export.createdAt,
```
pedroslopez: Hmm, this _probably_ finds just the export you want, but it might grab others if they have the same `createdAt` which would be an unexpected consequence to the user. I think a safer way would be to modify the action to accept an individual `exportId`.

core/src/modules/ops/export.ts
pedroslopez: Yeah, this could affect the existing "retry failed" functionality. After #2818, we are now canceling older exports if there are multiple for the same record/destination ([test case here](https://github.com/grouparoo/grouparoo/blob/eb40f33317cb9a003676f2bb74d358f55a15a9d8/core/__tests__/models/destination/plugins/exportRecords.ts#L695)). This means that these intentionally canceled ones would also be retried when doing it by date range.

Seeing this and the date range thing makes me think that perhaps this is a different method, `ExportOps.retryExportById` or similar. To ensure this new one and the existing one are retrying in the same way we could break out the actual updating of the export into a separate function and share that.

pedroslopez (changes requested):

ui/ui-components/components/export/List.tsx
krishnaglick:
```
  set(data: T) {
    this.publish(data);
  }
```
Apparently nothing!

krishnaglick: I'm not sure my `catch` code path will never be traversed since we eat all errors client-side.

ui/ui-components/pages/export/[id]/edit.tsx
Comment (krishnaglick) [reply ui/ui-components/pages/export/[id]/edit.tsx]: I think the best option may be to extract the meat of the function and just export that. Otherwise you'll end up passing in all the hook setters, etc. I'll move it, lemme know what you think!

core/src/models/Export.ts
```diff
@@ -314,7 +314,7 @@ export class Export extends CommonModel<Export> {
     }
   }
 
-  static async retryFailed(
+  static retryFailed(
```
krishnaglick, line 317: Doesn't need to be async since we're not awaiting anything.

ui/ui-components/client/client.ts
krishnaglick: We can use this to handle type inputs on the client anyway, but it'll be better.

core/src/actions/exports.ts
```diff
@@ -141,3 +141,23 @@ export class ExportsRetryFailed extends AuthenticatedAction {
...
+  inputs = {
+    exportId: { required: true },
+  };
+
+  async runWithinTransaction({
+    params,
+  }: {
+    params: ParamsFrom<ExportsRetryFailedById>;
+  }) {
+    console.log(params);
```
edmundito: console.log

edmundito (commented):
👍🏼

core/src/config/routes.ts
```diff
@@ -37,6 +37,7 @@ export const DEFAULT: { [namespace]: () => RoutesConfig } = {
         { path: "/v:apiVersion/exports", action: "exports:list" },
         { path: "/v:apiVersion/exports/totals", action: "exports:totals" },
         { path: "/v:apiVersion/exports/retryFailed", action: "exports:retryFailed" },
+        { path: "/v:apiVersion/exports/retryFailed/:exportId", action: "exports:retryFailedById" },
```
pedroslopez: I agree with what you said in the DevSync - these should be `POST`s! (including the existing `retryFailed` one) Could you move them over and make the relevant change in the client call from the frontend?

core/src/modules/ops/export.ts
```diff
@@ -184,22 +210,21 @@ export namespace ExportOps {
...
+    exportId: string,
+    saveExports = true
+  ) {
+    const where: WhereOptions<Export> = {
+      state: { [Op.in]: ["failed", "canceled"] },
+      id: exportId,
+    };
 
-    return Export.count({ where });
+    return retryExport(where, saveExports);
```
pedroslopez: nitpick-ish: this saveExports stuff is really meant for previewing how many exports the bulk action would affect, which isn't really needed for the single export case. When attempting to retry a single export, if nothing has been updated I would argue that should be an error thrown back to the action.

core/src/modules/ops/export.ts
```diff
@@ -184,22 +210,21 @@ export namespace ExportOps {
...
-      );
+    return retryExport(where, saveExports);
+  }
 
-      return count;
-    }
+  /**
+   * Finds finds and retries a failed or canceled export by id
+   */
+  export async function retryFailedExportById(
```
pedroslopez: Please add tests for this new method/action. There are some for the batch case in `__tests__/models/export.ts`.

edmundito:
@krishnaglick I feel that this PR is now doing too much: Both the feature change and the action cleanup.

krishnaglick:
> @krishnaglick I feel that this PR is now doing too much: Both the feature change and the action cleanup.

I can pull it out, but it did catch a bug. This is definitely a larger PR but I'm pretty sure it didn't break anything since it's just a type-inference fix.

core/src/config/routes.ts
```diff
@@ -122,6 +121,8 @@ export const DEFAULT: { [namespace]: () => RoutesConfig } = {
         { path: "/v:apiVersion/team", action: "team:create" },
         { path: "/v:apiVersion/team/initialize", action: "team:initialize" },
         { path: "/v:apiVersion/team/member", action: "teamMember:create" },
+        { path: "/v:apiVersion/exports/retryFailures", action: "exports:retryFailed" },
```
pedroslopez: the route and action names should probably match

ui/ui-components/client/client.ts
```diff
@@ -64,6 +66,16 @@ export class Client {
     } = () => ({})
   ) {}
 
+  // This will replace the request method below in the future
+  public async requestAction<A extends Action>(
```
pedroslopez, line 70: Should there be a story for this?

ui/ui-components/client/client.ts
```diff
@@ -64,6 +66,16 @@ export class Client {
...
+    verb: Method = "get",
+    path: string,
+    data: ParamsFrom<A>, // TODO: Make ParamsFrom handle required
```
pedroslopez: I think this `TODO` is done?

pedroslopez (commented):
Looks good, just one suggestion on keeping the names consistent.

pedroslopez (approved):

ui/ui-components/client/client.ts
krishnaglick: https://www.pivotaltracker.com/story/show/181297639

PR #2979, opened 2022-02-17
array properties can be mapped through non-unique properties

pedroslopez:
## Change description

In #2119 we introduced non-unique mappings, and added a restriction on being able to map array properties through non-unique mappings. After looking into this more, it seems there's no reason why we can't map array properties. If plugins return values per record ID for array properties then they will be set normally, and if not then `applyNonUniqueMappedResultsToAllRecords` will take care of fanning it out to all other records.

One thing I did notice, which may have led to the restriction in the first place is that there are some assumptions in the `applyNonUniqueMappedResultsToAllRecords` for the *mapping*, grabbing the first value, which would not work if the *mapping* was an array property. I believe this may have been the intention of the existing validation. However, there's already validations and tests that ensure that a source cannot map to an array property:

https://github.com/grouparoo/grouparoo/blob/d13529835b1e56c284b39bc37209749f531ff6a1/core/src/modules/ops/source.ts#L312-L320

(note the `values[0]`)

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/property/property.ts
+++ b/core/__tests__/models/property/property.ts
@@ -980,15 +980,6 @@ describe("models/property", () => {
           /Unique Property .+ cannot be mapped through a non-unique Property/
         );
       });
-
-      test("properties mapped through non-unique properties cannot be arrays", async () => {
-        await secondarySource.setMapping({ last_name: "lastName" });
-        await expect(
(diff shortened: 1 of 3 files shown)


evantahler (approved):
This is the best kind of fix - it already worked!
Very good writeup in the description.

PR #2984, opened 2022-02-21
OAuth and test connection improvements in App edit page

edmundito:
## Change description
While working on Hubspot OAuth, I noticed a few things that could be improved upon the App settings page:

1. Changes to getting token
    - Changed the "Sign in with OAuth" to "Request Token" because that's the ultimate action we're doing. 
    - The token field is only shown when there is a token, otherwise just shows the button
2. The "Test connection" button is now disabled when not all required fields are filled out

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/pages/app/[id]/edit.tsx
+++ b/ui/ui-components/pages/app/[id]/edit.tsx
@@ -188,6 +188,20 @@ const Page: NextPageWithInferredProps<typeof getServerSideProps> = (props) => {
     }
   };
 
+  const watchedFields = watch();
+
+  const hasAllRequiredFields = useMemo(() => {
+    return pluginOptions
+      .filter((opt) => opt.required)
+      .reduce(
+        (acc, opt) =>
+          acc &&
+          !!watchedFields.options[opt.key] &&
+          watchedFields.options[opt.key] !== "",
+        true
+      );
+  }, [pluginOptions, watchedFields]);
+
   return (
     <>
       <Head>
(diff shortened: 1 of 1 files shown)


pedroslopez (commented):
The reason why the field was being shown / editable is because you're still able to set the option through an environment variable. In this case you would need to type in `HUBSPOT_REFRESH_TOKEN` or whatever you set in your env.

You could do this manually by editing the config files, but I think we're trying to move away from manual edits to these in general.

ui/ui-components/pages/app/[id]/edit.tsx
```diff
@@ -360,23 +377,26 @@ const Page: NextPageWithInferredProps<typeof getServerSideProps> = (props) => {
...
-                                    required={opt.required}
-                                    type="password"
-                                    placeholder={opt.placeholder}
-                                    name={`options.${opt.key}`}
-                                    {...register(`options.${opt.key}`)}
-                                  />
-                                  <Form.Text className="text-muted">
-                                    {opt.description}
-                                  </Form.Text>
+                                  {hasValue && (
```
Comment (pedroslopez) [inline ui/ui-components/pages/app/[id]/edit.tsx]: This is still only being shown if it has a value. I think for filling out based on env vars you would need this shown in the empty state, too.

Comment (edmundito) [reply ui/ui-components/pages/app/[id]/edit.tsx]: ohh got it, Ok will undo that.

edmundito:
@pedroslopez Updated again. Looks like it's going to be just a copy change and disable the test button until all the required fields are filled in.

evantahler (approved):
This PR got a lot smaller!

pedroslopez (approved):

PR #2988, opened 2022-02-21
Validate source mappings and option values for lists / typeaheads

pedroslopez:
## Change description

Provides better validation around mappings and list/typeahead options. This is specially useful for validating that columns set on options and mappings actually exist in the source.

Previously, columns for table source properties required at least one record in the system to validate by attempting an import. There was no external validation for source mappings. 

Most of the changes in this PR are within tests to make the new validations not conflict with existing tests.

### Development

### Impact

### Code review


core/src/modules/mappingHelper.ts
```diff
@@ -150,4 +154,21 @@ export namespace MappingHelper {
...
+      const validKeys = Object.keys(sourcePreview[0]);
+      if (!validKeys.includes(remoteKey)) {
+        throw new Error(
+          `"${remoteKey}" is not a valid remote mapping key for ${modelName(
+            instance
+          )} ${instance.name}`
+        );
+      }
+    }
+  }
```
pedroslopez, line 173: We're using the source preview to know which columns are available to map to for sources. This is equivalent to what the frontend is doing to populate the column selection dropdown.

core/src/modules/optionHelper.ts
```diff
@@ -257,11 +266,21 @@ export namespace OptionHelper {
...
+
+      const val = options[k].toString();
+      const opts = optionOptions[k];
+      if (opts && opts.length > 0 && !opts.includes(val)) {
+        throw new Error(
+          `"${val}" is not a valid value for ${type} ${modelName<ModelWithOptions>(
+            instance
+          )} option "${k}"`
+        );
+      }
```
pedroslopez, line 284: For options that provide a list of options, we are now checking that the provided value is in that list of options.

pedroslopez (commented):
Here are the main changes:

core/__tests__/actions/apps.ts
```diff
@@ -90,12 +90,13 @@ describe("actions/apps", () => {
...
       expect(pluginOptions).toEqual([
         { key: "fileId", required: true },
+        { key: "environment", required: false },
         { key: "password", required: false },
         { key: "oAuthToken", type: "oauth-token", required: false },
         { key: "_failRemoteValidation", required: false },
       ]);
       expect(options).toEqual({
-        fileId: { options: ["a", "b"], type: "list" },
+        environment: { options: ["staging", "production"], type: "list" },
```
evantahler, line 99: This change captures the problem in a nutshell!

plugins/@grouparoo/spec-helper/src/lib/specHelper.ts
```diff
@@ -410,24 +423,25 @@ export namespace helper {
...
                 description: "the column to choose",
                 type: "list",
-                options: async () => {
-                  return [
-                    { key: "id", examples: [1, 2, 3] },
-                    { key: "fname", examples: ["mario", "luigi", "peach"] },
-                    { key: "lname", examples: ["mario", "mario", "toadstool"] },
-                  ];
+                options: async ({ sourceOptions }) => {
+                  const rows = TestPluginData[sourceOptions.tableWithOptions];
```
evantahler, line 436: 👍 nice

core/src/models/Schedule.ts
```diff
@@ -151,8 +151,8 @@ export class Schedule extends CommonModel<Schedule> {
     return OptionHelper.getOptions(this, sourceFromEnvironment);
   }
 
-  async setOptions(options: SimpleScheduleOptions) {
-    return OptionHelper.setOptions(this, options);
+  async setOptions(options: SimpleScheduleOptions, externallyValidate = true) {
+    return OptionHelper.setOptions(this, options, externallyValidate);
```
evantahler, line 155: lots more `externallyValidate(bool)` which makes sense

plugins/@grouparoo/mailchimp/__tests__/integration/mailchimpImport.ts
```diff
@@ -417,9 +417,9 @@ describe("integration/runs/mailchimp-import", () => {
 
         await run.updateTotals();
         expect(run.state).toBe("complete");
-        expect(run.importsCreated).toBe(10);
+        expect(run.importsCreated).toBe(20);
         expect(run.recordsCreated).toBe(0);
-        expect(run.recordsImported).toBe(5);
+        expect(run.recordsImported).toBe(10);
```
evantahler, line 422: The mailchimp importer was always sort of wonky, and dependent on the test data that we keep seeming to delete 😢

core/src/modules/optionHelper.ts
```diff
@@ -18,15 +18,15 @@ import { SourcesCache } from "../modules/caches/sourcesCache";
 
 export const ObfuscatedOptionString = "__ObfuscatedOption";
 
-type ModelWithMapping = Source | Destination | Schedule | Property | App;
+type ModelWithOptions = Source | Destination | Schedule | Property | App;
```
evantahler, line 21: Oops!

core/src/modules/optionHelper.ts
```diff
@@ -323,6 +342,57 @@ export namespace OptionHelper {
...
+    } else if (instance instanceof App) {
+      const appOptions = await instance.appOptions();
+      return optionOptionsFromObject(appOptions);
+    } else if (instance instanceof Source) {
+      const sourceConnectionOptions = await instance.sourceConnectionOptions(
+        options
+      );
+      return optionOptionsFromObject(sourceConnectionOptions);
+    }
+  }
```
evantahler: Now we've got 2 big case statements in this helper (this + validateOptions) that differ per model.  This sort of looks like we should be doing less in this helper and more per-model maybe?

evantahler (changes requested):
Lots of heroic test updates here! 

I think we should think about the optionHelper a little critically here - it might be doing too much work to support multiple models

plugins/@grouparoo/mailchimp/__tests__/integration/mailchimpImport.ts
pedroslopez: yeah... i was really confused when there was nothing there- same happened with snowflake, too 🤔 
There were also two missing custom fields in mailchimp 🤷🏼 

But I _think_ these should be the values for what the `.env.example` states and the sample data provided in  `data/profiles.csv`

core/src/modules/optionHelper.ts
pedroslopez: Yeah, that's a good point. These are _similar_ but they each have slight differences.

I refactored the `validateOptions` in dfb3edf / 5b82795 to be called from the model so we can get the data in the right format and then pass it to the common validation logic.

evantahler (approved):
Yeah, moving the `validateOptions` methods into the models is less convoluted!

Approved as-is, but a question:  Is it /always/ the case that if a validateOptions method exists, it should always be called as the first step in `OptionHelper.setOptions?`. If so, it might be nice to have that done automatically (vs in each models setOptions) method

PR #2989, opened 2022-02-22
Fix/hubspot company key

pauloouriques:
## Change description

This PR contains the code to fix the companyKey destination options issues when changing to another value.

### Development

### Impact

### Code review


plugins/@grouparoo/hubspot/src/lib/export-contacts/exportRecord.ts
```diff
@@ -180,6 +179,7 @@ async function syncAssociations(
   client: HubspotClient,
   contact: any,
   companyKey: string,
+  companyRecordField: string,
```
bleonard, line 182: but it's not actually using it?

bleonard (commented):
I think I need some more information on this. I can't actually figure out how the new field is being used.
Ideally, we shouldn't have to set this. We'd only set the `companyKey` and that would trigger the behavior of adding a required field of `Company.Name` (for example) and not `Company.hs_object_id` by the way.
Then maybe `exportRecord` would use `Company.X` to know to make the mapping and not the option itself. That would also know it's changed between old and new, I suppose.

plugins/@grouparoo/hubspot/src/lib/export/options.ts
```diff
@@ -65,34 +67,44 @@ class OptionsHandler {
...
+    }
+
+    const contactFields = await this.getObjectMatchNames("CONTACT");
+    out.companyRecordField.options = contactFields;
+    if (
+      !contactFields.includes(destinationOptions.companyRecordField?.toString())
+    ) {
+      destinationOptions.companyRecordField = null;
+    }
     return out;
```
edmundito, line 108: Maybe a more concise implementation of the function would be:

```ts
getContactOptions(...): Promise<DestinationOptionsMethodResponse> {
  const companyKey = searchableCompanyFields.includes(...) ? { ... } : null
  const contactFields = await ...
  const companyRecordField = contactFields.includes(...) ? { ... } : null
  
  return { companyKey, companyRecordField }
}
```

plugins/@grouparoo/hubspot/src/lib/export-contacts/destinationMappingOptions.ts
```diff
@@ -108,3 +109,26 @@ export const getUserFields = async (
...
+  if (!options) {
+    return;
+  }
+  const companyKeys = ["companyKey", "companyRecordField"];
+  // needs either none or all keys
+  let count = 0;
+  for (const key of companyKeys) {
+    const value = options ? (options[key] || "").toString().trim() : "";
+    if (value.length > 0) {
+      options[key] = value;
```
edmundito, line 123: Is it necessary to change the value of the options here? Is this function about sanitizing or checking that the values are there?

plugins/@grouparoo/hubspot/src/lib/export-contacts/destinationMappingOptions.ts
```diff
@@ -9,22 +9,23 @@ const importantFields = ["firstname", "lastname", "company"];
...
-          type = "phoneNumber";
+    const known = await getUserFields(client, appOptions);
+    if (destinationOptions?.companyRecordField) {
+      let index = -1;
+      for (let i = 0; i < known.length; i++) {
+        if (known[i].key === destinationOptions.companyRecordField) {
+          index = i;
           break;
+        }
+      }
```
edmundito, line 23:
```suggestion
      const index = known.findIndex(({ key }) => key === destinationOptions.companyRecordField);
```

plugins/@grouparoo/hubspot/src/lib/export-contacts/exportRecord.ts
pauloouriques: actually, the way I did this time it's not using `Company.Name`, it's using like: `oldRecordProperties[companyRecordField]`.

plugins/@grouparoo/hubspot/src/lib/export-contacts/destinationMappingOptions.ts
pauloouriques: I did both, just in case.

pauloouriques: I did both, just in case

PR #3004, opened 2022-02-24
Put email regex back to old, fixed tests

krishnaglick:
## Change description

Put email regex back to old, fixed tests

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/validators/isEmail.ts
+++ b/core/src/modules/validators/isEmail.ts
@@ -1,8 +1,6 @@
 /**
  * While there are many more robust ways to validate an email, we have seen our customers use long domains or suffixes to map to internal tools, deactivated users, etc.
  * We will only be validating that 1) there is an @ and 2) there is a . after the @.
- *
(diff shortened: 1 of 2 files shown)


core/src/modules/validators/isEmail.ts
```diff
@@ -5,6 +5,6 @@
  * We cannot use import isEmail from "validator/lib/isEmail" until https://github.com/validatorjs/validator.js/pull/1435 is merged in.
  */
 export default function isEmail(string: string) {
```
pedroslopez: This comment is waiting on something to be merged that has been merged at this point. We have `validator` installed... could we just use that?

If not, maybe this comment should be removed

krishnaglick: That'd be an @bleonard call.

evantahler: I'd remove that comment - we still don't really want to use the more strict email validator from validator.js - they won't like the leading/trailing spaces

evantahler (approved):

PR #3008, opened 2022-02-25
Postgres Query Source v2

pedroslopez:
## Change description

Postgres Query source can define a single query as a source option that will be used for schedules and importing properties.

image

### Development

### Impact

### Code review


core/src/models/Source.ts
```diff
@@ -145,8 +145,24 @@ export class Source extends CommonModel<Source> {
...
+    if (validateProperties) {
+      // check if properties are still valid
+      const properties = await this.$get("properties");
+      for (const property of properties) {
+        try {
+          await property.validateOptions();
+        } catch (err) {
+          throw new Error(`Error when validating properties: ${err}`);
+        }
+      }
```
pedroslopez, line 164: because validating the property options throws an error if the `column` is not valid, this prevents us from saving the source with a query that no longer returns the data for the properties.

Note that this would also affects changing the table on a table source with properties from the initial table. 

I think getting a validation error when updating the source is better than getting a bunch of errors down the line when trying to import records

plugins/@grouparoo/app-templates/src/source/queryv2/meta.ts
```diff
@@ -0,0 +1,138 @@
...
+  queryOptionDisplayName = null,
+  sourceOptions: additionalOptions,
+  getSampleRows,
+  getColumns,
+  getRows,
+  getPropertyValue,
+  getPropertyValues,
+  getRowCount,
+}) => {
+  const useAggregations = false;
```
pedroslopez, line 66: Query sources _can_ technically support aggregations, but we're disabling them since it's probably more efficient / natural for the user to do that within the query

plugins/@grouparoo/app-templates/src/source/shared/propertyOptions.ts
```diff
@@ -53,62 +55,68 @@ export const getPropertyOptions: GetPropertyOptionsMethod = async (
     type: "typeahead",
     options: async ({ connection, appOptions, appId, sourceOptions }) => {
       const tableName = sourceOptions[tableNameKey]?.toString();
+      const sourceQuery = sourceOptions[sourceQueryKey]?.toString();
```
pedroslopez, line 58: here's the main change in these app-templates: in addition to the `tableName` if it exists, we are also passing down the `sourceQuery` if it exists so that the plugin can do what it needs to in either scenario. The rest of the code is basically the same.

plugins/@grouparoo/app-templates/src/source/shared/recordProperty.ts
```diff
@@ -33,13 +36,14 @@ export const getRecordProperty: GetRecordPropertyMethod = ({
...
     const tableName = sourceOptions[tableNameKey]?.toString();
     const matchName = Object.keys(sourceMapping)[0]; // tableCol
     const recordPropertyMatch = Object.values(sourceMapping)[0];
     const columnName = propertyOptions[columnNameKey]?.toString();
-    const aggregationMethod = <AggregationMethod>(
-      propertyOptions[aggregationMethodKey]
-    );
+    const aggregationMethod = useAggregations
+      ? <AggregationMethod>propertyOptions[aggregationMethodKey]
+      : AggregationMethod.Exact;
```
pedroslopez: Query sources don't have an `aggregationMethod` option, but we still want to group all the properties so we can get them in one go. Because of this I'm now defaulting to `exact` in core when determining groupings. Because the plugin is still the one saying which aggregations can be grouped via `groupAggregations`, I think this is safe.

Initially I had the aggregation method option for query sources too, just with a single option to select, but this added an extra unnecessary step in the UI that I don't think should be there only for internal reasons.

plugins/@grouparoo/postgres/src/lib/shared/getPropertyValue.ts
```diff
@@ -58,8 +59,9 @@ export const getPropertyValue: GetPropertyValueMethod<
       throw new Error(`${aggregationMethod} is not a known aggregation method`);
   }
 
-  const params = [tableName];
-  let query = `SELECT ${aggSelect} as __result FROM %I WHERE`;
+  const params: string[] = [];
+  const from = makeFromClause({ tableName, sourceQuery }, params);
```
pedroslopez, line 63: On the postgres side, lots of the table methods are also reused, but they now use this `makeFromClause` helper any time we need to query their source

core/__tests__/models/source/source.ts
```diff
@@ -409,6 +409,36 @@ describe("models/source", () => {
...
+      );
+
+      await expect(
+        source.setOptions({
+          table: "admins",
+          tableWithOptions: "admins",
+        })
+      ).rejects.toThrow(
+        /Error when validating properties: Error: "email" is not a valid value/
+      );
```
evantahler, line 436: Nice

core/__tests__/tasks/recordProperty/enqueue/enqueue.ts
```diff
@@ -1,5 +1,6 @@
 import { helper } from "@grouparoo/spec-helper";
 import { api, task, specHelper } from "actionhero";
+import { AggregationMethod } from "../../../../dist";
```
evantahler: probably import this from src

core/src/modules/ops/recordProperty.ts
```diff
@@ -128,7 +128,9 @@ export namespace RecordPropertyOps {
     for (const property of properties) {
       const options = await property.getOptions();
       const filters = await property.getFilters();
-      const aggregationMethod = options.aggregationMethod as AggregationMethod;
+      const aggregationMethod =
+        (options.aggregationMethod as AggregationMethod) ??
+        AggregationMethod.Exact;
```
evantahler: A reasonable default... but might confuse us in the future

evantahler (commented):
This is really great!

I'm a little concerned that in a few places you have code that assumes a missing aggregationMethod implies "exact"

```
 const aggregationMethod = useAggregations
      ? <AggregationMethod>(
          propertyOptions[Object.keys(propertyOptions)[0]][aggregationMethodKey]
        )
      : AggregationMethod.Exact;
``` 

Maybe it would be better to hard-code only the Exact aggregation methods for these new query-sources?  I'm worried that there are other places in the code (grouping record property imports?) that need to be able to look this up

pedroslopez:
@evantahler should we add a `defaultAggregationMethod` to the connection configuration (alongside `groupAggregations`)? We could then reference this from core if it's set.

It's a bit odd because core is grouping property imports based on a property option that the plugin may or may not have. (and we need this so we can actually import these properties returned in our query at once)

The alternative that I was using earlier was actually having the aggregationMethod as an option with only one thing to select (`exact`), which works but then adds an extra step when creating properties and may confuse people having that there.

**Edit** implemented the `defaultAggregationMethod` solution in [8411f99](https://github.com/grouparoo/grouparoo/pull/3008/commits/8411f996497d64e9d290513bd7e55ee82029858d)

PR #3009, opened 2022-02-25
Add destination delivery mode, cancel duplicate exports when in "once" mode

rwfeather:
## Change description
This adds:

- Delivery mode for destinations. Options are [once, continual]
  - Continual - "Can deliver the same record multiple times" - default
  - Once - "Guaranteed to only deliver a record at-most-once"
- UI for the destination delivery mode

This changes:
- Add a check in `sendExports` that will refuse to run any export whose destination is _currently_ set to "once"

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/destination/plugins/exportRecord.ts
+++ b/core/__tests__/models/destination/plugins/exportRecord.ts
@@ -963,19 +963,6 @@ describe("models/destination - with custom exportRecord plugin", () => {
         await record.addOrUpdateProperties({
           email: ["newEmail@example.com"],
         });
-        // const group = await helper.factories.group();
-        // await GroupMember.create({ recordId: record.id, groupId: group.id });
-        // await destination.updateTracking("group", group.id);
-
-        // await destination.setMapping({
-        //   customer_email: "email",
-        // });
-
-        // const destinationGroupMemberships: Record<string, any> = {};
-        // destinationGroupMemberships[group.id] = group.name;
-        // await destination.setDestinationGroupMemberships(
-        //   destinationGroupMemberships
-        // );
 
         await destination.exportRecord(record, true);
       });
--- a/core/__tests__/models/destination/plugins/exportRecord.ts
+++ b/core/__tests__/models/destination/plugins/exportRecord.ts
@@ -955,6 +955,85 @@ describe("models/destination - with custom exportRecord plugin", () => {
       });
     });
 
+    describe("when destination delivery mode is", () => {
+      let record: GrouparooRecord = null;
+
+      beforeEach(async () => {
+        record = await helper.factories.record();
+        await record.addOrUpdateProperties({
+          email: ["newEmail@example.com"],
+        });
+        // const group = await helper.factories.group();
+        // await GroupMember.create({ recordId: record.id, groupId: group.id });
+        // await destination.updateTracking("group", group.id);
+
+        // await destination.setMapping({
+        //   customer_email: "email",
+        // });
+
+        // const destinationGroupMemberships: Record<string, any> = {};
+        // destinationGroupMemberships[group.id] = group.name;
+        // await destination.setDestinationGroupMemberships(
+        //   destinationGroupMemberships
+        // );
+
+        await destination.exportRecord(record, true);
+      });
+
+      afterEach(async () => {
+        await record.destroy();
+      });
+
+      test("continual, we can export a record multiple times", async () => {
+        destination.deliveryMode = "continual";
+
+        await destination.exportRecord(record, true);
+
+        const _exports = await Export.findAll({
+          where: { destinationId: destination.id },
+        });
+        expect(_exports).toHaveLength(2);
+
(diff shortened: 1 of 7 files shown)


core/src/models/Destination.ts
```diff
@@ -56,6 +56,9 @@ export interface SimpleDestinationOptions extends OptionHelper.SimpleOptions {}
 const SYNC_MODES = ["sync", "additive", "enrich"] as const;
 export type DestinationSyncMode = typeof SYNC_MODES[number];
 
+const DELIVERY_MODES = ["continual", "once"];
```
evantahler: Be sure to add `as const` here so typescript can validate the values of the strings, not just "string"

rwfeather:
Test is broken in CI but green locally ☹️ . Tracking it down now.

core/__tests__/models/destination/plugins/exportRecord.ts
```diff
@@ -955,6 +955,72 @@ describe("models/destination - with custom exportRecord plugin", () => {
       });
     });
 
+    describe("when destination delivery mode is", () => {
+      let record: GrouparooRecord = null;
```
evantahler, line 959: TS note: You don't need to set it to null.

```suggestion
      let record: GrouparooRecord;
```

This makes the record start with a value of `undefiend`

core/__tests__/models/destination/plugins/exportRecord.ts
```diff
@@ -955,6 +955,72 @@ describe("models/destination - with custom exportRecord plugin", () => {
...
+        await record.addOrUpdateProperties({
+          email: ["newEmail@example.com"],
+        });
+
+        await destination.exportRecord(record, true);
+      });
+
+      afterEach(async () => {
+        await record.destroy();
+      });
```
evantahler, line 972: You may want to `Export.truncate()` between tests as well, to ensure that there is no test polution

core/src/actions/destinations.ts
```diff
@@ -175,6 +175,7 @@ export class DestinationEdit extends AuthenticatedAction {
     state: { required: false },
     options: { required: false, formatter: APIData.ensureObject },
     mapping: { required: false, formatter: APIData.ensureObject },
+    deliveryMode: { required: false },
```
evantahler, line 178: This change implies that there should be an update to the Action's test as well, confirming that you can set the deliver mode

core/src/migrations/000102-addDeliveryMode.ts
```diff
@@ -0,0 +1,19 @@
...
+export default {
+  up: async (queryInterface: Sequelize.QueryInterface) => {
+    await queryInterface.addColumn("destinations", "deliveryMode", {
+      type: DataTypes.STRING(191),
+      allowNull: false,
+      defaultValue: "continual",
+    });
+    await queryInterface.changeColumn("destinations", "deliveryMode", {
+      type: DataTypes.STRING(191),
+      allowNull: false,
```
evantahler, line 12: 👍 we end up with a non-nullable column

core/src/models/Destination.ts
```diff
@@ -167,6 +170,11 @@ export class Destination extends CommonModel<Destination> {
   @Column(DataType.ENUM(...SYNC_MODES))
   syncMode: DestinationSyncMode;
 
+  @AllowNull(false)
+  @Default("continual")
+  @Column(DataType.ENUM(...DELIVERY_MODES))
+  deliveryMode: DestinationDeliveryMode;
```
evantahler, line 176: There's probably a test to add that `foo` can't be used as a delivery mode.  That's probably going to need a `@BeforeSave` validation check. 

This code here is about setting the types, but it won't validate the data coming in... I think.  The test will let us know for sure!

core/src/modules/ops/destination.ts
```diff
@@ -1005,7 +1012,31 @@ WHERE
...
+        }
+
+        const mostRecentExport = await Export.findOne({
+          where: {
+            destinationId: destination.id,
+            recordId: consideredExport.recordId,
+            state: "complete",
+          },
+          order: [["completedAt", "desc"]],
+        });
```
evantahler, line 1030: this method will run for every `(const consideredExport of exportsWithChanges)`, which could be a lot of SQL statements.  We'll want to load this data in one SQL statement before the for loop for every export we are considering.

ui/ui-components/pages/model/[modelId]/destination/[destinationId]/edit.tsx
```diff
@@ -1,3 +1,4 @@
+import { DestinationDeliveryMode } from "@grouparoo/core/src/models/Destination";
```
Comment (evantahler) [inline ui/ui-components/pages/model/[modelId]/destination/[destinationId]/edit.tsx:1]: We can't import models from core - this will cause compication problems.  The best we can do is import API responses via `import { Actions } from "../../../../../utils/apiData"`

evantahler (changes requested):

PR #3011, opened 2022-02-28
fix(hubspot): don't modify the options object when getting destination options

pedroslopez:
## Change description

This was causing validation to fail because of unexpected `null`s, meaning that hubspot destinations could not be created.

Because these were in here to clear out invalid values, this should no longer be necessary since core's validation introduced in #2988 will prevent these from being set in the first place. 

Also fixed up the `filterEmptyOptions` method to also clear `null`s and `undefined`s 

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/optionHelper.ts
+++ b/core/src/modules/optionHelper.ts
@@ -381,7 +381,8 @@ export namespace OptionHelper {
     const opts = Object.assign({}, options);
 
     Object.keys(opts).forEach((k) => {
-      if (opts[k] === "") delete opts[k];
+      if (typeof opts[k] === "undefined" || opts[k] === null || opts[k] === "")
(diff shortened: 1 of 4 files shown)


core/src/modules/optionHelper.ts
```diff
@@ -381,7 +381,8 @@ export namespace OptionHelper {
     const opts = Object.assign({}, options);
 
     Object.keys(opts).forEach((k) => {
-      if (opts[k] === "") delete opts[k];
+      if (typeof opts[k] === "undefined" || opts[k] === null || opts[k] === "")
+        delete opts[k];
```
evantahler, line 385: Nice

evantahler (approved):

PR #3012, opened 2022-02-28
Consolidated unique key/name method

krishnaglick:
## Change description

Moved the unique key/name method into Common model.
Dealt with the fallout.

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/models/mapping.ts
+++ b/core/__tests__/models/mapping.ts
@@ -52,7 +52,7 @@ describe("models/mapping", () => {
     test("a source cannot have 2 mappings with the same remoteKey", async () => {
       await expect(
         source.setMapping({ email: "userId", accountId: "userId" })
-      ).rejects.toThrow(/already in use in table Mapping/);
+      ).rejects.toThrow(/Validation error/);
     });
 
     test("a source cannot have 2 mappings with the same recordId", async () => {
@@ -66,7 +66,7 @@ describe("models/mapping", () => {
           propertyId: userId.id,
           remoteKey: "bar",
         })
-      ).rejects.toThrow(/already in use in table Mapping/);
+      ).rejects.toThrow(/Validation error/);
     });
 
     test("a source cannot be mapped to a property of another model", async () => {
--- a/core/__tests__/models/mapping.ts
+++ b/core/__tests__/models/mapping.ts
@@ -52,7 +52,7 @@ describe("models/mapping", () => {
     test("a source cannot have 2 mappings with the same remoteKey", async () => {
       await expect(
         source.setMapping({ email: "userId", accountId: "userId" })
-      ).rejects.toThrow(/There is already a Mapping for/);
+      ).rejects.toThrow(/already in use in table Mapping/);
     });
 
     test("a source cannot have 2 mappings with the same recordId", async () => {
@@ -66,7 +66,7 @@ describe("models/mapping", () => {
           propertyId: userId.id,
           remoteKey: "bar",
         })
-      ).rejects.toThrow(/There is already a Mapping for/);
+      ).rejects.toThrow(/already in use in table Mapping/);
     });
 
     test("a source cannot be mapped to a property of another model", async () => {
--- a/core/__tests__/models/mapping.ts
+++ b/core/__tests__/models/mapping.ts
@@ -3,11 +3,11 @@ import { Mapping, App, Source, Property, GrouparooModel } from "../../src";
 
 describe("models/mapping", () => {
   helper.grouparooTestServer({ truncate: true, enableTestPlugin: true });
-  let model: GrouparooModel;
-  let app: App;
-  let source: Source;
 
   describe("mappingHelper", () => {
+    let model: GrouparooModel;
+    let app: App;
+    let source: Source;
     beforeAll(async () => {
       app = await helper.factories.app();
       ({ model } = await helper.factories.properties());
@@ -19,6 +19,12 @@ describe("models/mapping", () => {
         modelId: model.id,
       });
       await source.setOptions({ table: "users" });
+    });
+
+    afterAll(async () => {
+      await Source.truncate({ restartIdentity: true, cascade: true });
+      await Property.truncate({ restartIdentity: true, cascade: true });
+      await App.truncate({ restartIdentity: true, cascade: true });
     });
 
     test("a source can have a mapping set", async () => {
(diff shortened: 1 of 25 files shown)


core/src/models/Schedule.ts
```diff
@@ -408,14 +408,8 @@ export class Schedule extends CommonModel<Schedule> {
...
-    const count = await Schedule.scope(null).count({
-      where: {
-        id: { [Op.ne]: instance.id },
-        name: instance.name,
-        state: { [Op.ne]: "draft" },
-      },
-    });
-    if (count > 0) throw new Error(`name "${instance.name}" is already in use`);
+    // Note: This may be problematic
+    await instance.ensureUnique(Schedule);
```
krishnaglick: I'm unsure how necessary `Schedule.scope(null)` is here since the tests are passing.

core/src/models/App.ts
```diff
@@ -327,14 +327,7 @@ export class App extends CommonModel<App> {
...
   static async ensureUniqueName(instance: App) {
-    const count = await App.count({
-      where: {
-        id: { [Op.ne]: instance.id },
-        name: instance.name,
-        state: { [Op.notIn]: ["draft", "deleted"] },
-      },
-    });
-    if (count > 0) throw new Error(`name "${instance.name}" is already in use`);
+    await instance.ensureUnique(App);
```
evantahler: I think the reason for adding the check for either `name` or `key` to common model is that so every instance of CommonModel could use it... which then implies you could move these `@BeforeSave` hooks into CommonModel as well?

core/src/models/Schedule.ts
evantahler: I don't think it matters - unscoped is getting replaced with `WHERE NOT DRAFT` later on.  Also, we don't let people name their schedules in the UI any more, so the risk here is low.

core/src/classes/commonModel.ts
```diff
@@ -112,4 +112,40 @@ export abstract class CommonModel<T> extends Model {
...
+      state: string;
+      name?: string;
+      key?: string;
+    }> = {
+      id: { [Op.ne]: this.id },
+      state: { [Op.notIn]: ["draft", "deleted"] },
+    };
+
+    if (this.name !== undefined) {
+      whereOpts.name = this.name;
```
evantahler: I *think* the goal of this story is to homogenize the unique check we are doing - so both `name` and `id` should be compared the same way - with lower-casing

core/src/models/App.ts
krishnaglick: I think we lose access to `.count` if we're not passing in the class.

core/bin/dev
```diff
@@ -23,6 +23,6 @@ for p in $GROUPAROO_PIDS; do
 done
 
 "$NODEMON" -e js,jsx,ts,tsx --signal SIGTERM --ignore dist --watch ./src --watch ../plugins/**/*.js \
---exec "$TS_NODE" --transpile-only --log-error src/grouparoo.ts \
+--exec node --inspect -r 'ts-node/register/transpile-only' src/grouparoo.ts \
```
krishnaglick, line 28: @evantahler 
We don't technically need to make this optional since if you don't open the inspector it doesn't do anything new. However I can make it based off a `DEBUG=1` or something if you'd like.

core/src/classes/commonModel.ts
```diff
@@ -28,6 +29,8 @@ export abstract class CommonModel<T> extends Model {
    */
   abstract idPrefix(): string;
 
+  uniqueIdentifier?: string | string[];
```
krishnaglick: Happy for some feedback on this pattern. Only 3 models needed it.

core/src/models/Property.ts
```diff
@@ -285,6 +285,8 @@ export class Property extends CommonModel<Property> {
     return this.idIsDefault() ? ConfigWriter.generateId(this.key) : this.id;
   }
 
+  uniqueIdentifier = ["key", "sourceId"];
```
krishnaglick: It was a little strange to me that the old version didn't need to check the source:key uniqueness, but this was necessary to make the app run under this new setup.

core/bin/dev
evantahler: I was under the impression that running `node --inspect` used more memory or was slower in some way... if that's not the case, sure!

krishnaglick: If we trust [this](https://stackoverflow.com/a/70950887) stackoverflow post, it's fine!

core/src/models/DestinationGroupMembership.ts
```diff
@@ -40,6 +40,8 @@ export class DestinationGroupMembership extends CommonModel<DestinationGroupMemb
   @BelongsTo(() => Destination)
   destination: Destination;
 
+  uniqueIdentifier = ["groupId", "destinationId", "remoteKey"];
```
krishnaglick, line 43: Previously `destinationId` + `groupId` was one unique pair, and `destinationId` + `remoteKey` another.
Unsure if this needs to be expanded into a `[["destinationId", "groupId"], ["destinationId", "remoteKey"]]`.

core/src/models/Mapping.ts
```diff
@@ -49,6 +49,9 @@ export class Mapping extends CommonModel<Mapping> {
   @BelongsTo(() => Property)
   property: Property;
 
+  // TODO: remoteKey can come into play here
+  uniqueIdentifier = ["ownerId", "ownerType", "propertyId"];
```
krishnaglick: Adding to my comment in `core/src/models/DestinationGroupMembership.ts`
This had `ownerId + ownerType + propertyId` and `ownerId + ownerType + remoteKey` as separate identifiers and combining them caused some tests to fail. Leaning more towards a `[[],[]]` style but hoping not to as the error messages will get even more obnoxious.

core/src/models/Source.ts
```diff
@@ -141,6 +141,8 @@ export class Source extends CommonModel<Source> {
   @BelongsTo(() => GrouparooModel)
   model: GrouparooModel;
 
+  uniqueIdentifier = ["appId", "name"];
```
krishnaglick, line 144: This differs from the previous functionality with the addition of `appId`, let me know if it may be a problem.

core/src/modules/errors.ts
```diff
@@ -1,8 +1,19 @@
+import { WhereOptions } from "sequelize/types";
+
 export namespace Errors {
   export const ERROR_LEVELS = ["error", "info"] as const;
   export type ErrorLevel = typeof ERROR_LEVELS[number];
 
-  export class InfoError extends Error {
+  export class GrouparooError extends Error {
```
krishnaglick, line 7: The properties defined here are in part based on the usage [here](https://github.com/grouparoo/grouparoo/blob/62de8fde7fbcf658ba30e22d4b23d1d45e5444f4/core/src/config/errors.ts#L147)

evantahler (approved):
This is well done!

core/src/models/Mapping.ts
evantahler: I think the reason for both of the checks was to prevent both of the following:

1. 
ownerType: property
ownerId: email
remotekey: email_column

^ So you can't have 2 properties for the same source column

2. 
ownerType: destination
ownerId: mailchimp
propertyId: email

^ So you can't export the email property to 2 different mailchimp fieilds.

I think your solution doesn't cover both cases (the first is missing at the moment)... BUT, good news, we want to [change how mappings work rather significantly](https://www.pivotaltracker.com/story/show/181244601) in the near future, and we would need to remove check (2) anyway. 

I think that it's OK to have only one check, but I'd suggest the remoteKey version.

```suggestion
  uniqueIdentifier = ["ownerId", "ownerType", "remoteKey"];
```

evantahler (changes requested):

evantahler (approved):

core/__tests__/models/mapping.ts
```diff
@@ -66,7 +72,7 @@ describe("models/mapping", () => {
           propertyId: userId.id,
           remoteKey: "bar",
         })
-      ).rejects.toThrow(/There is already a Mapping for/);
+      ).rejects.toThrow(/Validation error/);
```
krishnaglick, line 75: DB catches this now. I added a test below specifically for the new afterSave logic.

core/src/models/Source.ts
bleonard: Does this mean there can't be two sources with the same app? That's not right?
Or am I misreading the effect of this?

bleonard: I think right now the code is ANDing together everything from this list, right?

krishnaglick: Yeah, this is effectively just a composite unique index.

PR #3058, opened 2022-03-01
Replace text by binds on all queries

pauloouriques:
## Change description

This PR treats and tests the query values that could contain apostrophes. 

### Development

### Impact

### Code review


plugins/@grouparoo/snowflake/src/lib/export/exportRecords.ts
```diff
@@ -254,7 +254,11 @@ function formatPayloadValues(payloadValues: string[]) {
   return payloadValues.reduce((prev, curr, idx) => {
     return (
       prev +
-      (!curr || curr.length === 0 ? null : `'${curr}'`) +
+      (curr === null ||
+      curr === undefined ||
+      curr.toString().trim().length === 0
+        ? null
+        : `'${curr.toString().replace(new RegExp("'", "gi"), "")}'`) +
```
bleonard: anything that is trying to fix u pthe string is going to be a risk in general.
Above, it's doing `${values.join(", ")` that should be something more like [this](https://docs.snowflake.com/en/user-guide/nodejs-driver-use.html#binding-array-for-bulk-insert).

```
connection.execute({
  sqlText: 'insert into t(c1, c2, c3) values(?, ?, ?)',
  binds: [[1, 'string1', 2.0], [2, 'string2', 4.0], [3, 'string3', 6.0]]
});
```

each record would be a `?` and then the right thing would passed to `binds`.

pauloouriques: ok. I'll try this out.

pauloouriques: It only works with `select` (we're already using it) and `insert` (I changed to use it).

plugins/@grouparoo/snowflake/src/lib/export/exportRecords.ts
```diff
@@ -75,63 +75,37 @@ const updateByDestinationIds: BatchMethodUpdateByDestinationIds = async ({
...
-  columns = columns.join(", ");
-  const query = `UPDATE ${table} SET ${columnsToSet} FROM (VALUES ${values}) newRecord(${columns}) WHERE ${table}.${primaryKey} = newRecord.CURRENT_PRIMARY_KEY`;
-  await client.execute(query);
+  const singleValuePlaceholder = Array.from(
+    { length: columns.length },
+    () => "?"
+  );
+  const valuesPlaceholders = Array.from(
+    { length: values.length },
+    () => `(${singleValuePlaceholder})`
```
bleonard, line 97: I think i'd feel better if this was like `singleValuePlaceholder.join(", ")` instead of relying on the `toString` behavior.

plugins/@grouparoo/snowflake/src/lib/export/exportRecords.ts
```diff
@@ -434,24 +426,25 @@ const removeFromGroups: BatchMethodRemoveFromGroups = async ({
...
-    values.push(`(${cleanedValues})`);
-  }
+  const values = payloads.map((payload) => Object.values(payload));
+  const singleValuePlaceholder = Array.from(
+    { length: columns.length },
+    () => "?"
+  );
+  const valuesPlaceholders = Array.from(
+    { length: values.length },
+    () => `(${singleValuePlaceholder})`
```
bleonard, line 445: same here ton the `join`

bleonard (approved):
Looks good!

PR #3070, opened 2022-03-02
[FIX] Assigning Record to Export upon send

evantahler:
Fixes a bug when assigning the Record to the Export to send. 

### Development

### Impact

### Code review

### DIFF
--- a/core/src/modules/ops/destination.ts
+++ b/core/src/modules/ops/destination.ts
@@ -605,7 +605,7 @@ export namespace DestinationOps {
     });
 
     for (const _export of _exports) {
-      const record = records.find((r) => (r.id = _export.recordId));
+      const record = records.find((r) => r.id === _export.recordId);
       exportedRecords.push({
         record,
         recordId: record?.id,
--- a/plugins/@grouparoo/logger/__tests__/integration/exportRecords.ts
+++ b/plugins/@grouparoo/logger/__tests__/integration/exportRecords.ts
@@ -1,0 +1,314 @@
+import path from "path";
+import fs from "fs";
+import { EOL } from "os";
+process.env.GROUPAROO_INJECTED_PLUGINS = JSON.stringify({
+  "@grouparoo/logger": { path: path.join(__dirname, "..", "..") },
+});
+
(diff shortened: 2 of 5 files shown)


core/src/modules/ops/destination.ts
```diff
@@ -605,7 +605,7 @@ export namespace DestinationOps {
     });
 
     for (const _export of _exports) {
-      const record = records.find((r) => (r.id = _export.recordId));
+      const record = records.find((r) => r.id === _export.recordId);
```
evantahler, line 608: This was the problem

pedroslopez: Hmm- I wonder how this didn't fail any other tests. I believe this would have been finding the first record regardless of its ID

pedroslopez (approved):

plugins/@grouparoo/logger/__tests__/integration/exportRecords.ts
```diff
@@ -0,0 +1,314 @@
...
+}
+
+describe("integration/runs/logger", () => {
+  let session;
+  let csrfToken: string;
+  let app: AsyncReturnType<App["apiData"]>;
+  let recordA: GrouparooRecord;
+  let recordB: GrouparooRecord;
+  let recordC: GrouparooRecord;
+  let destination: AsyncReturnType<Destination["apiData"]>;
```
edmundito, line 48: Does ```Awaited<ReturnType<Destination["apiData"]>>``` work?

edmundito (approved):

core/src/modules/ops/destination.ts
evantahler: We don't look at the record's ID... ever... in the destinations.  It's something I noticed specifically in the logger, that's why I added a test there for it

plugins/@grouparoo/logger/__tests__/integration/exportRecords.ts
evantahler: Probably!  I wanted to match the other tests for now

PR #3077, opened 2022-03-03
Add basic eslint recommendations to `core/`

edmundito:
## Change description

Based on the issue in #3070 that could have been prevented with linting, this adds the most basic eslint rules to `core` and fixes a number of existing issues. The rules are not as strict as our plan because there's currently a lot to fix, and the plan is to incrementally enable rules and fix issues as we find the time to do it.

### Development

### Impact

### Code review


core/.eslintrc.js
```diff
@@ -0,0 +1,36 @@
+module.exports = {
+  env: {
+    es2021: true,
+    node: true,
+  },
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/eslint-recommended",
```
edmundito, line 8: Notice here that I'm using `@typescript-eslint/eslint-recommended` instead of `@typescript-eslint/recommended`. This extension only makes sure that typescript plays nicely with `eslint:recommended` but does not enable the typescript recommended rules. Currently, there are a gargantuan number of problems with `core` so doing it all once is a herculean effort. At the very least we're getting base safety.

core/.eslintrc.js
```diff
@@ -0,0 +1,36 @@
...
+  },
+  overrides: [
+    {
+      files: ["*.ts", "*.tsx"],
+    },
+  ],
+  globals: {
+    fetch: true, // For isomorphic-fetch
+  },
+  ignorePatterns: ["__tests__", "dist", "public"],
```
edmundito, line 24: Lots of problems in `__tests__` to fix, so it's being ignored for this PR.

core/src/modules/broadcastHelper.ts
```diff
@@ -11,5 +11,7 @@ export async function broadcastModel<T>(
       model: await instance.apiData(),
       verb,
     });
-  } catch {}
+  } catch {
+    // Ignore error
+  }
```
edmundito, line 16: In the recommendations we get the [no-empty](https://eslint.org/docs/rules/no-empty) rule that discourages empty blocks. The easiest way to fix it is to put a comment.

core/src/modules/cloudCli.ts
```diff
@@ -74,7 +74,7 @@ export namespace CloudCLI {
       );
 
       let lastState: string;
-      while (true) {
+      for (;;) {
```
edmundito, line 77: Fix for [no-constant-condition](https://eslint.org/docs/rules/no-constant-condition)

evantahler (commented):
Nice!

these look to me mostly stylistic changes… did you see any logical problems?

core/src/modules/configLoaders/schedule.ts
```diff
@@ -20,7 +20,7 @@ export async function loadSchedule(
 ): Promise<IdsByClass> {
   let isNew = false;
 
-  if (configObject.hasOwnProperty("confirmProfiles")) {
+  if (Object.prototype.hasOwnProperty.call(configObject, "confirmProfiles")) {
```
edmundito, line 23: [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins)

core/src/modules/ops/record.ts
```diff
@@ -260,6 +258,7 @@ export namespace RecordOps {
     recordProperties: {
       [key: string]: (string | number | boolean | Date)[] | any;
     }[],
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     toLock = true,
```
edmundito, line 263: This argument is not really being used right now, so I'm not sure whether to remove it and update all the instances that use it or leave it.

core/src/modules/ops/source.ts
```diff
@@ -156,8 +156,6 @@ export namespace SourceOps {
       });
 
       return response;
-    } catch (error) {
-      throw error;
```
edmundito, line 160: Unnecessary catches because they only rethrow the error

core/src/modules/reset.ts
```diff
@@ -28,6 +28,7 @@ export namespace Reset {
    * * clears the redis cache
    * * clears resque
    */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
edmundito, line 31: I left these here because the test is definitely passing a callerId. But callerId is not used in any of these functions. I could update the usage of these so that callerId is no longer used and leave it for some future use.

core/src/tasks/system/status/status.ts
```diff
@@ -81,6 +81,7 @@ export class StatusTask extends CLSTask {
     if (!toStop) return false;
 
     // do not await so the promise can end so the server can shut down
+    // eslint-disable-next-line no-async-promise-executor
```
edmundito, line 84: It really does not like returning a promise with an async call. Looks like this is intentional but could be dangerous otherwise. Doc explains why:

[no-async-promise-executor](https://eslint.org/docs/rules/no-async-promise-executor#no-async-promise-executor)

edmundito:
@evantahler The style issues are always easy to fix. I highlighted a few areas in the comments that point out potential logic issues.

core/.eslintrc.js
```diff
@@ -0,0 +1,36 @@
...
+  ],
+  plugins: ["@typescript-eslint"],
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    ecmaVersion: 12,
+    sourceType: "module",
+  },
+  overrides: [
+    {
+      files: ["*.ts", "*.tsx"],
```
krishnaglick, line 18: I think you could just add `*.js` to ignorePatterns and you shouldn't need this?

core/src/modules/cloudCli.ts
krishnaglick: Do you think we should disable that rule? We could rewrite this function to not need an infinite loop but I'm not it's worth the refactor time.

core/src/modules/oAuth.ts
```diff
@@ -34,41 +34,37 @@ export class oAuthAccessTokenGetter {
...
+        expirationSeconds: number;
+      } = await response.json();
 
-        this.accessToken = token;
-        this.expirationDate = Date.now() + expirationSeconds * 1000;
-      }
-    } catch (e) {
-      throw e;
+      this.accessToken = token;
+      this.expirationDate = Date.now() + expirationSeconds * 1000;
```
krishnaglick, line 67: Unwrapped 👍

core/src/tasks/system/status/status.ts
krishnaglick: The proimise isn't returned. Does:
```
await api.process.stop();
process.nextTick(() => process.exit(0));
```
Even need to be wrapped?

I feel like maybe the entire block could be `api.process.stop().then(() => process.nextTick(() => process.exit(0)));`
@evantahler Thoughts?

core/src/modules/statusReporters.ts
```diff
@@ -528,7 +527,7 @@ function mergeMetrics(metrics: StatusMetric[]) {
       mergedMetrics.push(item);
     } else if (idx !== null) {
       for (const k in Object.keys(item)) {
-        if (item.hasOwnProperty(k)) {
+        if (Object.prototype.hasOwnProperty.call(item, k)) {
```
krishnaglick, line 530: Don't love this rule

core/src/modules/reset.ts
krishnaglick: 🤷 Probably not worth removing

core/src/modules/ops/record.ts
krishnaglick: If there's enough of these I'd say icebox a ticket to remove.

krishnaglick (commented):
I expect some merge conflicts with my just-merged code.

core/.eslintrc.js
edmundito: This gets added by `@typescript:recommended` but not `@typescript:eslint-recommended` So I need to manually add this override otherwise I have to run eslint with `eslint . --ext ts,tsx`

core/src/modules/statusReporters.ts
edmundito: Yeah, good discussion point. Should we allow it? Should we not use `hasOwnProperty`?

krishnaglick: Looks like [hasOwn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) exists and we should maybe just use that instead?

edmundito: hasOwn is super new. It looks like we'll have to upgrade to target es2022, we're still targeting es2019.
https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/#target-es2022

krishnaglick: I don't object :P

PR #3084, opened 2022-03-04
Fix Import association when multiple Models share the same primary key values

evantahler:
A bug was introduced in https://github.com/grouparoo/grouparoo/pull/2968 which incorrectly associated Imports with the same values (e.g. userId=1 and adminId=1) to the same record.

## Change description

Description here

### Development

### Impact

### Code review

### DIFF
--- a/core/__tests__/tasks/import/associateRecords.ts
+++ b/core/__tests__/tasks/import/associateRecords.ts
@@ -275,6 +275,79 @@ describe("tasks/import:associateRecords", () => {
     await source.destroy();
   });
 
+  describe("with similar unique keys", () => {
+    let externalId: Property;
+
+    beforeAll(async () => {
+      externalId = await helper.factories.property(
+        primarySource,
+        {
+          type: "integer",
+          unique: true,
+          key: "externalId",
+        },
+        { column: "externalId" }
+      );
+      externalId.update({ state: "ready" });
+    });
+
+    afterAll(async () => {
+      await externalId.destroy();
+    });
+
+    beforeEach(async () => {
+      await api.resque.queue.connection.redis.flushdb();
+      await GrouparooRecord.truncate();
+      await RecordProperty.truncate();
+      await Import.truncate();
+    });
+
+    test("model selection works in batches when primary key values match", async () => {
+      const runA = await helper.factories.run(primarySchedule);
+
+      const _importA = await helper.factories.import(runA, {
+        userId: "1",
+        externalId: "2",
+      });
+      await specHelper.runTask("import:associateRecords", {});
+
+      const _importB = await helper.factories.import(runA, {
+        userId: "2",
+        externalId: "1",
+      });
+      await specHelper.runTask("import:associateRecords", {});
+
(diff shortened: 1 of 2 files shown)


core/src/modules/ops/record.ts
```diff
@@ -831,12 +831,16 @@ export namespace RecordOps {
     }
 
     const recordProperties = await RecordProperty.findAll({
-      where: { unique: true, rawValue: rawValues },
+      where: {
+        propertyId: uniqueProperties.map((p) => p.id),
```
evantahler, line 833: This addition will include only those existing records for the properties in this model/source now

core/__tests__/tasks/import/associateRecords.ts
```diff
@@ -272,4 +274,67 @@ describe("tasks/import:associateRecords", () => {
...
+      await RecordProperty.truncate();
+      await Import.truncate();
+    });
+
+    test("model selection works in batches when primary key values match", async () => {
+      const runA = await helper.factories.run(primarySchedule);
+      const runB = await helper.factories.run(otherSchedule);
+
+      const _importA = await helper.factories.import(runA, { userId: "1" });
+      const _importB = await helper.factories.import(runB, { adminId: "1" });
```
evantahler, line 394: Prior to this PR, these two imports would have made a single record

core/src/modules/ops/record.ts
```diff
@@ -831,12 +831,16 @@ export namespace RecordOps {
...
+        rawValue: rawValues,
+      },
       include: [GrouparooRecord],
     });
 
     const properties = await PropertiesCache.findAllWithCache(
-      undefined,
+      source instanceof Source ? source.modelId : undefined,
       "ready"
     );
```
pedroslopez, line 846: Could the same thing happen within the same model with two unique properties?

Say we have two unique id properties, user_id and external_id that are integers.

Record 1: user_id = 1, external_id=2
Record 2: user_id = 2, external_id = 3

I _think_ when matching this would look for any record property that has the value 2 or the value 3. Since we have a unique property (external_id) with the value 2, it would match that first record when it shouldn't.

Should the loop below that does the matching only match on recordProperties for a specific key?

pedroslopez (commented):
Could the same thing happen within the same model with two unique properties?

Say we have two unique id properties, user_id and external_id that are integers.

Record 1: user_id = 1, external_id=2
Record 2: user_id = 2, external_id = 3

I think when matching the second record this would look for any record property that has the value 2 or the value 3. Since we have a unique property (external_id) with the value 2, it would match that first record when it shouldn't.

Should the loop that does the matching only match on recordProperties for a specific key?

evantahler:
@pedroslopez that's an excellent find! https://github.com/grouparoo/grouparoo/pull/3084/commits/eaed2fbbeb399f02a78d4833d4da2074d80ba1bc adds a failing test for that case, and fixes it

edmundito (commented):
👍🏼 

Will defer to @pedroslopez for approval.

pedroslopez (approved):
👍🏼

PR #3092, opened 2022-03-10
Fix source edit page preview so that it always updates when source options change

edmundito:
## Change description

This fixes an issue when setting the options for a source, it would update the preview with the previous source options. The cause was because of effects related to the source.

When setting the options for the first time, it would fail to load a preview because the preview was loaded with the previous source state. The fix now uses the updated source to load the preview.

### Development

### Impact

### Code review

### DIFF
--- a/ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
+++ b/ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
@@ -237,8 +237,8 @@ const Page: NextPage<Props> = ({
   });
 
   const loadPreview = useCallback(
-    async (previewAvailable = source.previewAvailable) => {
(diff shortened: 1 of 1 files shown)


ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx
```diff
@@ -452,9 +452,11 @@ const Page: NextPage<Props> = ({
       setSource((src) => {
         src.options[optKey] = optValue;
         src.mapping = {};
+
+        loadPreview(src);
```
Comment (edmundito) [inline ui/ui-components/pages/model/[modelId]/source/[sourceId]/edit.tsx:456]: Not the best form to do this, but this results in a much simpler fix that avoids a complete refactor of how the page manages the form and the update state of the source.

evantahler (approved):

